PR #3, opened 2017-10-23
Add wizard and website creation code from vscode-azurewebservice

StephenWeatherford:
Needs tslint fixes before it can be checked in - might be easier to disable tslint temporarily, then fix after PR.

### DIFF
--- a/appservice/package.json
+++ b/appservice/package.json
@@ -1,7 +1,7 @@
 {
     "name": "vscode-azureappservice",
     "author": "Microsoft Corporation",
-    "version": "0.1.2",
+    "version": "0.2.0",
     "description": "Common tools for developing Azure App Service extensions for VS Code",
     "tags": [
         "azure",
--- a/appservice/package.json
+++ b/appservice/package.json
@@ -1,7 +1,7 @@
 {
     "name": "vscode-azureappservice",
     "author": "Microsoft Corporation",
-    "version": "0.1.1",
+    "version": "0.1.2",
     "description": "Common tools for developing Azure App Service extensions for VS Code",
     "tags": [
         "azure",
@@ -40,7 +40,8 @@
         "typescript": "^2.5.3",
         "tslint": "^5.7.0",
         "tslint-microsoft-contrib": "5.0.1",
-        "vscode": "^1.1.5"
+        "vscode": "^1.1.5",
+        "azure-arm-resource": "^2.0.0-preview"
     },
     "engines": {
         "vscode": "^1.16.0"
--- a/appservice/src/index.ts
+++ b/appservice/src/index.ts
@@ -3,10 +3,4 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-export * from './AzureAccountWrapper';
-export * from './errors';
 export * from './SiteWrapper';
-export * from './utils/ui';
-export * from './utils/random';
-export * from './WebsiteCreatorBase';
-export * from './wizard';
--- a/appservice/src/index.ts
+++ b/appservice/src/index.ts
@@ -3,4 +3,10 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+export * from './AzureAccountWrapper';
+export * from './errors';
 export * from './SiteWrapper';
+export * from './utils/ui';
+export * from './utils/random';
+export * from './WebsiteCreatorBase';
+export * from './wizard';
--- a/appservice/src/utils/random.ts
+++ b/appservice/src/utils/random.ts
@@ -5,8 +5,12 @@
 
 import * as crypto from "crypto";
 
-export function getRandomHexString(length: number): string {
-    const buffer = crypto.randomBytes(Math.ceil(length / 2));
-    let s = buffer.toString("hex").slice(0, length);
-    return s;
+export namespace randomUtils {
+
+    export function getRandomHexString(length: number): string {
+        const buffer = crypto.randomBytes(Math.ceil(length / 2));
+        let s = buffer.toString("hex").slice(0, length);
+        return s;
+    }
+
 }
--- a/appservice/src/utils/random.ts
+++ b/appservice/src/utils/random.ts
@@ -9,8 +9,7 @@ export namespace randomUtils {
 
     export function getRandomHexString(length: number): string {
         const buffer = crypto.randomBytes(Math.ceil(length / 2));
-        let s = buffer.toString("hex").slice(0, length);
-        return s;
+        return buffer.toString('hex').slice(0, length);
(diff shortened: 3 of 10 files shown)


appservice/src/index.ts
```diff
@@ -3,4 +3,10 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+export * from './AzureAccountWrapper';
```
EricJizbaMSFT: I think we should keep the 'api' surface as minimal as possible. Any change to any of the types you added becomes a breaking change and could cause issues. Ideally we just have one or two functions exported:
```
createAppService(credentials: ServiceClientCredentials, subscriptionId: string, appKind: AppKind)
```

or

```
createWebApp(credentials: ServiceClientCredentials, subscriptionId: string)
createFunctionApp(credentials: ServiceClientCredentials, subscriptionId: string)
```

You might need a few more parameters than that (a Memento for the persistence?) - but you get my point

appservice/src/utils/random.ts
```diff
@@ -0,0 +1,16 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as crypto from "crypto";
+
+export namespace randomUtils {
+
+    export function getRandomHexString(length: number): string {
```
EricJizbaMSFT: I don't think this is used ever?

appservice/src/index.ts
```diff
@@ -3,4 +3,10 @@
...
+export * from './errors';
 export * from './SiteWrapper';
+export * from './utils/ui';
+export * from './utils/random';
+export * from './WebsiteCreatorBase';
```
EricJizbaMSFT: It seems like you intend for each extension to extend WebsiteCreatorBase? As I said above, I would prefer to keep the api surface low and this would not do that

In addition, Chris mentioned he might have a 'Create App Service' entry point from the docker extension. It doesn't make sense for him to _also_ implement 'WebsiteCreator' when we could take care of it for him in this package

appservice/src/utils/ui.ts
```diff
@@ -0,0 +1,26 @@
...
+        const all: T[] = [];
+
+        for (let list = await first; list.length || list.nextLink; list = list.nextLink ? await client.listNext(list.nextLink) : []) {
+            all.push(...list);
+        }
+
+        return all;
+    }
+
+    export function getSignInCommandString(): string {
```
EricJizbaMSFT: Seems unnecessary to have an exported 'utility' function that returns a constant string and is only used once

appservice/src/wizard.ts
```diff
@@ -0,0 +1,197 @@
...
+    writeline(text: string) {
+        this.output.appendLine(text);
+    }
+
+    private onError(err: Error, step: WizardStep) {
+        if (err instanceof UserCancelledError) {
+            throw err;
+        }
+
+        this.writeline(`Error: ${err.message}`);
```
EricJizbaMSFT: I would let the extensions handle errors instead of printing it here. For example - what if the extension already prints multiline errors to the output window? (Like Nathan just implemented in app service) Then this would show up there twice

@nturinski we need to fix that because it's already an issue in app service haha

appservice/src/wizard.ts
```diff
@@ -0,0 +1,197 @@
...
+            try {
+                this.beforeExecute(step, i);
+                await this.steps[i].execute();
+            } catch (err) {
+                this.onError(err, step);
+            }
+        }
+
+        this._result = {
+            status: 'Completed',
```
EricJizbaMSFT: I don't think returning a WizardResult with status 'Completed' is very helpful (Same for 'PromptCompleted' above). The wizard would've thrown an error if it didn't complete, so you might as well return void and it would mean the same thing

appservice/src/wizard.ts
```diff
@@ -0,0 +1,197 @@
...
+            var previousId = this.persistenceState.get(persistenceKey);
+            var previousItem = previousId && resolvedItems.find(item => item.persistenceId === previousId);
+            if (previousItem) {
+                resolvedItems = ([previousItem]).concat(resolvedItems.filter(item => item !== previousItem));
+            }
+        }
+
+        const result = await vscode.window.showQuickPick(resolvedItems, options, token);
+        if (!result) {
+            throw new UserCancelledError();
```
EricJizbaMSFT: I like using UserCancelledError inside our own code, but I think we should make sure to catch it and return `undefined` at 'api' points. It's not really an _error_ and shouldn't be exposed like one

appservice/src/wizard.ts
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    async execute(): Promise<WizardResult> {
+        // Execute each step...
+        this.output.show(true);
+        for (var i = 0; i < this.steps.length; i++) {
+            const step = this.steps[i];
+
+            try {
+                this.beforeExecute(step, i);
```
EricJizbaMSFT: I don't see the value of 'beforeExecute'. Anything run in there could just be run in the first step

appservice/package.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "vscode-azureappservice",
     "author": "Microsoft Corporation",
-    "version": "0.1.1",
+    "version": "0.1.2",
```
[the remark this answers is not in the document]
StephenWeatherford: Agreed, done.  There's also a way to publish beta packages in npm, we'll probably want to use that when we're ready to publish and check in the changes to the extensions to use this.

appservice/src/index.ts
StephenWeatherford: I like that, will do.

appservice/src/utils/random.ts
StephenWeatherford: The reason I have this in here is because the security board wants people to only ever use security-strong random number generators (not the built-in JS versions), so I'd like to encourage their use by having them in the library.  Functions extension is the only one that currently uses it, so I could pull it into there, but personally I'd rather have it available in the library.  Thoughts?

appservice/src/utils/ui.ts
StephenWeatherford: Now we get to things that I simply copied from the wizard and didn't try to improve.  I'll take a look after the initial push.

StephenWeatherford:
Merging, but some comments still pending fixes, plus tslint.

appservice/src/utils/random.ts
EricJizbaMSFT: I see your point about 'encouraging best practices', but this function feels unrelated to app service. I don't want this to turn into a dumping ground for assorted utility functions like the Uitilties project in CAT

PR #6, opened 2017-10-27
 Leverage new zipdeploy instead of manual zip upload

### LINKED ISSUE #4: zip deploy should warn that it is a destructive action

EricJizbaMSFT:
This relies on PR #5 (the travis CI build will fail until that package is updated)

Fixes #4 
Fixes https://github.com/Microsoft/vscode-azureappservice/issues/116

The updated zipdeploy now supports much larger files, so I decided to remove our logic that excluded 'node_moduels'. Zipdeploy has intelligent file copying, etc., so I figured just better to stick with their defaults. I was able to deploy this:
screen shot 2017-10-26 at 7 27 30 pm
Resulting in a 1.4 GB site:
screen shot 2017-10-26 at 7 26 01 pm
On B1 app service plan:
screen shot 2017-10-26 at 7 37 24 pm

### DIFF
--- a/appservice/src/FileUtilities.ts
+++ b/appservice/src/FileUtilities.ts
@@ -53,8 +53,7 @@ export async function zipDirectory(folderPath: string): Promise<string> {
         zipper.pipe(zipOutput);
         zipper.glob('**/*', {
             cwd: folderPath,
-            dot: true,
-            ignore: 'node_modules{,/**}'
+            dot: true
         });
         void zipper.finalize();
     });
--- a/appservice/src/SiteWrapper.ts
+++ b/appservice/src/SiteWrapper.ts
@@ -10,7 +10,7 @@ import * as fs from 'fs';
 import { BasicAuthenticationCredentials } from 'ms-rest';
 import * as vscode from 'vscode';
 import KuduClient from 'vscode-azurekudu';
-import * as KuduModels from 'vscode-azurekudu/lib/models';
+import { DeployResult } from 'vscode-azurekudu/lib/models';
 import * as FileUtilities from './FileUtilities';
 
 export class SiteWrapper {
@@ -71,7 +71,6 @@ export class SiteWrapper {
 
     public async deployZip(fsPath: string, client: WebSiteManagementClient, outputChannel: vscode.OutputChannel): Promise<void> {
         outputChannel.show();
-        outputChannel.appendLine(`Starting deployment to '${this.appName}'...`);
         const kuduClient: KuduClient = await this.getKuduClient(client);
 
         let zipFilePath: string;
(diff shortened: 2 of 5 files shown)


appservice/src/FileUtilities.ts
```diff
@@ -53,8 +53,7 @@ export async function zipDirectory(folderPath: string): Promise<string> {
         zipper.pipe(zipOutput);
         zipper.glob('**/*', {
             cwd: folderPath,
-            dot: true,
-            ignore: 'node_modules{,/**}'
+            dot: true
```
nturinski: Kudu requires you to zip up the node modules too?

appservice/src/SiteWrapper.ts
```diff
@@ -70,8 +70,13 @@ export class SiteWrapper {
     }
 
     public async deployZip(fsPath: string, client: WebSiteManagementClient, outputChannel: vscode.OutputChannel): Promise<void> {
+        const yes: string = 'Yes';
+        const warning: string = `Are you sure you want to deploy to '${this.appName}'? This will overwrite any previous deployment and cannot be undone.`;
```
nturinski: Should we add this prompt to App Services?

EricJizbaMSFT: App services calls this function, so no need to change that repo

appservice/src/SiteWrapper.ts
```diff
@@ -80,47 +85,59 @@ export class SiteWrapper {
...
-            throw new Error(result.error);
-        } else if (result.output) {
-            outputChannel.appendLine(result.output);
+    private async waitForDeploymentToComplete(kuduClient: KuduClient, outputChannel: vscode.OutputChannel, pollingInterval: number = 5000): Promise<void> {
+        // Unfortunately, Kudu doesn't provide a unique id for a deployment right after it's started
+        // However, Kudu only supports one deployment at a time, so 'latest' will work in most cases
+        let deploymentId: string = 'latest';
+        let deployment: DeployResult = await kuduClient.deployment.getResult(deploymentId);
+        while (!deployment.complete) {
+            if (!deployment.isTemp && deployment.id) {
```
nturinski: Will DeployResult always conclude complete as true? What if there is no response or the client loses internet connection?

appservice/src/FileUtilities.ts
EricJizbaMSFT: They don't require it, but the default behavior of zipdeploy assumes a ready-to-run app. See [here](https://github.com/projectkudu/kudu/wiki/Deploying-from-a-zip-file) for more info, specifically in the part about SCM_DO_BUILD_DURING_DEPLOYMENT.

We could go against the default behavior and set  SCM_DO_BUILD_DURING_DEPLOYMENT to true, but I didn't think that was worth the added complexity. For example, do we overwrite the user's existing app setting if they have one? Do we put the app setting in a '.deployment' file? What if they already have a '.deployment' file?

Also - what if they don't have a standard nodejs project? Maybe they have a dotnet project with random files in a 'node_modules' folder. Maybe they have a nodejs project, but they install npm packages to a different folder.

I'm interested in what other people think, but I see my PR description about how zip deploy now supports larger files. I didn't think it was as bad to include the modules directly.

appservice/src/SiteWrapper.ts
EricJizbaMSFT: Either it will complete as true or an error will be thrown. Either way we exit the loop

appservice/src/FileUtilities.ts
EricJizbaMSFT: Discussed offline. I will keep this PR as-is and address at a later time. See issue #7

appservice/src/SiteWrapper.ts
whiskeyjay: Double quote in the warning message.

PR #10, opened 2017-11-08
Add createFunctionApp and createWebApp to exports

EricJizbaMSFT:
And fix all tslint/localize issues with previous code

Sorry it's a big review, but I don't believe I changed anything _functionally_ from Stephen's code.

I only tested with the Azure Functions extension. If we have time, I will also update the appservice extension, but I think that's low priority and possible cause for regressions before the next release.


appservice/src/SiteWrapper.ts
```diff
@@ -139,9 +141,9 @@ export class SiteWrapper {
     }
 
     private async getKuduClient(client: WebSiteManagementClient): Promise<KuduClient> {
-        const user: WebSiteModels.User = await this.getWebAppPublishCredential(client);
+        const user: User = await this.getWebAppPublishCredential(client);
         if (!user.publishingUserName || !user.publishingPassword) {
-            throw new Error('Invalid publishing credentials.');
+            throw new ArgumentError(user);
```
StephenWeatherford: nit: Personally not sure what you get out of using a specific error instead of simply Error, since it's rare that we actually handle errors based on ArgumentError, and since the output will be the same.

EricJizbaMSFT: The output is slightly different. Here's the code for the message:
`localize('azFunc.argumentError', 'Invalid {0}.', obj.constructor.name)`
Here are a few example messages:
`Invalid Site`, `Invalid ResourceGroup`, `Invalid User`

PR #13, opened 2017-11-08
Nturinski/feature9

### LINKED ISSUE #9: Share Local GIt Deploy code for Func Apps
The local git deployment method between App Services and Function Apps are the same, so we should include the local git deploy code in tools.

nturinski:
Fixes #9


appservice/thirdpartynotices.txt
```diff
@@ -36,3 +36,61 @@ OTHER DEALINGS IN THE SOFTWARE.
 END OF node-archiver NOTICES AND INFORMATION
 =========================================
 
+2. simple-git version 1.80.1 (https://github.com/steveukx/git-js)
```
EricJizbaMSFT: This isn't quite the format we're looking for. Take a look at this example: https://github.com/Microsoft/vscode-azurefunctions/blob/master/thirdpartynotices.txt

appservice/src/SiteWrapper.ts
```diff
@@ -116,6 +120,67 @@ export class SiteWrapper {
...
+            // tslint:disable-next-line:no-unsafe-any
+            if (err.message.indexOf('spawn git ENOENT') >= 0) {
+                throw new errors.GitNotInstalledError();
+            } else if (err.message.indexOf('error: failed to push') >= 0) { // tslint:disable-line:no-unsafe-any
+                const input: string | undefined = await vscode.window.showErrorMessage(pushReject, yes);
+                if (input === 'Yes') {
+                    await (<(a: string, b: string, c: object) => Promise<void>>localGit.push)(remote, 'HEAD:master', { '-f': true });
+                    // Ugly casting neccessary due to bug in simple-git. Issue filed
+                } else {
+                    throw new errors.UserCancelledError();
```
EricJizbaMSFT: It's fine to use this error internally, but we don't want to 'expose' it as an error outside of the npm package. (After all - it's not _technically_ an error).

The typical pattern is to return 'undefined' if the user canceled (In other cases, you could return the deployment id or some other useful piece of information)

appservice/src/SiteWrapper.ts
```diff
@@ -116,6 +120,67 @@ export class SiteWrapper {
         this.log(outputChannel, 'Deployment completed.');
     }
 
+    public async localGitDeploy(fsPath: string, client: WebSiteManagementClient, outputChannel: vscode.OutputChannel, servicePlan: string): Promise<void> {
+        vscode.window.showInformationMessage('It updated to 2.2');
```
EricJizbaMSFT: what's this message about?

appservice/src/SiteWrapper.ts
```diff
@@ -116,6 +120,67 @@ export class SiteWrapper {
...
+        let taskResults: [User, SiteConfigResource];
+        const kuduClient: KuduClient = await this.getKuduClient(client);
+        const yes: string = 'Yes';
+        const pushReject: string = 'Push rejected due to Git history diverging. Force push?';
+
+        if (!this.slotName) {
```
EricJizbaMSFT: You can simplify this code quite a bit (and completely get rid of the taskResults variable)
```
const [publishCredentials, config]: [User, SiteConfigResource] = await Promise.all([
    this.getWebAppPublishCredential(client),
    this.getSiteConfig(client)
]);
```

NOTE: You'll have to change the return type of `this.getSiteConfig` to `SiteConfigResource` - looks like it's incorrect at the moment? Not sure what the difference between `SiteConfig` and `SiteConfigResource` is

appservice/src/SiteWrapper.ts
```diff
@@ -116,6 +120,67 @@ export class SiteWrapper {
...
+            await localGit.push(remote, 'HEAD:master');
+        } catch (err) {
+            // tslint:disable-next-line:no-unsafe-any
+            if (err.message.indexOf('spawn git ENOENT') >= 0) {
+                throw new errors.GitNotInstalledError();
+            } else if (err.message.indexOf('error: failed to push') >= 0) { // tslint:disable-line:no-unsafe-any
+                const input: string | undefined = await vscode.window.showErrorMessage(pushReject, yes);
+                if (input === 'Yes') {
+                    await (<(a: string, b: string, c: object) => Promise<void>>localGit.push)(remote, 'HEAD:master', { '-f': true });
+                    // Ugly casting neccessary due to bug in simple-git. Issue filed
```
EricJizbaMSFT: Please add a link to the issue in the file

appservice/src/SiteWrapper.ts
```diff
@@ -143,11 +208,30 @@ export class SiteWrapper {
...
     }
+
+    private async updateScmType(client: WebSiteManagementClient, config: SiteConfigResource): Promise<void> {
+        const oldScmType: string = config.scmType;
+        const updateScm: string = `Deployment source for "${this.appName}" is set as "${oldScmType}".  Change to "LocalGit"?`;
+        const yes: string = 'Yes';
+        let input: string | undefined;
+
+        const updateConfig: SiteConfigResource = config;
+        updateConfig.scmType = 'LocalGit';
```
EricJizbaMSFT: Can you use a constant or enum for 'LocalGit'?

appservice/src/SiteWrapper.ts
```diff
@@ -143,11 +208,30 @@ export class SiteWrapper {
...
+        // to update one property, a complete config file must be sent
+        input = await vscode.window.showWarningMessage(updateScm, yes);
+        if (input === 'Yes') {
+            !this.slotName ?
```
EricJizbaMSFT: Please create a `SiteWrapper.updateConfiguration` method and call that

appservice/src/SiteWrapper.ts
nturinski: Whoops, sorry, that was just for testing. 😊

appservice/package.json
```diff
@@ -34,10 +34,12 @@
...
         "azure-arm-storage": "^3.1.0",
         "azure-arm-website": "^1.0.0-preview",
-        "vscode-azurekudu": "^0.1.2",
         "ms-rest": "^2.2.2",
-        "ms-rest-azure": "^2.4.0",
-        "vscode-nls": "^2.0.2"
+        "ms-rest-azure": "^2.4.4",
+        "vscode-nls": "^2.0.2",
+        "opn": "^5.1.0",
+        "simple-git": "^1.80.1",
```
EricJizbaMSFT: I just noticed `azure-arm-resource` is listed in both 'dependencies' and 'devDependencies'. Would you mind removing the dev one as a part of these changes?

nturinski:
Should we be localizing every string, even ones that only say 'Yes'?  And it's only strings that appear on the UI, correct?

appservice/src/SiteWrapper.ts
```diff
@@ -143,11 +208,27 @@ export class SiteWrapper {
...
+        const updateScm: string = localize('updateScm', `Deployment source for "${this.appName}" is set as "${oldScmType}".  Change to "${scmType}"?`);
+        const yes: string = 'Yes';
+        let input: string | undefined;
+
+        const updateConfig: SiteConfigResource = config;
+        updateConfig.scmType = scmType;
+        // to update one property, a complete config file must be sent
+        input = await vscode.window.showWarningMessage(updateScm, yes);
+        if (input === 'Yes') {
+            await this.updateConfiguration(client, updateConfig);
```
EricJizbaMSFT: For this function I would return 
```
const newConfig = await this.updateConfiguration(client, updateConfig);
return newConfig.scmType;
```
Otherwise return undefined

appservice/src/SiteWrapper.ts
```diff
@@ -143,11 +208,27 @@ export class SiteWrapper {
...
         return new KuduClient(cred, `https://${this.appName}.scm.azurewebsites.net`);
     }
+
+    private async updateScmType(client: WebSiteManagementClient, config: SiteConfigResource, scmType: string): Promise<string | undefined> {
+        const oldScmType: string = config.scmType;
+        const updateScm: string = localize('updateScm', `Deployment source for "${this.appName}" is set as "${oldScmType}".  Change to "${scmType}"?`);
+        const yes: string = 'Yes';
+        let input: string | undefined;
+
+        const updateConfig: SiteConfigResource = config;
```
EricJizbaMSFT: Why not use config directly?

appservice/src/SiteWrapper.ts
```diff
@@ -116,6 +126,61 @@ export class SiteWrapper {
...
+                    return undefined;
+                }
+            } else {
+                // tslint:disable-next-line:no-unsafe-any
+                throw new errors.LocalGitDeployError(err, servicePlan);
+            }
+        }
+
+        await this.waitForDeploymentToComplete(kuduClient, outputChannel);
+        return success;
```
EricJizbaMSFT: I would change waitForDeploymentToComplete to return DeployResult and just do:
```
return await this.waitForDeploymentToComplete(kuduClient, outputChannel);
```

 (rather than defining your own string for success)

appservice/src/SiteWrapper.ts
```diff
@@ -65,15 +69,21 @@ export class SiteWrapper {
...
+    public async updateConfiguration(client: WebSiteManagementClient, config: SiteConfigResource): Promise<SiteConfigResource> {
+        return this.slotName ?
+            await client.webApps.updateConfigurationSlot(this.resourceGroup, this.appName, config, this.slotName) :
+            await client.webApps.updateConfiguration(this.resourceGroup, this.appName, config);
+    }
+
     public async deployZip(fsPath: string, client: WebSiteManagementClient, outputChannel: vscode.OutputChannel): Promise<void> {
         const yes: string = 'Yes';
-        const warning: string = `Are you sure you want to deploy to "${this.appName}"? This will overwrite any previous deployment and cannot be undone.`;
+        const warning: string = localize('zipWarning', `Are you sure you want to deploy to "${this.appName}"? This will overwrite any previous deployment and cannot be undone.`);
```
EricJizbaMSFT: For strings like this you want to insert values slightly differently (It can't be translated the current way you have it):
```
localize('zipWarning', 'Are you sure you want to deploy to "{0}"? This will overwrite any previous deployment and cannot be undone.', this.appName);
```

.vscode/launch.json
```diff
@@ -0,0 +1,17 @@
+{
```
nturinski: I have no idea why this file was generated.  Any idea??

nturinski:
Tested with Azure App Services.  I was silly and forgot that I had changed branches after linking 😅

appservice/src/SiteWrapper.ts
```diff
@@ -116,7 +126,60 @@ export class SiteWrapper {
         this.log(outputChannel, 'Deployment completed.');
     }
 
-    private async waitForDeploymentToComplete(kuduClient: KuduClient, outputChannel: vscode.OutputChannel, pollingInterval: number = 5000): Promise<void> {
+    public async localGitDeploy(fsPath: string, client: WebSiteManagementClient, outputChannel: vscode.OutputChannel, servicePlan: string): Promise<DeployResult | undefined> {
```
EricJizbaMSFT: I don't think we should have servicePlan as a parameter since it's only used for telemetry purposes. The extension should take care of telemetry, not the npm package

EricJizbaMSFT: I wouldn't even include `LocalGitDeployError` in this package

appservice/src/errors.ts
```diff
@@ -23,3 +25,31 @@ export class ArgumentError extends Error {
         super(localize('azFunc.argumentError', 'Invalid {0}.', obj.constructor.name));
     }
 }
+
+// tslint:disable-next-line:max-classes-per-file
+export class GitNotInstalledError extends Error {
```
EricJizbaMSFT: I'm not sure we want to expose this error outside the package either. Maybe just show the install prompt and return undefined from the function

.vscode/launch.json
EricJizbaMSFT: You probably hit F5 on this workspace, even though you can't really 'F5' an npm package

appservice/src/SiteWrapper.ts
nturinski: How do we distinguish what caused the error if all errors return undefined?

appservice/src/SiteWrapper.ts
```diff
@@ -116,7 +127,60 @@ export class SiteWrapper {
...
+                const input: string | undefined = await vscode.window.showErrorMessage(pushReject, yes);
+                if (input === 'Yes') {
+                    await (<(remote: string, branch: string, options: object) => Promise<void>>localGit.push)(remote, 'HEAD:master', { '-f': true });
+                    // Ugly casting neccessary due to bug in simple-git. Issue filed:
+                    // https://github.com/steveukx/git-js/issues/218
+                } else {
+                    return undefined;
+                }
+            } else {
+                return undefined;
```
EricJizbaMSFT: Not all errors should return undefined. You should call `throw err` here. I was just saying don't throw a _custom_ error type because then you also have to export that error type

EricJizbaMSFT:
@nturinski Merged this so I can work on the 'Delete Function App' which is also in SiteWrapper. Feel free to make any leftover changes separately

PR #15, opened 2017-11-09
Add 'Delete Function App'

EricJizbaMSFT:
Copied the code over from appservice extension. Similar to createFunctionApp, I only tested this on Functions extension. We can update appservice extension after the next release

### DIFF
--- a/appservice/src/SiteWrapper.ts
+++ b/appservice/src/SiteWrapper.ts
@@ -130,9 +130,8 @@ export class SiteWrapper {
     }
 
     public async deployZip(fsPath: string, client: WebSiteManagementClient, outputChannel: vscode.OutputChannel): Promise<void> {
-        const yes: string = this._yes;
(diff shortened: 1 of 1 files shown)


appservice/src/SiteWrapper.ts
```diff
@@ -82,8 +91,46 @@ export class SiteWrapper {
...
+            await client.webApps.deleteSlot(this.resourceGroup, this.name, this.slotName);
+        } else {
+            await client.webApps.deleteMethod(this.resourceGroup, this.name, { deleteEmptyServerFarm: deletePlan });
+        }
+        outputChannel.appendLine(localize('DeleteSucceeded', 'Successfully deleted "{0}".', this.appName));
+    }
+
     public async deployZip(fsPath: string, client: WebSiteManagementClient, outputChannel: vscode.OutputChannel): Promise<void> {
-        const yes: string = 'Yes';
+        const yes: string = this._yes;
```
nturinski: Why define a const for the string if we have a class property already defined for localization??

EricJizbaMSFT: Haha just a 'find and replace' gone wrong

PR #16, opened 2017-11-09
Add vscode settings file to appservice folder

EricJizbaMSFT:
Previously I only had the settings here: https://github.com/Microsoft/vscode-azuretools/blob/master/vscode-azuretools.code-workspace

But I find myself opening _just_ the appservice folder a lot

### DIFF
--- a/appservice/.vscode/settings.json
+++ b/appservice/.vscode/settings.json
@@ -1,0 +1,13 @@
+{
+    "editor.formatOnSave": true,
+    "editor.insertSpaces": true,
+    "editor.tabSize": 4,
+    "files.insertFinalNewline": true,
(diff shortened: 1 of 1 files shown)


EricJizbaMSFT:
Not to completely ignore your comment on this one - but I just didn't feel like testing :)

I can look into it more after the release

jinglouMSFT:
What are you saying here?

EricJizbaMSFT:
@jinglouMSFT after I wrote my comment I knew you would say something ;)

Don't worry this change has absolutely zero effect on the user. It's for our internal dev use only (It's settings for our repo like 'format my code file when I click save')

PR #17, opened 2017-11-14
Add showScmPrompt to allow user to select scmType

nturinski:
Allows the users to change the scmType in the extension.  This is in preparation to implement one deploy entry point that will auto-detect the scmType and use the appropriate deployment method as discussed with @fiveisprime


appservice/src/SiteWrapper.ts
```diff
@@ -225,6 +234,20 @@ export class SiteWrapper {
         return await this.waitForDeploymentToComplete(kuduClient, outputChannel);
     }
 
+    public async showScmPrompt(currentScmType: string): Promise<string> {
+        const placeHolder: string = localize('scmPrompt', 'Current ScmType is "{0}".  Select a new deployment source.', currentScmType);
+        const scmQuickPicks: vscode.QuickPickItem[] = [];
+        // generate quickPicks to not include current type
+        for (const scmQuickPick of SCM_TYPES) {
```
nturinski: Rather than hiding the current active scmType, should I include all four types but just change the description/detail to indicate that it is the current active scmType?

appservice/src/SiteWrapper.ts
```diff
@@ -225,6 +220,27 @@ export class SiteWrapper {
...
+            if (scmQuickPick.label !== currentScmType) {
+                scmQuickPicks.push(scmQuickPick);
+            }
+        }
+
+        const quickPick: vscode.QuickPickItem = await vscode.window.showQuickPick(scmQuickPicks, { placeHolder: placeHolder });
+        return quickPick.label;
```
EricJizbaMSFT: Check if quickPick is undefined

appservice/src/SiteWrapper.ts
```diff
@@ -17,6 +17,15 @@ import { ArgumentError } from './errors';
 import * as FileUtilities from './FileUtilities';
 import { localize } from './localize';
 
+const noSupport: string = localize('noSupport', 'This deployment source is not currently supported and will be deployed by zip.');
```
EricJizbaMSFT: IMO the user should only see this message on deploy, instead of when they're changing the type

appservice/package.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "vscode-azureappservice",
     "author": "Microsoft Corporation",
-    "version": "0.4.0",
+    "version": "0.4.1",
```
EricJizbaMSFT: Adding `editScmType` feels more like a 'feature' than a 'bug fix' (in which case it should be '0.5.0')

but not a huge deal either way

nturinski: Okay, will we be including auto-detect deploying and Java/C# support to release 5.0.0 as well?

nturinski: Will we be including auto-detect deployment in version 5.0.0 as well?

appservice/src/SiteWrapper.ts
nturinski: Hmm, but I feel like as a user, I'd be annoyed to switch to GitHub, expecting to deploy, and finding out that it is not supported when I try to (possibly at a much later time).

appservice/package.json
EricJizbaMSFT: We can do whatever you want :)

appservice/src/SiteWrapper.ts
EricJizbaMSFT: Maybe we don't show GitHub here at all. It seems like the user will have to enter additional information anyways:
screen shot 2017-11-15 at 5 27 59 pm

Eventually we should prompt the user for the same info the portal asks for, but that seems like a separate feature

EricJizbaMSFT: (Either way the message needs to be 'not supported **in VS Code**' - your current message makes it feel like Azure App Service as a whole doesn't support Github)

nturinski: I'm fine with just removing GitHub and BitbucketGit until they are supported.

appservice/src/SiteWrapper.ts
```diff
@@ -17,6 +17,13 @@ import { ArgumentError } from './errors';
 import * as FileUtilities from './FileUtilities';
 import { localize } from './localize';
 
+const noSupport: string = localize('noSupport', 'This deployment source is not currently supported in VS Code and will be deployed by zip.');
```
StephenWeatherford: Where is this const used?

Can you add some screenshots of the experience?

nturinski: This was used for unsupported deployment sources when choosing a source type, but we have opted to remove them from the quickpick menu entirely.

PR #18, opened 2017-11-15
Add shared UI Package with AzureExplorer

EricJizbaMSFT:
For now, it only exports a common AzureTreeProvider implementation and UserCancelledError. In the future, it could also include things like the wizard

Take a look at the README I added for more info

(Quite a bit of this code is copied over from the functions extension, but I also modified it quite a bit to be more generic)


ui/src/explorer/AzureNode.ts
```diff
@@ -0,0 +1,85 @@
...
+        } else {
+            throw new ArgumentError(this);
+        }
+    }
+
+    public refresh(): void {
+        this.explorer.refresh(this.parent, false);
+    }
+
+    public openInPortal(): void {
```
nturinski: Does this belong in the SiteWrapper in the azureappservices package?  This doesn't seem like a UI action to me.

EricJizbaMSFT: This applies for all types of azure resources, though, not just app services (cosmosdb account or storage accounts would have the same logic)

EricJizbaMSFT:
Merged this so that I can start the process of initial publish. @StephenWeatherford feel free to review as you have time

I decided to go with the name @jinglouMSFT and I discussed. Its good enough for now and if need be we can always refactor into other packages down the line

ui/README.md
```diff
@@ -0,0 +1,87 @@
+# VSCode Azure SDK for Node.js - UI Tools
+
+This package provides common Azure UI elements for VS Code extensions.
+
+> NOTE: This package throws a `UserCancelledError` if the user cancels an operation. This error should be handle appropriately by your extension.
```
StephenWeatherford: handled

ui/index.d.ts
```diff
@@ -0,0 +1,108 @@
...
+    public refresh(node?: IAzureNode, clearCache?: boolean): void;
+    public loadMore(node: IAzureNode): Promise<void>;
+    public showNodePicker(expectedContextValue: string): Promise<IAzureNode>;
+    public dispose(): void;
+}
+
+/**
+ * The AzureExplorer returns instances of IAzureNode, which are wrappers for IAzureTreeItem with relevant context and functions from the explorer
+ */
+export interface IAzureNode<T extends IAzureTreeItem = IAzureTreeItem> {
```
StephenWeatherford: Can't find any documentation on the "extends IAzureTreeItem = IAzureTreeItem" notation - what does it mean?

EricJizbaMSFT: It basically just makes the generic type optional. I got tired of always writing `IAzureNode<ExampleTreeItem>`, especially because I don't always need to interact with the tree item.

By doing `= IAzureTreeItem` you can just write `IAzureNode` and it means the same thing as `IAzureNode<IAzureTreeItem>`

ui/index.d.ts
```diff
@@ -0,0 +1,108 @@
...
+
+    /**
+     * Refresh this node in the explorer
+     */
+    refresh(): void;
+
+    /**
+     * This class wraps IAzureTreeItem.deleteTreeItem and ensures the explorer is updated correctly when an item is deleted
+     */
+    deleteNode(): Promise<IAzureNode>;
```
StephenWeatherford: Why the return value?

ui/src/IUserInterface.ts
```diff
@@ -0,0 +1,32 @@
...
+export class Pick implements QuickPickItem {
+    public readonly description: string;
+    public readonly label: string;
+    constructor(label: string, description?: string) {
+        this.label = label;
+        this.description = description ? description : '';
+    }
+}
+
+export class PickWithData<T> extends Pick {
```
StephenWeatherford: How about get rid of PickWithData<T> and use something like:

class Pick<T = void> ...

?

StephenWeatherford: Or can you do  

class Pick<T> implement Pick

like you can in C#?

ui/src/errors.ts
```diff
@@ -0,0 +1,20 @@
...
+export class UserCancelledError extends Error { }
+
+export class NotImplementedError extends Error {
+    constructor(methodName: string, obj: object) {
+        super(localize('azFunc.notImplementedError', '"{0}" is not implemented on "{1}".', methodName, obj.constructor.name));
+    }
+}
+
+export class ArgumentError extends Error {
+    constructor(obj: object) {
```
StephenWeatherford: What if obj is null/undefined?

ui/src/errors.ts
```diff
@@ -0,0 +1,20 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { localize } from "./localize";
+
+export class UserCancelledError extends Error { }
```
StephenWeatherford: Should this have a default message in case it doesn't get handled?

ui/src/errors.ts
EricJizbaMSFT: This repo has 'strictNullChecks' on and this is an 'internal' class (aka it's not in index.d.ts at the root of the repo), so that should never happen

ui/src/IUserInterface.ts
EricJizbaMSFT: Filed an issue to address later:
https://github.com/Microsoft/vscode-azuretools/issues/23

PR #19, opened 2017-11-17
Move log stream settings to shared package

whiskeyjay:
So the JS diagnostics feature can reuse them.


appservice/src/SiteWrapper.ts
```diff
@@ -13,12 +13,14 @@ import * as git from 'simple-git/promise';
 import * as vscode from 'vscode';
 import KuduClient from 'vscode-azurekudu';
 import { DeployResult } from 'vscode-azurekudu/lib/models';
+import * as WebSiteModels from '../node_modules/azure-arm-website/lib/models';
```
EricJizbaMSFT: This is duplicate of line 8. I actually prefer importing the class directly instead of using `WebSiteModels`

VS Code will give you autocomplete for class/interface names from your node_modules even if they're not imported yet. Then if you hit 'tab' to autocomplete, it will automatically add that class to your import in the style of line 8

PR #21, opened 2017-11-20
Keep explorer in 'Loading...' state until subscriptions are ready

EricJizbaMSFT:
Before it was switching between 'Loading...' and 'No Subscriptions Found. Edit Filters...' even though I know I had filters selected

NOTE: We can't _completely_ ignore the onStatusChanged event because the onFiltersChanged event doesn't always work (for example when switching from 'LoggingIn' to 'LoggedOut' when a cached token has expired)

### DIFF
--- a/ui/src/treeDataProvider/AzureTreeDataProvider.ts
+++ b/ui/src/treeDataProvider/AzureTreeDataProvider.ts
@@ -6,7 +6,7 @@
 import * as path from 'path';
 import { Disposable, Event, EventEmitter, Extension, extensions, TreeDataProvider, TreeItem, TreeItemCollapsibleState } from 'vscode';
 import { IAzureNode, IChildProvider } from '../../index';
-import { AzureAccount, AzureSubscription } from '../azure-account.api';
+import { AzureAccount, AzureLoginStatus, AzureSubscription } from '../azure-account.api';
 import { ArgumentError } from '../errors';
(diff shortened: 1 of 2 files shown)


ui/src/treeDataProvider/AzureTreeDataProvider.ts
```diff
@@ -43,7 +43,13 @@ export class AzureTreeDataProvider implements TreeDataProvider<IAzureNode>, Disp
         }
 
         this._disposables.push(this._azureAccount.onFiltersChanged(() => this.refresh()));
```
JakeRadMSFT: This always fires on logged in?

EricJizbaMSFT: Sure looks like it to me. You can follow the trail of events in [this file](https://github.com/Microsoft/vscode-azure-account/blob/master/src/azure-account.ts):

1. `login` calls `updateSessions`
1. `updateSessions` fires `onSessionsChanged`, which triggers `updateSubscriptions`
1. `updateSubscriptions` fires `onSubscriptionsChanged`, which triggers `updateFilters`
1. `updateFilters` triggers `onFiltersChanged`

I tested opening VS Code with '0', '1', 'many', and 'all' filters selected. Also tested logging in/out in the same VS Code session. It was always fired

PR #24, opened 2017-11-21
 Clean up ui package based on feedback

EricJizbaMSFT:
Fixing feedback from @StephenWeatherford 

Also removed 'azFunc' from the localization ids (a prefix isn't necessary anyways, the translations are always stored inside the extension or package itself)

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -1,7 +1,7 @@
 {
     "name": "vscode-azureextensionui",
     "author": "Microsoft Corporation",
-    "version": "0.1.1",
+    "version": "0.2.0",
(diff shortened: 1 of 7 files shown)


ui/package.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "vscode-azureextensionui",
     "author": "Microsoft Corporation",
-    "version": "0.1.1",
+    "version": "0.2.0",
```
EricJizbaMSFT: Technically this is a breaking change to the 'contract', but since this package is in preview I'm just changing the minor version instead of bumping the major to 1.0

PR #27, opened 2017-11-23
Nturinski/bug22

### LINKED ISSUE #22: Use 'cancel' instead of 'close' on messages

nturinski:
Fixes #22


nturinski:
I was keeping the format consistent with azFunc localization tags.  No problem to get rid of them.

EricJizbaMSFT:
Yeah if we make a change, I would prefer to change the functions extension to remove 'azFunc'

PR #30, opened 2017-11-27
Allow tree providers to add root items

EricJizbaMSFT:
And fix a bug where 'clearCache' was never being passed

This is necessary for the 'Attached Mongo Server' node in the cosmosdb extension

### DIFF
--- a/ui/src/treeDataProvider/AzureTreeDataProvider.ts
+++ b/ui/src/treeDataProvider/AzureTreeDataProvider.ts
@@ -5,7 +5,7 @@
 
 import * as path from 'path';
 import { Disposable, Event, EventEmitter, Extension, extensions, TreeDataProvider, TreeItem, TreeItemCollapsibleState } from 'vscode';
-import { IAzureNode, IChildProvider } from '../../index';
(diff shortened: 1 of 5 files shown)


ui/src/treeDataProvider/AzureTreeDataProvider.ts
```diff
@@ -111,8 +115,10 @@ export class AzureTreeDataProvider implements TreeDataProvider<IAzureNode>, Disp
                         return new SubscriptionNode(this, this._resourceProvider, subscriptionInfo.subscription.subscriptionId, subscriptionInfo.subscription.displayName, subscriptionInfo);
                     }
                 });
-                return this._subscriptionNodes;
+                nodes = this._subscriptionNodes;
             }
+
+            return nodes.concat(this._rootNodes);
```
StephenWeatherford: should root nodes come first or last in the list?

EricJizbaMSFT: I chose last to be consistent with current behavior (The 'Attached Mongo Servers' node comes after subscriptions in cosmos db).

I think subscriptions take priority for an 'Azure Tree Provider'

PR #32, opened 2017-11-28
Show 'creating' node for new app services

EricJizbaMSFT:


appservice/src/wizard/WizardBase.ts
```diff
@@ -19,18 +19,18 @@ export abstract class WizardBase {
         this.persistence = persistence;
     }
 
-    public async run(): Promise<void> {
-        // Go through the prompts...
+    public async prompt(): Promise<void> {
```
nturinski: Will these changes affect the extensions?  azureappservice still has a wizard.ts file with a run function defined, for instance.

EricJizbaMSFT: Nope - we haven't exposed 'wizard' yet (it's not exported in `index.ts`) so the extensions aren't using this code. Ideally it should be exported in the 'ui' package, but nobody has gotten around to that work

PR #33, opened 2017-12-01
Implement auto-detect deploy

nturinski:
* Changed the deployment functions to be private as it should only be used by the siteWrapper as it should deploy according to scmType

### DIFF
--- a/appservice/src/SiteWrapper.ts
+++ b/appservice/src/SiteWrapper.ts
@@ -96,8 +96,8 @@ export class SiteWrapper {
 
     public async updateConfiguration(client: WebSiteManagementClient, config: SiteConfigResource): Promise<SiteConfigResource> {
         return this.slotName ?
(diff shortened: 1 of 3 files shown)


appservice/src/SiteWrapper.ts
```diff
@@ -217,50 +272,9 @@ export class SiteWrapper {
...
-    }
-
-    public async editScmType(client: WebSiteManagementClient): Promise<string | undefined> {
-        const config: SiteConfigResource = await this.getSiteConfig(client);
-        const newScmType: string = await this.showScmPrompt(config.scmType);
-        // returns the updated scmType
-        return await this.updateScmType(client, config, newScmType);
+        this.log(outputChannel, (localize('localGitDeploy', `Deploying Local Git repository to "${this.appName}"...`)));
+        await this.waitForDeploymentToComplete(kuduClient, outputChannel);
+        this.log(outputChannel, 'Deployment completed.');
```
EricJizbaMSFT: nit: looks like we missed a few more strings that should be in 'localize'

PR #42, opened 2017-12-11
Minor bug fixes related to deploy

EricJizbaMSFT:
Use `value` instead of `newAccountName` for storage account validation

Show outputchannel before local git deploy


appservice/src/createAppService/StorageAccountStep.ts
```diff
@@ -58,16 +58,16 @@ export class StorageAccountStep extends WizardStep {
...
+                if (!nameAvailabilityResult.nameAvailable) {
+                    return nameAvailabilityResult.message;
                 }
-            });
+                return undefined;
+            }
+        });
 
         this._account = {
             name: newAccountName.trim(),
```
StephenWeatherford: Thought you said you were using value here?

PR #43, opened 2017-12-11
Nturinski/baseEditor

nturinski:


ui/package.json
```diff
@@ -31,9 +31,11 @@
     },
     "dependencies": {
         "azure-arm-resource": "^3.0.0-preview",
+        "fs-extra": "^4.0.3",
         "ms-rest": "^2.2.2",
         "ms-rest-azure": "^2.4.4",
         "opn": "^5.1.0",
+        "path": "^0.12.7",
```
EricJizbaMSFT: Is this different than the 'path' that's a part of node?

ui/package.json
```diff
@@ -31,9 +31,11 @@
     },
     "dependencies": {
         "azure-arm-resource": "^3.0.0-preview",
+        "fs-extra": "^4.0.3",
```
EricJizbaMSFT: Add the license to thirdpartynotices.txt

ui/src/BaseEditor.ts
```diff
@@ -0,0 +1,166 @@
...
+    public abstract getSaveConfirmationText(context: ContextT): Promise<string>;
+
+    public async showEditor(context: ContextT): Promise<void> {
+        const size: number = await this.getSize(context);
+        const fileName: string = await this.getFilename(context);
+        const splitFileName: string[] = fileName.split('.');
+        const extension: string = splitFileName[splitFileName.length - 1];
+
+        this.appendToOutput(localize('opening', 'Opening "{0}" ...', fileName));
+        if (size > 50 /*Megabytes*/) {
```
EricJizbaMSFT: Why do we set a hard-coded limit of 50? Shouldn't Azure return a readable error if it's too large? Maybe different pricing tiers have different limits anyways

You should at least let the baseEditor overwrite the limit (I'm guessing cosmos has different limits than storage)

ui/src/BaseEditor.ts
```diff
@@ -0,0 +1,166 @@
...
+            const message: string = localize('tooLargeError', '"{0}" is too large to download.', fileName);
+
+            await vscode.window.showWarningMessage(message, dialogResponses.OK);
```
EricJizbaMSFT: I would throw an error rather than show a message for two reasons:
1. If you swallow the error like this, it's never tracked in telemetry
1. Our extensions already have logic to display errors to the user. We should just leverage that to be consistent

ui/src/BaseEditor.ts
```diff
@@ -0,0 +1,166 @@
...
+                // tslint:disable-next-line:no-unsafe-any
+                const localFilePath: string = await TemporaryFile.create(fileName);
+                const document: vscode.TextDocument = await vscode.workspace.openTextDocument(localFilePath);
+                this.fileMap[localFilePath] = [document, context];
+                const data: string = await this.getData(context);
+                const textEditor: vscode.TextEditor = await vscode.window.showTextDocument(document);
+                await this.updateEditor(data, textEditor);
+                this.appendLineToOutput(' Done.');
+            } catch (error) {
+                let details: string;
```
EricJizbaMSFT: Again, I would get rid of all the error handling in this file and let each extension handle it

ui/src/BaseEditor.ts
```diff
@@ -0,0 +1,166 @@
...
+                this.appendLineToOutput(localize('failed', " Failed."));
+                this.appendLineToOutput(localize('errorDetails', 'Error Details: {0}', details));
+
+                await vscode.window.showWarningMessage(localize('downloadError', 'Unable to download "{0}". Please check Output for more information.', fileName), dialogResponses.OK);
+            }
+
+        }
+    }
+
+    public async updateMatchingcontext(doc: vscode.Uri): Promise<void> {
```
EricJizbaMSFT: Fix the camelCase for context

ui/src/BaseEditor.ts
```diff
@@ -0,0 +1,166 @@
...
+    OK: localize('OK', "OK"),
+    DontShowAgain: localize('dontShow', "Don't Show Again")
+};
+
+// tslint:disable-next-line:no-unsafe-any
+export abstract class BaseEditor<ContextT> implements vscode.Disposable {
+    private fileMap: { [key: string]: [vscode.TextDocument, ContextT] } = {};
+    private ignoreSave: boolean = false;
+
+    constructor(readonly showSavePromptKey: string, readonly outputChanel?: vscode.OutputChannel) {
```
EricJizbaMSFT: Just my two cents, but I don't like parameter properties because it's easier for people to use incorrect signatures. For example, both of these should be private, correct?

I don't think people would've made that mistake if they were forced to write properties separately.

EricJizbaMSFT: outputChanel is spelled wrong

ui/src/BaseEditor.ts
```diff
@@ -0,0 +1,169 @@
...
+        if (size > 50 /*Megabytes*/) {
+            const message: string = localize('tooLargeError', '"{0}" is too large to download.', fileName);
+
+            await vscode.window.showWarningMessage(message, DialogResponses.OK);
+            this.appendLineToOutput(localize('failed', "Failed."));
+            this.appendLineToOutput(localize('errorDetails', 'Error Details: {0}', message));
+        } else if (extension === 'exe' || extension === 'img' || extension === 'zip') {
+            const message: string = localize('unsupportedError', '"{0}" has an unsupported file extension.', fileName);
+
+            await vscode.window.showWarningMessage(message, DialogResponses.OK);
```
EricJizbaMSFT: I would throw an error rather than show a message for two reasons:
1. If you swallow the error like this, it's never tracked in telemetry
1. Our extensions already have logic to display errors to the user. We should just leverage that to be consistent

ui/src/BaseEditor.ts
```diff
@@ -0,0 +1,169 @@
...
+        const extension: string = splitFileName[splitFileName.length - 1];
+
+        this.appendToOutput(localize('opening', 'Opening "{0}" ...', fileName));
+        if (size > 50 /*Megabytes*/) {
+            const message: string = localize('tooLargeError', '"{0}" is too large to download.', fileName);
+
+            await vscode.window.showWarningMessage(message, DialogResponses.OK);
+            this.appendLineToOutput(localize('failed', "Failed."));
+            this.appendLineToOutput(localize('errorDetails', 'Error Details: {0}', message));
+        } else if (extension === 'exe' || extension === 'img' || extension === 'zip') {
```
EricJizbaMSFT: Again - why are we hard-coding file extensions here? Wouldn't it be better to let Azure determine what can or cannot be downloaded?

(Again - cosmos will be different - it's always JSON files)

ui/src/BaseEditor.ts
nturinski: This is regarding the tslint rules, right?  

I remember this being discussed with the App Services Extension.  I thought we decided to disable this rule because people didn't like being forced to define the property and then assigning it in the constructor.

ui/src/BaseEditor.ts
nturinski: Well, I think this is more of a vscode.Editor limitation rather than an App Service Plan tier limitation.  From what I understand, @JakeRadMSFT tested the size limit of what the editor could open without crashing.

ui/src/BaseEditor.ts
nturinski: I believe this is also a text editor limitation.  It wouldn't make sense to allow any extension to attempt to open an exe, img, or zip file in the editor.

ui/src/BaseEditor.ts
nturinski: Good point. Should we create custom errors for this or would is this okay for a generic type?

ui/src/BaseEditor.ts
nturinski: @JakeRadMSFT I didn't actually see this function getting called anywhere.  Do we need to include this in the BaseEditor?

ui/src/BaseEditor.ts
EricJizbaMSFT: VS Code already has logic for files it cannot display:
screen shot 2017-12-11 at 5 02 01 pm

I just don't think we should have our own custom logic for this. At the end of the day it's not _our_ limitation - it's either Azure's or VS Code's.

ui/src/BaseEditor.ts
EricJizbaMSFT: A similar function is used in Cosmos DB if you have the 'Update to Azure' right click menu or button in the file.

However, I don't think Jake use's it in Storage because saving the file is his only entry point

ui/src/BaseEditor.ts
EricJizbaMSFT: Generic is fine IMO

ui/src/BaseEditor.ts
nturinski: I've attached what the logic is for a file that is too large.  I'm okay with removing these lines; I'm not sure that this is a great error message though.

image

nturinski: Okay-- talked to Jake.  50 MB is a limit due reduce wait time.  I'll allow an optional parameter to check for this.

ui/src/BaseEditor.ts
```diff
@@ -0,0 +1,146 @@
...
+
+    public async showEditor(context: ContextT, sizeLimit?: number /* in Megabytes */): Promise<void> {
+        const fileName: string = await this.getFilename(context);
+
+        this.appendToOutput(localize('opening', 'Opening "{0}" ...', fileName));
+        if (sizeLimit) {
+            const size: number = await this.getSize(context);
+            if (size > sizeLimit) {
+                const message: string = localize('tooLargeError', '"{0}" is too large to download.', fileName);
+                this.appendLineToOutput(localize('failed', " Failed."));
```
EricJizbaMSFT: Most of our extensions already print the error message to the output window - so I don't think we need to print 'Failed'.

If you really want to print 'Failed' - at least specify _which_ upload failed (`Failed to upload document ${fileName}`)

ui/src/BaseEditor.ts
```diff
@@ -0,0 +1,146 @@
...
+    public abstract getSize(context: ContextT): Promise<number>;
+    public abstract getSaveConfirmationText(context: ContextT): Promise<string>;
+
+    public async showEditor(context: ContextT, sizeLimit?: number /* in Megabytes */): Promise<void> {
+        const fileName: string = await this.getFilename(context);
+
+        this.appendToOutput(localize('opening', 'Opening "{0}" ...', fileName));
+        if (sizeLimit) {
+            const size: number = await this.getSize(context);
+            if (size > sizeLimit) {
```
EricJizbaMSFT: I don't think you have the if/else cases correct. For example - if this line is false, it should fall through to the else case, but I don't think it is

ui/src/BaseEditor.ts
```diff
@@ -0,0 +1,146 @@
...
+        }  else {
+            try {
+                const localFilePath: string = await TemporaryFile.create(fileName);
+                const document: vscode.TextDocument = await vscode.workspace.openTextDocument(localFilePath);
+                this.fileMap[localFilePath] = [document, context];
+                const data: string = await this.getData(context);
+                const textEditor: vscode.TextEditor = await vscode.window.showTextDocument(document);
+                await this.updateEditor(data, textEditor);
+                this.appendLineToOutput(' Done.');
+            } catch (error) {
```
EricJizbaMSFT: Just remove this entire 'catch' block and let the error propagate. Our extensions have more robust error parsing/handling already

ui/src/BaseEditor.ts
```diff
@@ -0,0 +1,146 @@
...
+        // tslint:disable-next-line:no-unsafe-any
+        Object.keys(this.fileMap).forEach(async (key: string) => await fse.remove(path.dirname(key)));
+    }
+
+    public async onDidSaveTextDocument(globalState: vscode.Memento, doc: vscode.TextDocument): Promise<void> {
+        const filePath: string | undefined = Object.keys(this.fileMap).find((fsPath: string) => path.relative(doc.uri.fsPath, fsPath) === '');
+        if (!this.ignoreSave && filePath) {
+            const context: ContextT = this.fileMap[filePath][1];
+            const showSaveWarning: boolean | undefined = vscode.workspace.getConfiguration().get(this.showSavePromptKey);
+            let shouldUpdateRemote: boolean = false;
```
EricJizbaMSFT: Remove the `shouldUpdateRemote` variable. (Jake has removed it in his code, too. He only had it initially because he wasn't throwing a `UserCancelledError`. That's the only place where `shouldUpdateRemote` is not set to true and it doesn't matter because the error is thrown)

ui/src/BaseEditor.ts
```diff
@@ -0,0 +1,146 @@
...
+            this.outputChannel.appendLine(value);
+            this.outputChannel.show(true);
+        }
+    }
+
+    private async updateRemote(context: ContextT, doc: vscode.TextDocument): Promise<void> {
+        const filename: string = await this.getFilename(context);
+        this.appendToOutput(localize('updating', 'Updating "{0}" ...', filename));
+        const updatedData: string = await this.updateData(context, doc.getText());
+        this.appendLineToOutput(localize('done', ' Done.'));
```
EricJizbaMSFT: Here's another case - don't write 'done' without saying which 'file' is done. Otherwise it gets confusing if I'm updating multiple files at a time

ui/src/BaseEditor.ts
```diff
@@ -0,0 +1,146 @@
...
+        await this.updateEditor(updatedData, vscode.window.activeTextEditor);
+    }
+
+    private async updateEditor(data: string, textEditor?: vscode.TextEditor): Promise<void> {
+        if (!!textEditor) {
+            await BaseEditor.writeToEditor(textEditor, data);
+            this.ignoreSave = true;
```
EricJizbaMSFT: Whenever you have "global" flags representing state like this, it's good practice to wrap it in a try/finally:
```
this.ignoreSave = true;
try {
    await textEditor.document.save();
} finally {
    this.ignoreSave = false;
}
```

That way ignoreSave is updated even if the save operation throws an exception

ui/src/utils/TemporaryFile.ts
```diff
@@ -0,0 +1,24 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as crypto from "crypto";
+import * as fse from 'fs-extra';
+import * as os from 'os';
+import * as path from 'path';
+
+// tslint:disable-next-line:no-stateless-class export-name
```
EricJizbaMSFT: Rather than disable the rule here, you could just make it a static class or a namespace

ui/src/utils/TemporaryFile.ts
```diff
@@ -0,0 +1,24 @@
...
+export class TemporaryFile {
+    private static randomFolderNameLength: number = 12;
+
+    // tslint:disable-next-line:function-name
+    public static async create(fileName: string): Promise<string> {
+        const buffer: Buffer = crypto.randomBytes(Math.ceil(TemporaryFile.randomFolderNameLength / 2));
+        const folderName: string = buffer.toString('hex').slice(0, TemporaryFile.randomFolderNameLength);
+        const filePath: string = path.join(os.tmpdir(), folderName, fileName);
+        // tslint:disable-next-line:no-unsafe-any
```
EricJizbaMSFT: Add `"@types/fs-extra"` to the devDependencies in package.json and you won't have to disable this tslint rule

ui/thirdpartynotices.txt
```diff
@@ -34,3 +35,24 @@ THE SOFTWARE.
...
+ furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED 'AS IS', WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE
+WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS
+OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
+ ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+END OF fs-extra NOTICES AND INFORMATION
```
EricJizbaMSFT: You're missing a row of equal signs

ui/package.json
EricJizbaMSFT: I still don't think you need this?

nturinski:
I think I did this right...

ui/src/BaseEditor.ts
```diff
@@ -0,0 +1,123 @@
...
+                    await globalState.update(this.showSavePromptKey, true);
+                } else if (result === DialogResponses.dontUpload) {
+                    throw new UserCancelledError();
+                }
+            }
+            await this.updateRemote(context, doc);
+        }
+    }
+
+    protected appendToOutput(value: string): void {
```
EricJizbaMSFT: I would remove this function. As discussed, we should always use appendLine because you never know what the state of the outputChannel is

EricJizbaMSFT:
Also - it wouldn't hurt to add documentation to the README - but you can do that later if you want

PR #44, opened 2017-12-11
Make Zip configurable

whiskeyjay:
Making Zip behavior configurable.

### DIFF
--- a/appservice/src/SiteWrapper.ts
+++ b/appservice/src/SiteWrapper.ts
@@ -137,14 +137,14 @@ export class SiteWrapper {
         outputChannel.appendLine(localize('DeleteSucceeded', 'Successfully deleted "{0}".', this.appName));
     }
 
-    public async deploy(fsPath: string, client: WebSiteManagementClient, outputChannel: vscode.OutputChannel, confirmDeployment: boolean = true): Promise<void> {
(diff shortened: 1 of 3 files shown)


appservice/src/SiteWrapper.ts
```diff
@@ -210,7 +210,13 @@ export class SiteWrapper {
         } else if (await FileUtilities.isDirectory(fsPath)) {
             createdZip = true;
             this.log(outputChannel, localize('zipCreate', 'Creating zip package...'));
-            zipFilePath = await FileUtilities.zipDirectory(fsPath);
+            const zipDeployConfig: vscode.WorkspaceConfiguration = vscode.workspace.getConfiguration('zipDeploy');
```
EricJizbaMSFT: By hard-coding the setting names in this npm package, you're essentially forcing each extension to add these settings to their package.json. What if instead the `deploy` method takes in an optional "zipOptions" parameter that has the globPattern and ignorePattern

whiskeyjay: If an extension doesn't define the settings, the patterns will just be "undefined"; when passed to the actual zip method, the glob pattern will be defaulted to '**/*' and ignore pattern remains undefined which is also default.

EricJizbaMSFT:
If you bump the version in package.json - I can get this package published for you right after you merge this PR

PR #50, opened 2017-12-14
Make completion messages more prominent

EricJizbaMSFT:
Add arrows and a blank line

screen shot 2017-12-14 at 2 34 52 pm


EricJizbaMSFT:
We've gotten feedback that the completion message was not obvious enough (for example https://github.com/Microsoft/vscode-azurefunctions/issues/111)

I tend to agree - hence the arrows

PR #52, opened 2017-12-18
Allow async refresh of label

EricJizbaMSFT:
We sometimes display state in the label and this would let us update that state asynchronously

### DIFF
--- a/ui/src/treeDataProvider/AzureTreeDataProvider.ts
+++ b/ui/src/treeDataProvider/AzureTreeDataProvider.ts
@@ -45,12 +45,12 @@ export class AzureTreeDataProvider implements TreeDataProvider<IAzureNode>, Disp
             this._azureAccount = azureAccountExtension.exports;
         }
 
-        this._disposables.push(this._azureAccount.onFiltersChanged(() => this.refresh()));
-        this._disposables.push(this._azureAccount.onStatusChanged((status: AzureLoginStatus) => {
+        this._disposables.push(this._azureAccount.onFiltersChanged(async () => await this.refresh()));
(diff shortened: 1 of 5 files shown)


ui/src/treeDataProvider/AzureTreeDataProvider.ts
```diff
@@ -65,7 +65,11 @@ export class AzureTreeDataProvider implements TreeDataProvider<IAzureNode>, Disp
         return this._onDidChangeTreeDataEmitter.event;
     }
 
-    public getTreeItem(node: IAzureNode): TreeItem {
+    public async getTreeItem(node: AzureNode): Promise<TreeItem> {
+        if (node.treeItem.refreshLabel) {
```
whiskeyjay: Since getTreeItem is the place where value of label will be read; does this imply that if refreshLabel is defined, one must call it before accessing the label property?

If that's true, can the two steps be merged into one getLabel method?

EricJizbaMSFT: refreshLabel is _not_ required before accessing the label property. For example, we access the label to sort the nodes before 'refreshLabel' is ever called.

I wanted to keep the 'label' property fast/synchronous so that you don't have to communicate with Azure every time you access it (hence `refreshLabel` being a separate method)

whiskeyjay: If the label is used by sorting before it is displayed, is it necessary to call refreshLabel here? If the label is expected to change, then sorting shouldn't use the old value; if not, then why refresh? Or is it up to each treeItem to decide whether to use a cached label value or do a trip to Azure?

PR #53, opened 2018-01-04
Add AzureActionHandler to UI package

### LINKED ISSUE #51: "Operation Canceled" error is shown on UI w/ app settings
Try Add or Edit app settings from either App Service extension or Functions extension, and cancel from the prompt.

Expected: Nothing happens

Actual:
image

EricJizbaMSFT:
The AzureActionHandler provides consistent error handling and telemetry across extensions. Most of this is not _new_ code, just code I pulled out from our extensions. However I did fix these issues:
* Fixes #51 ('Operation Cancelled' is sometimes displayed)
* Fixes telemetry tracking for events (we were tracking it way too much)

NOTE: I had to refactor the folder structure of build output a little bit so that I could get unit tests working

### DIFF
--- a/ui/README.md
+++ b/ui/README.md
@@ -1,8 +1,44 @@
 # VSCode Azure SDK for Node.js - UI Tools (Preview)
 
-This package provides common Azure UI elements for VS Code extensions.
+This package provides common Azure UI elements for VS Code extensions:
+* [AzureActionHandler](#azure-action-handler): Displays error messages and optionally adds telemetry to commands/events.
+* [AzureTreeDataProvider](#azure-tree-data-provider): Displays an Azure Explorer with Azure Subscriptions and child nodes of your implementation.
+* [AzureBaseEditor](#azure-base-editor): Displays a text editor with upload support to Azure.
 
-> NOTE: This package throws a `UserCancelledError` if the user cancels an operation. This error should be handled appropriately by your extension.
+> NOTE: This package throws a `UserCancelledError` if the user cancels an operation. If you do not use the AzureActionHandler, you must handle this exception in your extension.
+
+## Azure Action Handler
+
+Use the Azure Action Handler to consistently display error messages and track commands with telemetry. You should contruct the handler and register commands/events in your extension's `activate()` method. The simplest example is to register a command (in this case, refreshing a node):
+```typescript
+const actionHandler: AzureActionHandler = new AzureActionHandler(context, outputChannel, reporter);
+actionHandler.registerCommand('yourExtension.Refresh', (node: IAzureNode) => { node.refresh(); });
+```
+Here are a few of the benefits this provides:
+* Parses Azure errors of the form `{ "Code": "Conflict", "Message": "This is the actual message" }` and only displays the 'Message' property
+* Displays single line errors normally and multi-line errors in the output window
+* If you pass a TelemetryReporter, tracks multiple properties in addition to the [common extension properties](https://github.com/Microsoft/vscode-extension-telemetry#common-properties):
+  * result (Succeeded, Failed, or Canceled)
+  * duration
+  * error
+
+If you want to add custom telemetry proprties, use `registerCommandWithCustomTelemetry` and add your own properties or measurements:
+```typescript
+actionHandler.registerCommandWithCustomTelemetry('yourExtension.Refresh', async (properties: TelemetryProperties, measurements: TelemetryMeasurements) => {
+    properties.customProp = "example prop";
+    measurements.customMeas = 49;
+});
+```
+
+Finally, you can also register events. The main difference is that your callback must take in the `trackTelemetry` parameter. Events are not tracked by default (since they can happen very frequently). You must call `trackTelemetry()` to signal that this event is indeed handled by your extension. For example, if your extension only handles `json` files in the `onDidSaveTextDocument`, it might look like this:
+```typescript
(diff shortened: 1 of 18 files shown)


ui/src/AzureActionHandler.ts
```diff
@@ -0,0 +1,86 @@
...
+            properties.result = 'Succeeded';
+            let sendTelemetry: boolean = false;
+
+            try {
+                await Promise.resolve(callback(() => { sendTelemetry = true; }, properties, measurements, ...args));
+            } catch (error) {
+                errorData = parseError(error);
+                // NOTE: Intentionally not using 'error instanceof UserCancelledError' because that doesn't work if multiple versions of the UI package are used in one extension
+                // See https://github.com/Microsoft/vscode-azuretools/issues/51 for more info
+                if (errorData.errorType === 'UserCancelledError') {
```
StephenWeatherford: How about adding errorData.isUserCancelledError, which will encourage the correct usage (and doesn't require comments everywhere in the calling code)?

ui/README.md
```diff
@@ -1,8 +1,44 @@
...
+
+If you want to add custom telemetry proprties, use `registerCommandWithCustomTelemetry` and add your own properties or measurements:
+```typescript
+actionHandler.registerCommandWithCustomTelemetry('yourExtension.Refresh', async (properties: TelemetryProperties, measurements: TelemetryMeasurements) => {
+    properties.customProp = "example prop";
+    measurements.customMeas = 49;
+});
+```
+
+Finally, you can also register events. The main difference is that your callback must take in the `trackTelemetry` parameter. Events are not tracked by default (since they can happen very frequently). You must call `trackTelemetry()` to signal that this event is indeed handled by your extension. For example, if your extension only handles `json` files in the `onDidSaveTextDocument`, it might look like this:
```
StephenWeatherford: trackTelemetry sounds like a Boolean.  How about sendEventTelemetry or sendTelemetry?

EricJizbaMSFT: I'm worried 'sendTelemetry' implies it's actually sending the telemetry at that point in the code, so users will call it _after_ their own custom logic like this:
```
if (doc.fileExtension === 'json') {
    throw new Error('Oops I messed up');
    sendTelemetry();
}
```
In the above example, that means we don't catch their error in our telemetry because 'sendTelemetry' was never reached. If they did this, we would've tracked the error:
```
if (doc.fileExtension === 'json') {
    sendTelemetry();
    throw new Error('Oops I messed up');
}
```

I was kind of going for a boolean feeling with 'trackTelemetry' since it's basically flipping a switch saying "track telemetry for this event", but it's not actually _sending_ the telemetry until the command finishes

ui/README.md
```diff
@@ -1,8 +1,44 @@
...
+* [AzureActionHandler](#azure-action-handler): Displays error messages and optionally adds telemetry to commands/events.
+* [AzureTreeDataProvider](#azure-tree-data-provider): Displays an Azure Explorer with Azure Subscriptions and child nodes of your implementation.
+* [AzureBaseEditor](#azure-base-editor): Displays a text editor with upload support to Azure.
 
-> NOTE: This package throws a `UserCancelledError` if the user cancels an operation. This error should be handled appropriately by your extension.
+> NOTE: This package throws a `UserCancelledError` if the user cancels an operation. If you do not use the AzureActionHandler, you must handle this exception in your extension.
+
+## Azure Action Handler
+
+Use the Azure Action Handler to consistently display error messages and track commands with telemetry. You should contruct the handler and register commands/events in your extension's `activate()` method. The simplest example is to register a command (in this case, refreshing a node):
```
PrashanthCorp: You should _construct_ the handler

ui/src/AzureActionHandler.ts
```diff
@@ -0,0 +1,86 @@
...
+            const properties: TelemetryProperties = {};
+            const measurements: TelemetryMeasurements = {};
+            properties.result = 'Succeeded';
+            let sendTelemetry: boolean = false;
+
+            try {
+                await Promise.resolve(callback(() => { sendTelemetry = true; }, properties, measurements, ...args));
+            } catch (error) {
+                errorData = parseError(error);
+                // NOTE: Intentionally not using 'error instanceof UserCancelledError' because that doesn't work if multiple versions of the UI package are used in one extension
```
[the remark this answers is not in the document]
EricJizbaMSFT: You're right it's not the best practice. However, it's also unrealistic to expect that the versions will always match. For example, our vscode-azureappservice npm package depends on the ui npm package and we don't want to have to update that dependency version every single time.

ui/src/treeDataProvider/CreatingTreeItem.ts
```diff
@@ -26,8 +26,8 @@ export class CreatingTreeItem implements IAzureTreeItem {
 
     public get iconPath(): { light: string, dark: string } {
         return {
-            light: path.join(__filename, '..', '..', '..', 'resources', 'light', 'Loading.svg'),
-            dark: path.join(__filename, '..', '..', '..', 'resources', 'dark', 'Loading.svg')
+            light: path.join(__filename, '..', '..', '..', '..', 'resources', 'light', 'Loading.svg'),
```
PrashanthCorp: I don't this file or any of the image files moving. Why is there an extra `..`  in the path?

EricJizbaMSFT: I had to refactor the folder structure of build output a little bit so that I could get unit tests working. See the changes in tsconfig.json

PR #57, opened 2018-01-12
Improve node picker

### LINKED ISSUE #47: showNodePicker should allow for multiple types of nodes in the same quickpick
For example, Cosmos wants to be able to show MongoNodes and DocDb nodes for the OpenInPortal action

EricJizbaMSFT:
1. Allow for multiple context values to apply
1. Allow node picker to start at a specific node
1. Allow tree items to implement isAncestorOf

Fixes #47

### DIFF
--- a/ui/src/treeDataProvider/AzureParentNode.ts
+++ b/ui/src/treeDataProvider/AzureParentNode.ts
@@ -69,19 +69,21 @@ export class AzureParentNode<T extends IAzureParentTreeItem = IAzureParentTreeIt
             .sort((n1: AzureNode, n2: AzureNode) => n1.treeItem.label.localeCompare(n2.treeItem.label));
     }
 
-    public async pickChildNode(expectedContextValue: string, ui: IUserInterface): Promise<AzureNode> {
+    public async pickChildNode(expectedContextValues: string[], ui: IUserInterface): Promise<AzureNode> {
         if (this.treeItem.pickTreeItem) {
             const children: AzureNode[] = await this.getCachedChildren();
-            const treeItem: IAzureTreeItem | undefined = this.treeItem.pickTreeItem(expectedContextValue);
-            if (treeItem) {
-                const node: AzureNode | undefined = children.find((n: AzureNode) => n.treeItem.id === treeItem.id);
-                if (node) {
-                    return node;
+            for (const val of expectedContextValues) {
+                const treeItem: IAzureTreeItem | undefined = this.treeItem.pickTreeItem(val);
+                if (treeItem) {
+                    const node: AzureNode | undefined = children.find((n: AzureNode) => n.treeItem.id === treeItem.id);
+                    if (node) {
+                        return node;
+                    }
                 }
             }
         }
 
-        const pick: PickWithData<GetNodeFunction> = await ui.showQuickPick<GetNodeFunction>(this.getQuickPicks(), localize('azFunc.selectNode', 'Select a {0}', this.treeItem.childTypeLabel));
+        const pick: PickWithData<GetNodeFunction> = await ui.showQuickPick<GetNodeFunction>(this.getQuickPicks(expectedContextValues), localize('azFunc.selectNode', 'Select a {0}', this.treeItem.childTypeLabel));
         return await pick.data();
     }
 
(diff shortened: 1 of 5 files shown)


ui/src/treeDataProvider/AzureParentNode.ts
```diff
@@ -69,19 +69,21 @@ export class AzureParentNode<T extends IAzureParentTreeItem = IAzureParentTreeIt
...
+                    const node: AzureNode | undefined = children.find((n: AzureNode) => n.treeItem.id === treeItem.id);
+                    if (node) {
+                        return node;
+                    }
                 }
             }
         }
 
-        const pick: PickWithData<GetNodeFunction> = await ui.showQuickPick<GetNodeFunction>(this.getQuickPicks(), localize('azFunc.selectNode', 'Select a {0}', this.treeItem.childTypeLabel));
+        const pick: PickWithData<GetNodeFunction> = await ui.showQuickPick<GetNodeFunction>(this.getQuickPicks(expectedContextValues), localize('azFunc.selectNode', 'Select a {0}', this.treeItem.childTypeLabel));
```
nturinski: What will display to the user in the placeholder?  The first treeItem's childTypeLabel?

ui/src/treeDataProvider/AzureTreeDataProvider.ts
```diff
@@ -143,13 +143,23 @@ export class AzureTreeDataProvider implements TreeDataProvider<IAzureNode>, Disp
         }
     }
 
-    public async showNodePicker(expectedContextValue: string): Promise<IAzureNode> {
-        const picks: PickWithData<SubscriptionNode>[] = this._subscriptionNodes.map((n: SubscriptionNode) => new PickWithData<SubscriptionNode>(n, n.treeItem.label, n.subscription.subscriptionId));
-        let node: AzureNode = (await this._ui.showQuickPick<SubscriptionNode>(picks, localize('selectSubscription', 'Select a Subscription'))).data;
+    public async showNodePicker(expectedContextValues: string | string[], startingNode?: IAzureNode): Promise<IAzureNode> {
```
PrashanthCorp: Curious here, since I saw two choices of going about this : 
1. Accept a string or string array and do the conversion of string to string array inside the function. 
2. Ask all call sites to pass in a string array. 
I personally think approach two would be clearer to the users of this function.

ui/src/treeDataProvider/AzureTreeDataProvider.ts
```diff
@@ -143,13 +143,23 @@ export class AzureTreeDataProvider implements TreeDataProvider<IAzureNode>, Disp
...
+        if (startingNode) {
+            node = startingNode;
+        } else {
+            let picks: PickWithData<IAzureNode>[] = this._subscriptionNodes.map((n: SubscriptionNode) => new PickWithData(n, n.treeItem.label, n.subscription.subscriptionId));
+            picks = picks.concat(this._rootNodes.filter((n: AzureNode) => n.includeInNodePicker(<string[]>expectedContextValues)).map((n: AzureNode) => new PickWithData(n, n.treeItem.label)));
+            node = (await this._ui.showQuickPick<IAzureNode>(picks, localize('selectSubscription', 'Select a Subscription'))).data;
+        }
 
-        while (node.treeItem.contextValue !== expectedContextValue) {
+        while (!expectedContextValues.some((val: string) => node.treeItem.contextValue === val)) {
```
PrashanthCorp: Is this different from a `while(expectedContextValues.indexOf(node.treeItem.contextValue) === -1)` ?

ui/src/treeDataProvider/AzureParentNode.ts
EricJizbaMSFT: Not sure what you mean by 'the first'. There's only one childTypeLabel at this point in the code (`this.treeItem.childTypeLabel`). For example, if `this` is a Cosmos DB Account, the childTypeLabel is 'Database' and 'Select a Database' will be shown as the placeholder

For your other question, the Cosmos DB extension has several commands like 'Open In Portal' that apply to Cosmos DB Accounts that have different contextValues (mongoDatabaseAccount, graphDatabaseAccount, docDBDatabaseAccount, etc.) even though they're at the same level in the tree

nturinski: Oh, I misunderstood what treeType.childTypeLabel was.  That makes sense.

And also 👍 to your other answer.

ui/src/treeDataProvider/AzureTreeDataProvider.ts
EricJizbaMSFT: The biggest reason for number 2 was so that it would be backwards compatible (which is the best practice when you're publishing an npm package like this - and I just didn't want to go back and change every call of `showNodePicker` to an array)

We could debate which one is clearer (I happen to think 2 is clearer since most of the time you will pass in one value), but like I said that wasn't the main reason I did it this way

ui/src/treeDataProvider/AzureTreeDataProvider.ts
EricJizbaMSFT: Those look pretty much the same to me

PR #59, opened 2018-01-18
Refactor unpackBody to allow any field

nturinski:
The format of errors in the response from Kudu is also a Json string.  In the response, there is then a body property with a Json string containing a message.

I wasn't sure what a good name for the function/parameters would be though... Suggestions?

### DIFF
--- a/ui/src/parseError.ts
+++ b/ui/src/parseError.ts
@@ -12,7 +12,8 @@ export function parseError(error: any): IParsedError {
     let errorType: string = '';
     let message: string = '';
 
-    error = unpackErrorFromBody(error);
+    error = unpackErrorFromField(error, 'response');
+    error = unpackErrorFromField(error, 'body');
 
     if (typeof (error) === 'object' && error !== null) {
         if (error.constructor !== Object) {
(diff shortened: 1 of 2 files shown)


ui/src/parseError.ts
```diff
@@ -70,7 +71,7 @@ function getCode(o: any, defaultCode: string): string {
 
 function unpackErrorsInMessage(message: string): string {
     // Handle messages like this from Azure (just handle first error for now)
-    //   ["Errors":["The offer should have valid throughput ��",
+    //   ["Errors":["The offer should have valid throughput ��",
```
StephenWeatherford: Could you just get rid of the special characters?  Thx.

ui/src/parseError.ts
```diff
@@ -71,7 +71,7 @@ function getCode(o: any, defaultCode: string): string {
 
 function unpackErrorsInMessage(message: string): string {
     // Handle messages like this from Azure (just handle first error for now)
-    //   ["Errors":["The offer should have valid throughput ��",
+    //   ["Errors":["The offer should have valid throughput]]",
```
StephenWeatherford: Also, please add a testcase.

PR #64, opened 2018-02-01
Insert created node alphabetically

nturinski:
Insertion sort is worst case O(n).  While a Binary sort would be faster, I don't know if the size of nodes would ever be large enough to notice a performance drop.

Fixes https://github.com/Microsoft/vscode-azureappservice/issues/260


ui/src/utils/insertionSearch.ts
```diff
@@ -0,0 +1,18 @@
...
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { AzureNode } from '../treeDataProvider/AzureNode';
+
+// returns the index of where the node should be inserted
+export function insertionSearch(node: AzureNode, array: AzureNode[]): number {
+
+    // tslint:disable-next-line:no-increment-decrement
+    for (let i: number = 0; i < array.length; i++) {
```
EricJizbaMSFT: I think we should discourage putting stuff in 'utils' whenever possible (otherwise it becomes a dumping ground). No strong opinion, but I don't think this for-loop warrants its own file

nturinski: It felt a little messy to include it in the AzureParentNode.ts file to me as well though.  Where would you put it??

nturinski:
Tested with aisforapple (first), zisforzebra (last), and disforohhidoggy (normal)

ui/src/utils/insertionSearch.ts
StephenWeatherford: How about leaving it in the same folder as AzureParentNode.ts?

EricJizbaMSFT: It's only 6 lines of code - why not put it directly in addNodeToCache?

StephenWeatherford: Oh, wow, then agreed, unless you plan to test it separately.

PR #67, opened 2018-02-06
Add log streaming to appservice package

EricJizbaMSFT:
I copied most of this from the appservice extension, but I leveraged `AzureActionHandler.callWithTelemetry` from the ui package and `client.credentials.signRequest` from the new kudu package.

### DIFF
--- a/appservice/src/SiteWrapper.ts
+++ b/appservice/src/SiteWrapper.ts
@@ -203,7 +203,7 @@ export class SiteWrapper {
     /**
      * Starts the log-streaming service. Call 'dispose()' on the returned object when you want to stop the service.
      */
-    public async startStreamingLogs(client: KuduClient, actionHandler: AzureActionHandler, outputChannel: vscode.OutputChannel, path: string = ''): Promise<vscode.Disposable | undefined> {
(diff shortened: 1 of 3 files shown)


appservice/src/SiteWrapper.ts
```diff
@@ -195,6 +200,56 @@ export class SiteWrapper {
...
+                    resolve();
+                }
+            });
+        });
+
+        const requestApi: request.RequestAPI<request.Request, request.CoreOptions, {}> = request.defaults(httpRequest);
+        let disposable: vscode.Disposable | undefined;
+        // Intentionally setting up a separate telemetry event and not awaiting the result here since log stream is a long-running action
+        // tslint:disable-next-line:no-floating-promises
+        actionHandler.callWithTelemetry('appService.streamingLogs', async () => {
```
nturinski: Do we want all of our telemetry logs to come in with the label appsService.streamingLogs?  Can we make it context based?

EricJizbaMSFT: I mean both extensions are for app services, so the label is accurate. I think it's fine to be the same name. After all, it's the same code and we still have unique 'startStreamingLogs' events for each extension

PR #68, opened 2018-02-06
Fixes for log stream

EricJizbaMSFT:
1. Instead of returning just a Disposable, return an 'ILogStream' with the 'isConnected' property. That way we can prevent connecting to the same log stream twice
1. Add log config related methods. Unfortunately we can't just re-use the httpsLog code for functions because they use applicationLogs instead of httpsLogs

### DIFF
--- a/appservice/src/SiteWrapper.ts
+++ b/appservice/src/SiteWrapper.ts
@@ -18,6 +18,7 @@ import { DeployResult } from 'vscode-azurekudu/lib/models';
 import { DialogResponses } from './DialogResponses';
 import { ArgumentError } from './errors';
 import * as FileUtilities from './FileUtilities';
+import { ILogStream } from './ILogStream';
 import { localize } from './localize';
 
 // Deployment sources supported by Web Apps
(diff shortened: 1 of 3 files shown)


appservice/src/SiteWrapper.ts
```diff
@@ -216,18 +224,18 @@ export class SiteWrapper {
...
             await new Promise((resolve: () => void, reject: (err: Error) => void): void => {
                 const logsRequest: request.Request = requestApi(`${this.kuduUrl}/api/logstream/${path}`);
-                const disconnectMessage: string = localize('logStreamDisconnected', 'Disconnected from log-streaming service.');
-                disposable.dispose = (): void => {
+                logStream.dispose = (): void => {
                     logsRequest.removeAllListeners();
                     logsRequest.destroy();
                     outputChannel.show();
-                    outputChannel.appendLine(disconnectMessage);
+                    outputChannel.appendLine(localize('logStreamDisconnected', 'Disconnected from log-streaming service.'));
```
nturinski: Was this out of scope in the last PR?  Just wondering why you moved it

appservice/src/SiteWrapper.ts
```diff
@@ -158,12 +171,11 @@ export class SiteWrapper {
     }
 
     public async isHttpLogsEnabled(client: WebSiteManagementClient): Promise<boolean> {
```
nturinski: Where is this called?  Does it return the SiteLogsConfig now just to verify that the logs were updated?

EricJizbaMSFT: It's called in the app service extension. I mainly did it so we're consistent with client.webApps.updateDiagnosticLogsConfigSlot (which returns SiteLogsConfig)

appservice/src/SiteWrapper.ts
EricJizbaMSFT: `disconnectMessage` was used twice in the previous PR (once in `dispose` and once in `on('complete'`). I changed it so that `on('complete'` calls `dispose` and now the variable is only used once (aka no reason for a separate variable)

PR #69, opened 2018-02-07
Fix node refresh so that it keeps expansion state

EricJizbaMSFT:
VS Code released an [update today](https://code.visualstudio.com/updates/v1_20#_custom-views) which adds an 'id' property to tree items. This id tells VS Code that a node is the same even if the label changes.

I also made the id optional - we've had a bit of debate about whether or not to have id on our tree items and I think the best solution is to follow VS Code's pattern.

Bumped the minor version since this requires an update to the VS Code engine version.
Before:
refreshold

After:
refresh

### DIFF
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -71,9 +71,12 @@ export interface IAzureParentNode<T extends IAzureTreeItem = IAzureTreeItem> ext
  */
 export interface IAzureTreeItem {
     /**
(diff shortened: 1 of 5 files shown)


EricJizbaMSFT:
Yes, that is correct. Before, VS Code would collapse it because it thought it was a new node (because the label changed)

PR #73, opened 2018-02-13
Make context values more unique

StephenWeatherford:


ui/src/treeDataProvider/AzureParentNode.ts
```diff
@@ -126,7 +126,7 @@ export class AzureParentNode<T extends IAzureParentTreeItem = IAzureParentTreeIt
         if (this.treeItem.createChild && this.treeItem.childTypeLabel) {
             picks.unshift(new PickWithData<GetNodeFunction>(
                 async (): Promise<AzureNode> => await this.createChild(),
-                localize('azFunc.NodePickerCreateNew', '$(plus) Create New {0}', this.treeItem.childTypeLabel)
+                localize('azureextensionui.NodePickerCreateNew', '$(plus) Create New {0}', this.treeItem.childTypeLabel)
```
EricJizbaMSFT: We've actually been getting rid of these prefixes (azFunc, etc.) for localization because it shouldn't be necessary (I started the pattern - I know it's my fault lol). The id only needs to be unique on a per-file basis. You can see [this example](https://github.com/Microsoft/vscode-node-debug/tree/master/i18n/esn/out/node/extension) for what the localized file structure eventually looks like

ui/src/treeDataProvider/SubscriptionNode.ts
```diff
@@ -12,7 +12,7 @@ import { AzureSubscription } from '../azure-account.api';
 import { AzureParentNode } from './AzureParentNode';
 
 export class SubscriptionNode extends AzureParentNode {
-    public static readonly contextValue: string = 'azureSubscription';
+    public static readonly contextValue: string = 'azureextensionui.azureSubscription';
```
EricJizbaMSFT: This should be a minor version bump since it technically breaks functionality in extensions. For example:
* app service: https://github.com/Microsoft/vscode-azureappservice/blob/master/package.json#L262
* functions: https://github.com/Microsoft/vscode-azurefunctions/blob/master/package.json#L227

StephenWeatherford: @EricJizbaMSFT I'll take it then that you won't as republish 6.0.1.  I'll go ahead and update the version.

EricJizbaMSFT: I can't republish 0.6.1. You have to bump the version no matter what and I would prefer bumping the minor :)

StephenWeatherford: Realized what you meant after I wrote that. :-)

PR #74, opened 2018-02-16
Ping Function App so that logging works immediately

EricJizbaMSFT:
Fixes https://github.com/Microsoft/vscode-azurefunctions/issues/227


EricJizbaMSFT:
Fair point - but appservice _does_ use the defaultHostName [here](https://github.com/Microsoft/vscode-azureappservice/blob/master/src/explorer/SiteTreeItem.ts#L66). Ideally we switch that to siteWrapper.defaultHostName

PR #76, opened 2018-02-16
Add serializable id property to IAzureNode

EricJizbaMSFT:
This is useful for the Functions extension and the cosmos db extension.

@nturinski I also included updating the appservice package to 0.7.0 of the ui package in this change. Let me know if you think that'll cause issues

### DIFF
--- a/appservice/package.json
+++ b/appservice/package.json
@@ -1,7 +1,7 @@
 {
     "name": "vscode-azureappservice",
     "author": "Microsoft Corporation",
-    "version": "0.8.5",
+    "version": "0.9.0",
     "description": "Common tools for developing Azure App Service extensions for VS Code",
     "tags": [
         "azure",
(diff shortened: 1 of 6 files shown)


appservice/package.json
```diff
@@ -39,7 +39,7 @@
         "ms-rest-azure": "^2.4.4",
         "opn": "^5.1.0",
         "simple-git": "^1.80.1",
-        "vscode-azureextensionui": "~0.5.5",
+        "vscode-azureextensionui": "~0.7.0",
```
PrashanthCorp: The `ui/package.json` file says 0.7.1 while this is 0.7.0

EricJizbaMSFT: Yeah as long as the minor version matches, we should be fine. The Travis CI would never have passed if I used 0.7.1 here since I'm bumping the version of both packages in the same PR

PrashanthCorp: Scratch that, the requirement is an _at least_ 0.7.0

PR #81, opened 2018-02-26
Allow 204s for PutItem calls

EricJizbaMSFT:
Relevant Kudu PR here: https://github.com/EricJizbaMSFT/kudu/pull/4

PR #82, opened 2018-02-26
Add getFile and putFile calls to siteWrapper

EricJizbaMSFT:
We have to do some custom stuff when getting files, so I think it's worth having this in the shared package. Mostly copied this over from Nathan's implementation in appservice, but added a few things I needed for the functions repo.

This relies on one PR for the Kudu package: https://github.com/Microsoft/vscode-azuretools/pull/81

### DIFF
--- a/appservice/src/SiteWrapper.ts
+++ b/appservice/src/SiteWrapper.ts
@@ -7,11 +7,12 @@
 import WebSiteManagementClient = require('azure-arm-website');
 import { AppServicePlan, NameValuePair, Site, SiteConfigResource, SiteLogsConfig, SiteSourceControl, User } from 'azure-arm-website/lib/models';
 import * as fs from 'fs';
-import { BasicAuthenticationCredentials, ServiceClientCredentials, TokenCredentials, WebResource } from 'ms-rest';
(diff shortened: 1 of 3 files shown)


appservice/src/SiteWrapper.ts
```diff
@@ -291,6 +293,36 @@ export class SiteWrapper {
...
+    /**
+     * Returns the latest etag of the updated file
+     */
+    public async putFile(client: KuduClient, data: Readable | string, filePath: string, etag: string): Promise<string> {
+        let stream: Readable;
+        if (typeof data === 'string') {
+            stream = new Readable();
+            stream._read = function (_size: number): void {
+                this.push(data);
+                this.push(null);
```
nturinski: What's the purpose of pushing a null at the end of the data stream?  Does it just cut the stream?

EricJizbaMSFT: Per [the docs](https://nodejs.org/api/stream.html#stream_readable_push_chunk_encoding)

"Passing chunk as null signals the end of the stream (EOF), after which no more data can be written."

EricJizbaMSFT:
Filed an issue to fix the kudu builds on travis: https://github.com/Microsoft/vscode-azuretools/issues/83

I tried fixing it for a while, but couldn't get it to work. Since local builds still work, I think we're fine to merge PRs still. Let me know if you disagree

PR #87, opened 2018-02-28
Add additional runtimes for creating app

nturinski:


appservice/src/createAppService/SiteStep.ts
```diff
@@ -199,6 +199,34 @@ export class SiteStep extends WizardStep {
...
+                name: 'php|5.6',
+                displayName: 'PHP 5.6'
+            },
+            {
+                name: 'php|7.0',
+                displayName: 'PHP 7.0'
+            },
+            {
+                name: 'php|7.2',
+                displayName: 'PHP 7.2'
```
EricJizbaMSFT: I don't see 7.2 listed?
screen shot 2018-02-27 at 4 57 48 pm

nturinski: Portal app creation and Portal application settings don't match.  It seems that PHP 7.2 is supported though, so +1 for extension creation vs. portal creation.

image

EricJizbaMSFT: @nturinski if all of those node versions are supported - can you add them as well? And make 8.9 the LTS?

PR #88, opened 2018-02-28
Refactor AzureActionHandler

### LINKED ISSUE #78: Create custom error handling
Today there is no way to make a custom error as vscode-azureextensionui has automatic error handling.  We should make an error to allow for custom error handling while still tracking it in the telemetry,

EricJizbaMSFT:
* Removed the `withCustomTelemetry` methods and replaced it with a new parameter (`this: IActionContext`)  that can be consumed as needed. It feels simpler to me and allows us to add functionality in a backwards compatible format in the future (Since we can just add fields to `IActionContext`). For example I added 'displayError' to fix #78
* Changed `callWithTelemetry` to a static method per @StephenWeatherford's feedback that he didn't want to use an instance of the actionHandler directly.
* Added a method so that I could leverage `callWithTelemetry` _and_ get the error handling

NOTE: I tested these changes in most of our extensions, but didn't feel like sending out PRs until I get feedback on this first

### DIFF
--- a/ui/README.md
+++ b/ui/README.md
@@ -30,12 +30,12 @@ actionHandler.registerCommand('yourExtension.Refresh', function (this: IActionCo
 });
 ```
 
-Finally, you can also register events. By default, every event is tracked in telemetry. It is *highly recommended* to leverage the IActionContext.sendTelemetry parameter to filter only the events that apply to your extension. For example, if your extension only handles `json` files in the `onDidSaveTextDocument`, it might look like this:
+Finally, you can also register events. By default, every event is tracked in telemetry. It is *highly recommended* to leverage the IActionContext.suppressTelemetry parameter to filter only the events that apply to your extension. For example, if your extension only handles `json` files in the `onDidSaveTextDocument`, it might look like this:
 ```typescript
 actionHandler.registerEvent('yourExtension.onDidSaveTextDocument', vscode.workspace.onDidSaveTextDocument, async function (this: IActionContext, doc: vscode.TextDocument): Promise<void> {
-    this.sendTelemetry = false;
+    this.suppressTelemetry = true;
     if (doc.fileExtension === 'json') {
-        this.sendTelemetry = true;
+        this.suppressTelemetry = false;
         // custom logic here
     }
 });
--- a/ui/README.md
+++ b/ui/README.md
@@ -22,19 +22,20 @@ Here are a few of the benefits this provides:
   * duration
   * error
 
-If you want to add custom telemetry proprties, use `registerCommandWithCustomTelemetry` and add your own properties or measurements:
+If you want to add custom telemetry proprties, use the action's context and add your own properties or measurements:
 ```typescript
-actionHandler.registerCommandWithCustomTelemetry('yourExtension.Refresh', async (properties: TelemetryProperties, measurements: TelemetryMeasurements) => {
-    properties.customProp = "example prop";
-    measurements.customMeas = 49;
+actionHandler.registerCommand('yourExtension.Refresh', function (this: IActionContext): void {
+    this.properties.customProp = "example prop";
+    this.measurements.customMeas = 49;
 });
 ```
 
-Finally, you can also register events. The main difference is that your callback must take in the `trackTelemetry` parameter. Events are not tracked by default (since they can happen very frequently). You must call `trackTelemetry()` to signal that this event is indeed handled by your extension. For example, if your extension only handles `json` files in the `onDidSaveTextDocument`, it might look like this:
+Finally, you can also register events. By default, every event is tracked in telemetry. It is *highly recommended* to leverage the IActionContext.sendTelemetry parameter to filter only the events that apply to your extension. For example, if your extension only handles `json` files in the `onDidSaveTextDocument`, it might look like this:
 ```typescript
-actionHandler.registerEvent('yourExtension.onDidSaveTextDocument', vscode.workspace.onDidSaveTextDocument, (trackTelemetry: () => void, doc: vscode.TextDocument) => {
+actionHandler.registerEvent('yourExtension.onDidSaveTextDocument', vscode.workspace.onDidSaveTextDocument, async function (this: IActionContext, doc: vscode.TextDocument): Promise<void> {
+    this.sendTelemetry = false;
     if (doc.fileExtension === 'json') {
-        trackTelemetry();
+        this.sendTelemetry = true;
         // custom logic here
     }
 });
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -174,23 +174,31 @@ export declare class AzureActionHandler {
     registerCommand(commandId: string, callback: (this: IActionContext, ...args: any[]) => any): void;
 
     /**
-     * NOTE: By default, this sends a telemetry event every single time the event fires. It it recommended to use 'this.sendTelemetry' to only send events if they apply to your extension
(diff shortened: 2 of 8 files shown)


ui/README.md
```diff
@@ -22,19 +22,20 @@ Here are a few of the benefits this provides:
...
+    this.measurements.customMeas = 49;
 });
 ```
 
-Finally, you can also register events. The main difference is that your callback must take in the `trackTelemetry` parameter. Events are not tracked by default (since they can happen very frequently). You must call `trackTelemetry()` to signal that this event is indeed handled by your extension. For example, if your extension only handles `json` files in the `onDidSaveTextDocument`, it might look like this:
+Finally, you can also register events. By default, every event is tracked in telemetry. It is *highly recommended* to leverage the IActionContext.sendTelemetry parameter to filter only the events that apply to your extension. For example, if your extension only handles `json` files in the `onDidSaveTextDocument`, it might look like this:
 ```typescript
-actionHandler.registerEvent('yourExtension.onDidSaveTextDocument', vscode.workspace.onDidSaveTextDocument, (trackTelemetry: () => void, doc: vscode.TextDocument) => {
+actionHandler.registerEvent('yourExtension.onDidSaveTextDocument', vscode.workspace.onDidSaveTextDocument, async function (this: IActionContext, doc: vscode.TextDocument): Promise<void> {
```
StephenWeatherford: Why not an arrow function?

ui/index.d.ts
```diff
@@ -171,20 +171,26 @@ export declare abstract class BaseEditor<ContextT> implements Disposable {
...
-    registerEvent<T>(eventId: string, event: Event<T>, callback: (trackTelemetry: () => void, ...args: any[]) => any): void;
-    registerEventWithCustomTelemetry<T>(eventId: string, event: Event<T>, callback: (trackTelemetry: () => void, properties: TelemetryProperties, measurements: TelemetryMeasurements, ...args: any[]) => any): void;
+    registerEvent<T>(eventId: string, event: Event<T>, callback: (this: IActionContext, ...args: any[]) => any): void;
 
     /**
-     * Use this method for generic calls that you want to track telemetry for, but that aren't registered with VS Code
+     * Use these methods for generic calls that you want to handle, but that aren't registered with VS Code
      */
-    callWithTelemetry(callbackId: string, callback: (properties: TelemetryProperties, measurements: TelemetryMeasurements) => any): Promise<any>;
+    static callWithTelemetry(callbackId: string, telemetryReporter: TelemetryReporter | undefined, callback: (this: IActionContext) => any): Promise<any>;
```
StephenWeatherford: I still think these should be extracted from AzureActionHandler, they seem completely unrelated. I had it as a separate top-level function. Maybe a namespace?

StephenWeatherford: shouldn't the callback return Promise<void>?

StephenWeatherford: Thinking about this...  I like how it gets rid of callWithCustomTelemetry

ui/README.md
EricJizbaMSFT: You can't specify the type of "this" in an arrow function. For example, the [tslint rules](https://github.com/Microsoft/tslint-microsoft-contrib) explicitly call it out as an exception to their "no-function-expression" rule:
> Function expressions that contain a 'this' reference are allowed and will not create a failure.

[More docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions):
> An arrow function expression has a shorter syntax than a function expression and does not have its own this

StephenWeatherford: But why use this at all?  Isn't it better to be explicit about the parameters?

ui/index.d.ts
EricJizbaMSFT: Not sure we're going to agree on this one. They share 90% of the same code and this 'handles azure actions'

StephenWeatherford: I also don't think telemetry is just about actions. You could have telemetry about performance metrics that are unrelated to any specific action, for instance.

But it's not a big enough deal to worry about. Still an improvement.

ui/index.d.ts
```diff
@@ -171,20 +171,26 @@ export declare abstract class BaseEditor<ContextT> implements Disposable {
...
+    static callWithTelemetryAndErrorHandling(callbackId: string, telemetryReporter: TelemetryReporter | undefined, outputChannel: OutputChannel | undefined, callback: (this: IActionContext) => any): Promise<any>;
+}
+
+export interface IActionContext {
+    properties: TelemetryProperties;
+    measurements: TelemetryMeasurements;
+    sendTelemetry: boolean;
+    displayError: boolean;
```
StephenWeatherford: Is this also intended to be the solution to the DoNotShowToUserError?

ui/src/AzureActionHandler.ts
```diff
@@ -22,82 +22,75 @@ export class AzureActionHandler {
...
+    public static async callWithTelemetryAndErrorHandling(callbackId: string, telemetryReporter: TelemetryReporter | undefined, outputChannel: OutputChannel | undefined, callback: (this: IActionContext) => any): Promise<any> {
         const start: number = Date.now();
-        const properties: TelemetryProperties = {};
-        const measurements: TelemetryMeasurements = {};
-        properties.result = 'Succeeded';
-        let sendTelemetry: boolean = false;
+        const context: IActionContext = {
+            properties: {},
+            measurements: {},
+            sendTelemetry: true,
```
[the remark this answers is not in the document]
EricJizbaMSFT: Yeah you've mentioned this before and I guess I don't have a huge preference. Until you mentioned it, I'd never heard of the assumption that booleans default to false for optional parameters. This is the only case where I assume a boolean defaults to false:
```
class Example {
    x: boolean;
}
```
But again - that's distincly different than an optional parameter.

I'm also not the only one. If you look at [vscode.d.ts](https://github.com/Microsoft/vscode/blob/master/src/vs/vscode.d.ts) and search for "defaults to `false`" vs "defaults to `true`", you find 3 of each

ui/index.d.ts
EricJizbaMSFT: Yeah

ui/README.md
EricJizbaMSFT: The benefit of using 'this' IMO is that you can ignore `IActionContext` if you don't need to specify any custom properties, measurements, etc. (which I think is the more common case). I modeled it after the mocha tests where you can use ISuiteCallbackContext, etc. to do things like `this.setTimeout`

In other words, it was purely for the sake of convenience. If you're worried it'll be confusing or something (I was honestly on the fence about this) - I'm fine to make it a required parameter each time

ui/index.d.ts
EricJizbaMSFT: Not necessarily. You might call a synchronous method and still want telemetry on it

EricJizbaMSFT: If "Action" is the part you're worried about, then I'm happy to rename the class - but I do prefer to have the code together

ui/README.md
StephenWeatherford: I don't get an error when passing in () => {} when the IActionContext argument is explicitly specified. If you're getting a lint error from that, though, then I agree being able to leave it out is better.

ui/index.d.ts
EricJizbaMSFT: I wasn't feeling the "completely unrelated" argument, but you won me over with the "I don't like static methods" argument. Moved to separate file 😊

StephenWeatherford: 😆

PR #89, opened 2018-02-28
Update appservice package due to telemetry refactoring

EricJizbaMSFT:
Based on this PR: https://github.com/Microsoft/vscode-azuretools/pull/88

### DIFF
--- a/appservice/src/SiteWrapper.ts
+++ b/appservice/src/SiteWrapper.ts
@@ -15,7 +15,7 @@ import * as git from 'simple-git/promise';
 import { Readable } from 'stream';
 import { setInterval } from 'timers';
 import * as vscode from 'vscode';
-import { AzureActionHandler, IAzureNode, IParsedError, parseError, UserCancelledError } from 'vscode-azureextensionui';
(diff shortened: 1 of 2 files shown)


appservice/src/SiteWrapper.ts
```diff
@@ -246,18 +247,21 @@ export class SiteWrapper {
         const requestApi: request.RequestAPI<request.Request, request.CoreOptions, {}> = request.defaults(httpRequest);
         const logStream: ILogStream = { dispose: undefined, isConnected: true };
         // Intentionally setting up a separate telemetry event and not awaiting the result here since log stream is a long-running action
+        // tslint:disable-next-line:no-var-self
+        const me: SiteWrapper = this;
         // tslint:disable-next-line:no-floating-promises
-        actionHandler.callWithTelemetry('appService.streamingLogs', async () => {
+        callWithTelemetryAndErrorHandling('appService.streamingLogs', reporter, undefined, async function (this: IActionContext): Promise<void> {
+            this.suppressErrorDisplay = true;
```
StephenWeatherford: do you need to turn on rethrow?

EricJizbaMSFT: Nope - streaming logs prints the logs straight to the output window _before_ it throws the error. Also its intentionally a floating promise, so its not like we need to rethrow the error to break out of any following code

PR #90, opened 2018-02-28
Verify deployment has succeeded before returning from deploy function

nturinski:


nturinski:
Output now looks like:

image

appservice/src/SiteWrapper.ts
```diff
@@ -293,6 +298,43 @@ export class SiteWrapper {
         await requestP.get(`https://${this.defaultHostName}/admin/host/status`, requestOptions);
     }
 
+    public async validateDeployment(outputChannel: vscode.OutputChannel, pollingInterval: number = 5000, pollingAttempts: number = 12): Promise<responseStatus[]> {
+        // by default, try for a minute before we declare that the deployment failed
+        // tslint:disable-next-line:no-http-string
+        const uri: string = `${this.isSsl ? 'https://' : 'http://'}${this.defaultHostName}`;
+        const validatingDeployment: string = `Validating your ${this.isFunctionApp ? 'function' : 'web'} app successfully deployed.`;
```
EricJizbaMSFT: Use 'localize'

appservice/src/SiteWrapper.ts
```diff
@@ -293,6 +298,43 @@ export class SiteWrapper {
         await requestP.get(`https://${this.defaultHostName}/admin/host/status`, requestOptions);
     }
 
+    public async validateDeployment(outputChannel: vscode.OutputChannel, pollingInterval: number = 5000, pollingAttempts: number = 12): Promise<responseStatus[]> {
+        // by default, try for a minute before we declare that the deployment failed
```
EricJizbaMSFT: Now that I think about it more - I would prefer if it was based on time, not number of attempts. Using "number of attempts" would have inconsistent results dependent on how long it takes to poll the site each time. For example, if it takes 10 seconds to poll the site, the user is actually waiting (5 + 10)*12=3 minutes

I'd prefer something like this:
```typescript
const maxTime: number = Date.now() + timeoutInSeconds * 1000;
while (Date.now() < maxTime) {
    ...
}
```

appservice/src/SiteWrapper.ts
```diff
@@ -293,6 +298,43 @@ export class SiteWrapper {
         await requestP.get(`https://${this.defaultHostName}/admin/host/status`, requestOptions);
     }
 
+    public async validateDeployment(outputChannel: vscode.OutputChannel, pollingInterval: number = 5000, pollingAttempts: number = 12): Promise<responseStatus[]> {
```
EricJizbaMSFT: I don't think pollingInterval should be a parameter. It feels more like an internal implementation detail to me.

In any case, I think the pollingAttempts parameter is much more likely to be changed, so at least move that to be the first optional parameter

appservice/src/SiteWrapper.ts
```diff
@@ -45,6 +47,7 @@ export class SiteWrapper {
     public readonly id: string;
     public readonly defaultHostName: string;
     public readonly isFunctionApp: boolean;
+    public readonly isSsl: boolean;
```
EricJizbaMSFT: Do you think people will want to use "isSsl"? Or should this just be "defaultProtocol" that returns "http://" or "https://"?

appservice/src/SiteWrapper.ts
```diff
@@ -293,6 +298,43 @@ export class SiteWrapper {
...
+        const options: {} = {
+            method: 'GET',
+            uri: uri,
+            resolveWithFullResponse: true
+        };
+        let response: IncomingMessage;
+        const statuses: responseStatus[] = [];
+        for (let i: number = 0; i < pollingAttempts; i += 1) {
+            try {
+                await new Promise((resolve: () => void): void => { setTimeout(resolve, pollingInterval); });
```
EricJizbaMSFT: I'd move the pollingInterval code to the bottom of the loop. You don't want to force the user to wait 5 seconds if the site is ready immediately (although I realize its unlikely to be ready immediately 😂)

appservice/src/SiteWrapper.ts
```diff
@@ -293,6 +298,43 @@ export class SiteWrapper {
...
+                // wait for the polling interval to send a request
+                this.log(outputChannel, validatingDeployment);
+                // tslint:disable-next-line:no-unsafe-any
+                response = <IncomingMessage>(await requestP(options));
+                // request throws an error for 400-500 responses
+                if (response.statusCode >= 200 && response.statusCode < 400) {
+                    // a status code 200-300 indicates success
+                    statuses.push({ code: response.statusCode, message: response.statusMessage });
+                    return statuses;
```
EricJizbaMSFT: Based on the name of this function, I would expect it to return void and throw an error if it fails to validate. If you don't want to throw an error, then rename it to "tryValidateDeployment" ('try' usually indicates a function will not throw an error).

I'm also not sure what you expect the consumer of this method to do with the array of status codes? IMO only the last status code is useful

appservice/src/SiteWrapper.ts
```diff
@@ -293,6 +298,43 @@ export class SiteWrapper {
...
+                }
+            } catch (error) {
+                // tslint:disable-next-line:no-unsafe-any
+                statuses.push({ code: error.response.statusCode, message: error.response.statusMessage });
```
EricJizbaMSFT: Are you sure 'error' will always have a response? Or could this result in a dumb null reference exception?

I would prefer if you did the following:
1. Move as much code out of the 'try' block as you can (so that you know 'this.log' is not the code throwing an error)
1. Use parseError if possible (Does it handle 'statusCode' and 'statusMessage'? - If not I'm okay if you just fix the first thing)

EricJizbaMSFT:
Per offline discussion, closing in favor of Stephen's PR in app service repo: https://github.com/Microsoft/vscode-azureappservice/pull/317

PR #91, opened 2018-02-28
hideDeployConfirmation default to false

nturinski:


appservice/src/SiteWrapper.ts
```diff
@@ -168,7 +168,7 @@ export class SiteWrapper {
         outputChannel.appendLine(localize('DeleteSucceeded', 'Successfully deleted "{0}".', this.appName));
     }
 
-    public async deploy(fsPath: string, client: WebSiteManagementClient, outputChannel: vscode.OutputChannel, configurationSectionName: string, confirmDeployment: boolean = true): Promise<void> {
+    public async deploy(fsPath: string, client: WebSiteManagementClient, outputChannel: vscode.OutputChannel, configurationSectionName: string, hideDeployConfirmation: boolean = false): Promise<void> {
```
StephenWeatherford: I appreciate the sentiment, Nathan, but don't worry about it. This is less important for parameters than it used to be, given how good intellisense is, and we already have other places defaulting to true.

PR #92, opened 2018-03-01
Fix bug that would connect GitHub to web apps instead of slot

### LINKED ISSUE #80: GitHub connection always connects to the web app even if initialized on a slot
Forgot to check to see if the site is a slot or web app.

nturinski:
Fixes #80 

Added aka.ms link for GitHub documentation
Changed response from taking user to portal to taking them to documentation to set it up.

### DIFF
--- a/appservice/src/SiteWrapper.ts
+++ b/appservice/src/SiteWrapper.ts
@@ -134,6 +134,12 @@ export class SiteWrapper {
         return await client.appServicePlans.get(this.planResourceGroup, this.planName);
     }
 
+    public async updateSourceControl(client: WebSiteManagementClient, siteSourceControl: SiteSourceControl): Promise<SiteSourceControl> {
+        return this.slotName ?
+            await client.webApps.createOrUpdateSourceControlSlot(this.resourceGroup, this.name, siteSourceControl, this.slotName) :
+            await client.webApps.createOrUpdateSourceControl(this.resourceGroup, this.name, siteSourceControl);
+    }
+
     public async deleteSite(client: WebSiteManagementClient, outputChannel: vscode.OutputChannel): Promise<void> {
         const confirmMessage: string = localize('deleteConfirmation', 'Are you sure you want to delete "{0}"?', this.appName);
         if (await vscode.window.showWarningMessage(confirmMessage, DialogResponses.yes, DialogResponses.cancel) !== DialogResponses.yes) {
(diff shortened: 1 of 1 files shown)


appservice/src/SiteWrapper.ts
```diff
@@ -584,14 +590,18 @@ export class SiteWrapper {
             isMercurial: false
         };
 
-        this.log(outputChannel, `"${this.appName}" is being connected to GitHub repo. This may take several minutes...`);
+        this.log(outputChannel, `"${this.appName}" is being connected to the GitHub repo. This may take several minutes...`);
         try {
-            await client.webApps.createOrUpdateSourceControlWithHttpOperationResponse(this.resourceGroup, this.name, siteSourceControl);
+            await this.updateSourceControl(client, siteSourceControl);
```
EricJizbaMSFT: Do you think we'll always have to do a resync after updating source control? In other words, should updateSourceControl handle the resync or is it fine to have connectToGitHub handle the resync?

appservice/src/SiteWrapper.ts
```diff
@@ -602,12 +612,13 @@ export class SiteWrapper {
         }
     }
 
-    private async showGitHubAuthPrompt(node: IAzureNode): Promise<void> {
-        const goToPortal: string = localize('goToPortal', 'Go to Portal');
-        const setupGithub: string = localize('GitRequired', 'Authorize Azure for GitHub under "Deployment options."');
+    private async showGitHubAuthPrompt(): Promise<void> {
+        const goToPortal: string = localize('learnMore', 'Learn More');
```
EricJizbaMSFT: nit: rename the variable to learnMore

appservice/src/SiteWrapper.ts
```diff
@@ -602,12 +612,13 @@ export class SiteWrapper {
...
+        const setupGithub: string = localize('GitRequired', 'Connect your App to a GitHub repository to authorize Azure.');
```
EricJizbaMSFT: How about: "You must give Azure access to your Github account."

I think it sounds more like an error message and I think it makes it more obvious that you only have to do this once per account (rather than once per repository)

appservice/src/SiteWrapper.ts
```diff
@@ -584,14 +590,18 @@ export class SiteWrapper {
...
+        this.log(outputChannel, `"${this.appName}" is being connected to the GitHub repo. This may take several minutes...`);
         try {
-            await client.webApps.createOrUpdateSourceControlWithHttpOperationResponse(this.resourceGroup, this.name, siteSourceControl);
+            await this.updateSourceControl(client, siteSourceControl);
         } catch (err) {
             try {
                 // a resync will fix the first broken build
                 // https://github.com/projectkudu/kudu/issues/2277
-                await client.webApps.syncRepository(this.resourceGroup, this.name);
+                if (this.slotName) {
```
EricJizbaMSFT: nit: How about you move this into a SiteWrapper.syncRepository method

appservice/src/SiteWrapper.ts
nturinski: The resync seems to be specifically an issue for connecting to GitHub.  When we support other sourceControls (such as Bitbucket), I think we can see if that resync will be necessary for every updateSourceControl.

appservice/src/SiteWrapper.ts
nturinski: I'm not sure where we'd use this anywhere else (which is why I didn't create its own method).

PR #94, opened 2018-03-01
Add optional id parameter for openInPortal

nturinski:
https://github.com/Microsoft/vscode-azureappservice/issues/311

Portal changed the path for individual slots to include the parent's path anymore

### DIFF
--- a/ui/src/treeDataProvider/AzureNode.ts
+++ b/ui/src/treeDataProvider/AzureNode.ts
@@ -89,7 +89,7 @@ export class AzureNode<T extends IAzureTreeItem = IAzureTreeItem> implements IAz
     }
 
     public openInPortal(id?: string): void {
-        id = id === undefined ? id : this.id;
+        id = id === undefined ? this.id : id;
         (<(s: string) => void>opn)(`${this.environment.portalUrl}/${this.tenantId}/#resource${id}`);
     }
 
--- a/ui/src/treeDataProvider/AzureNode.ts
+++ b/ui/src/treeDataProvider/AzureNode.ts
@@ -89,7 +89,7 @@ export class AzureNode<T extends IAzureTreeItem = IAzureTreeItem> implements IAz
     }
 
     public openInPortal(id?: string): void {
-        id = id ? id : this.id;
+        id = id === undefined ? id : this.id;
         (<(s: string) => void>opn)(`${this.environment.portalUrl}/${this.tenantId}/#resource/${id}`);
     }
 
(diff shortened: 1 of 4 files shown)


nturinski:
The optional id allows two things:

1.) Opening a resource in Azure that does not necessarily live in the explorer
2.) Opening a resource in Azure where the id does not match the conventional parent/child id pattern

ui/src/treeDataProvider/AzureNode.ts
```diff
@@ -88,8 +88,10 @@ export class AzureNode<T extends IAzureTreeItem = IAzureTreeItem> implements IAz
         await this.treeDataProvider.refresh(this);
     }
 
-    public openInPortal(): void {
-        (<(s: string) => void>opn)(`${this.environment.portalUrl}/${this.tenantId}/#resource${this.id}`);
+    public openInPortal(id?: string): void {
+        id ?
```
EricJizbaMSFT: You can simplify by doing one of the following:
`id = id === undefined ? this.id : id`
`id = id || this.id`

That way you don't have duplicate the open and url logic

nturinski: This is why you are "boss".

ui/src/treeDataProvider/AzureNode.ts
```diff
@@ -88,8 +88,9 @@ export class AzureNode<T extends IAzureTreeItem = IAzureTreeItem> implements IAz
         await this.treeDataProvider.refresh(this);
     }
 
-    public openInPortal(): void {
-        (<(s: string) => void>opn)(`${this.environment.portalUrl}/${this.tenantId}/#resource${this.id}`);
+    public openInPortal(id?: string): void {
+        id = id === undefined ? id : this.id;
+        (<(s: string) => void>opn)(`${this.environment.portalUrl}/${this.tenantId}/#resource/${id}`);
```
EricJizbaMSFT: Did you intentionally add a slash?:
old: `resource${this.id}`
new: `resource/${id}`

nturinski: No, that was an accident from when I was calling opn twice.  Good catch.

ui/src/treeDataProvider/AzureNode.ts
```diff
@@ -88,8 +88,9 @@ export class AzureNode<T extends IAzureTreeItem = IAzureTreeItem> implements IAz
         await this.treeDataProvider.refresh(this);
     }
 
-    public openInPortal(): void {
-        (<(s: string) => void>opn)(`${this.environment.portalUrl}/${this.tenantId}/#resource${this.id}`);
+    public openInPortal(id?: string): void {
+        id = id === undefined ? id : this.id;
```
EricJizbaMSFT: Your conditionals are backwards

PR #97, opened 2018-03-02
Compare previous latest deployment id to current deployment id

nturinski:
Grabs the old latest deployment id before the deployment begins.

As I implemented this, I questioned how useful this actually is.  While this guarantees that deployments were successful, it doesn't guarantee that it was built successfully so deployments can still fail silently since technically, the deployment DID succeed, but the build didn't.

If a user tries to deploy a Local Git repo that is already deployed, it will also result in an error since the ids will be the same.  Should we consider that by design?

### DIFF
--- a/appservice/src/SiteWrapper.ts
+++ b/appservice/src/SiteWrapper.ts
@@ -185,16 +185,11 @@ export class SiteWrapper {
         const config: SiteConfigResource = await this.getSiteConfig(client);
         const kuduClient: KuduClient = await this.getKuduClient(client);
         const deploymentId: string = 'latest';
-        const expectedErrorMessage: string = 'Deployment \'latest\' not found.';
         let previousDeploymentId: string | undefined;
         try {
             previousDeploymentId = (await kuduClient.deployment.getResult(deploymentId)).id;
             // current active deployment used to verify new deployment was successful
         } catch (error) {
-            const parsedError: IParsedError = parseError(error);
-            if (!parsedError.message.includes(expectedErrorMessage)) {
-                throw new Error(parsedError.message);
-            }
             // swallow the error-- getResult('latest') should only fail if this is the first deployment
         }
 
--- a/appservice/src/SiteWrapper.ts
+++ b/appservice/src/SiteWrapper.ts
@@ -185,11 +185,16 @@ export class SiteWrapper {
         const config: SiteConfigResource = await this.getSiteConfig(client);
         const kuduClient: KuduClient = await this.getKuduClient(client);
         const deploymentId: string = 'latest';
-        let previousDeploymentId: string;
+        const expectedErrorMessage: string = 'Deployment \'latest\' not found.';
+        let previousDeploymentId: string | undefined;
         try {
             previousDeploymentId = (await kuduClient.deployment.getResult(deploymentId)).id;
             // current active deployment used to verify new deployment was successful
         } catch (error) {
+            const parsedError: IParsedError = parseError(error);
+            if (!parsedError.message.includes(expectedErrorMessage)) {
+                throw new Error(parsedError.message);
+            }
             // swallow the error-- getResult('latest') should only fail if this is the first deployment
         }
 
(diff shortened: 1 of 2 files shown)


appservice/src/SiteWrapper.ts
```diff
@@ -183,14 +183,24 @@ export class SiteWrapper {
 
     public async deploy(fsPath: string, client: WebSiteManagementClient, outputChannel: vscode.OutputChannel, configurationSectionName: string, confirmDeployment: boolean = true): Promise<void> {
         const config: SiteConfigResource = await this.getSiteConfig(client);
+        const kuduClient: KuduClient = await this.getKuduClient(client);
+        const deploymentId: string = 'latest';
+        let previousDeploymentId: string;
+        try {
+            previousDeploymentId = (await kuduClient.deployment.getResult(deploymentId)).id;
+            // current active deployment used to verify new deployment was successful
+        } catch (error) {
```
StephenWeatherford: Perhaps you should check that the error is what you're expecting for this scenario?

appservice/src/SiteWrapper.ts
```diff
@@ -183,14 +183,24 @@ export class SiteWrapper {
 
     public async deploy(fsPath: string, client: WebSiteManagementClient, outputChannel: vscode.OutputChannel, configurationSectionName: string, confirmDeployment: boolean = true): Promise<void> {
         const config: SiteConfigResource = await this.getSiteConfig(client);
+        const kuduClient: KuduClient = await this.getKuduClient(client);
+        const deploymentId: string = 'latest';
+        let previousDeploymentId: string;
```
StephenWeatherford: string|undefined

appservice/src/SiteWrapper.ts
```diff
@@ -537,7 +547,9 @@ export class SiteWrapper {
             await new Promise((resolve: () => void): void => { setTimeout(resolve, pollingInterval); });
             deployment = await kuduClient.deployment.getResult(deploymentId);
         }
-
+        if (deployment.id === previousDeploymentId) {
+            throw new Error(localize('unsuccessfulDeployment', 'Deployment was unsuccessful and fell back to last active deployment.'));
```
StephenWeatherford: How about something like:

  The deployment was unsuccessful and fell back to the last active deployment. If this is a local git deployment, it could also mean this repo has already been deployed.

EricJizbaMSFT: The deployment didn't actually fall back to the last deployment, though. If you navigate to the web app's url, you will see the broken app that you just deployed (you will _not_ see the last successful deployment).

The only thing that "fell back" is the deployment metadata reported by kudu when you query for 'latest'. But that's just an internal implementation detail to kudu and the user doesn't care about it.

In other words, I don't think we should mention "falling back" at all

appservice/src/SiteWrapper.ts
EricJizbaMSFT: @StephenWeatherford I just talked to Nathan offline and I think we should swallow all errors here. No matter if or how it fails, we should still move on and try to deploy IMO

appservice/src/SiteWrapper.ts
```diff
@@ -537,7 +547,9 @@ export class SiteWrapper {
             await new Promise((resolve: () => void): void => { setTimeout(resolve, pollingInterval); });
             deployment = await kuduClient.deployment.getResult(deploymentId);
         }
-
+        if (deployment.id === previousDeploymentId) {
```
EricJizbaMSFT:
```
if (previousDeploymentId !== undefined && deployment.id === previousDeploymentId)
```

appservice/src/SiteWrapper.ts
nturinski: Discussing with Eric, it may not be best to check for the specific error message we're expecting.

If the message ever changes or is localized, this will break the deployment.
If we get an error at this point of the deployment, this shouldn't break the deployment.  This is only for validation and we account for undefined ids.

If there is an error, it's better for it to be caught in the specific deployment command rather than during the check for a latest id.

appservice/src/SiteWrapper.ts
nturinski: Is this necessary?  If the previousDeploymentId is undefined, this will not be true unless deployment.id is also undefined which would probably be a failure anyway.

EricJizbaMSFT: True - I didn't think about it that way. I might still prefer it for the sake of readability, but no strong preference

nturinski: 👑

PR #99, opened 2018-03-02
Add more detailed type information to telemetry properties

EricJizbaMSFT:


ui/index.d.ts
```diff
@@ -201,8 +201,22 @@ export interface IActionContext {
     rethrowError: boolean;
 }
 
-export type TelemetryProperties = { [key: string]: string; };
-export type TelemetryMeasurements = { [key: string]: number };
+export interface TelemetryProperties {
```
nturinski: nit: If it's an interface, shouldn't we name is ITelemetryProperties?

EricJizbaMSFT: I'm going to leave it as-is. I don't want to change this everywhere in all our extensions and I don't think its as "required" to have "I" in front of interface names in typescript. I'll fix this later if more people complain

PR #100, opened 2018-03-02
Add supported node runtimes

nturinski:


appservice/src/createAppService/SiteStep.ts
```diff
@@ -200,6 +200,22 @@ export class SiteStep extends WizardStep {
...
+            {
+                name: 'node|8.2',
+                displayName: 'Node.js 8.2'
+            },
+            {
+                name: 'node|8.8',
+                displayName: 'Node.js 8.8'
+            },
+            {
+                name: 'node|8.9',
```
EricJizbaMSFT: Move "(LTS - Recommended for new apps)" down to version 8.9

It's the LTS listed on nodejs.org

nturinski: Don't we want to stay consistent with the portal?

Issue #101, opened 2018-03-02
Azure UI - How to control order of nodes?


philliphoff:
It looks like the ordering of nodes is based on each's `label` property (i.e. the nodes are in alphabetical order).  If some nodes have indicators in their label (e.g. a `*` as a prefix to indicate selected), currently the node ordering changes as the indicators change, which may be confusing.

Is there a way to keep the nodes in a specific order irrespective of their `label`?

EricJizbaMSFT:
@philliphoff just curious what you're using the ui package for? Are you creating a new extension for VS Code?

In addition, I'm not sure what you mean by "selected". Could you give more detail on why it would have a `*` prefix?

philliphoff:
@EricJizbaMSFT Just a little prototype of something related to the VS tooling I've been working on.  Happy to show you a demo if you want to stop by.

@StephenWeatherford No, sorry, I should have been more clear.  This is strictly about wanting to control the order of the nodes (somewhat) independently of the content of their labels.

A number of CLIs have a concept of an ambient "selected" context (e.g. the Azure CLI with subscriptions, the Kubernetes CLI with clusters, etc.).  If I used a `*`prefix in a label of a node to indicate that it was the currently selected context, VS Code would suddenly shuffle that node to the top of its parent (as the `*` prefix causes the node to sort "higher").  The effect is a little jarring, so I'd like to keep the nodes in the same order regardless of which node has the `*` prefix.

(You could argue that it might be better to use the node icon to indicate selection, and I'd agree, but not being a graphic design artist, a `*` was going to be as good as it gets until I could beg UX to draw me a custom icon.)

#103 is the PR, if you want to see exactly what I'm proposing.

EricJizbaMSFT:
Yeah I'm honestly not sure how I feel about a star at the beginning of the the label to indicate 'selected', but I can totally get on board with the main issue at hand (to allow custom sorting of nodes).

philliphoff:
(The `*` prefix actually mimics what's displayed when using the CLI command to list the items, so it's at least consistent until I get a real UX resource.)

Issue #83, opened 2018-02-26
Fix kudu builds on travis


EricJizbaMSFT:
About a week ago, kudu builds on travis started failing with the following:
```
Error: Cannot resolve pipeline stage 'openapi-document/identity'.
    at resolvePipelineStageName (/home/travis/.autorest/@microsoft.azure_autorest-core@2.0.4147/node_modules/@microsoft.azure/autorest-core/dist/lib/pipeline/pipeline.js:142: 15)
    at inputs.map (/home/travis/.autorest/@microsoft.azure_autorest-core@2.0.4147/node_modules/@microsoft.azure/autorest-core/dist/lib/pipeline/pipeline.js:159: 108)
    at Array.map (<anonymous>)
    at createNodesAndSuffixes (/home/travis/.autorest/@microsoft.azure_autorest-core@2.0.4147/node_modules/@microsoft.azure/autorest-core/dist/lib/pipeline/pipeline.js:159: 97)
    at BuildPipeline (/home/travis/.autorest/@microsoft.azure_autorest-core@2.0.4147/node_modules/@microsoft.azure/autorest-core/dist/lib/pipeline/pipeline.js:205: 9)
    at Object.RunPipeline (/home/travis/.autorest/@microsoft.azure_autorest-core@2.0.4147/node_modules/@microsoft.azure/autorest-core/dist/lib/pipeline/pipeline.js:256: 22)
    at <anonymous>
```

This error does not reproduce locally

PR #103, opened 2018-03-03
Add custom sort of child nodes

philliphoff:
Enables `IChildProvider` implementations the option of custom-sorting their nodes. If a child provider implements `compareChildren()`, the tree will use that comparison function instead of its default alphabetical sort.

Related to #101.

### DIFF
--- a/appservice/src/createAppService/SiteStep.ts
+++ b/appservice/src/createAppService/SiteStep.ts
@@ -190,7 +190,7 @@ export class SiteStep extends WizardStep {
             },
             {
                 name: 'node|6.11',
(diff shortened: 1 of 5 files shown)


philliphoff:
A component of the PR build is failing, but it appears to be unrelated to my changes.

EricJizbaMSFT:
Yeah the build failures are a known issue: https://github.com/Microsoft/vscode-azuretools/issues/83
As long as the 'ui' part of the build is passing, that's all that matters.

We're actually in the middle of a release of several of our extensions right now - might not get to this for a few days

philliphoff:
@EricJizbaMSFT Great! Bumped patch version.

PR #107, opened 2018-03-06
Add telemetry for the tree

EricJizbaMSFT:
This will track the following:
1. Success or failure of getChildren
1. Which nodes users expand
1. Where users are running common commands from (like "openInPortal or "refresh")

Also display error node (we had this functionality for attached nodes in Cosmos - but I think it would be nice for everywhere)

This is another PR I suggest opening in Codeflow - a lot of the change is just whitespace

### DIFF
--- a/ui/src/treeDataProvider/AzureTreeDataProvider.ts
+++ b/ui/src/treeDataProvider/AzureTreeDataProvider.ts
@@ -5,11 +5,14 @@
 
 import * as path from 'path';
 import { Disposable, Event, EventEmitter, Extension, extensions, TreeDataProvider, TreeItem, TreeItemCollapsibleState } from 'vscode';
-import { IAzureNode, IAzureParentTreeItem, IChildProvider } from '../../index';
+import TelemetryReporter from 'vscode-extension-telemetry';
+import { IActionContext, IAzureNode, IAzureParentTreeItem, IChildProvider } from '../../index';
 import { AzureAccount, AzureLoginStatus, AzureSubscription } from '../azure-account.api';
+import { callWithTelemetryAndErrorHandling } from '../callWithTelemetryAndErrorHandling';
 import { ArgumentError } from '../errors';
 import { IUserInterface, PickWithData } from '../IUserInterface';
 import { localize } from '../localize';
+import { parseError } from '../parseError';
 import { VSCodeUI } from '../VSCodeUI';
 import { AzureNode } from './AzureNode';
 import { AzureParentNode } from './AzureParentNode';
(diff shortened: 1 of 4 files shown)


ui/src/treeDataProvider/AzureTreeDataProvider.ts
```diff
@@ -81,49 +86,75 @@ export class AzureTreeDataProvider implements TreeDataProvider<IAzureNode>, Disp
...
+                let result: IAzureNode[];
+
+                if (node !== undefined) {
+                    actionContext.properties.contextValue = node.treeItem.contextValue;
+
+                    result = node.creatingNodes
+                        .concat(await node.getCachedChildren())
+                        .concat(node.treeItem.hasMoreChildren() ? new AzureNode(node, new LoadMoreTreeItem(thisTree._loadMoreCommandId)) : []);
+                } else { // Root of tree
+                    actionContext.properties.isActivationEvent = 'true';
```
StephenWeatherford: What does isActivationEvent mean again?

ui/src/treeDataProvider/AzureTreeDataProvider.ts
```diff
@@ -81,49 +86,75 @@ export class AzureTreeDataProvider implements TreeDataProvider<IAzureNode>, Disp
...
                     }
-                });
-                nodes = this._subscriptionNodes;
-            }
 
-            return nodes.concat(this._rootNodes);
+                    result = nodes.concat(thisTree._rootNodes);
+                }
+
+                this.measurements.childCount = result.length;
```
StephenWeatherford: Do you care about whether it has more children, or whether this request is a result of load more children?

ui/src/treeDataProvider/AzureTreeDataProvider.ts
```diff
@@ -81,49 +86,75 @@ export class AzureTreeDataProvider implements TreeDataProvider<IAzureNode>, Disp
...
+                return result;
+            });
+        } catch (error) {
+            return [new AzureNode(node, {
+                label: localize('errorNode', 'Error: {0}', parseError(error).message),
+                contextValue: 'azureextensionui.error'
+            })];
         }
```
StephenWeatherford: Consider extracting some of this code into separate functions.

ui/src/treeDataProvider/AzureTreeDataProvider.ts
EricJizbaMSFT: It means that this event happens as a part of extension activation - aka the event is sent even if they don't interact with the explorer or our extension at all (which we shouldn't be counting as 'usage' of our extension).

In this case, it means I don't want to count 'getChildren' at the root of the explorer as 'usage' for our extension

ui/src/treeDataProvider/AzureTreeDataProvider.ts
EricJizbaMSFT: I'll add something about 'hasMoreChildren'. Not really interested in load more since we already have a command tracking loadMore

PR #108, opened 2018-03-07
Add deployment src/dest hashes to telemetry

StephenWeatherford:


appservice/src/SiteWrapper.ts
```diff
@@ -181,7 +182,12 @@ export class SiteWrapper {
         outputChannel.appendLine(localize('DeleteSucceeded', 'Successfully deleted "{0}".', this.appName));
     }
 
-    public async deploy(fsPath: string, client: WebSiteManagementClient, outputChannel: vscode.OutputChannel, configurationSectionName: string, confirmDeployment: boolean = true): Promise<void> {
+    public async deploy(fsPath: string, client: WebSiteManagementClient, outputChannel: vscode.OutputChannel, configurationSectionName: string, confirmDeployment: boolean = true, telemetryProperties?: { [key: string]: string }): Promise<void> {
```
EricJizbaMSFT: nit: You could use the `TelemetryProperties` type defined in the ui package. Are you intentionally using a more generic type here?

appservice/package.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "vscode-azureappservice",
     "author": "Microsoft Corporation",
-    "version": "0.11.0",
+    "version": "0.11.1",
```
EricJizbaMSFT: No need to bump the version. Its already bumped on master and we'll include this change before publishing

appservice/src/SiteWrapper.ts
StephenWeatherford: Done

PR #109, opened 2018-03-07
Add scm type to telemetry

nturinski:


appservice/src/SiteWrapper.ts
```diff
@@ -183,12 +183,13 @@ export class SiteWrapper {
     }
 
     public async deploy(fsPath: string, client: WebSiteManagementClient, outputChannel: vscode.OutputChannel, configurationSectionName: string, confirmDeployment: boolean = true, telemetryProperties?: TelemetryProperties): Promise<void> {
+        const config: SiteConfigResource = await this.getSiteConfig(client);
         if (telemetryProperties) {
             telemetryProperties.sourceHash = randomUtils.getPseudononymousStringHash(fsPath);
             telemetryProperties.destHash = randomUtils.getPseudononymousStringHash(this.appName);
+            telemetryProperties.scmType = config.scmType;
```
StephenWeatherford: Anything else we want to add that's easily available?

nturinski: I just pushed up slots

EricJizbaMSFT: The confirmation for zip deployments is in this file - so if you want to track where people drop off - maybe add a boolean there?

StephenWeatherford: Add it as a cancelStep property, e.g.:

telemetryProperties.cancelStep = 'confirmDeployment';

PR #116, opened 2018-03-08
Saw/url

StephenWeatherford:
Display deployed website URl and add some more deployment telemetry values

### DIFF
--- a/appservice/src/SiteWrapper.ts
+++ b/appservice/src/SiteWrapper.ts
@@ -185,10 +185,24 @@ export class SiteWrapper {
     public async deploy(fsPath: string, client: WebSiteManagementClient, outputChannel: vscode.OutputChannel, configurationSectionName: string, confirmDeployment: boolean = true, telemetryProperties?: TelemetryProperties): Promise<void> {
         const config: SiteConfigResource = await this.getSiteConfig(client);
         if (telemetryProperties) {
-            telemetryProperties.sourceHash = randomUtils.getPseudononymousStringHash(fsPath);
-            telemetryProperties.destHash = randomUtils.getPseudononymousStringHash(this.appName);
-            telemetryProperties.scmType = config.scmType;
-            telemetryProperties.isSlot = this.slotName ? 'true' : 'false';
+            try {
+                telemetryProperties.sourceHash = randomUtils.getPseudononymousStringHash(fsPath);
+                telemetryProperties.destHash = randomUtils.getPseudononymousStringHash(this.appName);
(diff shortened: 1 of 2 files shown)


appservice/src/SiteWrapper.ts
```diff
@@ -185,10 +185,24 @@ export class SiteWrapper {
...
+                telemetryProperties.alwaysOn = config.alwaysOn ? 'true' : 'false';
+                telemetryProperties.linuxFxVersion = config.linuxFxVersion;
+                telemetryProperties.nodeVersion = config.nodeVersion;
+                telemetryProperties.pythonVersion = config.pythonVersion;
+                telemetryProperties.hasCors = config.cors ? 'true' : 'false';
+                telemetryProperties.hasIpSecurityRestrictions = config.ipSecurityRestrictions && config.ipSecurityRestrictions.length > 0 ? 'true' : 'false';
+                telemetryProperties.javaVersion = config.javaVersion;
+
+                // Do this last in case it fails
+                telemetryProperties.state = await this.getState(client);
```
EricJizbaMSFT: My only worry is that this might cause a delay before the "Are you sure you want to deploy" message pops up. Thoughts? Do you think it will be fast?

appservice/src/SiteWrapper.ts
```diff
@@ -202,7 +216,13 @@ export class SiteWrapper {
                 break;
         }
 
-        outputChannel.appendLine(localize('deployComplete', '>>>>>> Deployment to "{0}" completed. <<<<<<', this.appName));
+        const displayUrl: string = this.isFunctionApp ? '' : this.defaultHostName;
+        outputChannel.appendLine(
+            displayUrl ?
```
EricJizbaMSFT: nit: displayUrl doesn't seem necessary. You could just do something like:
```
!this.isFunctionApp ?
    localize('deployCompleteWithUrl', '>>>>>> Deployment to "{0}" completed: {1} <<<<<<', this.appName, `https://${this.defaultHostName}`) :
    localize('deployComplete', '>>>>>> Deployment to "{0}" completed. <<<<<<', this.appName)
```

appservice/src/SiteWrapper.ts
StephenWeatherford: Good point. No, it's not fast enough. I put it in the background.

appservice/src/SiteWrapper.ts
StephenWeatherford: I could, I just think this makes it more explicit what I'm trying to do.

PR #118, opened 2018-03-12
Split siteWrapper into multiple files

### LINKED ISSUE #112: Split SiteWrapper into separate files
The files is getting pretty large. We could at least move the deploy logic out.

I've also always been unhappy with the name "SiteWrapper". Not sure what the name should be, but the current name doesn't tell us any helpful information about what it is.

EricJizbaMSFT:
I basically renamed SiteWrapper to SiteClient and removed anything that wasn't a simple wrapper of a WebSiteManagementClient method. See the updated README for more info.

Fixes #112


appservice/src/editScmType.ts
```diff
@@ -0,0 +1,189 @@
...
+    const quickPick: vscode.QuickPickItem = await vscode.window.showQuickPick(scmQuickPicks, { placeHolder: placeHolder });
+    if (quickPick === undefined || quickPick.description === currentSource) {
+        // if the user clicks the current source, treat it as a cancel
+        throw new UserCancelledError();
+    } else {
+        return quickPick.label;
+    }
+}
+
+async function connectToGitHub(node: IAzureNode, client: SiteClient, outputChannel: vscode.OutputChannel): Promise<void> {
```
nturinski: I feel like all the GitHub related code could be in its own folder or at least its own file.  Seems a little odd to throw it in editScmType

EricJizbaMSFT:
Yeah I think its worth categorizing them, but I'd rather do that separately for two reasons:
1. I'm not sure what the categorization should be and I don't want to delay a large change like this while we figure that out (For example, I don't think 'commands' has the same meaning in an npm package as it does in a VS Code extension)
1. I think that will touch several files that were not originally a part of siteWrapper - aka slightly out of scope for the original issue

appservice/README.md
```diff
@@ -1,6 +1,15 @@
-# VSCode Azure SDK for Node.js - App Service Tools
+# VSCode Azure SDK for Node.js - App Service Tools (Preview)
+
+This package provides common Azure App Service elements for VS Code extensions.
+
+All functionality for this package is built around the `SiteClient`, a wrapper of a `WebSiteManagementClient` for use with a specific Site. It reduces the number of arguments needed for every call and automatically ensures the 'slot' method is called when appropriate. Support for all methods on the `WebSiteManagementClient` are being added on an as-needed basis.
```
StephenWeatherford: Thanks!

PR #119, opened 2018-03-15
Add support for common user input

### LINKED ISSUE #23: Expose IUserInterface in ui package
It has a few benefits:
1. It consistently throws UserCancelledError
    1. helpful for telemetry purposes
    1. Easier than constantly checking `if (result !== undefined)`
1. It allows us to 'mock' the UI for unit tests (see [TestUI](https://github.com/Microsoft/vscode-azuretools/blob/master/ui/src/TestUI.ts))

### LINKED ISSUE #114: telemetry reporter, telemetry properties, etc should be required args (but should accept null/undefined)
... to make sure clients pass these in or make a specific judgment to pass in null/undefined.

EricJizbaMSFT:
I copied most of this code from several parts of several of our repos. First of all, this is meant to reduce the amount of copied code and increase shared code. Its also just nice to have a consistent implementation. When I was doing this, I tried to keep the interfaces as similar to VS Code interfaces as possible - so that we can easily leverage their improvements in the future.

There's still work left to do, but I felt this was definitely enough to start. For example, I didn't do any documentation in the README yet, but I did add it to the index.d.ts file. I also want to expose several of the common wizard steps (like resource group) but I'll do that in a separate PR.

Fixes #114 
Fixes #23


ui/index.d.ts
```diff
@@ -234,3 +236,139 @@ export interface IParsedError {
...
+    public showWarningMessage(message: string, ...items: MessageItem[]): Promise<MessageItem>;
+    public showOpenDialog(options: OpenDialogOptions): Promise<Uri[]>;
+}
+
+/**
+ * Provides additional options for QuickPickItems used in Azure Extensions
+ */
+export interface IAzureQuickPickItem<T = undefined> extends QuickPickItem {
+    /**
+     * An optional id to uniqely identify this item items across sessions, used in persisting previous selections
```
StephenWeatherford: item items

Did I do that?  :-)

StephenWeatherford: uniqely

ui/index.d.ts
```diff
@@ -234,3 +236,139 @@ export interface IParsedError {
...
+     * If not specified, a hash of the label will be used
+     */
+    id?: string;
```
StephenWeatherford: You think we should rename this persistenceId?

ui/src/DialogResponses.ts
```diff
@@ -7,8 +7,14 @@ import { MessageItem } from 'vscode';
...
+    export const yes: MessageItem = { title: localize('yes', 'Yes') };
+    export const no: MessageItem = { title: localize('no', 'No') };
+    export const cancel: MessageItem = { title: localize('cancel', 'Cancel'), isCloseAffordance: true };
+    export const deleteResponse: MessageItem = { title: localize('delete', 'Delete') };
+    export const learnMore: MessageItem = { title: localize('learnMore', 'Learn more') };
+    export const dontWarnAgain: MessageItem = { title: localize('dontWarnAgain', 'Don\'t warn again') };
+    export const skipForNow: MessageItem = { title: localize('skipForNow', 'Skip for now') };
     export const upload: MessageItem = { title: localize('upload', "Upload") };
-    export const dontWarn: MessageItem = { title: localize('dontwarn', "Upload, don't warn again") };
+    export const uploadDontWarnAgain: MessageItem = { title: localize('dontwarn', "Upload, don't warn again") };
```
StephenWeatherford: Can we change to "Always upload"?

ui/src/DialogResponses.ts
```diff
@@ -7,8 +7,14 @@ import { MessageItem } from 'vscode';
...
     export const dontUpload: MessageItem = { title: localize('dontUpload', "Don't Upload"), isCloseAffordance: true };
```
StephenWeatherford: Don't capitalize Upload (none of the others capitalize multiple words).

ui/index.d.ts
EricJizbaMSFT: I chose just `id` to be consistent with how its done for TreeItems. See the comment here: https://github.com/Microsoft/vscode/blob/master/src/vs/vscode.d.ts#L5233

In both TreeItems and Picks, its used to persist some sort of state, but at the end of the day its just an id for that thing. In other words, we could potentially use it for something other than persistence in the future

ui/src/TestUserInput.ts
```diff
@@ -3,52 +3,51 @@
...
                 // Use default value if input is undefined
-                result = value;
+                result = options.value;
             }
 
             if (result) {
-                if (validateInput) {
-                    const msg: string | null | undefined = validateInput(result);
+                if (options.validateInput) {
+                    const msg: string | null | undefined = await Promise.resolve(options.validateInput(result));
```
StephenWeatherford: await will wait on both the inner and outer promises if validateInput is a promise?

EricJizbaMSFT: Promise.resolve will just return the original promise if the argument is already a promise. There will be no "inner" and "outer" promise in that case. From [the docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/resolve):


> The Promise.resolve(value) method returns a Promise object that is resolved with the given value. If the value is a thenable (i.e. has a "then" method), the returned promise will "follow" that thenable, adopting its eventual state; if the value was a promise, that object becomes the result of the call to Promise.resolve; otherwise the returned promise will be fulfilled with the value.

ui/src/AzureUserInput.ts
```diff
@@ -0,0 +1,104 @@
...
+        this._persistence = persistence;
+    }
+
+    public async showQuickPick<T extends QuickPickItem>(items: T[] | Thenable<T[]>, options: QuickPickOptions): Promise<T> {
+        if (options.ignoreFocusOut === undefined) {
+            options.ignoreFocusOut = true;
+        }
+
+        let persistenceKey: string | undefined;
+        const hashData: string | undefined = (<IAzureQuickPickOptions>options).id || options.placeHolder;
```
StephenWeatherford: We should add something from the wizard itself into this hash, otherwise different wizards asking for resource groups will all persistent the same choice.

ui/src/AzureUserInput.ts
```diff
@@ -0,0 +1,104 @@
...
+        items = await Promise.resolve(items);
+
+        if (persistenceKey) {
+            const previousValue: string | undefined = this._persistence.get(persistenceKey);
+            if (previousValue) {
+                const index: number = items.findIndex((item: T) => getPersistenceValue(item) === previousValue);
+                if (index !== -1) {
+                    const previousItem: T = items.splice(index, 1)[0];
+
+                    const recentlyUsed: string = localize('recentlyUsed', '(recently used)');
```
StephenWeatherford: I like the recently used label

ui/src/AzureUserInput.ts
EricJizbaMSFT: Two things:
1. IMO that's a good thing. If I select a resource group for one wizard, I want that to also be 'recently used' in another wizard
1. GlobalState is on a per-extension basis, so this already only works on a per-extension basis

StephenWeatherford: the name "hashData" is very confusing, I kept thinking it was the actual persistence data. How about "unhashedKey"?

StephenWeatherford: 2. is good, but I'm not convinced about 1.  Sometimes it will be good, sometimes not. If you don't want it for a particular one you can change the behavior by manually giving an id, right? So I'm willing to wait and see how it works out.

ui/src/DialogResponses.ts
EricJizbaMSFT: I don't have a preference - but if someone else complains about the change I'm pointing them directly to you :P

ui/src/AzureUserInput.ts
EricJizbaMSFT: Yes you can change the behavior by specifying an id

PR #123, opened 2018-03-17
Run from zip

nturinski:


appservice/src/deploy/deployRunFromZip.ts
```diff
@@ -0,0 +1,165 @@
...
+    const storageAccountKeys: StorageAccountListKeysResult = await client.listStorageAccountKeys(saResourceGroup, sa.name);
+    return azureStorage.createBlobService(sa.name, storageAccountKeys.keys[0].value);
+
+}
+
+async function createBlobFromZip(blobService: azureStorage.BlobService, zipFilePath: string): Promise<string> {
+    const containerName: string = 'azureappservice-run-from-zip';
+    const blobName: string = azureStorage.date.secondsFromNow(0).toLocaleString().replace(/[.,\/#!$%\^&\*;:{}=\-_`~()]/g, '').replace(/\s/g, ''); // get rid of any special characters and spaces in the timestamp
+    await new Promise<void>((resolve: () => void, reject: (err: Error) => void): void => {
+        blobService.createContainerIfNotExists(containerName, { publicAccessLevel: 'blob' }, (err: Error) => {
```
EricJizbaMSFT: I don't think we should make this blob publicly accessible. Thats not required for run-from-zip is it?

appservice/src/deploy/deployRunFromZip.ts
```diff
@@ -0,0 +1,165 @@
...
+    const resourceGroups: string = 'resourceGroups';
+    const saResourceGroup: string = parsedId[resourceGroups];
+    const storageAccountKeys: StorageAccountListKeysResult = await client.listStorageAccountKeys(saResourceGroup, sa.name);
+    return azureStorage.createBlobService(sa.name, storageAccountKeys.keys[0].value);
+
+}
+
+async function createBlobFromZip(blobService: azureStorage.BlobService, zipFilePath: string): Promise<string> {
+    const containerName: string = 'azureappservice-run-from-zip';
+    const blobName: string = azureStorage.date.secondsFromNow(0).toLocaleString().replace(/[.,\/#!$%\^&\*;:{}=\-_`~()]/g, '').replace(/\s/g, ''); // get rid of any special characters and spaces in the timestamp
```
EricJizbaMSFT: It feels like the date will be illegible once you remove all the special characters and spaces. We should investigate the issue you ran into because blob names should support some special characters

appservice/src/deploy/deployRunFromZip.ts
```diff
@@ -0,0 +1,165 @@
...
+            } else {
+                resolve();
+            }
+        });
+    });
+    const sasToken: string = blobService.generateSharedAccessSignature(containerName, blobName, <azureStorage.common.SharedAccessPolicy>{ AccessPolicy: {
+        Permissions: azureStorage.BlobUtilities.SharedAccessPermissions.READ + azureStorage.BlobUtilities.SharedAccessPermissions.LIST,
+        Start: azureStorage.date.secondsFromNow(-10),
+        // for clock desync
+        Expiry: azureStorage.date.minutesFromNow(60),
```
EricJizbaMSFT: Does this mean their deployment will only work for 60 minutes?

appservice/src/deploy/deployRunFromZip.ts
```diff
@@ -0,0 +1,165 @@
...
+import { TelemetryProperties } from 'vscode-azureextensionui';
+import * as FileUtilities from '../FileUtilities';
+import { localize } from '../localize';
+import { SiteClient } from '../SiteClient';
+import { uiUtils } from '../utils/uiUtils';
+import { IQuickPickItemWithData } from '../wizard/IQuickPickItemWithData';
+import { formatDeployLog } from './formatDeployLog';
+
+export async function deployRunFromZip(client: SiteClient, fsPath: string, outputChannel: vscode.OutputChannel, telemetryProperties?: TelemetryProperties): Promise<void> {
+    // if (confirmDeployment) {
```
EricJizbaMSFT: Remove the commented code. Your current behavior is "Create a new blob every time" which is not destructive. We should also support "Replace the blob every time", but I'm okay addressing that separately

appservice/src/ScmType.ts
```diff
@@ -7,5 +7,6 @@
 export enum ScmType {
     None = 'None', // default scmType
     LocalGit = 'LocalGit',
-    GitHub = 'GitHub'
+    GitHub = 'GitHub',
+    RunFromZip = 'Run-From-Zip'
```
EricJizbaMSFT: I don't like creating our own 'deployment type'.  I'm not convinced that Kudu will always allow us to put whatever we want here. We definitely need to discuss with the partner team because it feels like they should already define a "Run from zip" deployment type for us

nturinski: Agreed, I'm uneasy about it as well.  I'll ask them what their plans are-- I wanted to confirm whether or not a deployment was created after a run-from-zip before I commented on that thread.

appservice/src/deploy/deployRunFromZip.ts
```diff
@@ -0,0 +1,165 @@
...
+    const storageAccounts: StorageAccountListResult = await client.listStorageAccounts();
+    const storageAccountQuickPicks: IQuickPickItemWithData<StorageAccount>[] = storageAccounts.map((sa: StorageAccount) => {
+        return {
+            label: sa.name,
+            description: '',
+            data: sa
+        };
+    });
+
+    const storageAccount: StorageAccount = (await uiUtils.showQuickPickWithData(storageAccountQuickPicks, { placeHolder: 'Choose a storage account to host the zip file.', ignoreFocusOut: true })).data;
```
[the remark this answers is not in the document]
nturinski: Agreed.  Should that be a workspace setting so that the user can change it?

appservice/src/deploy/deployRunFromZip.ts
StephenWeatherford: Don't use toLocateString() - the output will vary depending on the machine/user's locale, and you could end up with cases you don't expect.  I'd suggest toISOString() - 2011-10-05T14:48:00.000Z.

You could replace the colons with hyphens or something but otherwise the format will be consistent.

EricJizbaMSFT:
@nturinski I assume you can close this in favor of https://github.com/Microsoft/vscode-azurefunctions/pull/320?

PR #124, opened 2018-03-19
Use data instead of description for editScmType

EricJizbaMSFT:
Because of my recent changes related to persistence, we sometimes display "(recently used)" as the description, which means we should use the data filed instead of the description field to get the value


EricJizbaMSFT:
@nturinski Yeah I noticed you did that in your other PR, but I didn't realize why. I'll let you take care of that since it only becomes an issue with run from zip

PR #125, opened 2018-03-19
Add common wizard steps

EricJizbaMSFT:
Added the following common steps:
1. Resource Group
1. Location
1. Generic "name" step
1. Storage Account step

See the index.d.ts file for more documentation.

This code was mostly copied from the appservice package, but I took bits and pieces from other repos as well. For example, I liked the resource group name validation in the Cosmos DB extension the best because it had the most readable error messages.

### DIFF
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -414,7 +414,7 @@ export interface IRelatedNameWizardContext {
 
 /**
  * A generic class for a step that specifies the name of a new resource, used to generate a related name for other new resources.
- * You must implement `isNameAvailable` and assign `wizardContext.relatedNameTask` to the result of `generateRelatedName`
+ * You must implement `isRelatedNameAvailable` and assign `wizardContext.relatedNameTask` to the result of `generateRelatedName`
  */
 export declare abstract class AzureNameStep<T extends IRelatedNameWizardContext> extends AzureWizardStep<T> {
     /**
@@ -422,7 +422,7 @@ export declare abstract class AzureNameStep<T extends IRelatedNameWizardContext>
      * @param wizardContext The context of the wizard.
      * @param name The name that will be checked.
      */
-    protected abstract isNameAvailable(wizardContext: T, name: string): Promise<boolean>;
+    protected abstract isRelatedNameAvailable(wizardContext: T, name: string): Promise<boolean>;
 
     /**
      * Generates a related name for new resources
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -369,19 +369,33 @@ export interface ISubscriptionWizardContext {
 
 export interface ILocationWizardContext extends ISubscriptionWizardContext {
     /**
-     * You may specify the defaultLocationName if you don't want the `LocationStep` to prompt for a location
-     * For example, if the user selects an existing resource, you might want to use that location as the default for the wizard's other resources
-     */
-    defaultLocationName?: string;
-
-    /**
      * The location to use for new resources
-     * This value will be defined after `LocationStep.prompt` occurs.
+     * This value will be defined after `LocationStep.prompt` occurs or after you call `LocationStep.setLocation`
      */
     location?: Location;
+
+    /**
+     * The task used to get locations.
+     * By specifying this in the context, we can ensure that Azure is only queried once for the entire wizard
+     */
+    locationsTask?: Promise<Location[]>;
 }
 
 export declare class LocationStep<T extends ILocationWizardContext> extends AzureWizardStep<T> {
+    /**
+     * This will set the wizard context's location (in which case the user will _not_ be prompted for location)
+     * For example, if the user selects an existing resource, you might want to use that location as the default for the wizard's other resources
+     * @param wizardContext The context of the wizard
+     * @param name The name or display name of the location
+     */
+    public static setLocation<T extends ILocationWizardContext>(wizardContext: T, name: string): Promise<void>;
+
+    /**
+     * Used to get locations. By passing in the context, we can ensure that Azure is only queried once for the entire wizard
+     * @param wizardContext The context of the wizard.
+     */
+    public static getLocations<T extends ILocationWizardContext>(wizardContext: T): Promise<Location[]>;
+
     public prompt(wizardContext: T, ui: IAzureUserInput): Promise<T>;
     public execute(wizardContext: T, outputChannel: OutputChannel): Promise<T>;
 }
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -4,11 +4,13 @@
  *--------------------------------------------------------------------------------------------*/
 
 
-import { Subscription } from 'azure-arm-resource/lib/subscription/models';
+import { Subscription, Location } from 'azure-arm-resource/lib/subscription/models';
 import { ServiceClientCredentials } from 'ms-rest';
 import { AzureEnvironment } from 'ms-rest-azure';
 import { Uri, TreeDataProvider, Disposable, TreeItem, Event, OutputChannel, Memento, InputBoxOptions, QuickPickItem, QuickPickOptions, TextDocument, ExtensionContext, MessageItem, OpenDialogOptions } from 'vscode';
 import TelemetryReporter from 'vscode-extension-telemetry';
+import { ResourceGroup } from 'azure-arm-resource/lib/resource/models';
+import { StorageAccount, CheckNameAvailabilityResult } from 'azure-arm-storage/lib/models';
 
 export declare class AzureTreeDataProvider implements TreeDataProvider<IAzureNode>, Disposable {
     public static readonly subscriptionContextValue: string;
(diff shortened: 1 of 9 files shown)


ui/index.d.ts
```diff
@@ -360,6 +362,128 @@ export declare abstract class AzureWizardStep<T> {
...
+    credentials: ServiceClientCredentials;
+    subscription: Subscription;
+}
+
+export interface ILocationWizardContext extends ISubscriptionWizardContext {
+    /**
+     * You may specify the defaultLocationName if you don't want the `LocationStep` to prompt for a location
+     * For example, if the user selects an existing resource, you might want to use that location as the default for the wizard's other resources
+     */
+    defaultLocationName?: string;
```
StephenWeatherford: This is confusing - default implies you're going to ask the user but ENTER will select this. Would it make sense to just set location instead?  Or maybe defaultLocationNoPrompt?

ui/index.d.ts
```diff
@@ -360,6 +362,128 @@ export declare abstract class AzureWizardStep<T> {
...
+ * A generic class for a step that specifies the name of a new resource, used to generate a related name for other new resources.
+ * You must implement `isRelatedNameAvailable` and assign `wizardContext.relatedNameTask` to the result of `generateRelatedName`
+ */
+export declare abstract class AzureNameStep<T extends IRelatedNameWizardContext> extends AzureWizardStep<T> {
+    /**
+     * This method will by called by `generateRelatedName` when trying to find a unique suffix for the related name
+     * @param wizardContext The context of the wizard.
+     * @param name The name that will be checked.
+     */
+    protected abstract isRelatedNameAvailable(wizardContext: T, name: string): Promise<boolean>;
```
nturinski: I'm hesitant to create our own isRelatedNameAvailable functions.  The WebSiteManagementClient has its own checkNameAvailability function that should handle checking ResouceNameAvailability.  It seems like only the storage step is leveraging this.

The portal isn't super transparent about their naming restrictions so we may unintentionally miss/block resource names by creating our own rules.  Or if they alter their naming rules without alerting us.

ui/index.d.ts
```diff
@@ -360,6 +362,128 @@ export declare abstract class AzureWizardStep<T> {
...
+    public execute(wizardContext: T, outputChannel: OutputChannel): Promise<T>;
+}
+
+export interface IAzureNamingRules {
+    minLength: number;
+    maxLength: number;
+
+    /**
+     * A RegExp specifying the invalid characters.
+     * For example, /[^a-z0-9]/ would specify that only lowercase, alphanumeric characters are allowed.
```
StephenWeatherford: Don't resource groups have other naming rules?  E.g. I think hyphens are allowed, but not at the end and not two in a row.

ui/index.d.ts
```diff
@@ -360,6 +362,128 @@ export declare abstract class AzureWizardStep<T> {
     public abstract execute(wizardContext: T, outputChannel: OutputChannel): Promise<T>;
 }
 
+export interface ISubscriptionWizardContext {
+    credentials: ServiceClientCredentials;
+    subscription: Subscription;
+}
+
+export interface ILocationWizardContext extends ISubscriptionWizardContext {
```
StephenWeatherford: I appreciate all the explicit comments.

ui/src/wizard/ResourceGroupStep.ts
```diff
@@ -0,0 +1,105 @@
...
+        if (wizardContext.resourceGroupsTask === undefined) {
+            // tslint:disable-next-line:no-non-null-assertion
+            const client: ResourceManagementClient = new ResourceManagementClient(wizardContext.credentials, wizardContext.subscription.subscriptionId!);
+            wizardContext.resourceGroupsTask = uiUtils.listAll(client.resourceGroups, client.resourceGroups.list());
+        }
+
+        return await wizardContext.resourceGroupsTask;
+    }
+
+    public static async isNameAvailable<T extends IResourceGroupWizardContext>(wizardContext: T, name: string): Promise<boolean> {
```
nturinski: The resourceManagementClient has a function called checkExistence that does most of this name validation.   
https://github.com/Azure/azure-sdk-for-node/blob/1640f6952c3510547c3e02aac1c663d0b7354736/lib/services/resourceManagement/lib/resource/operations/resourceGroups.js#L40-L80

ui/src/wizard/AzureNameStep.ts
```diff
@@ -0,0 +1,57 @@
...
+        let preferredName: string = namingRules.some((n: IAzureNamingRules) => !!n.lowercaseOnly) ? name.toLowerCase() : name;
+
+        for (const invalidCharsRegExp of namingRules.map((n: IAzureNamingRules) => n.invalidCharsRegExp)) {
+            preferredName = preferredName.replace(invalidCharsRegExp, '');
+        }
+
+        const minLength: number = Math.max(...namingRules.map((n: IAzureNamingRules) => n.minLength));
+        const maxLength: number = Math.min(...namingRules.map((n: IAzureNamingRules) => n.maxLength));
+
+        const maxTries: number = 1000;
```
StephenWeatherford: 1000 seems excessive since each try is an Azure call

ui/src/wizard/AzureNameStep.ts
```diff
@@ -0,0 +1,57 @@
...
+                return newName;
+            }
+            count += 1;
+        }
+
+        return undefined;
+    }
+
+    private generateSuffixedName(preferredName: string, i: number, minLength: number, maxLength: number): string {
+        const suffix: string = i === 1 ? '' : i.toString();
```
StephenWeatherford: Yay!  I hate auto-namings that start with suffixing "1".  'Course this may have come from my code... :-)

ui/index.d.ts
EricJizbaMSFT: I see your point and I think we should make this decision on a case-by-case basis. I agree that ideally we can use `checkNameAvailability`, but frankly its pretty terrible sometimes. For example this is what the WebSiteManagementClient.checkNameAvailability message looks like:
screen shot 2018-03-22 at 11 45 44 am

The naming rules actually don't change that often and there's usually documentation outside of the portal. For example:
https://docs.microsoft.com/en-us/azure/architecture/best-practices/naming-conventions#storage

ui/index.d.ts
```diff
@@ -360,6 +362,128 @@ export declare abstract class AzureWizardStep<T> {
...
+    relatedNameTask?: Promise<string | undefined>;
+}
+
+/**
+ * A generic class for a step that specifies the name of a new resource, used to generate a related name for other new resources.
+ * You must implement `isRelatedNameAvailable` and assign `wizardContext.relatedNameTask` to the result of `generateRelatedName`
+ */
+export declare abstract class AzureNameStep<T extends IRelatedNameWizardContext> extends AzureWizardStep<T> {
+    /**
+     * This method will by called by `generateRelatedName` when trying to find a unique suffix for the related name
```
StephenWeatherford: What was the difference between isRelatedNameAvailable and isNameAvailable?

ui/index.d.ts
nturinski: Fair enough.  We could always intercept the error message and output actual human readable errors too, but not have to write/enforce our own rules.

But anyway, I don't consider this blocking as it's not something that would be hard to change later.

StephenWeatherford: We can't intercept their messages because that would assume English.  This is something unfortunately they've never done but should, but at least they document them.

ui/src/wizard/ResourceGroupStep.ts
EricJizbaMSFT: I didn't know that existed, but again I'm not a huge fan of those error messages. I think they're worded weirdly, but more importantly I hate error messages that include regular expressions. I just don't think regular expressions are easily readable.

For those reasons, I think I'll keep it as-is for this PR. But like you said it would be easy to change later and I'm open to the idea

ui/src/wizard/AzureNameStep.ts
EricJizbaMSFT: Yep this was from your code :)

ui/index.d.ts
EricJizbaMSFT: isNameAvailable is for this resource (for example the naming rules for an App Service when creating an App Service)

isRelatedNameAvailable is used when generating a related name for other resources (aka it would check existing resources groups, app service plans, etc. to see if the generated name is available)

ui/index.d.ts
EricJizbaMSFT: I believe you're thinking of blob containers: https://docs.microsoft.com/en-us/rest/api/storageservices/naming-and-referencing-containers--blobs--and-metadata#container-names

Here's the docs I found for resource groups:
https://docs.microsoft.com/en-us/azure/architecture/best-practices/naming-conventions#naming-rules-and-restrictions

ui/src/wizard/ResourceGroupStep.ts
StephenWeatherford: Perhaps we should complain to the SDK and give them suggestions. I agree they stink.

ui/src/wizard/AzureNameStep.ts
StephenWeatherford: ... Or was this using the pre-fetched ones? If so, never mind.

EricJizbaMSFT: This is generic and its up to implementers on whether its an Azure call or whether is using pre-fetched ones. I can bring it down. How about 100?

StephenWeatherford: Sounds good for now.

ui/index.d.ts
EricJizbaMSFT: I see what you mean. Refactored this so that you call `LocationStep.setLocation` and removed the defaultLocationName property

PR #127, opened 2018-03-20
Minor fix for subscription id

EricJizbaMSFT:
Finally figured out the difference between `subscription.id` and `subscription.subscriptionId`. The first has "/subscriptions/" as a prefix and the second is just the guid. We want the prefix in this case because its used for OpenInPortal

### DIFF
--- a/ui/src/treeDataProvider/AzureTreeDataProvider.ts
+++ b/ui/src/treeDataProvider/AzureTreeDataProvider.ts
@@ -4,11 +4,11 @@
  *--------------------------------------------------------------------------------------------*/
 
 import * as path from 'path';
-import * as vscode from 'vscode';
 import { Disposable, Event, EventEmitter, Extension, extensions, QuickPickOptions, TreeDataProvider, TreeItem, TreeItemCollapsibleState } from 'vscode';
+import * as vscode from 'vscode';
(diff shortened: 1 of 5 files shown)


ui/src/treeDataProvider/AzureTreeDataProvider.ts
```diff
@@ -249,10 +249,10 @@ export class AzureTreeDataProvider implements TreeDataProvider<IAzureNode>, Disp
             nodes = [new AzureNode(undefined, { label: commandLabel, commandId: 'azure-account.selectSubscriptions', contextValue: 'azureCommandNode', id: 'azure-account.selectSubscriptions' })];
```
StephenWeatherford: While you're in this file, could you rename subscriptionInfo -> subscription?

StephenWeatherford: How about some comments, name changes, or property getters to make this clearer to everyone?

EricJizbaMSFT:
@StephenWeatherford changed this to only expose the guid form of the subscription id. I think that's the only one we should be using on a regular basis. Let me know what you think

This also has the side-effect of fewer non-null assertion operators because we can specify the type as `string` instead of `string | undefined`

PR #130, opened 2018-03-26
Use EventEmitter to detect when node is created from picker

nturinski:


ui/src/treeDataProvider/AzureParentNode.ts
```diff
@@ -30,7 +31,7 @@ export class AzureParentNode<T extends IAzureParentTreeItem = IAzureParentTreeIt
         this._cachedChildren = undefined;
     }
 
-    public async createChild(userOptions?: {}): Promise<AzureNode> {
+    public async createChild(userOptions?: {}, onNodeCreateEmitter?: EventEmitter<IAzureNode>): Promise<AzureNode> {
```
EricJizbaMSFT: createChild can be called from outside the npm package, but users should never have access to the event emitter that needs to be passed in. You'll have to refactor this. Not sure what the best way is - but we can discuss if you need ideas

PR #132, opened 2018-03-26
Fix subscription bug

EricJizbaMSFT:
Okay the subscription stuff from https://github.com/Microsoft/vscode-azuretools/pull/127 still isn't working 😅. In my defense, it doesn't break anyone unless they specifically bumped the version within the last hour lol.

Actually did testing and it should be working now :) Here's the error I was getting:
screen shot 2018-03-26 at 3 37 01 pm

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -1,7 +1,7 @@
 {
     "name": "vscode-azureextensionui",
     "author": "Microsoft Corporation",
-    "version": "0.11.0",
+    "version": "0.11.1",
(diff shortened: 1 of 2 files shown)


EricJizbaMSFT:
Honestly its possible we found a typescript bug. I'm guessing the compiler sees the 'subscriptionId' implementation on the base class and assumes `public readonly subscriptionId: string` is redundant (which is why its not included in the .js code). But in this case its not redundant because I need the setter in the constructor

PR #135, opened 2018-03-28
Add runWithTemporaryDescription to IAzureNode

### LINKED ISSUE #104: Show indication of 'deleting' when node is being deleted
This should follow the same format as the 'creating' node

EricJizbaMSFT:
I already had this in the functions extension, but I think it's useful for other extensions as well. There's two differences from that implementation:
1. Rename from 'runWithTemporaryState' to 'runWithTemporaryDescription' since extensions like Cosmos DB don't display state - they display API
1. Show the loading icon in addition to changing the description

Fixes #104

### DIFF
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -66,6 +66,11 @@ export interface IAzureNode<T extends IAzureTreeItem = IAzureTreeItem> {
      * This method combines the environment.portalLink and IAzureTreeItem.id to open the resource in the portal. Optionally, an id can be passed to manually open nodes that may not be in the explorer.
      */
     openInPortal(id?: string): void;
+
+    /**
+     * Displays a 'Loading...' icon and temporarily changes the node's description while `callback` is being run
+     */
+    runWithTemporaryDescription(description: string, callback: () => Promise<void>): Promise<void>;
 }
 
 export interface IAzureParentNode<T extends IAzureTreeItem = IAzureTreeItem> extends IAzureNode<T> {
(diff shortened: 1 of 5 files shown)


ui/index.d.ts
```diff
@@ -66,6 +66,11 @@ export interface IAzureNode<T extends IAzureTreeItem = IAzureTreeItem> {
      * This method combines the environment.portalLink and IAzureTreeItem.id to open the resource in the portal. Optionally, an id can be passed to manually open nodes that may not be in the explorer.
      */
     openInPortal(id?: string): void;
+
+    /**
+     * Displays a 'Loading...' icon and temporarily changes the node's description while `callback` is being run
```
StephenWeatherford: fix comments

ui/src/treeDataProvider/AzureTreeDataProvider.ts
```diff
@@ -79,13 +79,13 @@ export class AzureTreeDataProvider implements TreeDataProvider<IAzureNode>, Disp
         return this._onNodeCreateEmitter.event;
     }
 
-    public getTreeItem(node: IAzureNode): TreeItem {
+    public getTreeItem(node: AzureNode): TreeItem {
```
StephenWeatherford: what's this change for?

ui/index.d.ts
```diff
@@ -66,6 +66,11 @@ export interface IAzureNode<T extends IAzureTreeItem = IAzureTreeItem> {
...
+     */
+    runWithTemporaryDescription(description: string, callback: () => Promise<void>): Promise<void>;
```
StephenWeatherford: nice

ui/src/treeDataProvider/AzureTreeDataProvider.ts
EricJizbaMSFT: I don't think node.iconPath or node.label are worth exposing outside the ui package, so I changed it from `IAzureNode` (external interface) to `AzureNode` (internal class)

ui/index.d.ts
EricJizbaMSFT: Uh anything in particular you want me to fix?

PR #136, opened 2018-03-28
Add 'Report issue' button when displaying errors

### LINKED ISSUE #129: Direct users to Github on error to report issues

EricJizbaMSFT:
Fixes #129

### DIFF
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -192,7 +192,12 @@ export declare class AzureActionHandler {
     registerEvent<T>(eventId: string, event: Event<T>, callback: (this: IActionContext, ...args: any[]) => any): void;
 }
 
-export declare function callWithTelemetryAndErrorHandling(callbackId: string, telemetryReporter: TelemetryReporter | undefined, outputChannel: OutputChannel | undefined, callback: (this: IActionContext) => any): Promise<any>;
+export declare function callWithTelemetryAndErrorHandling(callbackId: string, telemetryReporter: TelemetryReporter | undefined, outputChannel: OutputChannel | undefined, callback: (this: IActionContext) => any, extensionPath?: string): Promise<any>;
+
+/**
+ * Used to open the browser to the "New Issue" page on GitHub with relevant context pre-filled in the issue body
+ */
+export declare function reportIssue(parsedError: IParsedError, extensionPath?: string): void;
 
 export interface IActionContext {
     properties: TelemetryProperties;
@@ -523,4 +528,5 @@ export declare namespace DialogResponses {
     export const upload: MessageItem;
     export const alwaysUpload: MessageItem;
     export const dontUpload: MessageItem;
+    export const reportIssue: MessageItem;
 }
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -192,12 +192,7 @@ export declare class AzureActionHandler {
     registerEvent<T>(eventId: string, event: Event<T>, callback: (this: IActionContext, ...args: any[]) => any): void;
 }
 
-export declare function callWithTelemetryAndErrorHandling(callbackId: string, telemetryReporter: TelemetryReporter | undefined, outputChannel: OutputChannel | undefined, callback: (this: IActionContext) => any, extensionPath?: string): Promise<any>;
-
-/**
- * Used to open the browser to the "New Issue" page on GitHub with relevant context pre-filled in the issue body
- */
-export declare function reportIssue(parsedError: IParsedError, extensionPath?: string): void;
+export declare function callWithTelemetryAndErrorHandling(callbackId: string, telemetryReporter: TelemetryReporter | undefined, outputChannel: OutputChannel | undefined, callback: (this: IActionContext) => any): Promise<any>;
 
 export interface IActionContext {
     properties: TelemetryProperties;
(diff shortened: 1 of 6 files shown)


ui/src/callWithTelemetryAndErrorHandling.ts
```diff
@@ -3,15 +3,17 @@
...
 import { IActionContext } from '../index';
 import { IParsedError } from '../index';
+import { DialogResponses } from './DialogResponses';
 import { localize } from './localize';
 import { parseError } from './parseError';
+import { reportIssue } from './reportIssue';
 
 // tslint:disable-next-line:no-any
-export async function callWithTelemetryAndErrorHandling(callbackId: string, telemetryReporter: TelemetryReporter | undefined, outputChannel: OutputChannel | undefined, callback: (this: IActionContext) => any): Promise<any> {
+export async function callWithTelemetryAndErrorHandling(callbackId: string, telemetryReporter: TelemetryReporter | undefined, outputChannel: OutputChannel | undefined, callback: (this: IActionContext) => any, extensionPath?: string): Promise<any> {
```
StephenWeatherford: See https://github.com/Microsoft/vscode-azurestorage/blob/32c9f8f1f06c6eb66902eb090b95693582cf1601/src/components/telemetry/reporter.ts#L28

ui/src/DialogResponses.ts
```diff
@@ -17,4 +17,5 @@ export namespace DialogResponses {
     export const upload: MessageItem = { title: localize('upload', "Upload") };
     export const alwaysUpload: MessageItem = { title: localize('alwaysUpload', "Always upload") };
     export const dontUpload: MessageItem = { title: localize('dontUpload', "Don't upload"), isCloseAffordance: true };
+    export const reportIssue: MessageItem = { title: localize('reportIssue', "Report issue") };
```
StephenWeatherford: I wonder if "Report an Issue" or "Report as Issue" would be better. This dialog will show for many things which are not bugs, and "Report Issue" to me implies we're saying it's an issue and you should report it.

ui/index.d.ts
```diff
@@ -192,7 +192,12 @@ export declare class AzureActionHandler {
     registerEvent<T>(eventId: string, event: Event<T>, callback: (this: IActionContext, ...args: any[]) => any): void;
 }
 
-export declare function callWithTelemetryAndErrorHandling(callbackId: string, telemetryReporter: TelemetryReporter | undefined, outputChannel: OutputChannel | undefined, callback: (this: IActionContext) => any): Promise<any>;
+export declare function callWithTelemetryAndErrorHandling(callbackId: string, telemetryReporter: TelemetryReporter | undefined, outputChannel: OutputChannel | undefined, callback: (this: IActionContext) => any, extensionPath?: string): Promise<any>;
```
StephenWeatherford: Show-off!  Got to this first!

Very nice, though, should really help.

ui/src/DialogResponses.ts
EricJizbaMSFT: I like "Report an issue"

ui/src/callWithTelemetryAndErrorHandling.ts
EricJizbaMSFT: Anything in particular I'm looking at? Feels like 6 of one, half dozen of another

StephenWeatherford: I.e. you should be able to get extensionPath automatically, right?  Or did I not look at it closely enough?

EricJizbaMSFT: I mean its just a question of if we want to pass in extensionPath or extensionContext. I'm fine with either

StephenWeatherford: Sorry, missed that. I don't want to have to pass in either to every call to callWithTelemetry, although it seems like context would be more at hand if we need to do one, right?

What about calling a registration method in the extension's initialization to set this up once?

EricJizbaMSFT: If you don't want to pass stuff in every time, then I would prefer if this method was a part of AzureActionHandler because we already pass extensionContext, outputChannel, and telemetryReporter to its constructor.

That's how we used to have it and I _believe_ you didn't like that because you had to pass `actionHandler` around all the time. But I was trying a [different thing](https://github.com/Microsoft/vscode-azureeventgrid/blob/master/src/extension.ts#L14) in the eventGrid extension where I make global 'extension' variables for things like the outputChannel so that they don't have to be passed around all the time. We could do the same with an instance of the actionHandler. [The reporter](https://github.com/Microsoft/vscode-azurestorage/blob/32c9f8f1f06c6eb66902eb090b95693582cf1601/src/components/telemetry/reporter.ts#L9) that you linked to works in a similar way (aka just a variable that you import rather than passing it around all the time)

Then it would just be the following:
```typescript
import { actionHandler } from './extension'; // we could export actionHandler from a different file if necessary

function example() {
    actionHandler.callWithTelemetryAndErrorHandling('callbackId', () => {
        ...
    });
}
```

Thoughts? Was there some other reason you didn't like having it in AzureActionHandler?

EricJizbaMSFT: Filed an issue to address this separately: https://github.com/Microsoft/vscode-azuretools/issues/137 In terms of this PR, I got rid of most of my changes in index.d.ts so that these changes are purely internal. That means 'reportAnIssue' won't work perfectly if you use callWithTelemetryAndErrorHandling, but I'm okay with that for now.

I agree with your points above, but I just don't want to work on engineering stuff at this point in the release cycle.

PR #146, opened 2018-04-04
Fix 'isFunctionApp' for linux function apps

EricJizbaMSFT:
Linux shows up in the portal as an option for Function Apps so people are trying it out, but the 'kind' looks like this: "functionapp,linux,container"

screen shot 2018-04-03 at 5 50 46 pm

I'm not saying linux function apps will have great support in our extension, but we should at least make sure they show up in the _correct_ extension


EricJizbaMSFT:
Merging so I can get a build for CTI

PR #147, opened 2018-04-04
Fix 'findNode' for nodes at the same level with similar names

EricJizbaMSFT:
Fixes https://github.com/Microsoft/vscode-cosmosdb/issues/488

### DIFF
--- a/ui/src/treeDataProvider/AzureTreeDataProvider.ts
+++ b/ui/src/treeDataProvider/AzureTreeDataProvider.ts
@@ -171,15 +171,24 @@ export class AzureTreeDataProvider implements TreeDataProvider<IAzureNode>, Disp
     }
 
     public async findNode(id: string): Promise<IAzureNode | undefined> {
-        let children: IAzureNode[] = await this.getChildren();
+        let nodes: IAzureNode[] = await this.getChildren();
 
         // tslint:disable-next-line:no-constant-condition
         while (true) {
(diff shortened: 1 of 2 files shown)


ui/src/treeDataProvider/AzureTreeDataProvider.ts
```diff
@@ -171,15 +171,24 @@ export class AzureTreeDataProvider implements TreeDataProvider<IAzureNode>, Disp
...
-            } else if (node instanceof AzureParentNode) {
-                children = await node.getCachedChildren();
+            let ancestorNode: AzureParentNode | undefined;
+            for (const node of nodes) {
+                if (node.id === id) {
+                    return node;
+                } else if (id.startsWith(`${node.id}/`) && node instanceof AzureParentNode) {
+                    // Append '/' to 'node.id' when checking 'startsWith' to ensure its actually an ancenstor, rather than a node at the same level that _happens_ to start with the same id
+                    // For example, two databases named 'test' and 'test1' as described in this issue: https://github.com/Microsoft/vscode-cosmosdb/issues/488
+                    ancestorNode = node;
```
StephenWeatherford: nit: seems you could get rid of ancestorNode and just do nodes = await... right after the check against starts With.

EricJizbaMSFT: I still need some way to know when I _didn't_ find an ancestor, but you right I can call `nodes = await ...` here. Refactored a bit

PR #156, opened 2018-04-17
 Refactor wizard to support individual use of "List" steps

EricJizbaMSFT:
This is required for https://github.com/Microsoft/vscode-azureeventgrid/issues/3 and https://github.com/Microsoft/vscode-azurefunctions/issues/139

Today, the StorageAccountStep cannot be used by itself because the 'Create' option requires at least a resource group and location step (Storage is an example, but this is a general problem). I refactored so that steps can now specify a subWizard that has all the steps they need. NOTE: This might result in duplicate steps in the wizard, which is why each step must check the wizardContext to see if it's already run

I also refactored into discrete 'prompt' and 'execute' steps so that we can fine-tune the order of steps.


ui/src/wizard/StorageAccountListStep.ts
```diff
@@ -73,10 +73,10 @@ export class StorageAccountListStep<T extends IStorageAccountWizardContext> exte
     }
 
     public async prompt(wizardContext: T, ui: IAzureUserInput): Promise<T> {
-        if (!wizardContext.storageAccount) {
+        if (!wizardContext.storageAccount && !wizardContext.newStorageAccountName) {
             const client: StorageManagementClient = new StorageManagementClient(wizardContext.credentials, wizardContext.subscriptionId);
 
-            const quickPickOptions: IAzureQuickPickOptions = { placeHolder: 'Select a storage account that supports blobs, queues and tables.', id: `StorageAccountListStep/${wizardContext.subscriptionId}` };
+            const quickPickOptions: IAzureQuickPickOptions = { placeHolder: 'Select a storage account.', id: `StorageAccountListStep/${wizardContext.subscriptionId}` };
```
StephenWeatherford: FYI, there are storage accounts that only support blob containers.  So I think whoever put that in originally did have that in mind.

ui/index.d.ts
```diff
@@ -473,8 +476,8 @@ export declare abstract class AzureNameStep<T extends IRelatedNameWizardContext>
 export interface IResourceGroupWizardContext extends ILocationWizardContext, IRelatedNameWizardContext {
     /**
      * The resource group to use for new resources.
-     * If an existing resource group is picked, this value will be defined after `ResourceGroupStep.prompt` occurs
-     * If a new resource group is picked, this value will be defined after `ResourceGroupStep.execute` occurs
+     * If an existing resource group is picked, this value will be defined after `ResourceGroupListStep.prompt` occurs
+     * If a new resource group is picked, this value will be defined after the `execute` phase of the 'create' subwizard
      */
     resourceGroup?: ResourceGroup;
```
StephenWeatherford: Ideally these related entries would be inside an object e.g.:

   resourceGroup: {
      existing?: ResourceGroup;
      newName?: string;
   }

ui/src/wizard/StorageAccountListStep.ts
```diff
@@ -19,57 +22,73 @@ export const storageAccountNamingRules: IAzureNamingRules = {
...
-                    sku: newSku,
-                    kind: 'Storage',
-                    location: newLocation
-                }
-            );
-            outputChannel.appendLine(localize('CreatedStorageAccount', 'Successfully created storage account "{0}".', this._newName));
+                await LocationListStep.setLocation(wizardContext, wizardContext.storageAccount.location!);
+            } else {
+                this.subWizard = new AzureWizard(
+                    [new StorageAccountNameStep(), new ResourceGroupListStep(), new LocationListStep()],
```
StephenWeatherford: Is there a reason it's a subwizard instead of a set of substeps?

ui/src/wizard/AzureWizard.ts
```diff
@@ -20,14 +20,18 @@ export class AzureWizard<T> {
         this._wizardContext = wizardContext;
     }
 
-    public async prompt(actionContext: IActionContext, ui: IAzureUserInput): Promise<T> {
+    public async prompt(actionContext: IActionContext, ui: IAzureUserInput, parentWizard?: AzureWizard<T>): Promise<T> {
         for (const step of this._promptSteps) {
+            if (parentWizard && parentWizard._promptSteps.some((parentStep: AzureWizardPromptStep<T>) => parentStep.constructor.name === step.constructor.name)) {
```
StephenWeatherford: So this is to remove multiple instances of the same step?  Is it possible for this to be intended?  E.g. two location steps, one for a resource group and one for something else?

StephenWeatherford: Might this need to traverse into the parent wizard's parent wizard's steps?

EricJizbaMSFT: Yes that's the point. With the current UI, I think we should encourage the least amount of prompts possible. Also two of the same steps won't work today with how the wizardContext is setup (Since it's a single `wizardContext` object being passed around, the second prompt would just overwrite the value from the first prompt).

I'm fine re-visiting this behavior once VS Code has better wizard support, but I think this is by-design for now.

StephenWeatherford: Make sense.

ui/index.d.ts
EricJizbaMSFT: I agree grouping the values makes more sense, but I also think there's more improvements to be made around the wizardContext. Filed an issue to address separately: https://github.com/Microsoft/vscode-azuretools/issues/159

ui/src/wizard/AzureWizard.ts
EricJizbaMSFT: Fixed

PR #170, opened 2018-05-01
Load More... quick pick for GitHub responses with more than 30 repositories

nturinski:
https://github.com/Microsoft/vscode-azurefunctions/issues/283

### DIFF
--- a/appservice/src/connectToGitHub.ts
+++ b/appservice/src/connectToGitHub.ts
@@ -6,19 +6,26 @@
 import { NameValuePair, SiteSourceControl } from 'azure-arm-website/lib/models';
 import { TokenCredentials, WebResource } from 'ms-rest';
 import * as opn from 'opn';
-import * as requestP from 'request-promise';
+import { Response } from 'request';
+import * as request from 'request-promise';
 import * as vscode from 'vscode';
 import { IAzureNode, IAzureQuickPickItem, IParsedError, parseError, UserCancelledError } from 'vscode-azureextensionui';
 import { localize } from './localize';
 import { signRequest } from './signRequest';
 import { SiteClient } from './SiteClient';
 
-export async function connectToGitHub(node: IAzureNode, client: SiteClient, outputChannel: vscode.OutputChannel): Promise<void> {
-    type gitHubOrgData = { repos_url?: string };
-    type gitHubReposData = { repos_url?: string, url?: string, html_url?: string };
+type gitHubOrgData = { repos_url?: string };
+type gitHubReposData = { repos_url?: string, url?: string, html_url?: string };
+// tslint:disable-next-line:no-reserved-keywords
+type gitHubWebResource = WebResource & { resolveWithFullResponse?: boolean, nextLink?: string, lastLink?: string, type?: string };
 
-    const requestOptions: WebResource = new WebResource();
-    requestOptions.headers = { ['User-Agent']: 'vscode-azureappservice-extension' };
+export async function connectToGitHub(node: IAzureNode, client: SiteClient, outputChannel: vscode.OutputChannel): Promise<void> {
+    const requestOptions: gitHubWebResource = new WebResource();
+    let repoSelected: boolean = false;
+    requestOptions.resolveWithFullResponse = true;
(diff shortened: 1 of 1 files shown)


appservice/src/connectToGitHub.ts
```diff
@@ -110,9 +151,10 @@ async function getJsonRequest(url: string, requestOptions: WebResource, node: IA
  * @param label Property of JSON that will be used as the QuickPicks label
  * @param description Optional property of JSON that will be used as QuickPicks description
  * @param data Optional property of JSON that will be used as QuickPicks data saved as a NameValue pair
+ * @param quickPicks Optional property of QuickPickItems array that will be added to rather than returning a new one (side-effect)
  */
-function createQuickPickFromJsons(jsons: Object[], label: string, description?: string, data?: string[]): IAzureQuickPickItem<{}>[] {
-    const quickPicks: IAzureQuickPickItem<{}>[] = [];
+function createQuickPickFromJsons(jsons: Object[], label: string, description?: string, data?: string[], quickPicks?: IAzureQuickPickItem<{}>[]): IAzureQuickPickItem<{}>[] {
```
EricJizbaMSFT: nit: It feels odd to pass quickPicks here. Can you just append the result of this function to quickPicks outside of this function instead?

appservice/src/connectToGitHub.ts
```diff
@@ -78,14 +111,22 @@ async function showGitHubAuthPrompt(): Promise<void> {
...
     // Note: blank after user implies look up authorized user
     try {
         // tslint:disable-next-line:no-unsafe-any
-        const gitHubResponse: string = await requestP.get(url, <WebResource>requestOptions);
-        return <Object[]>JSON.parse(gitHubResponse);
+        const gitHubResponse: Response = await request(requestOptions).promise();
+        if (gitHubResponse.headers.link) {
+            const link: string = <string>gitHubResponse.headers.link;
+            // regex to find the next and last links in the string
+            const linkUrls: string[] = link.match(/https\:\/\/[a-zA-Z0-9\-\.]+\.[a-zA-Z]{2,3}(\/\S*)?[0-9]/g);
```
EricJizbaMSFT: What exactly does 'link' look like? Why do you have to use this regex? There's no easier way to get this information?

It feels like you're making assumptions about the ordering, etc. and I'm just worried about another issue like the customDomains thing

nturinski: Link is attached to the header of the response if there are additional repos.  This is what the response looks like:

"<https://api.github.com/organizations/6154722/repos?page=2>; rel="prev", <https://api.github.com/organizations/6154722/repos?page=4>; rel="next", <https://api.github.com/organizations/6154722/repos?page=59>; rel="last", <https://api.github.com/organizations/6154722/repos?page=1>; rel="first" "

For whatever reason, it comes as a string so I have to parse it like that to retrieve the urls.  That being said, it looks like you probably made me find a bug because I never paid attention to rel="prev" part before.

EricJizbaMSFT: Yeah I would leverage the next/prev/last things somehow rather than making an assumption about the order. Also if you can be less strict about the url regex I think that might be safer. It feels like the format is always `<url>; rel="<rel>"`. For example - maybe you split on "," and then use a regex like `(.*);[\s]*rel="(.*)"`

nturinski:
Fix for this issue https://github.com/Microsoft/vscode-azurefunctions/issues/283

appservice/src/connectToGitHub.ts
```diff
@@ -110,6 +155,7 @@ async function getJsonRequest(url: string, requestOptions: WebResource, node: IA
  * @param label Property of JSON that will be used as the QuickPicks label
  * @param description Optional property of JSON that will be used as QuickPicks description
  * @param data Optional property of JSON that will be used as QuickPicks data saved as a NameValue pair
+ * @param quickPicks Optional property of QuickPickItems array that will be added to rather than returning a new one (side-effect)
```
EricJizbaMSFT: nit: remove this parameter from the comment

PR #172, opened 2018-05-02
Allow filtering storage accounts

EricJizbaMSFT:
Both the Event Grid and Functions extensions have specific types of storage accounts that are allowed.

Here's what it looks like:
screen shot 2018-05-02 at 3 25 39 pm

### DIFF
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -562,13 +562,13 @@ export declare enum StorageAccountReplication {
     RAGRS = "RAGRS",
 }
 
-export interface IStorageAccountCreateOptions {
+export interface INewStorageAccountDefaults {
     kind: StorageAccountKind;
     performance: StorageAccountPerformance;
     replication: StorageAccountReplication;
 }
 
-export interface IStorageAccountFilterOptions {
+export interface IStorageAccountFilters {
     kind?: StorageAccountKind[];
     performance?: StorageAccountPerformance[];
     replication?: StorageAccountReplication[];
@@ -585,7 +585,7 @@ export declare class StorageAccountListStep<T extends IStorageAccountWizardConte
      * @param createOptions Default options to use when creating a Storage Account
      * @param filterOptions Optional filters used when listing Storage Accounts
      */
-    public constructor(createOptions: IStorageAccountCreateOptions, filterOptions?: IStorageAccountFilterOptions);
+    public constructor(createOptions: INewStorageAccountDefaults, filterOptions?: IStorageAccountFilters);
 
     public static isNameAvailable<T extends IStorageAccountWizardContext>(wizardContext: T, name: string): Promise<boolean>;
 
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -533,41 +533,59 @@ export interface IStorageAccountWizardContext extends IResourceGroupWizardContex
 }
 
 export declare enum StorageAccountKind {
-    Storage,
-    StorageV2,
-    BlobStorage,
+    Storage = "Storage",
+    StorageV2 = "StorageV2",
+    BlobStorage = "BlobStorage",
 }
 
 export declare enum StorageAccountPerformance {
-    Standard,
-    Premium,
+    Standard = "Standard",
+    Premium = "Premium",
 }
 
 export declare enum StorageAccountReplication {
     /**
      * Locally redundant storage
      */
-    LRS,
+    LRS = "LRS",
     /**
      * Zone-redundant storage
      */
-    ZRS,
+    ZRS = "ZRS",
     /**
      * Geo-redundant storage
      */
-    GRS,
+    GRS = "GRS",
     /**
      * Read-access geo-redundant storage
      */
-    RAGRS,
+    RAGRS = "RAGRS",
+}
+
+export interface IStorageAccountCreateOptions {
+    kind: StorageAccountKind;
+    performance: StorageAccountPerformance;
+    replication: StorageAccountReplication;
+}
+
+export interface IStorageAccountFilterOptions {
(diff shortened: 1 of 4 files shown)


ui/index.d.ts
```diff
@@ -533,41 +533,59 @@ export interface IStorageAccountWizardContext extends IResourceGroupWizardContex
...
-    GRS,
+    GRS = "GRS",
     /**
      * Read-access geo-redundant storage
      */
-    RAGRS,
+    RAGRS = "RAGRS",
+}
+
+export interface IStorageAccountCreateOptions {
```
nturinski: I'm not sure how I feel about the name IStorageAccountCreateOptions.  I suppose that you are choosing them, but I feel like they are more like properties?

ui/src/wizard/StorageAccountListStep.ts
```diff
@@ -77,15 +79,21 @@ export class StorageAccountListStep<T extends IStorageAccountWizardContext> exte
             const client: StorageManagementClient = new StorageManagementClient(wizardContext.credentials, wizardContext.subscriptionId);
 
             const quickPickOptions: IAzureQuickPickOptions = { placeHolder: 'Select a storage account.', id: `StorageAccountListStep/${wizardContext.subscriptionId}` };
-            wizardContext.storageAccount = (await ui.showQuickPick(this.getQuickPicks(client.storageAccounts.list()), quickPickOptions)).data;
+            const result: StorageAccount | undefined | string = (await ui.showQuickPick(this.getQuickPicks(client.storageAccounts.list()), quickPickOptions)).data;
+            if (isString(result)) {
```
nturinski: I'm assuming this is to the open the "Learn more..." page, but I feel like the conditional feels very vague.  Maybe a comment would be clarifying?

nturinski: Also, do we necessarily want to close the quick pick menu after opening a new page for the user?  I feel like it'd be more helpful to keep the quick pick list active so that when they navigate back to VS Code, they can resume where they left off rather than having to repeat whatever initial steps.

ui/src/wizard/StorageAccountListStep.ts
```diff
@@ -94,22 +102,48 @@ export class StorageAccountListStep<T extends IStorageAccountWizardContext> exte
         return wizardContext;
     }
 
-    private async getQuickPicks(storageAccountsTask: Promise<StorageAccount[]>): Promise<IAzureQuickPickItem<StorageAccount | undefined>[]> {
-        const picks: IAzureQuickPickItem<StorageAccount | undefined>[] = [{
+    private async getQuickPicks(storageAccountsTask: Promise<StorageAccount[]>): Promise<IAzureQuickPickItem<StorageAccount | undefined | string>[]> {
```
nturinski: nit: Can we list undefined last in type defs?  I dunno why that bothers me 😆

ui/src/wizard/StorageAccountListStep.ts
EricJizbaMSFT: Yeah I thought about doing that, but eventually decided it wasn't worth the effort/code-complexity. I'll fix in a separate PR if you think it's worth it

ui/src/wizard/StorageAccountListStep.ts
```diff
@@ -94,22 +102,48 @@ export class StorageAccountListStep<T extends IStorageAccountWizardContext> exte
...
+        const kindRegExp: RegExp = new RegExp(`^${convertFilterToPattern(this._filterOptions.kind)}$`, 'i');
+        const skuRegExp: RegExp = new RegExp(`^${convertFilterToPattern(this._filterOptions.performance)}_${convertFilterToPattern(this._filterOptions.replication)}$`, 'i');
+
+        let hasFilteredAccounts: boolean = false;
         const storageAccounts: StorageAccount[] = await storageAccountsTask;
-        return picks.concat(storageAccounts.map((sa: StorageAccount) => {
-            return {
+        for (const sa of storageAccounts) {
+            // tslint:disable:strict-boolean-expressions
+            if (!sa.kind || !sa.kind.match(kindRegExp) || !sa.sku || !sa.sku.name.match(skuRegExp)) {
```
nturinski: Isn't this more of a 'select' rather than a 'filter'?  If I'm understanding this correctly, the filterOptions are actually selecting which accounts CAN be used rather than filtering them out?

ui/src/wizard/StorageAccountListStep.ts
nturinski: Definitely not blocking, but could you make a P3 issue on it? I can see this being a minor annoyance.

ui/src/wizard/StorageAccountListStep.ts
EricJizbaMSFT: Mmm good point. I'll keep 'filter' in the name, but flip the logic. I think it makes more sense to specify which accounts _can't_ be used rather than _can_ be used.

ui/index.d.ts
EricJizbaMSFT: No preference on my side. I assume you have the same feedback for 'Filter'? So:
`IStorageAccountCreateProperties` and `IStorageAccountFilterProperties`

ui/src/wizard/StorageAccountListStep.ts
EricJizbaMSFT: https://github.com/Microsoft/vscode-azuretools/issues/176

ui/index.d.ts
EricJizbaMSFT: How about `INewStorageAccountDefaults ` and `IStorageAccountFilters `

PR #185, opened 2018-05-24
add war deploy support

### LINKED ISSUE #184: Add war deploy support in appservice
This is also a part of tasks in: https://github.com/Microsoft/vscode-azureappservice/issues/394

jdneo:
fix #184


jdneo:
@EricJizbaMSFT Please help review.

PR #188, opened 2018-06-01
Implementation for automated azure testing.

nturinski:


ui/src/TestAzureTreeDataProvider.ts
```diff
@@ -0,0 +1,237 @@
+// CTI testing key: du44tQEltLOBZ91c5zAeL1isGNNRFfEX4mAJdh4ViVM=
```
StephenWeatherford: remove

ui/src/TestAzureTreeDataProvider.ts
```diff
@@ -0,0 +1,237 @@
...
+        const secret: string | undefined = process.env.SERVICE_PRINCIPAL_SECRET;
+        const domain: string | undefined = process.env.SERVICE_PRINCIPAL_DOMAIN;
+
+        if (!clientId || !secret || !domain) {
+            throw new Error('Tests can only be run on Travis.');
+        }
+        const credentials: servicePrincipalCredentials = <servicePrincipalCredentials>(await loginWithServicePrincipalSecret(clientId, secret, domain));
+        const subscriptionClient: SubscriptionClient = new SubscriptionClient(credentials);
+        const subscriptions: SubscriptionListResult = await subscriptionClient.subscriptions.list();
+        // returns an array withy subscriptionId, displayName
```
StephenWeatherford: "withy"?  Is that like "truthy" and "falsey"?

ui/index.d.ts
```diff
@@ -32,6 +32,32 @@ export declare class AzureTreeDataProvider implements TreeDataProvider<IAzureNod
     public findNode(id: string): Promise<IAzureNode | undefined>;
     public dispose(): void;
 }
+/**
+ * The TestAzureTreeDataProvider is an implementation of the IAzureTreeDataProvider that automatically signs into
+ * an Azure account via service principals with credentials stored on Travis.  Because signing in is an interactive
+ * experience, this circumvents that to do automated testing with Azure resources.
+ */
+export declare class TestAzureTreeDataProvider implements IAzureTreeDataProvider<IAzureNode>, Disposable {
```
StephenWeatherford: Can you point to code that uses this?

ui/src/TestAzureTreeDataProvider.ts
nturinski: :ups

ui/index.d.ts
nturinski: This class isn't used directly by the azureextensionui package, but is leveraged in azure-functions to automate the login experience for the testing suite.

StephenWeatherford: If you push them to a branch on GitHub, they will no longer be just on your local machine...

nturinski: Oh, actually I had pushed up the branch here:
https://github.com/Microsoft/vscode-azurefunctions/tree/nturinski/funcTest

But I haven't created a PR yet as it is still incomplete.  I'm focused on some features for our milestones, so it'll probably be a couple of weeks before I can get time to finish it.

StephenWeatherford: It would be much better to share the bulk of this code.  I'm thinking having a test version of the azure account extension interface might be the best way to do it.  I just fixed some bugs in this code, for instance, we shouldn't need to make such changes in two places.

ui/src/TestAzureTreeDataProvider.ts
```diff
@@ -0,0 +1,233 @@
...
+    private _customRootNodes: AzureNode[];
+    private _telemetryReporter: TelemetryReporter | undefined;
+
+    private _subscriptionNodes: IAzureNode[] = [];
+    private _disposables: Disposable[] = [];
+
+    constructor(resourceProvider: IChildProvider, loadMoreCommandId: string, ui: IAzureUserInput, telemetryReporter: TelemetryReporter | undefined, rootTreeItems?: IAzureParentTreeItem[]) {
+        this._resourceProvider = resourceProvider;
+        this._loadMoreCommandId = loadMoreCommandId;
+        this._ui = ui;
```
StephenWeatherford: What is IAzureUserInput used for here?  Not sure I like the reassignInputs() method.

nturinski: Today there is no way to add more inputs to a TestUserInput once it is created.  I wanted to implement this as a way to add/alter fake user inputs for tests if I need to dynamically add inputs.  The only other alternative to that today is to create a new TestUserInput object.

ui/index.d.ts
nturinski: I think that this makes way more sense.  I'll try to implement it in that way and resubmit a PR.

PR #189, opened 2018-06-04
suppressTelemetryIfNotFailure shouldn't suppress on failures/cancels

StephenWeatherford:


ui/index.d.ts
```diff
@@ -203,6 +203,13 @@ export declare class AzureActionHandler {
     registerEvent<T>(eventId: string, event: Event<T>, callback: (this: IActionContext, ...args: any[]) => any): void;
 }
 
+/**
+ * Runs the given code. If it succeeds or is canceled, nothing else happens. If it fails, telemetry is sent (with id "UnexpectedError"),
+ * and the error is also sent to the outputChannel if defined.
+ * @param errorContext A string describing the context of the code in case of an error.
+ */
+export declare function callWithUnxpectedErrorTelemetry(errorContext: string, telemetryReporter: TelemetryReporter | undefined, outputChannel: OutputChannel | undefined, callback: (this: IActionContext) => any, extensionContext: ExtensionContext): Promise<any>;
```
EricJizbaMSFT: So this goes outside our established pattern for error telemetry and I think I'd need more explanation to get on board.

For example, one of the biggest ways I prioritize error telemetry is _percentage_ of failed events. If you never report 'succeeded' or 'canceled', then there is no percentage. Even if you look at raw number of failed events (instead of percentage), you still have all of the events lumped into 'UnexpectedError'. We could modify our query to use 'errorContext' instead of the propertyName, but I just don't see the value in changing the norm

StephenWeatherford:
Some of these things run frequently, e.g. on every keystroke. I don't think sending telemetry on every keystroke is reasonable.

also don't want to have to create a new telemetry event id every place where we might hit unexpected exceptions.

EricJizbaMSFT:
I'm about to be OOF so I'll let other people finish reviewing this PR. I will say that if you keep 'callWithUnxpectedErrorTelemetry', at least add the missing 'e' to make it 'callWithUnexpectedErrorTelemetry' haha

StephenWeatherford:
The reasons for sending the PR in this direction were mostly these:
1) I need to be able in some cases to suppress successful telemetry
2) When telemetry is only sent on exceptions, that means the first time an exception is seen, we have to approve the telemetry in Nova, causing a delay of up to 1-2 weeks before actually seeing the data (whereas if there's simply an UnhandledError event, it only needs to be approved once)

We've discussed and agreed that we will solve #1 by changing the current suppressTelemetry to actually send telemetry on errors, and just suppress success.
#2 will only be an issue for those events that are suppressing, which should be few.

So, I'll implement #1 and remove the new function.

StephenWeatherford:
Superseded by #191

PR #191, opened 2018-06-06
suppressTelemetryIfNotFailure shouldn't suppress on failures/cancels

StephenWeatherford:


ui/src/callWithTelemetryAndErrorHandling.ts
```diff
@@ -18,6 +18,7 @@ export async function callWithTelemetryAndErrorHandling(callbackId: string, tele
     const context: IActionContext = {
         properties: {
             isActivationEvent: 'false',
+            cancelStep: '',
```
PrashanthCorp: `cancelStep` has the same value as `error`. Is this intended and easily differentiated at the telemetry side?

StephenWeatherford: It just defaults to blank.  It's already in use in many places, I'm just adding it to the definitions to it's easier to find and encouraged to use when something is canceled.

PR #192, opened 2018-06-07
Service Plan Picker: Re-order tiers and add free tier

burkeholland:
This PR does two things...

1. Reorders tiers so that Basic (B) comes first (which makes more logical sense in the ordering to me - low power to high power - low price to high price) 
2. Add in the Free tier

Number 2 requires that the information about what SKU is supported on what OS is available somewhere. For lack of a better place, I stuck it on the `capabilities` node of each SKU. I'm not sure if that's a safe place to write, but it beats changing the interface or maintaining two lists of skus.

### DIFF
--- a/appservice/src/createAppService/AppServicePlanSkuStep.ts
+++ b/appservice/src/createAppService/AppServicePlanSkuStep.ts
@@ -3,7 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { SkuDescription } from 'azure-arm-website/lib/models';
+import { Capability, SkuDescription } from 'azure-arm-website/lib/models';
 import { AzureWizardPromptStep, IAzureUserInput } from 'vscode-azureextensionui';
 import { IAzureQuickPickItem } from 'vscode-azureextensionui';
 import { localize } from '../localize';
@@ -12,7 +12,7 @@ import { IAppServiceWizardContext } from './IAppServiceWizardContext';
 export class AppServicePlanSkuStep extends AzureWizardPromptStep<IAppServiceWizardContext> {
     public async prompt(wizardContext: IAppServiceWizardContext, ui: IAzureUserInput): Promise<IAppServiceWizardContext> {
         if (!wizardContext.newPlanSku) {
-            const pricingTiers: IAzureQuickPickItem<SkuDescription>[] = this.getPlanSkus().map((s: SkuDescription) => {
+            const pricingTiers: IAzureQuickPickItem<SkuDescription>[] = this.getPlanSkus(wizardContext.newSiteOS).map((s: SkuDescription) => {
                 return {
                     label: s.name,
                     description: s.tier,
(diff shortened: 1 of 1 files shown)


appservice/src/createAppService/AppServicePlanSkuStep.ts
```diff
@@ -26,50 +26,71 @@ export class AppServicePlanSkuStep extends AzureWizardPromptStep<IAppServiceWiza
...
+                capabilities: [{ name: 'supportedOn', value: 'windows' }]
             },
             {
                 name: 'B1',
                 tier: 'Basic',
                 size: 'B1',
                 family: 'B',
-                capacity: 1
+                capacity: 1,
+                capabilities: [{ name: 'supportedOn', value: 'windows,linux' }]
```
StephenWeatherford: Why not use an array of strings for value?

StephenWeatherford: Actually preferably an array of WebsiteOS.

appservice/src/createAppService/AppServicePlanSkuStep.ts
```diff
@@ -26,50 +26,71 @@ export class AppServicePlanSkuStep extends AzureWizardPromptStep<IAppServiceWiza
         return wizardContext;
     }
 
-    private getPlanSkus(): SkuDescription[] {
-        return [
+    private getPlanSkus(os: string): SkuDescription[] {
```
StephenWeatherford: os: WebsiteOS

StephenWeatherford: I think adding to the preexisting capabilities is probably not a good idea. @nturinski is planning some changes in this area, so I'll leave the rest up to him.  Thanks.

burkeholland: Thanks for the update!

nturinski: Hi @burkeholland 

Just wanted to follow-up on what Stephen was saying.

Totally agree with point 1!

We are working on streamlining the web app creation process to default the resource group and the asp name and SKU.  We are defaulting to the B1 plan due to its 30-day trial and ability to work on both platforms.  

Do you mind telling me what the advantages of using F1 would be over B1?  Should we default to F1 for Windows-based apps?

Regarding using capabilities property, I have some reservations about relying on that as we aren't clear what Azure uses those fields for.

Ideally, we'd be able to generate our PlanSKUS dynamically from the Azure REST API, but if we continue to hardcode it, would you be okay with having a conditional statement that pushes the F1 plan into the array of PlanSKUs if the WebsiteOS is Windows rather than relying on the capabilities property?

Thanks!

burkeholland: I think that funneling people through VS Code for tutorials is probably out best developer experience. This means that it will show up a lot in tutorials. That would indicate that people would probably want the free tier. That was my thinking here.

nturinski: @burkeholland-- wanted to keep you in the loop.

With the next release of https://github.com/Microsoft/vscode-azureappservice/, we will have the new webappUp style web-app creation experience.  This will automatically select the F1 plan for users who select Windows and B1 for Linux.

The SKU plans will only be exposed if users opt for the advancedCreation scenario.  I want to expose the F1 tier, but I don't feel confident creating our own setting in the SKU to filter off of.

Would you mind if I close this PR and open a new PR that pushes the F1 tier into the array of SKU plans based on whether or not the user is creating a Windows ASP?

burkeholland: I understand the problems with using the "capabilities" node. That makes sense. That said, I still think that free tiers should be listed in the SKU picker - even if they choose the advanced creation scenario.

I think this is particularly important for our onboarding story. Developers coming to the platform from a tutorial or article should be able to get to the free tier no matter what.

nturinski: 100% agree!  Would you be okay with my proposed solution above as long as the free tier is exposed?

nturinski:
Closing this PR as it is resolved by this https://github.com/Microsoft/vscode-azuretools/pull/226

PR #197, opened 2018-06-12
Nturinski/faster creation

nturinski:
This PR is the bulk of what will be leveraged to solve this issue:
https://github.com/Microsoft/vscode-azureappservice/issues/453

Changes:
1. If the improvedCreation boolean is true, the OS/runtime will automatically be selected.  Otherwise, users will be prompted
1.  If the improvedCreation boolean is true, the Resource Group and App Service Plan names will be auto-generated based on the OS and the location.  This is to emulate the behavior of az webapp up.
1. If the improvedCreation boolean is false, the order of prompts will be changed, but other than that should remain unchanged.
1. If there is only one active Subscription Node, it will automatically be selected rather than prompting the user.

### DIFF
--- a/appservice/src/createAppService/AppServicePlanListStep.ts
+++ b/appservice/src/createAppService/AppServicePlanListStep.ts
@@ -49,6 +49,14 @@ export class AppServicePlanListStep extends AzureWizardPromptStep<IAppServiceWiz
                     wizardContext
                 );
             }
+        } else if (!wizardContext.plan && wizardContext.newPlanName) {
+            if (await AppServicePlanListStep.isNameAvailable(wizardContext, wizardContext.newPlanName, wizardContext.newResourceGroupName)) {
+                this.subWizard = new AzureWizard([], [new AppServicePlanCreateStep()], wizardContext);
+            } else {
+                wizardContext.plan = (await AppServicePlanListStep.getPlans(wizardContext)).find((asp: AppServicePlan) => {
+                    return (asp.name === wizardContext.newPlanName);
+                });
+            }
         }
 
         return wizardContext;
--- a/appservice/src/createAppService/SetWizardContextPropertyStep.ts
+++ b/appservice/src/createAppService/SetWizardContextPropertyStep.ts
@@ -1,0 +1,28 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { AzureWizardPromptStep } from 'vscode-azureextensionui';
+import { IAppServiceWizardContext } from './IAppServiceWizardContext';
+
+export class SetWizardContextPropertyStep extends AzureWizardPromptStep<IAppServiceWizardContext> {
+    public readonly property: string;
+    // tslint:disable-next-line:no-any
+    public readonly value: any;
+    // tslint:disable-next-line:no-any
+    public constructor(property: string, value: any) {
+        super();
+        this.property = property;
+        this.value = value;
+
+    }
+    public async prompt(wizardContext: IAppServiceWizardContext): Promise<IAppServiceWizardContext> {
+        if (typeof this.value === 'function') {
+            wizardContext[this.property] = this.value();
+        } else {
+            wizardContext[this.property] = this.value;
+        }
+        return wizardContext;
+    }
+}
--- a/appservice/src/createAppService/createAppService.ts
+++ b/appservice/src/createAppService/createAppService.ts
@@ -3,13 +3,15 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { Site } from 'azure-arm-website/lib/models';
+import { Site, SkuDescription } from 'azure-arm-website/lib/models';
+import * as find from 'find';
 import { ServiceClientCredentials } from 'ms-rest';
 import { OutputChannel } from 'vscode';
 import { AzureWizard, AzureWizardPromptStep, IActionContext, IAzureUserInput, LocationListStep, ResourceGroupListStep, StorageAccountKind, StorageAccountListStep, StorageAccountPerformance, StorageAccountReplication } from 'vscode-azureextensionui';
 import { AppKind, WebsiteOS } from './AppKind';
 import { AppServicePlanListStep } from './AppServicePlanListStep';
 import { IAppServiceWizardContext } from './IAppServiceWizardContext';
+import { SetWizardContextPropertyStep } from './SetWizardContextPropertyStep';
 import { SiteCreateStep } from './SiteCreateStep';
 import { SiteNameStep } from './SiteNameStep';
 import { SiteOSStep } from './SiteOSStep';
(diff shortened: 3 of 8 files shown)


appservice/package.json
```diff
@@ -31,10 +31,12 @@
         "test": "echo 'Test is not enabled for this package'"
     },
     "dependencies": {
+        "@types/find": "^0.2.0",
         "archiver": "^2.0.3",
         "azure-arm-resource": "^3.0.0-preview",
         "azure-arm-storage": "^3.1.0",
         "azure-arm-website": "^1.0.0-preview",
+        "find": "^0.2.9",
```
StephenWeatherford: Have you e-mailed Matt to add dependence?

appservice/package.json
```diff
@@ -31,10 +31,12 @@
         "test": "echo 'Test is not enabled for this package'"
     },
     "dependencies": {
+        "@types/find": "^0.2.0",
```
StephenWeatherford: This should be under devDependencies

appservice/src/createAppService/AppServicePlanListStep.ts
```diff
@@ -49,6 +49,14 @@ export class AppServicePlanListStep extends AzureWizardPromptStep<IAppServiceWiz
                     wizardContext
                 );
             }
+        } else if (!wizardContext.plan && wizardContext.newPlanName) {
```
StephenWeatherford: Add comments - not sure what the purpose is here

appservice/src/createAppService/SetWizardContextPropertyStep.ts
```diff
@@ -0,0 +1,28 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { AzureWizardPromptStep } from 'vscode-azureextensionui';
+import { IAppServiceWizardContext } from './IAppServiceWizardContext';
+
+export class SetWizardContextPropertyStep extends AzureWizardPromptStep<IAppServiceWizardContext> {
```
StephenWeatherford: Even less clue what this is about

appservice/src/createAppService/createAppService.ts
```diff
@@ -24,13 +26,67 @@ export async function createAppService(
     credentials: ServiceClientCredentials,
     subscriptionId: string,
     subscriptionDisplayName: string,
-    showCreatingNode?: (label: string) => void): Promise<Site> {
+    showCreatingNode?: (label: string) => void,
+    fsPath?: string,
+    improvedCreation: boolean = false): Promise<Site> {
```
StephenWeatherford: How about streamlinedSteps?  Also could use a comment.

appservice/src/createAppService/createAppService.ts
```diff
@@ -24,13 +26,67 @@ export async function createAppService(
...
+    };
 
     const promptSteps: AzureWizardPromptStep<IAppServiceWizardContext>[] = [];
     promptSteps.push(new SiteNameStep());
-    promptSteps.push(new ResourceGroupListStep());
+    promptSteps.push(new LocationListStep());
     promptSteps.push(new SiteOSStep());
     promptSteps.push(new SiteRuntimeStep());
+    if (improvedCreation) {
+        // function used to set name property after location and OS has been selected
```
StephenWeatherford: nit: "have" been selected

appservice/src/createAppService/createAppService.ts
```diff
@@ -24,13 +26,67 @@ export async function createAppService(
...
+        const setResourceGroupProperty: Function = (): string => {
```
StephenWeatherford: TODO

StephenWeatherford: TODO

appservice/src/createAppService/createAppService.ts
```diff
@@ -79,3 +128,20 @@ export async function createAppService(
 
     return wizardContext.site;
 }
+
+function setDefaultsForImprovedConfiguration(wizardContext: IAppServiceWizardContext, fsPath: string): void {
```
StephenWeatherford: ...ForStreamlinedConfiguration

appservice/src/createAppService/createAppService.ts
```diff
@@ -79,3 +128,20 @@ export async function createAppService(
...
+    const files: string[] = find.fileSync(fsPath);
+    for (const file of files) {
+        if (file.indexOf('package.json') >= 0) {
```
StephenWeatherford: Should always use platform built-in functions for file manipulation.  Too many edge cases.

var path = require('path');
path.name('index.html'); // I think (?)

appservice/src/createAppService/createAppService.ts
```diff
@@ -79,3 +128,20 @@ export async function createAppService(
...
+            wizardContext.newSiteOS = WebsiteOS.linux;
+            wizardContext.newSiteRuntime = 'node|8.9';
+            break;
+        }
+
+        if (file.split('.').pop() === 'csproj') {
```
StephenWeatherford: path.extname // I think

appservice/src/createAppService/createAppService.ts
StephenWeatherford: Also that would keep you from matching mypackage.json and package.json.txt etc.

ui/src/treeDataProvider/AzureTreeDataProvider.ts
```diff
@@ -224,7 +224,7 @@ export class AzureTreeDataProvider implements TreeDataProvider<IAzureNode>, Disp
             .map((n: AzureNode) => { return { data: n, description: '', label: n.treeItem.label }; }));
 
         const options: QuickPickOptions = { placeHolder: localize('selectSubscription', 'Select a Subscription') };
-        const result: AzureNode | string = (await this._ui.showQuickPick(picks, options)).data;
+        const result: AzureNode | string = picks.length === 1 ? picks[0].data : (await this._ui.showQuickPick(picks, options)).data;
```
StephenWeatherford: Shouldn't this be handled in the ui package?

appservice/src/createAppService/AppServicePlanListStep.ts
```diff
@@ -49,6 +49,14 @@ export class AppServicePlanListStep extends AzureWizardPromptStep<IAppServiceWiz
...
+            if (await AppServicePlanListStep.isNameAvailable(wizardContext, wizardContext.newPlanName, wizardContext.newResourceGroupName)) {
+                this.subWizard = new AzureWizard([], [new AppServicePlanCreateStep()], wizardContext);
+            } else {
+                wizardContext.plan = (await AppServicePlanListStep.getPlans(wizardContext)).find((asp: AppServicePlan) => {
```
StephenWeatherford: nit:

.find((asp: AppServicePlan) => asp.name === wizardContext.newPlanName);

ui/src/wizard/ResourceGroupListStep.ts
```diff
@@ -48,6 +48,14 @@ export class ResourceGroupListStep<T extends IResourceGroupWizardContext> extend
                     wizardContext
                 );
             }
+        } else if (!wizardContext.resourceGroup && wizardContext.newResourceGroupName) {
+            if (await ResourceGroupListStep.isNameAvailable(wizardContext, wizardContext.newResourceGroupName)) {
+                this.subWizard = new AzureWizard([], [new ResourceGroupCreateStep()], wizardContext);
+            } else {
+                wizardContext.resourceGroup = (await ResourceGroupListStep.getResourceGroups(wizardContext)).find((rg: ResourceGroup) => {
```
StephenWeatherford: ditto about find

appservice/src/createAppService/createAppService.ts
nturinski: The functions below are passed into the SetWizardPropertySteps to be executed later to set the rg and asp names which this comment is referring to.  I rewrote it to be more clarifying.

ui/src/treeDataProvider/AzureTreeDataProvider.ts
nturinski: I'm not sure where I could handle this in the ui package.  The SubscriptionNode file itself and the azureaccounts-extension retrieves all of the subscriptions.

I believe that the AzureDataTreeProvider is the only place where we can control the subscriptionNode pick.

appservice/src/createAppService/SetWizardContextPropertyStep.ts
```diff
@@ -0,0 +1,38 @@
...
+
+import { AzureWizardPromptStep } from 'vscode-azureextensionui';
+import { IAppServiceWizardContext } from './IAppServiceWizardContext';
+
+/**
+ * This step is used to alter the WizardContext's properties without having to recreate the wizard or
+ * stop the prompts.  This is neccessary when the value of a property relies on previously prompted
+ * steps i.e. the resource group's name relies on the app location so a location must be selected first
+ */
+export class SetWizardContextPropertyStep extends AzureWizardPromptStep<IAppServiceWizardContext> {
```
EricJizbaMSFT: Yeah I know you added comments to this class to make it less confusing, but I still don't get why this class is necessary. Maybe you can explain it to me offline and we can brainstorm ways to possibly get rid of it

appservice/src/createAppService/createAppService.ts
```diff
@@ -24,13 +27,67 @@ export async function createAppService(
     credentials: ServiceClientCredentials,
     subscriptionId: string,
     subscriptionDisplayName: string,
-    showCreatingNode?: (label: string) => void): Promise<Site> {
+    showCreatingNode?: (label: string) => void,
+    fsPath?: string,
+    streamlinedSteps: boolean = false /* enables streamlined app creation that emulates az webapp up's behavior */): Promise<Site> {
```
EricJizbaMSFT: We want 'streamlined' to be the default user experience now, right? If so I would flip all of the naming to match that. For example in this case you would having something like 'showAllSteps' or 'ignoreDefaults' or 'beReallyAnnoying' idk. Honestly this should probably match whatever user setting we decide on that lets users 'not' streamline stuff

ui/src/treeDataProvider/AzureTreeDataProvider.ts
nturinski: So you're okay with this behavior here?

StephenWeatherford: Ah...   didn't notice.  Good.

appservice/src/createAppService/createAppService.ts
```diff
@@ -79,3 +129,20 @@ export async function createAppService(
 
     return wizardContext.site;
 }
+
+function setDefaultsForStreamlinedConfiguration(wizardContext: IAppServiceWizardContext, fsPath: string): void {
+    const files: string[] = find.fileSync(fsPath);
```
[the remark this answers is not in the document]
nturinski: Yeah, I guess looking at the root directory would be enough to search for project files

appservice/src/createAppService/createAppService.ts
```diff
@@ -79,3 +129,20 @@ export async function createAppService(
...
+    for (const file of files) {
+        if (path.basename(file) === 'package.json') {
+            wizardContext.newSiteOS = WebsiteOS.linux;
+            wizardContext.newSiteRuntime = 'node|8.9';
+            break;
+        }
+
+        if (path.extname(file) === 'csproj') {
```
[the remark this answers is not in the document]
nturinski: Is there a concept of runtimes in Functions?  I thought it default to windows OS.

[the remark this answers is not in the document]
EricJizbaMSFT: Yeah I don't think the 'runtimes' matters as much - just the OS. And yes right now it defaults to Windows, but eventually we might have more people on Linux

Issue #111, opened 2018-03-07
Turn 'strictNullChecks' back on for appservice package


EricJizbaMSFT:
Not sure when this was turned off. It just says "TODO: Fix issues in copied code and turn back on" in tsconfig.json

EricJizbaMSFT:
Fixed in https://github.com/Microsoft/vscode-azuretools/pull/274

PR #202, opened 2018-06-21
Streamline web app creation

nturinski:
Refactored the streamlined creation with PR fixes from
https://github.com/Microsoft/vscode-azuretools/pull/197

Relies on #203

### DIFF
--- a/appservice/src/createAppService/createAppService.ts
+++ b/appservice/src/createAppService/createAppService.ts
@@ -3,11 +3,12 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { Site } from 'azure-arm-website/lib/models';
+import { Site, SkuDescription } from 'azure-arm-website/lib/models';
 import { ServiceClientCredentials } from 'ms-rest';
-import { OutputChannel } from 'vscode';
-import { AzureWizard, AzureWizardPromptStep, IActionContext, IAzureUserInput, LocationListStep, ResourceGroupListStep, StorageAccountKind, StorageAccountListStep, StorageAccountPerformance, StorageAccountReplication } from 'vscode-azureextensionui';
+import { OutputChannel, Uri, workspace } from 'vscode';
+import { AzureWizard, AzureWizardExecuteStep, AzureWizardPromptStep, IActionContext, IAzureUserInput, LocationListStep, ResourceGroupCreateStep, ResourceGroupListStep, StorageAccountKind, StorageAccountListStep, StorageAccountPerformance, StorageAccountReplication } from 'vscode-azureextensionui';
 import { AppKind, WebsiteOS } from './AppKind';
+import { AppServicePlanCreateStep } from './AppServicePlanCreateStep';
 import { AppServicePlanListStep } from './AppServicePlanListStep';
 import { IAppServiceWizardContext } from './IAppServiceWizardContext';
 import { SiteCreateStep } from './SiteCreateStep';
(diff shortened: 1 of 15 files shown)


nturinski:
The build will break because of the ui package (which might be a good reason to make two separate PRs).  I'm guessing lint is complaining because line 85 is 
`executeSteps.push(new ResourceGroupCreateStep());`

which wouldn't be a type until the index files in the ui package are updated.

appservice/src/createAppService/createAppService.ts
```diff
@@ -52,21 +64,31 @@ export async function createAppService(
...
             break;
         case AppKind.app:
+            if (advancedCreation) {
+                promptSteps.push(new ResourceGroupListStep());
+                promptSteps.push(new SiteOSStep());
+                promptSteps.push(new SiteRuntimeStep());
+                promptSteps.push(new AppServicePlanListStep());
+                promptSteps.push(new LocationListStep());
+            } else {
+                if (workspace.workspaceFolders.length === 1) {
```
EricJizbaMSFT: strictNullChecks is turned off in the appservice package. Until we fix it (#111) you need to be extra careful. For example, I think workspaceFolders can be undefined

appservice/src/createAppService/createAppService.ts
```diff
@@ -75,7 +97,30 @@ export async function createAppService(
...
+        wizardContext.newPlanSku = wizardContext.newSiteOS === WebsiteOS.windows ? freePlanSku : basicPlanSku;
+    }
     wizardContext = await wizard.execute(actionContext, outputChannel);
 
     return wizardContext.site;
 }
+
+async function setWizardContextDefaults(wizardContext: IAppServiceWizardContext): Promise<void> {
+    await workspace.findFiles('package.json').then((files: Uri[]) => {
+        if (files.length > 0) {
```
EricJizbaMSFT: Is 'files' always defined?

appservice/src/createAppService/createAppService.ts
```diff
@@ -52,21 +64,31 @@ export async function createAppService(
                     learnMoreLink: 'https://aka.ms/Cfqnrc'
                 }
             ));
+            promptSteps.push(new AppServicePlanListStep());
```
EricJizbaMSFT: Function Apps don't need `AppServicePlanListStep`

appservice/src/createAppService/createAppService.ts
```diff
@@ -52,21 +64,31 @@ export async function createAppService(
...
+                promptSteps.push(new SiteRuntimeStep());
+                promptSteps.push(new AppServicePlanListStep());
+                promptSteps.push(new LocationListStep());
+            } else {
+                if (workspace.workspaceFolders.length === 1) {
+                    // can make smart defaults if only one workspace is opened
+                    await setWizardContextDefaults(wizardContext);
+                }
+                promptSteps.push(new LocationListStep());
+                promptSteps.push(new SiteOSStep()); // will be skipped with a smart default
```
EricJizbaMSFT: I'm confused by this comment. Shouldn't it be "will be skipped _if there is_ a smart default"? I assume these steps are only included just in case you can't figure out the defaults...

appservice/src/createAppService/createAppService.ts
```diff
@@ -75,7 +97,30 @@ export async function createAppService(
...
+        wizardContext.newPlanName = `appsvc_asp_${wizardContext.newSiteOS}_${wizardContext.location.name}`;
+        // Free tier is only available for Windows
+        wizardContext.newPlanSku = wizardContext.newSiteOS === WebsiteOS.windows ? freePlanSku : basicPlanSku;
+    }
     wizardContext = await wizard.execute(actionContext, outputChannel);
 
     return wizardContext.site;
 }
+
+async function setWizardContextDefaults(wizardContext: IAppServiceWizardContext): Promise<void> {
```
EricJizbaMSFT: The name of this function makes me think it's being used for App Services _and_ Function Apps. I suggest you just move this code up into createAppService so that the flow is more obvious.

I suppose you could also rename this function to make it more clear (that it's only used for Web Apps), but I couldn't come up with a short enough name that made sense.

appservice/src/createAppService/createAppService.ts
```diff
@@ -24,15 +25,26 @@ export async function createAppService(
     credentials: ServiceClientCredentials,
     subscriptionId: string,
     subscriptionDisplayName: string,
-    showCreatingNode?: (label: string) => void): Promise<Site> {
+    showCreatingNode?: (label: string) => void,
+    advancedCreation: boolean = false): Promise<Site> {
```
EricJizbaMSFT: I don't see you ever passing 'advancedCreation' as true?

appservice/src/createAppService/createAppService.ts
nturinski: Correct

appservice/src/createAppService/createAppService.ts
nturinski: advancedCreation will be passed as true in the vscode-azureappservice extension.

appservice/src/createAppService/createAppService.ts
nturinski: Yeah, if findFIles doesn't return any files, then it will return an empty array,

appservice/src/createAppService/createAppService.ts
```diff
@@ -52,21 +65,30 @@ export async function createAppService(
...
+                promptSteps.push(new SiteRuntimeStep());
+                promptSteps.push(new AppServicePlanListStep());
+                promptSteps.push(new LocationListStep());
+            } else {
+                if (workspace.workspaceFolders && workspace.workspaceFolders.length === 1) {
+                    // can make smart defaults if only one workspace is opened
+                    await setWizardContextDefaults(wizardContext);
+                }
+                promptSteps.push(new LocationListStep());
+                promptSteps.push(new SiteOSStep()); // will be skipped with if there is a smart default
```
EricJizbaMSFT: nit: 'with if' -> 'if'

PR #208, opened 2018-06-29
Register common variables rather than passing them in all the time

### LINKED ISSUE #137: Refactor callWithTelemetryAndErrorHandling so that you only have to pass in callbackId and callback
We should do some sort of 'preregistration' of the telemetryReporter, outputChannel, and extensionContext rather than forcing people to pass them in every time

EricJizbaMSFT:
Before I did the work for the user-agent (which will require extensionContext to be passed a bunch of places to get the package version), I decided to fix #137. This allows us to use common variables without having to pass them in everywhere.

I implemented this all the way through for the functions extension to make sure it would work with tests, etc. Assuming you approve, I'll start fixing in the other extensions as well.

### DIFF
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -30,7 +30,7 @@ export declare class AzureTreeDataProvider implements TreeDataProvider<IAzureNod
      * @param telemetryReporter Optionally used to track telemetry for the tree
      * @param rootTreeItems Any nodes other than the subscriptions that should be shown at the root of the explorer
      */
-    constructor(resourceProvider: IChildProvider, loadMoreCommandId: string, ui: IAzureUserInput, telemetryReporter: TelemetryReporter | undefined, rootTreeItems?: IAzureParentTreeItem[]);
+    constructor(resourceProvider: IChildProvider, loadMoreCommandId: string, rootTreeItems?: IAzureParentTreeItem[]);
     public getTreeItem(node: IAzureNode): TreeItem;
     public getChildren(node?: IAzureParentNode): Promise<IAzureNode[]>;
     public refresh(node?: IAzureNode, clearCache?: boolean): Promise<void>;
@@ -58,7 +58,6 @@ export interface IAzureNode<T extends IAzureTreeItem = IAzureTreeItem> {
     readonly tenantId: string;
     readonly userId: string;
     readonly environment: AzureEnvironment;
-    readonly ui: IAzureUserInput;
 
     /**
      * Refresh this node in the tree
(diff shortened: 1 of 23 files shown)


nturinski:
This is replacing extensionVariables in some of our extensions then?  I'm pro this.

EricJizbaMSFT:
Okay I've sent out all PRs to give a full example using the functions extension: https://github.com/Microsoft/vscode-azurefunctions/pull/434

I would say this _leverages_ the extensionVariables, rather than _replaces_ the extensionVariables. I wanted to keep the ability for each extension/package to have different variables that they use. That's why 'ext' is not exposed from the ui package. You can register the exact same 'ext', but the ui package will only use the 4 variables that it needs from within the package.

ui/index.d.ts
```diff
@@ -620,3 +615,18 @@ export declare namespace DialogResponses {
...
+ */
+export declare function registerUIExtensionVariables(extVars: UIExtensionVariables): void;
+
+/**
+ * Interface for common extension variables used throughout the UI package.
+ */
+export interface UIExtensionVariables {
+    context: ExtensionContext;
+    outputChannel: OutputChannel;
+    ui: IAzureUserInput;
```
StephenWeatherford: Tests will need to change the IAzureUserInput, right?  Or do they refill it with data?

ui/src/AzureActionHandler.ts
```diff
@@ -3,54 +3,37 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { commands, Event, ExtensionContext, OutputChannel } from 'vscode';
-import TelemetryReporter from 'vscode-extension-telemetry';
+import { commands, Event } from 'vscode';
 import { IActionContext } from '../index';
 import { callWithTelemetryAndErrorHandling } from './callWithTelemetryAndErrorHandling';
+import { ext } from './extensionVariables';
```
StephenWeatherford: This does seem a bit confusing with extension vars, is there a better name?

ui/index.d.ts
StephenWeatherford: Oh, you're not making a copy, you're keeping a reference...

ui/src/AzureActionHandler.ts
StephenWeatherford: Never mind, makes more sense now.

ui/index.d.ts
EricJizbaMSFT: Yep it's easy for tests to change it since it's a reference

PR #210, opened 2018-06-29
Add common function for user agent

EricJizbaMSFT:
I realize the line is a little vague here in terms of "Is this UI related" for the "ui" package, but it shares some code with 'report an issue' and every azure extension needs it. I'm just saying I'm okay with it being here, but let me know if you really care otherwise.

### DIFF
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -4,7 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { Location } from 'azure-arm-resource/lib/subscription/models';
-import { ServiceClientCredentials } from 'ms-rest';
+import { ServiceClientCredentials, ServiceClient } from 'ms-rest';
 import { AzureEnvironment } from 'ms-rest-azure';
 import { Uri, TreeDataProvider, Disposable, TreeItem, Event, OutputChannel, Memento, InputBoxOptions, QuickPickItem, QuickPickOptions, TextDocument, ExtensionContext, MessageItem, OpenDialogOptions, MessageOptions } from 'vscode';
 import TelemetryReporter from 'vscode-extension-telemetry';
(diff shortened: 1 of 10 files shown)


ui/index.d.ts
```diff
@@ -630,3 +630,8 @@ export interface UIExtensionVariables {
     ui: IAzureUserInput;
     reporter: TelemetryReporter | undefined;
 }
+
+/**
+ * Adds a user agent specific to the VS Code extension, of the form `${extensionName}/${extensionVersion}`
+ */
+export declare function addExtensionUserAgentInfo(client: ServiceClient): void;
```
StephenWeatherford: nit: I'd say the "Info" part is unnecessary

ui/src/getPackageInfo.ts
```diff
@@ -0,0 +1,27 @@
...
+
+    // tslint:disable-next-line:strict-boolean-expressions
+    const extensionName: string = (packageJson && packageJson.name) || 'vscode-azuretools';
+    // tslint:disable-next-line:strict-boolean-expressions
+    const extensionVersion: string = (packageJson && packageJson.version) || 'Unknown';
+    return [extensionName, extensionVersion];
+}
+
+interface IPackageJson {
+    version?: string;
```
StephenWeatherford: Why are these fields optional?

ui/src/getPackageInfo.ts
```diff
@@ -0,0 +1,27 @@
...
+
+import { ext } from "./extensionVariables";
+
+export function getPackageInfo(): [string, string] {
+    let packageJson: IPackageJson | undefined;
+    try {
+        // tslint:disable-next-line:non-literal-require
+        packageJson = <IPackageJson>require(ext.context.asAbsolutePath('package.json'));
+    } catch (error) {
+        // ignore errors
```
StephenWeatherford: telemetry on error?

ui/index.d.ts
EricJizbaMSFT: Fine with me. I was just being consistent with `client.addUserAgentInfo`

PR #211, opened 2018-06-29
Nturinski/runtime step

nturinski:
As discussed offline, we are not going to auto-detect any runtimes yet.  If we find that an overwhelming amount of users are creating node projects, we may revisit this.

Also put node 8.9 at the top of the runtime quickpick.


EricJizbaMSFT:
@nturinski the other part of this change is we need to make sure we have telemetry on which runtime they're selecting. Do we already have that? If not, please fix or file an issue. Not sure if that would be in the shared package or the extension itself

nturinski:
I think the azuretools has been updated to include telemetry from the wizard.  I don't see many data points with it, but I think I can assume it'll be that way moving forward?

PR #212, opened 2018-07-02
Add appendExtensionUserAgent 

StephenWeatherford:
Can't use addExtensionUserAgent as is for Cosmos' DocumentClient


ui/index.d.ts
```diff
@@ -631,7 +631,17 @@ export interface UIExtensionVariables {
...
+ * and appends it to the given user agent string, if it isn't already in the string.
+ */
+export declare function appendExtensionUserAgent(userAgent: string | undefined): string;
+
 /**
- * Adds a user agent specific to the VS Code extension, of the form `${extensionName}/${extensionVersion}`
+ * Adds the extension user agent to the given ServiceClient or other object support AddUserAgentInfo
  */
-export declare function addExtensionUserAgent(client: ServiceClient): void;
+export declare function addExtensionUserAgent(client: IAddUserAgent): void;
```
nturinski: The function definition doesn't match with extensionUserAgent.ts anymore.  Will that matter?

StephenWeatherford: Fixed, thanks.

nturinski:
I would want @EricJizbaMSFT to take a look at this since I don't really know exactly how he intended this code to be used.  Is the DocumentClient from 'documentdb' part of the Azure REST API like the WebsiteManagementClient or is it operated by somebody else completely?

ui/test/extensionUserAgent.test.ts
```diff
@@ -0,0 +1,35 @@
+// tslint:disable:no-useless-files
+
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+ // tslint:disable:no-any max-func-body-length
+
+// Depends on vscode - need to figure out how to enable
```
EricJizbaMSFT: You probably just need to call `registerUIExtensionVariables` with a fake extensionContext (I'm guessing that asAbsolutePath is failing [here](https://github.com/Microsoft/vscode-azuretools/blob/master/ui/src/getPackageInfo.ts#L18). I believe you already figured it out, but these tests are just run using mocha, not VS Code's test runner. It's because they're an npm pacakge and not a VS Code extension and I don't think we'll easily be able to get around that.

StephenWeatherford: Actually, I'm getting an error that the vscode module can't be found (whenever tries to import 'vscode', which callWithTelemetryAndErrorHandling.ts does).  Haven't tried very hard to figure this out, may have to do some stubbing/etc.

PR #214, opened 2018-07-02
Nturinski/no show output

nturinski:
Remove outputChannel.show() calls and add long-running notifications for a lot of actions. 
 Part of the effort to resolve this issue: 
https://github.com/Microsoft/vscode-azureappservice/issues/455

There are changes in both appservice and ui.

I thought it was okay to remove the ugly `>>>>>> <<<<<<` output messages since we shouldn't need to draw attention to it anymore.


appservice/src/connectToGitHub.ts
```diff
@@ -88,7 +88,14 @@ export async function connectToGitHub(node: IAzureNode, client: SiteClient): Pro
     ext.outputChannel.show(true);
```
EricJizbaMSFT: Remove outputChannel.show?

ui/src/wizard/AzureWizard.ts
```diff
@@ -42,8 +41,6 @@ export class AzureWizard<T> {
     }
 
     public async execute(actionContext: IActionContext): Promise<T> {
-        ext.outputChannel.show(true);
```
EricJizbaMSFT: Interesting - so this is going to affect every single wizard. Do we have a plan for addressing that? I'm most concerned about Cosmos since that'll probably ship the soonest. Might be easy enough to fix the one or two wizards

EricJizbaMSFT: I think it's fine. The wizard isn't used all that often anyways. Just submit a PR to Cosmos to call outputChannel.show

PR #217, opened 2018-07-05
Fix "Cannot read property 'concat' of undefined"

### LINKED ISSUE #175: Collapsing and expanding a node while it is refreshing causes duplicate nodes
Repro steps:
1.) Refresh a node on the explorer
2.) While it is refreshing, collapse and expand a node
3.) ???
4.) Profit

EricJizbaMSFT:
Currently there's a race condition where `_cachedChildren` can be undefined when we're trying to call `concat` in loadMoreChildren, which leads to hard-to-reproduce bugs. This should fix it in the following two ways:
1. Change the type of `_cachedChildren` so that it can't be undefined
1. Restrict `loadMoreChildren` to only run once at a time

Fixes #175 
Fixes https://github.com/Microsoft/vscode-cosmosdb/issues/745
Fixes https://github.com/Microsoft/vscode-azurefunctions/issues/333
FIxes https://github.com/Microsoft/vscode-azurefunctions/issues/158

NOTE: The build will fail because of tslint's 'member-ordering' rule, but I left it how it is for the sake of review.

### DIFF
--- a/ui/src/treeDataProvider/AzureParentNode.ts
+++ b/ui/src/treeDataProvider/AzureParentNode.ts
@@ -13,20 +13,20 @@ import { CreatingTreeItem } from './CreatingTreeItem';
 import { LoadMoreTreeItem } from './LoadMoreTreeItem';
 
 export class AzureParentNode<T extends IAzureParentTreeItem = IAzureParentTreeItem> extends AzureNode<T> implements IAzureParentNodeInternal {
-    private _cachedChildren: AzureNode[] | undefined;
+    private _cachedChildren: AzureNode[] = [];
     private _creatingNodes: AzureNode[] = [];
     private _onNodeCreateEmitter: EventEmitter<IAzureNode>;
-
+    private _clearCache: boolean = true;
+    private _loadMoreChildrenTask: Promise<void> | undefined;
     private _initChildrenTask: Promise<void> | undefined;
 
     public constructor(parent: IAzureParentNodeInternal | undefined, treeItem: T, onNodeCreateEmitter: EventEmitter<IAzureNode>) {
         super(parent, treeItem);
         this._onNodeCreateEmitter = onNodeCreateEmitter;
-
     }
 
     public async getCachedChildren(): Promise<AzureNode[]> {
-        if (this._cachedChildren === undefined) {
+        if (this._clearCache) {
             this._initChildrenTask = this.loadMoreChildren();
         }
 
@@ -34,7 +34,7 @@ export class AzureParentNode<T extends IAzureParentTreeItem = IAzureParentTreeIt
             await this._initChildrenTask;
         }
 
-        return this._cachedChildren ? this._cachedChildren : [];
+        return this._cachedChildren;
     }
 
     public get creatingNodes(): AzureNode[] {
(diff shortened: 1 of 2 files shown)


ui/src/treeDataProvider/AzureParentNode.ts
```diff
@@ -75,21 +75,33 @@ export class AzureParentNode<T extends IAzureParentTreeItem = IAzureParentTreeIt
     }
 
     public async loadMoreChildren(): Promise<void> {
-        let clearCache: boolean = false;
-        if (this._cachedChildren === undefined) {
+        if (this._loadMoreChildrenTask) {
+            await this._loadMoreChildrenTask;
```
PrashanthCorp: 1. On a first read, it looks like line 84's `finally` obviates this branch. 
Is this a guard against the race condition, where `loadMoreChildren` is called again before the `finally` is executed? 

2. line 83 (identical) has error catching while this line does not. Can you fill me in on what I'm missing here?

ui/src/treeDataProvider/AzureParentNode.ts
```diff
@@ -116,28 +128,24 @@ export class AzureParentNode<T extends IAzureParentTreeItem = IAzureParentTreeIt
...
-        if (this._cachedChildren) {
-            const index: number = this._cachedChildren.indexOf(node);
-            if (index !== -1) {
-                this._cachedChildren.splice(index, 1);
-                await this.treeDataProvider.refresh(this, false);
-            }
+        const index: number = this._cachedChildren.indexOf(node);
+        if (index !== -1) {
+            this._cachedChildren.splice(index, 1);
+            await this.treeDataProvider.refresh(this, false);
```
PrashanthCorp: Is this code mainly because the Set data structure is not sorted by default? (TIL it stores values in insertion order :P)

ui/src/treeDataProvider/AzureParentNode.ts
```diff
@@ -75,21 +75,33 @@ export class AzureParentNode<T extends IAzureParentTreeItem = IAzureParentTreeIt
...
 
     public async loadMoreChildren(): Promise<void> {
-        let clearCache: boolean = false;
-        if (this._cachedChildren === undefined) {
+        if (this._loadMoreChildrenTask) {
+            await this._loadMoreChildrenTask;
+        } else {
+            this._loadMoreChildrenTask = this.loadMoreChildrenInternal();
+            try {
+                await this._loadMoreChildrenTask;
```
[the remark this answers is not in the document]
EricJizbaMSFT: This is a `try/finally` block - not a `try/catch` block. Finally blocks don't swallow errors - they just ensure their code is run even if an error occurs.

ui/src/treeDataProvider/AzureParentNode.ts
EricJizbaMSFT: 1. Yes. loadMoreChildren can be called multiple times at once and the first block of the if statement will only be entered if two calls are happening at the same time
2. See other comment. It has a `finally` block, not a `catch` block. There is no error handling.

ui/src/treeDataProvider/AzureParentNode.ts
EricJizbaMSFT: Yes

PR #218, opened 2018-07-09
Pass function app settings from the extension

EricJizbaMSFT:
`FUNCTIONS_EXTENSION_VERSION` and `WEBSITE_NODE_DEFAULT_VERSION` vary based on runtime, so I'll pass them in from the functions extension instead of hard-coding them here.

Also fix a few minor things with the new app service creation:
1. It was using the appsvc_rg_... naming for function apps
1. Add a 'View Logs' button to the final notification

### DIFF
--- a/appservice/src/createAppService/SiteCreateStep.ts
+++ b/appservice/src/createAppService/SiteCreateStep.ts
@@ -9,7 +9,7 @@ import { StorageAccountListKeysResult } from 'azure-arm-storage/lib/models';
 // tslint:disable-next-line:no-require-imports
 import WebSiteManagementClient = require('azure-arm-website');
 import { SiteConfig } from 'azure-arm-website/lib/models';
-import { ProgressLocation, window } from 'vscode';
+import { MessageItem, ProgressLocation, window } from 'vscode';
 import { addExtensionUserAgent, AzureWizardExecuteStep } from 'vscode-azureextensionui';
 import { ext } from '../extensionVariables';
 import { localize } from '../localize';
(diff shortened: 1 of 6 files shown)


appservice/src/createAppService/SiteCreateStep.ts
```diff
@@ -32,11 +39,20 @@ export class SiteCreateStep extends AzureWizardExecuteStep<IAppServiceWizardCont
...
+                const createdNewApp: string = localize('CreatedNewApp', 'Created new {0} "{1}": {2}', getAppKindDisplayName(wizardContext.newSiteKind), wizardContext.site.name, `https://${wizardContext.site.defaultHostName}`);
                 ext.outputChannel.appendLine(createdNewApp);
                 ext.outputChannel.appendLine('');
-                window.showInformationMessage(createdNewApp);
+                const viewLogs: MessageItem = {
+                    title: localize('viewLogs', 'View Logs')
+                };
+
+                // Note: intentionally not waiting for the result of this before returning
+                window.showInformationMessage(createdNewApp, viewLogs).then((result: MessageItem | undefined) => {
```
StephenWeatherford: Did you want the response in telemetry?

appservice/src/createAppService/SiteCreateStep.ts
```diff
@@ -76,23 +92,22 @@ export class SiteCreateStep extends AzureWizardExecuteStep<IAppServiceWizardCont
...
                     name: 'WEBSITE_CONTENTSHARE',
                     value: fileShareName
-                },
-                {
-                    name: 'WEBSITE_NODE_DEFAULT_VERSION',
-                    value: '6.5.0'
                 }
             ];
+
+            for (const key of Object.keys(this._functionAppSettings)) {
```
StephenWeatherford: Object.assign?

EricJizbaMSFT: Was not aware of `Object.assign` - that's nifty. However I don't think I can use it in this case since I'm changing from `{ exampleKey: exampleValue }` to `{ name: exampleKey, value: exampleValue }`

appservice/src/createAppService/SiteCreateStep.ts
EricJizbaMSFT: It would take some changes in the UI package to pass along telemetry stuff down through the wizard and I don't think it's worth it for this response. Filed an issue to fix later, though: https://github.com/Microsoft/vscode-azuretools/issues/219

PR #222, opened 2018-07-12
Add browse website button to deploy confirmation

EricJizbaMSFT:


appservice/src/deploy/deploy.ts
```diff
@@ -68,20 +69,30 @@ export async function deploy(client: SiteClient, fsPath: string, configurationSe
...
+        }
 
         // Note: intentionally not waiting for the result of this before returning
-        window.showInformationMessage(deployComplete, viewLogs).then((result: MessageItem | undefined) => {
+        window.showInformationMessage(deployComplete, ...buttons).then(async (result: MessageItem | undefined) => {
             if (result === viewLogs) {
                 ext.outputChannel.show();
+            } else if (result === browseWebsite) {
+                // tslint:disable-next-line:no-unsafe-any
+                opn(client.defaultHostUrl);
```
nturinski: Is it easy to use the treeItem's browse website function?  If not, don't worry about it.

EricJizbaMSFT: I don't have access to the treeItem from the shared package

PR #225, opened 2018-07-16
dismiss error display when it is a Maven error

jdneo:
related with: https://github.com/Microsoft/vscode-azurefunctions/pull/448

### DIFF
--- a/ui/src/callWithTelemetryAndErrorHandling.ts
+++ b/ui/src/callWithTelemetryAndErrorHandling.ts
@@ -47,7 +47,9 @@ export async function callWithTelemetryAndErrorHandling(callbackId: string, call
 
         if (!context.suppressErrorDisplay) {
             // Always append the error to the output channel, but only 'show' the output channel for multiline errors
-            ext.outputChannel.appendLine(localize('outputError', 'Error: {0}', errorData.message));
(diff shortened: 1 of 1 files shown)


ui/src/callWithTelemetryAndErrorHandling.ts
```diff
@@ -47,7 +47,9 @@ export async function callWithTelemetryAndErrorHandling(callbackId: string, call
 
         if (!context.suppressErrorDisplay) {
             // Always append the error to the output channel, but only 'show' the output channel for multiline errors
-            ext.outputChannel.appendLine(localize('outputError', 'Error: {0}', errorData.message));
+            if (errorData.errorType !== 'MavenExecutionError') {
```
StephenWeatherford: Sounds questionable to me although I have no context.  I would expect though to at least have reasoning in comments in the code.  Thx.

EricJizbaMSFT:
Yeah this shared package should be kept agnostic of any specific extension (functions) or language (java/maven).

Have you tried using the `suppressErrorDisplay` property?

jdneo:
@EricJizbaMSFT Do you mean setting the `context.suppressErrorDisplay` to true when `errorData.errorType === 'MavenExecutionError'`?

By suppressing the error, then the outputChannel will not automatically show.

PR #226, opened 2018-07-19
Reorder tiers and add F1 for Windows only

nturinski:
Reorder tiers so that F1 and B1-3 appear on the top.  Remove F1 from the list if the WebsiteOS is Linux as F1 is not supported for Linux ASPs.


appservice/src/createAppService/AppServicePlanSkuStep.ts
```diff
@@ -21,6 +22,11 @@ export class AppServicePlanSkuStep extends AzureWizardPromptStep<IAppServiceWiza
                 };
             });
 
+            if (wizardContext.newSiteOS === WebsiteOS.linux) {
+                // Free tier is not supported for Linux asp's
+                pricingTiers.shift();
```
EricJizbaMSFT: I'm not sure I'm the biggest fan of just 'shift'. It seems prone to bugs. For example - what if we re-order this again, or what if there are multiple free tiers at some point? I'd prefer to explicitly filter out all items that have 'free' as the tier

PR #228, opened 2018-08-01
Update to latest telemetry package

EricJizbaMSFT:
I thought we would get telemetry package updates automatically, but turns out we're not. A lot of places say the caret matches "The most recent major version", including VS Code:
screen shot 2018-08-01 at 2 30 21 pm

But it turns out the caret means the following:
> Allows changes that do not modify the left-most non-zero digit in the [major, minor, patch] tuple. In other words, this allows patch and minor updates for versions 1.0.0 and above, patch updates for versions 0.X >=0.1.0, and no updates for versions 0.0.X.

https://docs.npmjs.com/misc/semver#caret-ranges-123-025-004

### DIFF
--- a/appservice/package.json
+++ b/appservice/package.json
@@ -1,7 +1,7 @@
 {
     "name": "vscode-azureappservice",
     "author": "Microsoft Corporation",
-    "version": "0.19.0",
+    "version": "0.19.1",
     "description": "Common tools for developing Azure App Service extensions for VS Code",
     "tags": [
         "azure",
@@ -41,9 +41,9 @@
         "request": "^2.83.0",
         "request-promise": "^4.2.2",
         "simple-git": "~1.92.0",
-        "vscode-azureextensionui": "~0.16.0",
-        "vscode-azurekudu": "~0.1.8",
-        "vscode-extension-telemetry": "^0.0.15",
+        "vscode-azureextensionui": "^0.16.0",
+        "vscode-azurekudu": "^0.1.8",
+        "vscode-extension-telemetry": "^0.0.18",
         "vscode-nls": "^2.0.2",
         "websocket": "^1.0.25"
     },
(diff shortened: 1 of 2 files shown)


appservice/package.json
```diff
@@ -41,9 +41,9 @@
         "request": "^2.83.0",
         "request-promise": "^4.2.2",
         "simple-git": "~1.92.0",
-        "vscode-azureextensionui": "~0.16.0",
-        "vscode-azurekudu": "~0.1.8",
-        "vscode-extension-telemetry": "^0.0.15",
+        "vscode-azureextensionui": "^0.16.0",
+        "vscode-azurekudu": "^0.1.8",
+        "vscode-extension-telemetry": "^0.0.18",
```
StephenWeatherford: Don't you want ~0.0.18 for telemetry?  Using ^ means you're stuck with 0.0.18 always, right?  Or if you're trying not to pick up 0.1.00, maybe we should use ">=0.0.18 <0.1.0"?

EricJizbaMSFT: Now that I know more about how the caret works, I think it makes sense that we would have to manually update this version every once in a while. After all, every update could contain breaking changes.

But I realize that's kind of annoying to manually update, so I don't have a strong preference. I honestly wish they used the minor/patch system like we did and I could also file an issue on their repo

StephenWeatherford: Oh, they're not using semantic versioning?  I had assumed they were.

EricJizbaMSFT: According to https://semver.org/, 
> Major version zero (0.y.z) is for initial development. Anything may change at any time. The public API should not be considered stable.

Which sounds like "anything goes" to me. We use a system where we upgrade the minor for breaking changes and the patch for most other things. I think that's a common way to do it for pre-1.0.0, but I don't think that's an "official" part of semver. Looks like the telemetry package decided to just update the patch version for every release and ignore the minor version, which also technically seems valid according to the above guidance.

StephenWeatherford: I'm not convinced we should be using 0.x.y

PR #229, opened 2018-08-02
Debug telemetry 

StephenWeatherford:


ui/package.json
```diff
@@ -33,6 +33,7 @@
     "dependencies": {
         "azure-arm-resource": "^3.0.0-preview",
         "azure-arm-storage": "^3.1.0",
+        "debug": "^3.1.0",
```
EricJizbaMSFT: So this debug package will now be included with every extension we ship. Are we just saying it's small enough that it doesn't matter?

ui/src/createTelemetryReporter.ts
```diff
@@ -0,0 +1,65 @@
...
+        log(`** TELEMETRY("${this._packageInfo.name}/${eventName}", ${this._packageInfo.version}) properties=${propertiesString}, measures=${measuresString}`);
+    }
+}
+
+export function createTelemetryReporter(ctx: vscode.ExtensionContext): ITelemetryReporter {
+    const packageInfo: IPackageInfo = getPackageInfo(ctx);
+    let reporter: ITelemetryReporter & vscode.Disposable;
+
+    if (log.enabled) {
+        reporter = new DebugReporter(ctx);
```
EricJizbaMSFT: You already have packageInfo. Just pass that in to DebugReporter's constructor instead of passing in the ctx

ui/src/createTelemetryReporter.ts
```diff
@@ -0,0 +1,65 @@
...
+        // tslint:disable-next-line:strict-boolean-expressions
+        const propertiesString: string = JSON.stringify(properties || {});
+        // tslint:disable-next-line:strict-boolean-expressions
+        const measuresString: string = JSON.stringify(measures || {});
+        log(`** TELEMETRY("${this._packageInfo.name}/${eventName}", ${this._packageInfo.version}) properties=${propertiesString}, measures=${measuresString}`);
+    }
+}
+
+export function createTelemetryReporter(ctx: vscode.ExtensionContext): ITelemetryReporter {
+    const packageInfo: IPackageInfo = getPackageInfo(ctx);
```
EricJizbaMSFT: As mentioned offline, I would prefer if this had some error handling. I feel like interacting with the file system is always a possible source of errors and IMO we should be extra careful to make sure telemetry never breaks a user

ui/src/createTelemetryReporter.ts
```diff
@@ -0,0 +1,65 @@
...
+    return reporter;
+}
+
+interface IPackageInfo {
+    name: string;
+    version: string;
+    aiKey: string;
+}
+
+function getPackageInfo(context: vscode.ExtensionContext): IPackageInfo {
```
EricJizbaMSFT: Can we merge this with https://github.com/Microsoft/vscode-azuretools/blob/master/ui/src/getPackageInfo.ts?

ui/package.json
StephenWeatherford: It's pretty small and I do intend to be using it in more places eventually. However, node.js actually has similar functionality built in which I could use instead (uses the NODE_DEBUG variable), although I see debug being used instead in many places, including vscode. I assume the only reason is to avoid the NODE_DEBUG env var.

No strong opinion, I could use the built-in version for now if you'd prefer (https://nodejs.org/api/util.html#util_util_debuglog_section).

EricJizbaMSFT: If VS Code is using it, then I say go ahead and use it

ui/src/createTelemetryReporter.ts
StephenWeatherford: Done.

ui/package.json
StephenWeatherford: Okay, big change in direction.  Thinking about this, I realize I'm using the debug module wrong. Normally if a component is enabled in DEBUG, then it starts getting debug info output to console. If you set DEBUG=*, that would enable it for all modules.  The way I've got it, then would also turn off telemetry and switch to debug telemetry, which is a behavioral change.

I think I should remove debug for now and simply check an environment variable instead, like DEBUGTELEMETRY=1.

ui/src/extensionVariables.ts
```diff
@@ -32,7 +32,9 @@ class UninitializedExtensionVariables implements UIExtensionVariables {
  * Container for common extension variables used throughout the UI package. They must be initialized with registerUIExtensionVariables
  */
 export let ext: UIExtensionVariables = new UninitializedExtensionVariables();
+export let extInitialized: boolean = false;
```
EricJizbaMSFT: We already set `ext` to `UninitializedExtensionVariables` the line above this, which means you will get the following error if you try to use `ext` before you're supposed to: https://github.com/Microsoft/vscode-azuretools/blob/master/ui/src/extensionVariables.ts#L12

Your work with `extInitialized` feels redundant

ui/package.json
```diff
@@ -45,9 +45,9 @@
         "@types/mocha": "^2.2.32",
```
EricJizbaMSFT: FYI you're gonna have to bump the version of the ui package again because I already published a 0.16.5 version this morning

PR #230, opened 2018-08-02
Test Azure Account to be used for automated Azure testing

nturinski:
With a squash, all the history that used to contain hardcoded credentials should hopefully be lost to time 😅


appservice/src/deleteSite.ts
```diff
@@ -22,15 +22,15 @@ export async function deleteSite(client: SiteClient): Promise<void> {
         plan = await client.getAppServicePlan();
     }
 
-    if (!client.isSlot && plan.numberOfSites < 2) {
+    if (!client.isSlot && plan && plan.numberOfSites < 2) {
```
nturinski: Believe this might be the reason for the numberOfSites error I was seeing.  Once again, Eric's point about non-null assertion stands.  My guess is that client.getAppServicePlan was returning 'null' because of some timing issue between creating the asp and then immediately trying to get it.  Which would explain why sometimes it would work and other times it would not.

EricJizbaMSFT:
Nothing is ever gone from the internet haha. You should delete whatever was hard-coded and create new credentials

ui/src/treeDataProvider/AzureTreeDataProvider.ts
```diff
@@ -39,14 +40,16 @@ export class AzureTreeDataProvider implements TreeDataProvider<IAzureNode>, Disp
...
-    constructor(resourceProvider: IChildProvider, loadMoreCommandId: string, rootTreeItems?: IAzureParentTreeItem[]) {
+    constructor(resourceProvider: IChildProvider, loadMoreCommandId: string, rootTreeItems?: IAzureParentTreeItem[], testAccount?: TestAzureAccount) {
         this._resourceProvider = resourceProvider;
         this._loadMoreCommandId = loadMoreCommandId;
         this._customRootNodes = rootTreeItems ? rootTreeItems.map((treeItem: IAzureParentTreeItem) => new RootNode(this, treeItem, this._onNodeCreateEmitter)) : [];
 
         // Rather than expose 'AzureAccount' types in the index.ts contract, simply get it inside of this npm package
         const azureAccountExtension: Extension<AzureAccount> | undefined = extensions.getExtension<AzureAccount>('ms-vscode.azure-account');
-        if (!azureAccountExtension) {
+        if (testAccount) {
```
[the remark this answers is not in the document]
nturinski: The only issue I'm having with using a debug env variable is trying to create a TestAzureAccount in the constructor since I can't call async functions within it.  I'm not entirely sure how to remedy this without passing the account directly to the AzureTreeDataProvider.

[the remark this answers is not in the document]
StephenWeatherford: There can actually be advantages to having the TestAzureAccount around - it can provide additional testing functionality, such as being told to log in and out, etc.  

Like Eric, I'd love to see the client PR for this, too, though.

ui/src/TestAzureAccount.ts
```diff
@@ -0,0 +1,77 @@
...
+        this.subscriptions = [];
+        this.status = 'Initializing';
+        this.onStatusChangedEmitter = new EventEmitter<AzureLoginStatus>();
+        this.onStatusChanged = this.onStatusChangedEmitter.event;
+        this.onFiltersChangedEmitter = new EventEmitter<void>();
+        this.onFiltersChanged = this.onFiltersChangedEmitter.event;
+        this.filters = [];
+    }
+
+    public async getTestSubscription(): Promise<void> {
```
StephenWeatherford: I think the name is confusing.  How about testLogIn()?

ui/src/TestAzureAccount.ts
```diff
@@ -0,0 +1,77 @@
...
+    public waitForSubscriptions: () => Promise<boolean>;
+    public filters: AzureResourceFilter[];
+    public onFiltersChanged: Event<void>;
+    public waitForFilters: () => Promise<boolean>;
+    private onStatusChangedEmitter: EventEmitter<AzureLoginStatus>;
+    private onFiltersChangedEmitter: EventEmitter<void>;
+
+    public constructor() {
+        this.subscriptions = [];
+        this.status = 'Initializing';
```
StephenWeatherford: Not sure it should stay in the Initializing state, I think it should switch to LoggedOut just like the real one.  Probably after a very small delay (just make sure testLogIn() waits for initialization to succeed first so that test code that immediately asks for login doesn't fail - this mimics the fact that the real one won't let users log in until it's done initializing).

.vscode/launch.json
```diff
@@ -0,0 +1,17 @@
+{
+    // Use IntelliSense to learn about possible attributes.
```
EricJizbaMSFT: Delete this file. I'm guess it was generated on accident?

kudu/src/index.ts
```diff
@@ -0,0 +1,7 @@
+/*---------------------------------------------------------------------------------------------
```
EricJizbaMSFT: Likewise - why are there changes in the kudu folder for this PR? Please revert

ui/index.d.ts
```diff
@@ -29,8 +30,9 @@ export declare class AzureTreeDataProvider implements TreeDataProvider<IAzureNod
      * @param ui Used to get input from the user
      * @param telemetryReporter Optionally used to track telemetry for the tree
      * @param rootTreeItems Any nodes other than the subscriptions that should be shown at the root of the explorer
+     * @param testAccount A test Azure Account validated with a service principal that is leveraged for CI testing.
```
EricJizbaMSFT: I don't understand "validated with a service principal"

How about "A test Azure Account that leverages a service principal instead of interactive login"

ui/index.d.ts
```diff
@@ -10,6 +10,7 @@ import { Uri, TreeDataProvider, Disposable, TreeItem, Event, OutputChannel, Meme
 import TelemetryReporter from 'vscode-extension-telemetry';
 import { ResourceGroup } from 'azure-arm-resource/lib/resource/models';
 import { StorageAccount, CheckNameAvailabilityResult } from 'azure-arm-storage/lib/models';
+import { TestAzureAccount } from './src';
```
EricJizbaMSFT: Don't import TestAzureAccount like this. You actually have to add the typing to this file so that it's accessible outside the package.

When you add the typing, please add documentation on the exact environment variables that need to be set for it to work

ui/src/TestAzureAccount.ts
```diff
@@ -0,0 +1,77 @@
...
+ *--------------------------------------------------------------------------------------------*/
+import { SubscriptionClient } from 'azure-arm-resource';
+import { SubscriptionListResult, TenantListResult } from 'azure-arm-resource/lib/subscription/models';
+import { ApplicationTokenCredentials, AzureEnvironment, loginWithServicePrincipalSecret } from 'ms-rest-azure';
+import { Event, EventEmitter } from 'vscode';
+import { AzureAccount, AzureLoginStatus, AzureResourceFilter, AzureSession, AzureSubscription } from './azure-account.api';
+import { ArgumentError } from './errors';
+import { localize } from './localize';
+
+export class TestAzureAccount implements AzureAccount {
```
EricJizbaMSFT: As mentioned in the Functions PR, I think this class should provide a way to get subscription information (at least subscription id and credentials) so that people can validate Azure stuff without going through our tree

ui/src/TestAzureAccount.ts
```diff
@@ -0,0 +1,77 @@
...
+            const tenantId: string = <string>tenants[0].id;
+            const session: AzureSession = {
+                environment: credentials.environment,
+                userId: '',
+                tenantId: tenantId,
+                credentials: credentials
+            };
+
+            // tslint:disable-next-line:strict-boolean-expressions
+            if (subscriptions && subscriptions[0].id && subscriptions[0].displayName && subscriptions[0].subscriptionId) {
```
EricJizbaMSFT: nit: Why do you need to disable `strict-boolean-expressions` here? I'm guessing you can remove `subscriptions` from the if clause because it's probably always defined

ui/src/TestAzureAccount.ts
EricJizbaMSFT: I vote for just 'logIn'. The 'test' part feels redundant in the _Test_AzureAccount

I also just get confused by a lot of our `test` naming (my bad, I know), since it feels like `testLogIn` would actually run a test, but this isn't actually testing anything. I'm starting to think we should've named this stuff `MockAzureAccount` and `MockUserInput` or something like that.

ui/index.d.ts
StephenWeatherford: I've discovered I don't like using npm link to test packages - it won't catch things like this until you publish and discover it doesn't actually work.

I prefer to use 'npm pack' and then use 'npm install <path-to-tgz>' which should test exactly the same as if it were published.

EricJizbaMSFT: Not sure what you mean by "it won't catch things like this". I linked his package as-is, added a reference to TestAzureAccount in a random file in functions, and now I see this build error:
screen shot 2018-08-09 at 10 32 59 am

All I had to do was add `export declare class TestAzureAccount {}` to index.d.ts and remove `import { TestAzureAccount } from './src';` and now the error is gone:
screen shot 2018-08-09 at 10 32 14 am

I will admit that the code will run if he ignores build errors, but we shouldn't be ignoring build errors :P

kudu/src/index.ts
nturinski: I keep running into this issue and honestly have no idea why it's happening 😕

nturinski:
I just tested it and there aren't any build errors after removing the `implements AzureAccount` in the `index.d.ts`.  I wasn't thinking, but the only place that the `TestAzureAccount` must be recognized as an `AzureAccount` type is when we assign it in the `AzureTreeDataProvider` which is all within the same package, so the `index.d.ts` isn't affecting it at all.  😌

ui/src/TestAzureAccount.ts
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    public signOut(): void {
+        this.changeStatus('LoggedOut');
+        this.changeFilter();
+        this.subscriptions = [];
+    }
+
+    public getSubscriptionCredentials(): ServiceClientCredentials {
+        return this.subscriptions[0].session.credentials;
```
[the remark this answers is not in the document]
nturinski: Do you think an ArgumentError would be helpful enough or should I suggest to the user to either signIn first or that an error likely happened during the signIn?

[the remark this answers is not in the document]
EricJizbaMSFT: I might just say "No subscription found."

nturinski:
@StephenWeatherford Just wanted to ping you to see if there were any requested changes that I've missed

PR #236, opened 2018-08-14
Default to centralus for automated creation experience

nturinski:
In reference to this issue: https://github.com/Microsoft/vscode-azureappservice/issues/536

### DIFF
--- a/appservice/src/createAppService/createAppService.ts
+++ b/appservice/src/createAppService/createAppService.ts
@@ -3,6 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import { Location } from 'azure-arm-resource/lib/subscription/models';
 import { Site, SkuDescription } from 'azure-arm-website/lib/models';
 import { ServiceClientCredentials } from 'ms-rest';
 import { workspace } from 'vscode';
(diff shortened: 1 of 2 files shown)


appservice/src/createAppService/createAppService.ts
```diff
@@ -92,6 +93,15 @@ export async function createAppService(
     // Ideally actionContext should always be defined, but there's a bug with the NodePicker. Create a 'fake' actionContext until that bug is fixed
     // https://github.com/Microsoft/vscode-azuretools/issues/120
     actionContext = actionContext || <IActionContext>{ properties: {}, measurements: {} };
+
+    // az webapp up defaults to centralus so for we are defaulting to that location as well
+    if (wizardContext.newSiteKind === AppKind.app && !advancedCreation) {
+        const locations: Location[] = await LocationListStep.getLocations(wizardContext);
```
EricJizbaMSFT: You can just use `LocationListStep.setLocation`

appservice/src/createAppService/createAppService.ts
```diff
@@ -92,6 +93,15 @@ export async function createAppService(
...
+        wizardContext.location = locations.find((location: Location) => {
```
EricJizbaMSFT: I feel like it might make more sense to put this in `setWizardContextDefaults`

appservice/src/createAppService/createAppService.ts
```diff
@@ -74,8 +73,8 @@ export async function createAppService(
                 promptSteps.push(new AppServicePlanListStep());
                 promptSteps.push(new LocationListStep());
             } else {
-                if (workspace.workspaceFolders && workspace.workspaceFolders.length === 1) {
-                    // can make smart defaults if only one workspace is opened
+                if (wizardContext.newSiteKind === AppKind.app && !advancedCreation) {
```
EricJizbaMSFT: So you're already in a switch case for `AppKind.app` and an else case for `!advancedCreation` so I don't think you need this `if` at all

PR #237, opened 2018-08-17
Add Java SE runtime add support zip single file to deploy

SummerSun:


appservice/src/FileUtilities.ts
```diff
@@ -12,6 +12,14 @@ export function getFileExtension(fsPath: string): string | undefined {
     return fsPath.split('.').pop();
 }
 
+export async function isfilePathExist(fsPath: string): Promise<boolean> {
```
EricJizbaMSFT: Please leverage `fse.pathExists` instead of writing your own function. See here for more info: https://github.com/Microsoft/vscode-azurefunctions/blob/3c65d5596bc7c0ff8f2ac106793690610052a5af/src/utils/fs.ts#L38

I thought we were already using 'fs-extra' in all of our repos, but I guess we haven't started using it here yet. You'll have to add it to the package.json

appservice/package.json
```diff
@@ -35,6 +35,7 @@
         "azure-arm-resource": "^3.0.0-preview",
         "azure-arm-storage": "^3.1.0",
         "azure-arm-website": "^3.0.0-preview",
+        "fs-extra": "^4.0.2",
```
EricJizbaMSFT: I forgot to mention, but you need to add the license to thirdpartynotices when adding a new package https://github.com/Microsoft/vscode-azuretools/blob/master/appservice/thirdpartynotices.txt

Just follow the format used in that file

PR #243, opened 2018-08-23
Register ui extension variables as a part of app service package

EricJizbaMSFT:
@nturinski and I run into a problem when linking the app service package during development. Basically, the linked package is using it's own copy of the ui package, which means the extension variables aren't registered for that second copy of the ui package. This should fix that (And I don't think it will have any affect on users/functionality)


EricJizbaMSFT:
Closing in favor of https://github.com/Microsoft/vscode-azuretools/pull/245

PR #245, opened 2018-08-23
Move more log streaming logic into shared package

EricJizbaMSFT:
There are two bugs related to log streaming that affect both app service and functions. I moved some more of the logic into the shared package so that I could fix both of them here.

Fixes https://github.com/Microsoft/vscode-azureappservice/issues/525
Fixes https://github.com/Microsoft/vscode-azurefunctions/issues/483

Definitely view this PR with whitespace changes off haha

### DIFF
--- a/appservice/src/extensionVariables.ts
+++ b/appservice/src/extensionVariables.ts
@@ -3,8 +3,8 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { OutputChannel } from "vscode";
-import { IAzureUserInput } from 'vscode-azureextensionui';
+import { ExtensionContext, OutputChannel } from "vscode";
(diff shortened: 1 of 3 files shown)


appservice/src/extensionVariables.ts
```diff
@@ -35,6 +40,7 @@ export let ext: IAppServiceExtensionVariables = new UninitializedExtensionVariab
 /**
  * Call this to register common variables used throughout the AppService package.
  */
-export function registerAppServiceExtensionVariables(extVars: IAppServiceExtensionVariables): void {
+export function registerAppServiceExtensionVariables(extVars: IAppServiceExtensionVariables & UIExtensionVariables): void {
     ext = extVars;
+    registerUIExtensionVariables(extVars);
```
nturinski: Is this only to fix the npm link issue or is there another reason for this?

EricJizbaMSFT: Yep I did it to fix the npm link issue and I can't think of any downside

PR #248, opened 2018-08-28
Add common function to create Azure clients

EricJizbaMSFT:
Based on comment by @StephenWeatherford here: https://github.com/Microsoft/vscode-azureeventgrid/pull/31#discussion_r212778173

Tested in the EventGrid extension and seems to be working. Let me know what you think.

### DIFF
--- a/ui/src/createAzureClient.ts
+++ b/ui/src/createAzureClient.ts
@@ -1,0 +1,25 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { ServiceClientCredentials } from "ms-rest";
+import { AzureEnvironment, AzureServiceClientOptions } from "ms-rest-azure";
+import { IAddUserAgent } from "../index";
+import { addExtensionUserAgent } from "./extensionUserAgent";
+
+export function createAzureClient<T extends IAddUserAgent>(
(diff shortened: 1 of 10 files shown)


ui/src/createAzureClient.ts
```diff
@@ -0,0 +1,25 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+import { ServiceClientCredentials } from "ms-rest";
+import { AzureEnvironment, AzureServiceClientOptions } from "ms-rest-azure";
+import { IAddUserAgent } from "../index";
+import { addExtensionUserAgent } from "./extensionUserAgent";
+
+export function createAzureClient<T extends IAddUserAgent>(
+    clientInfo: { credentials: ServiceClientCredentials; subscriptionId: string; environment: AzureEnvironment; },
+    clientType: { new(credentials: ServiceClientCredentials, subscriptionId: string, baseUri?: string, options?: AzureServiceClientOptions): T }): T {
```
nturinski: Would it be useful to have an enum for the clientTypes that we are expecting?  What happens if the user passes in an invalid clientType?

EricJizbaMSFT: What do you mean by invalid? I have the type specified, so you should get build errors if the constructor for the client doesn't match what we expect. That's what all that `new(credentials...` stuff is about. It's the type of the constructor

nturinski: Ah, right.  Sorry, when I saw the way you were invoking the method, I was thinking you were passing a string (the name of the client) but I realize that's the actual class type now.

EricJizbaMSFT: forgive
you got it

PR #249, opened 2018-08-29
Add 'remove private information before submitting'

StephenWeatherford:


ui/src/reportAnIssue.ts
```diff
@@ -15,6 +15,8 @@ export function reportAnIssue(actionId: string, parsedError: IParsedError): void
     const { extensionName, extensionVersion } = getPackageInfo();
 
     const body: string = `
+<Please be sure to remove any private information before submitting.>
```
EricJizbaMSFT: `<` shows up in the 'Write' tab of GitHub comments, but doesn't show up in the 'Preview' tab or after publishing. Maybe change it to `&lt;`? Can you fix the ones below, too?

screen shot 2018-08-29 at 2 05 19 pm

screen shot 2018-08-29 at 2 05 22 pm

StephenWeatherford: I had wondered, but assumed the current ones were working and thus correct.  Will do.

PR #251, opened 2018-08-30
Remove deploy message from shared package

EricJizbaMSFT:
This logic already had several `isFunctionApp` checks and it would've just gotten more complicated when I added a 'Stream Logs' button as described here: https://github.com/Microsoft/vscode-azureappservice/issues/530

I thought it would be simpler to remove it from shared package


EricJizbaMSFT:
Not that I'm aware of. In any case, that's why I bumped the minor version of the package to indicate that this was a breaking change

PR #253, opened 2018-08-31
Programmatic RG

nturinski:
For the following issue: https://github.com/Microsoft/vscode-azurefunctions/issues/471
Works in conjunction with this PR: https://github.com/Microsoft/vscode-azurefunctions/pull/551

### DIFF
--- a/appservice/src/createAppService/createAppService.ts
+++ b/appservice/src/createAppService/createAppService.ts
@@ -3,6 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import { ResourceGroup } from 'azure-arm-resource/lib/resource/models';
 import { Site, SkuDescription } from 'azure-arm-website/lib/models';
 import { AzureWizard, AzureWizardExecuteStep, AzureWizardPromptStep, IActionContext, IAzureNode, LocationListStep, ResourceGroupCreateStep, ResourceGroupListStep, StorageAccountKind, StorageAccountListStep, StorageAccountPerformance, StorageAccountReplication } from 'vscode-azureextensionui';
 import { AppKind, WebsiteOS } from './AppKind';
@@ -14,7 +15,6 @@ import { SiteCreateStep } from './SiteCreateStep';
 import { SiteNameStep } from './SiteNameStep';
 import { SiteOSStep } from './SiteOSStep';
 import { SiteRuntimeStep } from './SiteRuntimeStep';
-import { ResourceGroup } from 'azure-arm-resource/lib/resource/models';
 
 export async function createAppService(
     appKind: AppKind,
--- a/appservice/src/createAppService/createAppService.ts
+++ b/appservice/src/createAppService/createAppService.ts
@@ -38,9 +38,9 @@ export async function createAppService(
 
     if (actionContext.properties.resourceGroup) {
         // if a rg was passed in, use that as the default
-        const progamaticResourceGroup: ResourceGroup = (await ResourceGroupListStep.getResourceGroups(wizardContext))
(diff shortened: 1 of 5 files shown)


appservice/src/createAppService/createAppService.ts
```diff
@@ -35,6 +36,13 @@ export async function createAppService(
         environment: node.environment
     };
 
+    if (actionContext.properties.resourceGroup) {
```
EricJizbaMSFT: Same comment as other PR:

`actionContext.properties` is used for telemetry and we don't want to collect this in telemetry

We'll have to find a different way to pass the resource group along

appservice/src/createAppService/createAppService.ts
```diff
@@ -35,6 +37,16 @@ export async function createAppService(
         environment: node.environment
     };
 
+    if (resourceGroup) {
```
EricJizbaMSFT: Feels like you should just set `wizardContext.newResourceGroupName = resourceGroup` like you do [here](https://github.com/Microsoft/vscode-azuretools/blob/master/appservice/src/createAppService/createAppService.ts#L96)

That would be simpler, plus I think it makes sense to create the resource group if it doesn't exist

ui/index.d.ts
```diff
@@ -133,8 +133,9 @@ export interface IChildProvider {
     /**
      * Implement this if you want the 'create' option to show up in the node picker
      * @param options User-defined options that are passed to the IAzureParentTreeItem.createChild call
+     * @param resourceGroup The name of the resource group the app will be created under programmatically
      */
-    createChild?(node: IAzureNode, showCreatingNode: (label: string) => void, userOptions?: any): Promise<IAzureTreeItem>;
+    createChild?(node: IAzureNode, showCreatingNode: (label: string) => void, userOptions?: any, resourceGroup?: string): Promise<IAzureTreeItem>;
```
EricJizbaMSFT: You shouldn't be changing anything in the ui package for this. `resourceGroup` should be a _part_ of userOptions (that's the whole point of it being type `any`)

appservice/src/createAppService/createAppService.ts
nturinski: Yeah, I suppose it makes sense to just create the rg for them then.

nturinski: So that logic relies on the `advancedCreation` being `false`.

The step is being added here https://github.com/Microsoft/vscode-azuretools/blob/master/appservice/src/createAppService/createAppService.ts#L78.  This is neccessary to execute the rg create step which will find/create the rg if neccessary.  In order to make it work for Functions, we'd have to add some more logic.  

Thinking it might be better to just go back to the find logic, and if it can't find it, it'll be undefined and the user will just be prompted for an rg name.

nturinski: So my thoughts are I'll go ahead and merge this because:
1. The use case seems to be for a predetermined rg, so creating it for them seems like a nice to have.
1. This way touches less code and the consequence for incorrectly writing the rg name is to be prompted rather than creating a new rg with the wrong name.
1. If we were to add the creation, we'd have to add some logic to `ResourceListStep()` to create if it only detects a `newResourceGroupName` which could have potential regressions/unexpected side effects.

If we don't want this, let me know and I can fix it up before we release.

EricJizbaMSFT: I think creating it for them is more than just a "nice to have". I think the scenario is less about using an existing resource group and more about putting a bunch of new resources in the same resource group. So depending on which order they create those resources in, it's entirely possible the rg won't exist yet

I'm intimately familiar with the wizard code haha and I'm confident you can just add `executeSteps.push(new ResourceGroupCreateStep());` to the function app case you referenced above and it will work. I definitely think the risk is worth the value considering CTI hasn't started testing yet

PR #256, opened 2018-09-07
Check for linuxConsumption plans

nturinski:


appservice/src/SiteClient.ts
```diff
@@ -60,15 +61,19 @@ export class SiteClient {
...
         this.planName = matches[3];
 
         this.defaultHostName = site.defaultHostName;
         this.defaultHostUrl = `https://${this.defaultHostName}`;
-        this.kuduHostName = site.hostNameSslStates.find((h: HostNameSslState) => h.hostType && h.hostType.toLowerCase() === 'repository').name;
-        this.kuduUrl = `https://${this.kuduHostName}`;
-        this.gitUrl = `${this.kuduHostName}:443/${site.repositorySiteName}.git`;
+        const kuduRepositoryUrl: HostNameSslState = site.hostNameSslStates.find((h: HostNameSslState) => h.hostType && h.hostType.toLowerCase() === 'repository');
+        if (kuduRepositoryUrl) {
+            this.kuduHostName = kuduRepositoryUrl.name;
```
EricJizbaMSFT: You should make the type of these members `| undefined`

(This package still has strictNullChecks turned off so nothing will be enforced, but other repos referencing SiteClient will benefit)

appservice/src/getKuduClient.ts
```diff
@@ -8,10 +8,15 @@ import { BasicAuthenticationCredentials } from 'ms-rest';
 import { addExtensionUserAgent } from 'vscode-azureextensionui';
 import KuduClient from 'vscode-azurekudu';
 import { ArgumentError } from './errors';
+import { localize } from './localize';
 import { SiteClient } from './SiteClient';
 
 export async function getKuduClient(client: SiteClient): Promise<KuduClient> {
     const user: User = await client.getWebAppPublishCredential();
+    if (client.isLinuxConsumptionPlan) {
+        const notSupportedLinux: string = localize('notSupportedLinux', 'This action is not supported with Linux Consumption Plans.');
```
EricJizbaMSFT: I don't want to base _whether we throw an error_ off of the plan type. That should be based off of `if (!this.kuduUrl)` or something like that. Specifically, this logic wouldn't handle these two cases:
1. The app service team adds support for kudu down the line on linux consumption
1. They add another plan type that _doesn't_ support kudu.

Maybe the message could be something like this where you just insert site.kind and plan.sku.tier:
```
`This operation is not supported by App Service Plans with kind "{...}" and sku "{...}"`
```

appservice/src/getKuduClient.ts
```diff
@@ -8,10 +8,15 @@ import { BasicAuthenticationCredentials } from 'ms-rest';
 import { addExtensionUserAgent } from 'vscode-azureextensionui';
 import KuduClient from 'vscode-azurekudu';
 import { ArgumentError } from './errors';
+import { localize } from './localize';
 import { SiteClient } from './SiteClient';
 
 export async function getKuduClient(client: SiteClient): Promise<KuduClient> {
     const user: User = await client.getWebAppPublishCredential();
```
EricJizbaMSFT: Move the error above `await client.getWebAppPublishCredential();`. No need to wait for that operation

appservice/src/SiteClient.ts
```diff
@@ -60,15 +61,19 @@ export class SiteClient {
         this.kind = site.kind;
         this.initialState = site.state;
         this.isFunctionApp = site.kind && site.kind.includes('functionapp');
+        this.isLinuxConsumptionPlan = site.kind && site.kind.includes('functionapp,linux,container');
```
EricJizbaMSFT: So I don't think this check is correct. To check if it's linux it would just be:
```
site.kind && site.kind.includes('linux')
```

And to check if it's consumption, you'd have to check if the plan's sku was "Dynamic"

appservice/src/SiteClient.ts
```diff
@@ -31,15 +31,16 @@ export class SiteClient {
     public readonly kind: string;
     public readonly initialState: string;
     public readonly isFunctionApp: boolean;
+    public readonly isLinuxConsumptionPlan: boolean;
```
EricJizbaMSFT: Remove this property since it's no longer used

appservice/src/getKuduClient.ts
```diff
@@ -3,14 +3,20 @@
...
 import { addExtensionUserAgent } from 'vscode-azureextensionui';
 import KuduClient from 'vscode-azurekudu';
 import { ArgumentError } from './errors';
+import { localize } from './localize';
 import { SiteClient } from './SiteClient';
 
 export async function getKuduClient(client: SiteClient): Promise<KuduClient> {
+    if (!client.kuduHostName) {
+        const asp: AppServicePlan = await client.getAppServicePlan();
+        const notSupportedLinux: string = localize('notSupportedLinux', 'This operation is not supported by App Service Plans with kind "{0}" and sku tier "{1}"', client.kind, asp.sku.tier);
```
EricJizbaMSFT: Nit: Add a period at the end of the message and change "Plans" to "plans" (here's an example doc that uses lowercase "plans" all over: https://docs.microsoft.com/en-us/azure/app-service/azure-web-sites-web-hosting-plans-in-depth-overview)

PR #260, opened 2018-09-10
showQuickPick - enable multiselect (don't persist choices in this case)

### LINKED ISSUE #259: showQuickPick should also be able to return an array of items
If the canPickMany flag is set to True, the return type can be an array of items. For example, [here](
https://github.com/Microsoft/vscode/blob/554c9b1460786899a79591e10f3dde16bfbe5b7d/src/vs/workbench/api/node/extHostQuickOpen.ts#L39).

PrashanthCorp:
Fixes #259

### DIFF
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -3,13 +3,12 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { Location } from 'azure-arm-resource/lib/subscription/models';
-import { ServiceClientCredentials, ServiceClient } from 'ms-rest';
-import { AzureEnvironment, AzureServiceClientOptions } from 'ms-rest-azure';
-import { Uri, TreeDataProvider, Disposable, TreeItem, Event, OutputChannel, Memento, InputBoxOptions, QuickPickItem, QuickPickOptions, TextDocument, ExtensionContext, MessageItem, OpenDialogOptions, MessageOptions } from 'vscode';
-import TelemetryReporter from 'vscode-extension-telemetry';
 import { ResourceGroup } from 'azure-arm-resource/lib/resource/models';
-import { StorageAccount, CheckNameAvailabilityResult } from 'azure-arm-storage/lib/models';
+import { Location } from 'azure-arm-resource/lib/subscription/models';
+import { StorageAccount } from 'azure-arm-storage/lib/models';
+import { ServiceClientCredentials } from 'ms-rest';
+import { AzureEnvironment, AzureServiceClientOptions } from 'ms-rest-azure';
+import { Disposable, Event, ExtensionContext, InputBoxOptions, Memento, MessageItem, MessageOptions, OpenDialogOptions, OutputChannel, QuickPickItem, QuickPickOptions, TextDocument, TreeDataProvider, TreeItem, Uri } from 'vscode';
 
 export type OpenInPortalOptions = {
     /**
@@ -279,14 +278,24 @@ export interface IParsedError {
  */
 export interface IAzureUserInput {
     /**
-     * Shows a selection list.
-     * Automatically persists the 'recently used' item and displays that at the top of the list
-     *
-     * @param items An array of items, or a promise that resolves to an array of items.
-     * @param options Configures the behavior of the selection list.
-     * @throws `UserCancelledError` if the user cancels.
-     * @return A promise that resolves to the item the user picked.
-     */
+    * Shows a selection list.
+    *
+    * @param items An array of items, or a promise that resolves to an array of items.
+    * @param options Configures the behavior of the selection list.
+    * @throws `UserCancelledError` if the user cancels.
+    * @return A promise that resolves to an array of items the user picked.
+    */
+    showQuickPick<T extends QuickPickItem>(items: T[] | Thenable<T[]>, options: QuickPickOptions & { canPickMany: true }): Promise<T[]>;
+
+    /**
+      * Shows a selection list.
+      * Automatically persists the 'recently used' item and displays that at the top of the list
+      *
+      * @param items An array of items, or a promise that resolves to an array of items.
+      * @param options Configures the behavior of the selection list.
+      * @throws `UserCancelledError` if the user cancels.
+      * @return A promise that resolves to the item the user picked.
+      */
     showQuickPick<T extends QuickPickItem>(items: T[] | Thenable<T[]>, options: QuickPickOptions): Promise<T>;
 
     /**
@@ -339,6 +348,7 @@ export declare class AzureUserInput implements IAzureUserInput {
      */
     public constructor(persistence: Memento);
 
+    public showQuickPick<T extends QuickPickItem>(items: T[] | Thenable<T[]>, options: QuickPickOptions & { canPickMany: true }): Promise<T[]>
     public showQuickPick<T extends QuickPickItem>(items: T[] | Thenable<T[]>, options: QuickPickOptions): Promise<T>;
     public showInputBox(options: InputBoxOptions): Promise<string>;
     public showWarningMessage<T extends MessageItem>(message: string, ...items: T[]): Promise<T>;
--- a/ui/src/AzureUserInput.ts
+++ b/ui/src/AzureUserInput.ts
@@ -3,8 +3,8 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { Memento, MessageItem, MessageOptions, QuickPickItem, QuickPickOptions } from 'vscode';
 import * as vscode from 'vscode';
+import { Memento, MessageItem, MessageOptions, QuickPickItem, QuickPickOptions } from 'vscode';
 import { IAzureQuickPickItem, IAzureQuickPickOptions, IAzureUserInput } from '../index';
 import { DialogResponses } from './DialogResponses';
 import { UserCancelledError } from './errors';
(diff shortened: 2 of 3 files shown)


ui/src/AzureUserInput.ts
```diff
@@ -18,23 +18,27 @@ export class AzureUserInput implements IAzureUserInput {
         this._persistence = persistence;
     }
 
-    public async showQuickPick<T extends QuickPickItem>(items: T[] | Thenable<T[]>, options: QuickPickOptions): Promise<T> {
+    public async showQuickPick<T extends QuickPickItem>(items: T[] | Thenable<T[]>, options: QuickPickOptions): Promise<T>;
```
StephenWeatherford: Remove this one, you only need the full version, since the overloads are in index.d.ts

ui/src/AzureUserInput.ts
```diff
@@ -18,23 +18,27 @@ export class AzureUserInput implements IAzureUserInput {
...
+    public async showQuickPick<T extends QuickPickItem>(items: T[] | Thenable<T[]>, options: QuickPickOptions & { canPickMany: true }): Promise<T[]> {
```
StephenWeatherford: { canPickMany: Boolean }   - you're handling both cases in this code.

ui/src/AzureUserInput.ts
```diff
@@ -18,23 +18,27 @@ export class AzureUserInput implements IAzureUserInput {
...
         let persistenceKey: string | undefined;
         const unhashedKey: string | undefined = (<IAzureQuickPickOptions>options).id || options.placeHolder;
-        if (unhashedKey) {
+        if (unhashedKey && !options.canPickMany) {
             persistenceKey = `showQuickPick.${randomUtils.getPseudononymousStringHash(unhashedKey)}`;
         }
 
-        const result: T | undefined = await vscode.window.showQuickPick(this.getOrderedItems(items, persistenceKey, (<IAzureQuickPickOptions>options).suppressPersistence), options);
-        if (result === undefined) {
+        const result: T[] | undefined = await vscode.window.showQuickPick(this.getOrderedItems(items, persistenceKey, (<IAzureQuickPickOptions>options).suppressPersistence), options);
```
StephenWeatherford: const result: T | T[] | undefined

ui/src/AzureUserInput.ts
StephenWeatherford: returns T | T[] | undefined

This doesn't have compile errors?

ui/src/AzureUserInput.ts
```diff
@@ -18,23 +18,27 @@ export class AzureUserInput implements IAzureUserInput {
...
+        if (result === undefined || result.length === 0) {
             throw new UserCancelledError();
         }
+        if (Array.isArray(result)) {
+            return result;
+        }
 
-        if (persistenceKey && !(<IAzureQuickPickItem><{}>result).suppressPersistence) {
+        if (persistenceKey && !(<IAzureQuickPickItem><{}>result).suppressPersistence && !Array.isArray(result)) {
```
StephenWeatherford: Not needed - you've already returned in this case.

PrashanthCorp: I put that in so that the compiler knew that result being used in the next line was not an array. It felt like code that informs the compiler rather than function.

ui/src/AzureUserInput.ts
```diff
@@ -18,23 +18,27 @@ export class AzureUserInput implements IAzureUserInput {
...
         const unhashedKey: string | undefined = (<IAzureQuickPickOptions>options).id || options.placeHolder;
-        if (unhashedKey) {
+        if (unhashedKey && !options.canPickMany) {
             persistenceKey = `showQuickPick.${randomUtils.getPseudononymousStringHash(unhashedKey)}`;
         }
 
-        const result: T | undefined = await vscode.window.showQuickPick(this.getOrderedItems(items, persistenceKey, (<IAzureQuickPickOptions>options).suppressPersistence), options);
-        if (result === undefined) {
+        const result: T[] | undefined = await vscode.window.showQuickPick(this.getOrderedItems(items, persistenceKey, (<IAzureQuickPickOptions>options).suppressPersistence), options);
+        if (result === undefined || result.length === 0) {
```
EricJizbaMSFT: I feel like an empty array is a valid response. As a simple example, it could be like when you're trying to unsubscribe from emails and rather than giving you a single "unsubscribe" button, they give you a list like this:
[ ] Send weekly newsletter
[ ] Send coupons
[ ] Send notifications

And you have to unselect all of them to stop getting those annoying emails haha

ui/src/AzureUserInput.ts
EricJizbaMSFT: I agree with @StephenWeatherford feels like this shouldn't matter even for the compiler

ui/index.d.ts
```diff
@@ -339,6 +348,7 @@ export declare class AzureUserInput implements IAzureUserInput {
      */
     public constructor(persistence: Memento);
 
+    public showQuickPick<T extends QuickPickItem>(items: T[] | Thenable<T[]>, options: QuickPickOptions & { canPickMany: true }): Promise<T[]>
```
EricJizbaMSFT: nit: Add a semicolon at the end of this

ui/src/AzureUserInput.ts
StephenWeatherford: I'm not convinced.  We want UserCancelled to happen automatically, otherwise telemetry is messed up.  In all the likely use cases I can think of, not selecting anything would be a cancel. The unsubscribe scenario isn't realistic for us.

EricJizbaMSFT: UserCancelled will still happen automatically if they hit "escape" correct? Won't that return undefined? I'm assuming an empty array only happens if they select 'enter'. Does VS Code support "validateInput" on quick picks? Can you force the quick pick to stay open if they haven't selected anything? If not, I still feel like you should throw an actual error message like "This field is required." rather than a UserCancelledError which has no feedback to the user.

ui/src/AzureUserInput.ts
```diff
@@ -18,21 +18,25 @@ export class AzureUserInput implements IAzureUserInput {
         this._persistence = persistence;
     }
 
-    public async showQuickPick<T extends QuickPickItem>(items: T[] | Thenable<T[]>, options: QuickPickOptions): Promise<T> {
+    public async showQuickPick<T extends QuickPickItem>(items: T[] | Thenable<T[]>, options: QuickPickOptions & { canPickMany: boolean }): Promise<T | T[]> {
```
EricJizbaMSFT: nit: Remove `{ canPickMany: boolean }`. It's already a part of QuickPickOptions and unlike index.d.ts, you're not setting it to a specific value in this case

ui/src/AzureUserInput.ts
```diff
@@ -18,21 +18,25 @@ export class AzureUserInput implements IAzureUserInput {
...
         let persistenceKey: string | undefined;
         const unhashedKey: string | undefined = (<IAzureQuickPickOptions>options).id || options.placeHolder;
-        if (unhashedKey) {
+        if (unhashedKey && !options.canPickMany) {
             persistenceKey = `showQuickPick.${randomUtils.getPseudononymousStringHash(unhashedKey)}`;
         }
 
-        const result: T | undefined = await vscode.window.showQuickPick(this.getOrderedItems(items, persistenceKey, (<IAzureQuickPickOptions>options).suppressPersistence), options);
+        const result: T | T[] | undefined = await vscode.window.showQuickPick(this.getOrderedItems(items, persistenceKey, (<IAzureQuickPickOptions>options).suppressPersistence), options);
+        // tslint:disable-next-line:no-any
```
EricJizbaMSFT: nit: I don't think this tslint disable is necessary?

ui/src/AzureUserInput.ts
PrashanthCorp: Talking offline to Stephen, I agree with him now. Response inline. 


---
> UserCancelled will still happen automatically if they hit "escape" correct? Won't that return undefined? I'm assuming an empty array only happens if they select 'enter'. 

Yes. 
> Does VS Code support "validateInput" on quick picks? 

I don't see a validateInput (makes sense since the only input is a choice of options already presented to them). 
>Can you force the quick pick to stay open if they haven't selected anything? 

Aside from ignoreFocusOut, I don't see a way currently. Knowing they haven't selected anything seems to happen after the fact.

> If not, I still feel like you should throw an actual error message like "This field is required." rather than a UserCancelledError which has no feedback to the user.

I think this would be more nuanced. In my case a user choosing no option would mean the user is explicit about not wanting to do anything. Hitting "Enter" not undoable anyway. 

In some cases, the user might prefer a headsup, in which case, the caller would decide what to show and the actions on that info message. Example:  

"You have not chosen the registry to which you deploy these containers. What would you like to do? " Option 1: "Create a more suitable registry" Option 2: "Redo this step" Option 3: "Skip this step hereon"

---

I feel having a flag to indicate whether we cancel on empty responses would be ideal. We default to not cancelling.

EricJizbaMSFT:
FYI I fixed the build breaks in my most recent PR, so they should pass once you merge with master. See this commit for more info: https://github.com/Microsoft/vscode-azuretools/commit/d1f26cdd85336730777d7253acb37b8293413a3d

I believe `travis_terminate` used to default to an exit code of "0", but seems like now we have to explicitly specify it

StephenWeatherford:
We'll need to discuss.

PrashanthCorp:
Ping. The change you request (not throwing UserCancelled unless user hits Escape) is the current state of the PR. Are there other requested changes I didn't pick up?

ui/index.d.ts
```diff
@@ -279,14 +278,24 @@ export interface IParsedError {
...
     /**
-     * Shows a selection list.
-     * Automatically persists the 'recently used' item and displays that at the top of the list
-     *
-     * @param items An array of items, or a promise that resolves to an array of items.
-     * @param options Configures the behavior of the selection list.
-     * @throws `UserCancelledError` if the user cancels.
-     * @return A promise that resolves to the item the user picked.
-     */
+    * Shows a selection list.
```
StephenWeatherford: Shows a multi-selection list

PR #262, opened 2018-09-13
 Move linux consumption deployment into shared package

EricJizbaMSFT:
I split this up into two commits. The first commit just copies Nathan's code from the func repo. The second commit contains everything that I actually changed:
https://github.com/Microsoft/vscode-azuretools/commit/8036a8f5febc268014d9d6c7a62afed48435625b

Nathan and I discussed this refactoring before he left, but he didn't have time to get to it. This scenario is specific to functions, but it makes sense to have the code here so that we can leverage all the existing telemetry/logging/etc. that lives with this deploy code

### DIFF
--- a/appservice/src/deploy/deploy.ts
+++ b/appservice/src/deploy/deploy.ts
@@ -17,6 +17,8 @@ import { localGitDeploy } from './localGitDeploy';
 
 export async function deploy(client: SiteClient, fsPath: string, configurationSectionName: string, telemetryProperties?: TelemetryProperties): Promise<void> {
     const config: SiteConfigResource = await client.getSiteConfig();
+    // We use the AppServicePlan in a few places, but we don't want to delay deployment, so start the promise now and save as a const
+    const aspPromise: Promise<AppServicePlan> = client.getAppServicePlan();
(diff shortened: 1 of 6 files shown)


appservice/src/deploy/deploy.ts
```diff
@@ -37,7 +39,7 @@ export async function deploy(client: SiteClient, fsPath: string, configurationSe
                 () => {
                     // ignore
                 });
-            client.getAppServicePlan().then(
+            aspPromise.then(
```
StephenWeatherford: Don't you need to have failure code, or else do an await?

EricJizbaMSFT: Lol you wrote this "// ignore" failure code

We decided to ignore errors because it's just being used for telemetry in this case and we don't want telemetry to block deployment

EricJizbaMSFT: When I use `aspPromise` elsewhere (aka when it's not just being used for telemetry), I use an await

StephenWeatherford: Ah, I see it now.

PR #265, opened 2018-09-19
Add createTreeItemsWithErrorHandling

EricJizbaMSFT:
See documentation I added to the function in 'index.d.ts' for more info.

Mostly copied from functions repo, but made more generic and improved based on this comment: https://github.com/Microsoft/vscode-azurefunctions/pull/611#discussion_r218642240

Here are some screenshots from functions:
screen shot 2018-09-19 at 9 04 47 am
screen shot 2018-09-19 at 9 05 08 am
screen shot 2018-09-19 at 9 08 30 am


StephenWeatherford:
Yay!  Been wishing for something like this.  Do you intend to check this in today?  Wouldn't mind spending some time playing with it.

EricJizbaMSFT:
Yeah was planning on checking in today. Was going to send out a PR for functions shortly

ui/index.d.ts
```diff
@@ -155,6 +155,20 @@ export interface IAzureParentTreeItem extends IAzureTreeItem, IChildProvider {
     pickTreeItem?(expectedContextValue: string): IAzureTreeItem | undefined;
 }
 
+/**
+ * Used to ensure a single invalid object does not prevent display of other valid objects
+ * Invalid objects will be shown with the error and the object's name. If the name cannot be determined for any invalid objects, a TreeItem will be added to the end with a generic label like "Some items cannot be displayed"
+ * @param sourceArray The collection of source objects before converting to TreeItems
+ * @param invalidContextValue The context value to use for invalid source objects
+ * @param createTreeItem A function that converts a source object to a TreeItem
```
StephenWeatherford: Note that returning undefined from createTreeItem will simply skip that item (not added to tree)

ui/src/treeDataProvider/createTreeItemsWithErrorHandling.ts
```diff
@@ -0,0 +1,77 @@
...
+            } else if (error && !unknownError) {
+                unknownError = error;
+            }
+        }
+    }));
+
+    if (unknownError) {
+        // Display a generic error node if there are any unknown items. Only the first error will be displayed
+        const message: string = localize('cantShowItems', 'Some items could not be displayed');
+        treeItems.push(new InvalidTreeItem(message, unknownError, `${invalidContextValue}Unknown`, ''));
```
StephenWeatherford: I don't think we should be inventing a new context value, they might have code expecting only certain ones.  You could always special-case this in telemetry if you wanted.  It would actually be good (maybe later) to special-case InvalidTreeItem in telemetry so we can automatically identify error nodes rather than needing to look at an unknown set of context values.

EricJizbaMSFT: Since InvalidTreeItem throws the error in it's implementation of 'getChildren', this is already tracked as a part of our normal error telemetry for the "AzureTreeDataProvider.getChildren" event [here](https://github.com/Microsoft/vscode-azuretools/blob/master/ui/src/treeDataProvider/AzureTreeDataProvider.ts#L101). Admittedly, this only gets tracked if they expand the invalid node, but I thought that would be easier than special casing anything.

PR #270, opened 2018-09-24
Improve parseError

EricJizbaMSFT:
- Handle the case where `response.body` has the actual error and `error.message` is just a failure to parse `response.body`
- Check for 'statusCode' property
- Ensure 'errorType' is actually a string. I also changed 'equal' to 'strictEqual' in the tests because previously `401` (a number) was passing the test even though `"401"` (a string) was expected

Fixes https://github.com/Microsoft/vscode-azurefunctions/issues/648

### DIFF
--- a/ui/src/parseError.ts
+++ b/ui/src/parseError.ts
@@ -67,7 +67,7 @@ export function parseError(error: any): IParsedError {
 
 function convertCodeToError(errorType: string | undefined): string | undefined {
     if (errorType) {
-        const code: number = Number(errorType);
+        const code: number = parseInt(errorType, 10);
         if (!isNaN(code)) {
             return localize('failedWithCode', 'Failed with code "{0}".', code);
         }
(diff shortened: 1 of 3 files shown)


ui/src/parseError.ts
```diff
@@ -78,11 +89,12 @@ function parseIfJson(o: any): any {
 }
 
 function getMessage(o: any, defaultMessage: string): string {
-    return (o && (o.message || o.Message)) || defaultMessage;
+    return (o && (o.message || o.Message || (typeof parseIfJson(o.body) === 'string' && o.body))) || defaultMessage;
```
StephenWeatherford: Number('') = 0
Number(null) = 0
etc.

I think you want parseInt

EricJizbaMSFT: Shouldn't matter for those two cases since we already check `if (errorType) {` the line above, but changed anyways since I don't have a strong pref.

StephenWeatherford: There are probably other cases, the point was it tries to convert into a number so is probably overpermissive.

PR #274, opened 2018-10-01
Turn on strict null checks in app service package

EricJizbaMSFT:


appservice/src/SiteClient.ts
```diff
@@ -84,89 +82,89 @@ export class SiteClient {
     }
 
     public async stop(): Promise<void> {
-        this.isSlot ?
+        this.slotName ?
```
StephenWeatherford: So why the change to checking slotName instead of isSlot?  Don't like that there are two members dependent on each other but there is no assertion verifying this dependence.

EricJizbaMSFT: I can change `isSlot` to be `!!this.slotName`. There's no need to have different logic.

The change was so that the compiler wouldn't complain when passing `this.slotName` to `this._client.webApps.startSlot`, which doesn't allow `undefined`

appservice/src/connectToGitHub.ts
```diff
@@ -61,15 +62,15 @@ export async function connectToGitHub(node: IAzureNode, client: SiteClient): Pro
...
 
         if (repoQuickPick.url === requestOptions.nextLink) {
-            requestOptions.url = requestOptions.nextLink;
+            requestOptions.url = nonNullProp(requestOptions, 'nextLink');
             // remove the stale Load More quick pick
             repoQuickPicks.pop();
             const moreGitHubRepos: Object[] = await getJsonRequest(requestOptions, node);
             repoQuickPicks = repoQuickPicks.concat(createQuickPickFromJsons(moreGitHubRepos, 'name', undefined, ['url', 'html_url']));
         } else {
             repoSelected = true;
```
StephenWeatherford: Where's repoSelected  defined?

EricJizbaMSFT: Second line of this `connectToGitHub` function. I changed it from a 'while' to a 'do/while' because I was getting this error:
screen shot 2018-10-01 at 11 49 54 am

Even though the while loop would've always been run once since `repoSelected` is set to false by default.

appservice/src/getKuduClient.ts
```diff
@@ -7,22 +7,18 @@ import { AppServicePlan, User } from 'azure-arm-website/lib/models';
...
-        const asp: AppServicePlan = await client.getAppServicePlan();
-        const notSupportedLinux: string = localize('notSupportedLinux', 'This operation is not supported by App Service plans with kind "{0}" and sku tier "{1}".', client.kind, asp.sku.tier);
-        throw new Error (notSupportedLinux);
+        const asp: AppServicePlan | undefined = await client.getAppServicePlan();
+        const notSupportedLinux: string = localize('notSupportedLinux', 'This operation is not supported by App Service plans with kind "{0}" and sku tier "{1}".', client.kind, asp && asp.sku && asp.sku.tier);
+        throw new Error(notSupportedLinux);
     }
     const user: User = await client.getWebAppPublishCredential();
-    if (!user.publishingUserName || !user.publishingPassword) {
-        throw new ArgumentError(user);
```
StephenWeatherford: Should we change nonNullProp to throw ArgumentError?  Does anybody depend on that?

PR #280, opened 2018-10-11
Remove creating resource prompts

nturinski:
Created a progress property on wizardContext that is leveraged to update one information window reducing the total number of prompts.

image

image

### DIFF
--- a/appservice/src/createAppService/AppServicePlanCreateStep.ts
+++ b/appservice/src/createAppService/AppServicePlanCreateStep.ts
@@ -5,7 +5,6 @@
 
 import { WebSiteManagementClient } from 'azure-arm-website';
 import { AppServicePlan } from 'azure-arm-website/lib/models';
-import { ProgressLocation, window } from 'vscode';
 import { addExtensionUserAgent, AzureWizardExecuteStep } from 'vscode-azureextensionui';
 import { ext } from '../extensionVariables';
 import { localize } from '../localize';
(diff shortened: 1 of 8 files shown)


appservice/src/createAppService/AppServicePlanCreateStep.ts
```diff
@@ -21,29 +20,28 @@ export class AppServicePlanCreateStep extends AzureWizardExecuteStep<IAppService
...
                     ext.outputChannel.appendLine(creatingAppServicePlan);
-                    window.showInformationMessage(creatingAppServicePlan);
                     wizardContext.plan = await client.appServicePlans.createOrUpdate(rgName, newPlanName, {
                         kind: getAppServicePlanModelKind(wizardContext.newSiteKind, nonNullProp(wizardContext, 'newSiteOS')),
                         sku: nonNullProp(wizardContext, 'newPlanSku'),
                         location: nonNullValueAndProp(wizardContext.location, 'name'),
                         reserved: wizardContext.newSiteOS === WebsiteOS.linux  // The secret property - must be set to true to make it a Linux plan. Confirmed by the team who owns this API.
                     });
-                    window.showInformationMessage(createdAppServicePlan);
+                    if (wizardContext.progress) {wizardContext.progress.report({message: createdAppServicePlan}); }
```
EricJizbaMSFT: Can you do a gif of this experience rather than just a screenshot? Or maybe just demo for me when you're back in the office?

I'm guessing several of the messages (like `createdAppServicePlan`) flash up quickly without ever being readable. As a general rule, if the message doesn't end in `...`, I don't think it should be included in 'progress'.

appservice/src/createAppService/IAppServiceWizardContext.ts
```diff
@@ -58,4 +59,9 @@ export interface IAppServiceWizardContext extends IResourceGroupWizardContext, I
      * By specifying this in the context, we can ensure that Azure is only queried once for the entire wizard
      */
     plansTask?: Promise<AppServicePlan[]>;
+    /**
+     * Progress to report the status of the wizard. By leveraging progress.report,
+     * the information window can be reused rather than creating a new prompt per message.
+     */
+    progress?: Progress<{message: string}>;
```
EricJizbaMSFT: This shouldn't be necessary as long as you depend on v0.18.2 of the ui package

appservice/src/createAppService/SiteCreateStep.ts
```diff
@@ -29,11 +29,11 @@ export class SiteCreateStep extends AzureWizardExecuteStep<IAppServiceWizardCont
     public async execute(wizardContext: IAppServiceWizardContext): Promise<IAppServiceWizardContext> {
         if (!wizardContext.site) {
             const creatingNewApp: string = localize('CreatingNewApp', 'Creating {0} "{1}"...', getAppKindDisplayName(wizardContext.newSiteKind), wizardContext.newSiteName);
-            await window.withProgress({ location: ProgressLocation.Notification, title: creatingNewApp }, async (): Promise<void> => {
-                ext.outputChannel.appendLine(creatingNewApp);
-                const client: WebSiteManagementClient = new WebSiteManagementClient(wizardContext.credentials, wizardContext.subscriptionId, wizardContext.environment.resourceManagerEndpointUrl);
-                addExtensionUserAgent(client);
-                wizardContext.site = await client.webApps.createOrUpdate(nonNullValueAndProp(wizardContext.resourceGroup, 'name'), nonNullProp(wizardContext, 'newSiteName'), {
+            if (wizardContext.progress) { wizardContext.progress.report({message: creatingNewApp}); }
```
EricJizbaMSFT: Did you make sure to try out the functions scenario? I feel like there's some function-specific steps that you didn't change (storage account?)

Also make sure to try 'appService.advancedCreation' if you didn't already.

nturinski:
Do you think we can just ditch a majority of these prompts for the non-advanced creation scenario?  I feel like it just distracts the user from things that they probably do not care about if they're not using advanced create.

EricJizbaMSFT:
No preference on prompting for all of them - but I think it should still be in the output channel

nturinski:
Closing in favor of: https://github.com/Microsoft/vscode-azuretools/pull/301

PR #283, opened 2018-10-15
Minor fixes to tree found while updating Cosmos extension

EricJizbaMSFT:
Also add icons to the "Load More..." actions

In tree:
screen shot 2018-10-15 at 9 24 08 am

In treeItemPicker:
screen shot 2018-10-15 at 9 23 55 am

I realize they're slightly different icons, but one has to use [octicons](https://octicons.github.com/) and one has to use normal svgs.


StephenWeatherford:
Nice.

PR #284, opened 2018-10-15
Added getParent method to AzureTreeDataProvider

Omelianenko:
Need this for reveal command


ui/index.d.ts
```diff
@@ -68,6 +68,17 @@ export declare class AzureTreeDataProvider<TRoot = ISubscriptionRoot> implements
      * Should not be called directly
      */
     public dispose(): void;
+
+    /**
+     * Optional method to return the parent of `element`.
+     * Return `null` or `undefined` if `element` is a child of root.
```
StephenWeatherford: FYI, you've got it typed as undefined (which is what I'd expect), not null or undefined

PR #287, opened 2018-10-17
Allow '0' to turn off DEBUGTELEMETRY

StephenWeatherford:


ui/src/createTelemetryReporter.ts
```diff
@@ -12,7 +12,7 @@ import { ITelemetryReporter } from '../index';
 import { extInitialized } from './extensionVariables';
 import { getPackageInfo } from './getPackageInfo';
 
-const debugTelemetryEnabled: boolean = !!process.env.DEBUGTELEMETRY;
+const debugTelemetryEnabled: boolean = (process.env.DEBUGTELEMETRY || '') === 'true' || parseInt(process.env.DEBUGTELEMETRY || '', 10) > 0;
```
EricJizbaMSFT: parseInt might throw an error, right? Nathan did this for a similar ENV variable, which I kinda like better: https://github.com/Microsoft/vscode-azurefunctions/blob/master/test/global.test.ts#L26
```
!/^(false|0)?$/i.test(process.env.ENABLE_LONG_RUNNING_TESTS);
```

StephenWeatherford: parseInt() returns NaN instead of throwing errors, but I think I like Nathan's better, too.

PR #289, opened 2018-10-18
Nat/switch to local git

nturinski:


appservice/src/deploy/localGitDeploy.ts
```diff
@@ -58,3 +59,23 @@ export async function localGitDeploy(client: SiteClient, fsPath: string): Promis
...
+}
+
+async function verifyNoRunFromPackageSetting(client: SiteClient): Promise<void> {
+    let updateSettings: boolean = false;
+    const applicationSettings: StringDictionary = await client.listApplicationSettings();
+    for (const key of Object.keys(RunFromPackageAliases)) {
+        const runFromPackageSettingName: RunFromPackageAliases = <RunFromPackageAliases>RunFromPackageAliases[key];
+        if (applicationSettings.properties && applicationSettings.properties[runFromPackageSettingName]) {
+            delete applicationSettings.properties[runFromPackageSettingName];
+            updateSettings = true;
```
EricJizbaMSFT: nit: Add:
```
ext.outputChannel.appendLine(formatDeployLog(client, ...
```
with a message that you're removing the setting

appservice/src/deploy/localGitDeploy.ts
```diff
@@ -58,3 +59,23 @@ export async function localGitDeploy(client: SiteClient, fsPath: string): Promis
...
+enum RunFromPackageAliases {
+    fromPackage = 'WEBSITE_RUN_FROM_PACKAGE',
+    fromZip = 'WEBSITE_RUN_FROM_ZIP'
+}
+
+async function verifyNoRunFromPackageSetting(client: SiteClient): Promise<void> {
+    let updateSettings: boolean = false;
+    const applicationSettings: StringDictionary = await client.listApplicationSettings();
+    for (const key of Object.keys(RunFromPackageAliases)) {
+        const runFromPackageSettingName: RunFromPackageAliases = <RunFromPackageAliases>RunFromPackageAliases[key];
```
EricJizbaMSFT: nit: rather than doing all this stuff with Object.keys and whatever, you could just make it an array:
```
const runFromPackageAliases: string[] = ...
```

appservice/src/deploy/localGitDeploy.ts
```diff
@@ -26,6 +26,7 @@ export async function localGitDeploy(client: SiteClient, fsPath: string): Promis
     const password: string = nonNullProp(publishCredentials, 'publishingPassword');
     const remote: string = `https://${username}:${password}@${client.gitUrl}`;
     const localGit: git.SimpleGit = git(fsPath);
+    await verifyNoRunFromPackageSetting(client);
```
EricJizbaMSFT: Can we move this after we check for uncommitted changes? Feels unnecessary to update the application settings unless we know they're actually going to deploy

appservice/src/deploy/localGitDeploy.ts
nturinski: It's not worth promping the user, right?

PR #292, opened 2018-10-23
Add CODEOWNERS

EricJizbaMSFT:
See https://github.com/Microsoft/vscode-azurefunctions/pull/725 for more info. Seems to have been working in the functions repo

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+* @Microsoft/vscodeazuretools


nturinski:
Do you need to add `.github` to the `.gitignore`?

EricJizbaMSFT:
There is no `.vscodeignore` in this repo. There are `.npmignore` files, but those are in each subfolder so I don't have to worry about that since this is in the root

PR #293, opened 2018-10-23
Allow undefined in telemetry properties

EricJizbaMSFT:
It's bothered me for a while that telemetry properties were required. For example some times in testing it would be easier to mock IActionContext if these were optional: https://github.com/Microsoft/vscode-azurefunctions/blob/master/test/createNewProject.test.ts#L168

Also some times we set telemetry properties to a value that might be undefined. It would be easier if we didn't have to cast that to a string: https://github.com/Microsoft/vscode-azurefunctions/blob/master/src/commands/createNewProject/validateFunctionProjects.ts#L37

### DIFF
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -72,9 +72,9 @@ export declare class AzureTreeDataProvider<TRoot = ISubscriptionRoot> implements
     /**
      * Optional method to return the parent of `element`.
      * Return `null` or `undefined` if `element` is a child of root.
-     * 
+     *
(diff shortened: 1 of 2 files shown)


nturinski:
I'd like to see @StephenWeatherford's opinion on this, but I agree that having properties be optional sounds useful, especially for testing as you mentioned.

StephenWeatherford:
There was a reason for this but I don't remember what it was, and I hate the current situation, too, so I'm all for this.

PR #294, opened 2018-10-24
Logic to display gitUrl

nturinski:
Fixes https://github.com/Microsoft/vscode-azureappservice/issues/619

The publishing credentials we use for local git deploy is on a web app level and is guaranteed.

The credentials shown as the azure remote (and the one Burke asks about) is a user level credential that can be used across all apps and is what is displayed.  User level passwords do not work for the web-app level credentials.

This username is set in Deployment Options and once set, cannot be removed (but can be altered).  I haven't tested it with somebody who has clean credentials, so I'm not sure what it would return (but I'm assuming either undefined or "")

Web-app level:
image

User level:
image

### DIFF
--- a/appservice/src/SiteClient.ts
+++ b/appservice/src/SiteClient.ts
@@ -231,4 +231,8 @@ export class SiteClient {
             }
         }
     }
+
+    public async getPublishingUser(): Promise<User> {
+        return await this._client.getPublishingUser({});
+    }
 }
--- a/appservice/src/editScmType.ts
+++ b/appservice/src/editScmType.ts
@@ -3,7 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { SiteConfigResource } from 'azure-arm-website/lib/models';
+import { WebSiteManagementModels } from 'azure-arm-website';
 import { AzureTreeItem, IAzureQuickPickItem, IAzureQuickPickOptions, UserCancelledError } from 'vscode-azureextensionui';
 import { connectToGitHub } from './connectToGitHub';
 import { ext } from './extensionVariables';
@@ -13,7 +13,7 @@ import { SiteClient } from './SiteClient';
 import { nonNullProp } from './utils/nonNull';
 
 export async function editScmType(client: SiteClient, node: AzureTreeItem): Promise<string | undefined> {
-    const config: SiteConfigResource = await client.getSiteConfig();
+    const config: WebSiteManagementModels.SiteConfigResource = await client.getSiteConfig();
     const newScmType: string = await showScmPrompt(nonNullProp(config, 'scmType'));
     if (newScmType === ScmType.GitHub) {
         if (config.scmType !== ScmType.None) {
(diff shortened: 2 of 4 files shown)


appservice/src/SiteClient.ts
```diff
@@ -231,4 +231,8 @@ export class SiteClient {
             }
         }
     }
+
+    public async getPublishingUser(): Promise<User> {
+        return await this._client.getPublishingUser({});
```
EricJizbaMSFT: Why are you passing an empty object? Is there a slot specific method for this?

appservice/src/editScmType.ts
```diff
@@ -23,10 +23,18 @@ export async function editScmType(client: SiteClient, node: AzureTreeItem): Prom
         await connectToGitHub(node, client);
     } else {
         config.scmType = newScmType;
-        // to update one property, a complete config file must be sent
+        // to update one property, a complete config file must be sentnpm
```
EricJizbaMSFT: "sentnpm"

appservice/src/editScmType.ts
```diff
@@ -23,10 +23,18 @@ export async function editScmType(client: SiteClient, node: AzureTreeItem): Prom
...
         await client.updateConfiguration(config);
     }
     ext.outputChannel.appendLine(localize('deploymentSourceUpdated,', 'Deployment source for "{0}" has been updated to "{1}".', client.fullName, newScmType));
+    if (newScmType === ScmType.LocalGit) {
+        const user: WebSiteManagementModels.User = await client.getPublishingUser();
+        if (user.publishingUserName) {
+            // first time users must set up deployment credentials via the Portal or they will not have a UserName
+            const gitCloneUri: string = `https://${user.publishingUserName}@${client.gitUrl}`;
```
EricJizbaMSFT: What's the relationship between this uri and publishCredentials.scmUri?

appservice/src/editScmType.ts
```diff
@@ -23,10 +23,18 @@ export async function editScmType(client: SiteClient, node: AzureTreeItem): Prom
...
+            ext.outputChannel.appendLine(localize('gitCloneUri', 'Git Clone Uri for "{0}": "{1}"', client.fullName, gitCloneUri));
```
EricJizbaMSFT: Is https the only supported uri or could users use an ssh remote uri?

appservice/src/SiteClient.ts
nturinski: It requires an options object, all options are optional.  This returns the publishing User associated to Deployment Options (like in the screen shot) rather than the app specific username.

appservice/src/editScmType.ts
nturinski: The username from `getPublishingUser` is universally accepted on a subscription level.  The `publishCredentials` username and scmUri are both on the app level.  If users use that uri and enter the credentials they set on the portal for user level access, it'll fail authentication so we have to provide them with the proper username.

appservice/src/editScmType.ts
nturinski: As far as I've seen, every uri has been https for the git remote url, but I could be wrong about that.

EricJizbaMSFT: If it supports ssh I'd prefer if we printed that as well

nturinski: I only see SSH support if you deploy via DevOps: https://docs.microsoft.com/en-us/azure/devops/repos/git/use-ssh-keys-to-authenticate?view=vsts

But I don't think that'd be trivial to implement.

appservice/src/SiteClient.ts
nturinski: Oh, and no, there is no slot option as this is a `client` function, not `client.webapp`

PR #295, opened 2018-10-24
Add versioning to extension APIs

EricJizbaMSFT:
Waiting for a few of Andrii's PRs to get merged before I actually test this - but wanted to get eyes on this as soon as possible. We ideally won't change the `AzureExtensionProvider` contract ever in the future so I want to make sure multiple people agree on it.

### DIFF
--- a/ui/gulpfile.js
+++ b/ui/gulpfile.js
@@ -1,0 +1,19 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+const gulp = require('gulp');
+const path = require('path');
+const cp = require('child_process');
+
+gulp.task('test', (cb) => {
+    const env = process.env;
+    env.DEBUGTELEMETRY = 1;
+    env.CODE_EXTENSIONS_PATH = path.join(__dirname, 'test', 'extension');
+    const cmd = cp.spawn('node', ['./node_modules/vscode/bin/test'], { stdio: 'inherit', env });
+    cmd.on('close', (code) => {
+        cb(code);
+    });
+});
+
--- a/ui/src/callWithTelemetryAndErrorHandling.ts
+++ b/ui/src/callWithTelemetryAndErrorHandling.ts
@@ -44,7 +44,7 @@ export function callWithTelemetryAndErrorHandlingSync<T>(callbackId: string, cal
         handleError(context, callbackId, error);
         return undefined;
     } finally {
-        handleFinally(context, callbackId, start);
+        handleTelemetry(context, callbackId, start);
     }
 }
 
@@ -57,7 +57,7 @@ export async function callWithTelemetryAndErrorHandling<T>(callbackId: string, c
         handleError(context, callbackId, error);
         return undefined;
     } finally {
-        handleFinally(context, callbackId, start);
+        handleTelemetry(context, callbackId, start);
     }
 }
 
@@ -99,7 +99,7 @@ function handleError(context: IActionContext, callbackId: string, error: any): v
     }
 }
 
-function handleFinally(context: IActionContext, callbackId: string, start: number): void {
+function handleTelemetry(context: IActionContext, callbackId: string, start: number): void {
     if (ext.reporter) {
         // For suppressTelemetry=true, ignore successful results
         if (!(context.suppressTelemetry && context.properties.result === 'Succeeded')) {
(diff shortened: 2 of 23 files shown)


ui/src/wrapApiWithVersioning.ts
```diff
@@ -0,0 +1,49 @@
...
+export function wrapApiWithVersioning(azExts: AzureExtension[]): AzureExtensionProvider {
+    for (const azExt of azExts) {
+        if (!semver.valid(azExt.apiVersion)) {
+            throw new Error(localize('invalidVersion', 'Invalid semver "{0}".', azExt.apiVersion));
+        }
+    }
+
+    const extensionId: string = getPackageInfo().extensionId;
+    return {
+        getExtension: async <T extends AzureExtension>(apiVersion: string): Promise<T> => getExtensionInternal<T>(azExts, extensionId, apiVersion)
```
StephenWeatherford: getApi?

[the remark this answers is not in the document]
EricJizbaMSFT: Probably but it won't be simple. Specifically I want to make sure we're not changing the api in any way when we wrap it with telemetry. It might be a bit thorny figuring out what is a property vs. function vs. async function and handling each appropriately. To that end - I've added unit tests to make sure I don't break stuff when I add telemetry (preferably in a separate PR)

EricJizbaMSFT:
A summary of the changes I just pushed:
1. Change return type of `getApi` from `Promise<T>` to `T`. I prefer to keep this as generic/simple as possible - which IMO means it shouldn't be async.
1. Add typing to `callWithTelemetryAndErrorHandling` and add a "Sync" version (needed to support the first bullet point)
    > NOTE: Make sure to review with whitespace off for this change
1. Enable unit tests that require "vscode"
1. Add unit tests for `wrapApiWithVersioning` and `callWithTelemetryAndErrorHandling`

ui/gulpfile.js
```diff
@@ -0,0 +1,19 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+const gulp = require('gulp');
+const path = require('path');
+const cp = require('child_process');
+
+gulp.task('test', (cb) => {
```
StephenWeatherford: Curious why you switched to gulp?

ui/src/callWithTelemetryAndErrorHandling.ts
```diff
@@ -33,54 +32,82 @@ export async function callWithTelemetryAndErrorHandling(callbackId: string, call
...
             }
+        });
+    }
+
+    if (context.rethrowError) {
+        throw error;
+    }
+}
+
+function handleFinally(context: IActionContext, callbackId: string, start: number): void {
```
StephenWeatherford: How about handleTelemetry?

ui/test/callWithTelemetryAndErrorHandling.test.ts
```diff
@@ -0,0 +1,66 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as assert from 'assert';
```
StephenWeatherford: Last time I tried doing this I have some issues because the vscode module wasn't defined, since this wasn't an actual extension. What did you do to solve this?

ui/gulpfile.js
EricJizbaMSFT: 1. I find it easier to work with gulp than scripts. For example I was trying to pass in "CODE_EXTENSIONS_PATH" as "./test/extension", but it wasn't working unless I passed in the _full_ path. Wasn't sure how to do that for a script, but it's easier to do (and debug) in javascript (with `path.join(__dirname ...`)
1. I also just find it more readable. Scripts start to get out of hand when you have multiple environment variables.

ui/test/callWithTelemetryAndErrorHandling.test.ts
EricJizbaMSFT: I just created a super simple "test" extension. That's what these files are for:
ui/test/extension/extension.js
ui/test/extension/package.json

EricJizbaMSFT: And then I use the typical "launch.json" that we use in the extension repos, except I pass "test/extension" in as `CODE_EXTENSIONS_PATH`

PR #298, opened 2018-10-29
Moved validateNewKeyInput from class to file

### LINKED ISSUE #297: Make validateNewKeyInput method in AppSettingsTreeItem open for external usage
Need this for AppService connection adding scenario

Omelianenko:
Fix #297

### DIFF
--- a/appservice/src/tree/AppSettingsTreeItem.ts
+++ b/appservice/src/tree/AppSettingsTreeItem.ts
@@ -10,7 +10,7 @@ import { ext } from '../extensionVariables';
 import { AppSettingTreeItem } from './AppSettingTreeItem';
 import { ISiteTreeRoot } from './ISiteTreeRoot';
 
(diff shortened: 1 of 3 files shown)


Omelianenko:
Doesn't create any AppSetting:
image
Compile error from appservice:
image

Omelianenko:
Should I copy-paste this code in AppService then? Here I have exactly what I need and can't see any other way to run it

appservice/src/tree/AppSettingsTreeItem.ts
```diff
@@ -10,6 +10,23 @@ import { ext } from '../extensionVariables';
 import { AppSettingTreeItem } from './AppSettingTreeItem';
 import { ISiteTreeRoot } from './ISiteTreeRoot';
 
+export function validateNewKeyInput(settings: StringDictionary, newKey?: string, oldKey?: string): string | undefined {
```
EricJizbaMSFT: Can you rename this to `validateAppSettingKey`

PR #301, opened 2018-10-30
Remove information window for creating rg/asp prompts

nturinski:
The prompts were kind of cluttering the UI according to Chris: https://github.com/Microsoft/vscode-azureappservice/issues/623

I was considering showing the informationWindow for the advancedCreate scenario, but it didn't seem worth the trade-off for the added complexity it'd add to the wizard steps.


nturinski:
Tested with Cosmos extension-- doesn't seem to affect creating accounts, databases, collections, or documents.

PR #304, opened 2018-11-06
Make webpack-friendly

StephenWeatherford:
Webpack works best when there are no dynamic imports (which getPackageInfo does). Instead, have the extension pass it in.
Also clean up the initialization of extension variables.


StephenWeatherford:
What?!  Looks like a nail to me!

StephenWeatherford:
Superseded by #305

PR #305, opened 2018-11-06
Make webpack friendly and make registering vars more logical

StephenWeatherford:


ui/package.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "vscode-azureextensionui",
     "author": "Microsoft Corporation",
-    "version": "0.18.5",
+    "version": "0.19.0",
```
EricJizbaMSFT: I don't think you have to do a "breaking" change anymore, right? Also I published 0.18.6 just a bit ago - so you'd have to do 0.18.7

EricJizbaMSFT: okeydokey

PR #312, opened 2018-11-12
Fix "instanceof" when linking both ui and appservice package

EricJizbaMSFT:
The full repro steps are hard to describe but @nturinski knows what I'm talking about, right?

### DIFF
--- a/ui/src/errors.ts
+++ b/ui/src/errors.ts
@@ -6,8 +6,16 @@
 import { localize } from "./localize";
 
 export class UserCancelledError extends Error {
+    private readonly _isUserCancelledError: boolean = true;
+
     constructor() {
         super(localize('userCancelledError', 'Operation cancelled.'));
+    }
+
+    // Workaround to fix "instanceof UserCancelledError" when testing package with "npm link"
+    // tslint:disable-next-line:function-name no-any
+    public static [Symbol.hasInstance](instance: any): boolean {
+        return instance !== null && typeof instance === 'object' && !!(<UserCancelledError>instance)._isUserCancelledError;
     }
 }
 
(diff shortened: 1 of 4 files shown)


nturinski:
Right. 👍

ui/src/treeDataProvider/AzureParentTreeItem.ts
```diff
@@ -24,6 +24,14 @@ export abstract class AzureParentTreeItem<TRoot = ISubscriptionRoot> extends Azu
     private _loadMoreChildrenTask: Promise<void> | undefined;
     private _initChildrenTask: Promise<void> | undefined;
 
+    private readonly _isAzureParentTreeItem: boolean = true;
```
nturinski: Should this also have `_isAzureTreeItem = true`?  since it extends `AzureTreeItem`?

ui/src/errors.ts
```diff
@@ -6,9 +6,17 @@
...
 export class UserCancelledError extends Error {
+    private readonly _isUserCancelledError: boolean = true;
+
     constructor() {
         super(localize('userCancelledError', 'Operation cancelled.'));
     }
+
+    // Workaround to fix "instanceof UserCancelledError" when testing package with "npm link"
+    // tslint:disable-next-line:function-name no-any
+    public static [Symbol.hasInstance](instance: any): boolean {
```
nturinski: Can you explain what this is?? I'm guessing it's overloading, but just curious as to how.

EricJizbaMSFT: Just overwrites default behavior of "instanceof"
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/hasInstance

> The Symbol.hasInstance well-known symbol is used to determine if a constructor object recognizes an object as its instance. The instanceof operator's behavior can be customized by this symbol.

ui/src/treeDataProvider/AzureParentTreeItem.ts
EricJizbaMSFT: I mean it does already have that property... because it extends AzureTreeItem...

nturinski: 🦆

PR #314, opened 2018-11-13
Deployments implementation

nturinski:
Moved all shared code from this PR (https://github.com/Microsoft/vscode-azureappservice/pull/697/files) to `vscode-azuretools` incorporating the PR comments.


appservice/src/tree/DeploymentTreeItem.ts
```diff
@@ -0,0 +1,103 @@
...
+    public active: boolean;
+    public receivedTime: Date;
+    public parent: DeploymentsTreeItem;
+    private _kuduClient: KuduClient;
+    private _deployResult: DeployResult;
+
+    constructor(parent: DeploymentsTreeItem, deployResult: DeployResult, kuduClient: KuduClient) {
+        super(parent);
+        this._deployResult = deployResult;
+        if (!this._deployResult.receivedTime || !this._deployResult.message || this._deployResult.active === undefined) {
```
EricJizbaMSFT: Now that you're in this repo, you can leverage 'nonNullProp' and 'nonNullValue' like here: https://github.com/Microsoft/vscode-azuretools/blob/master/appservice/src/SiteClient.ts#L51

I plan to move that over to other repos, but hasn't made it to app service repo yet.

appservice/src/tree/DeploymentsTreeItem.ts
```diff
@@ -0,0 +1,80 @@
...
+import { ext } from '../extensionVariables';
+import { getKuduClient } from '../getKuduClient';
+import { ScmType } from '../ScmType';
+import { DeploymentTreeItem } from './DeploymentTreeItem';
+import { ISiteTreeRoot } from './ISiteTreeRoot';
+
+export class DeploymentsTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
+    public static contextValueConnected: string = 'deploymentsConnected';
+    public static contextValueUnconnected: string = 'deploymentsUnconnected';
+    public contextValue: string = DeploymentsTreeItem.contextValueUnconnected;
```
EricJizbaMSFT: nit: Don't set a default value for contextValue if you're just going to overwrite it in the constructor

appservice/src/tree/DeploymentsTreeItem.ts
```diff
@@ -0,0 +1,80 @@
...
+    public parent: AzureParentTreeItem<ISiteTreeRoot>;
+    public readonly label: string = 'Deployments';
```
EricJizbaMSFT: localize

appservice/src/tree/DeploymentTreeItem.ts
```diff
@@ -0,0 +1,103 @@
...
+            throw new Error('Invalid Deployment Result.');
+        }
+        this._kuduClient = kuduClient;
```
EricJizbaMSFT: Don't store something like kuduClient. The credentials could've changed so you should create a new client each time

appservice/src/tree/DeploymentsTreeItem.ts
```diff
@@ -0,0 +1,80 @@
...
+        return false;
+    }
+
+    public async loadMoreChildrenImpl(_clearCache: boolean): Promise<DeploymentTreeItem[] | GenericTreeItem<ISiteTreeRoot>[]> {
+        const siteConfig: SiteConfig = await this.root.client.getSiteConfig();
+        if (siteConfig.scmType === ScmType.GitHub || siteConfig.scmType === ScmType.LocalGit) {
+            const kuduClient: KuduClient = await getKuduClient(this.root.client);
+            const deployments: DeployResult[] = await kuduClient.deployment.getDeployResults();
+            return deployments.map((deployResult: DeployResult) => {
+                return new DeploymentTreeItem(this, deployResult, kuduClient);
```
EricJizbaMSFT: Probably worth leveraging `createTreeItemsWithErrorHandling` so that one invalid deployResult doesn't prevent the user from viewing all valid deployResults (and because I just don't trust kudu to always return valid stuff)

appservice/src/tree/DeploymentTreeItem.ts
```diff
@@ -0,0 +1,103 @@
...
+
+    public async showDeploymentLogs(): Promise<void> {
+        const logData: string = await this.getDeploymentLogs();
+        const logDocument: TextDocument = await workspace.openTextDocument({ content: logData, language: 'log' });
+        await window.showTextDocument(logDocument);
+    }
+
+    private parseLogEntry(logEntry: LogEntry): string {
+        if (logEntry.logTime && logEntry.message) {
+            return `${logEntry.logTime.toISOString()} - ${logEntry.message} ${os.EOL}`;
```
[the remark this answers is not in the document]
nturinski: Got it                                                                                         boss

appservice/src/tree/DeploymentTreeItem.ts
```diff
@@ -21,28 +21,24 @@ export class DeploymentTreeItem extends AzureTreeItem<ISiteTreeRoot> {
...
+            message = `${message.substring(0, 50)}...`;
         }
+        this.label = `${this.id.substring(0, 7)} - ${message}`;
 
-        if (this.active) {
+        if (active) {
             this.description = 'Active';
-        }
-        if (!this._deployResult.lastSuccessEndTime && this._deployResult.complete) {
+        } else if (!this._deployResult.lastSuccessEndTime && this._deployResult.complete) {
```
EricJizbaMSFT: :P One of the points of "else if" is that you don't have to do `this._deployResult.complete` multiple times. Although probably need to switch the order of "Deploying..." and "Failed" cases

nturinski:
The UI (including description labels)
image

image

image

PR #318, opened 2018-11-15
Add TestOutputChannel to dev package

EricJizbaMSFT:
Re-routes output to the console instead of a VS Code output channel (which disappears after a test run has finished)

### DIFF
--- a/dev/src/TestOutputChannel.ts
+++ b/dev/src/TestOutputChannel.ts
@@ -1,0 +1,34 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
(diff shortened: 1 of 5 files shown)


dev/src/TestOutputChannel.ts
```diff
@@ -0,0 +1,34 @@
...
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { OutputChannel } from "vscode";
+
+export class TestOutputChannel implements OutputChannel {
+    public name: string = 'Extension Test Output';
+
+    public append(value: string): void {
+        console.log(value);
```
StephenWeatherford: Technically this is wrong, but I guess it's good enough for now.

EricJizbaMSFT: I assume because of the new line?

StephenWeatherford: yeah

PR #319, opened 2018-11-15
Update DeployStatus logic to be more accurate

nturinski:
Kudu has status codes on DeployResult that are represented as numbers.  We mapped their DeployStatus enum from here: https://github.com/projectkudu/kudu/blob/a13592e6654585d5c2ee5c6a05fa39fa812ebb84/Kudu.Contracts/Deployment/DeployStatus.cs to update our descriptions rather than relying on (inaccurate) custom logic.

Also added a refresh to the label to refresh during redeploying.  Intentionally does not have a context menu since I don't really think it'd be valuable for users.

### DIFF
--- a/appservice/src/tree/DeploymentTreeItem.ts
+++ b/appservice/src/tree/DeploymentTreeItem.ts
@@ -11,9 +11,19 @@ import KuduClient from 'vscode-azurekudu';
 import { DeployResult, LogEntry } from 'vscode-azurekudu/lib/models';
 import { ext } from '../extensionVariables';
 import { getKuduClient } from '../getKuduClient';
+import { localize } from '../localize';
 import { nonNullProp } from '../utils/nonNull';
 import { DeploymentsTreeItem } from './DeploymentsTreeItem';
 import { ISiteTreeRoot } from './ISiteTreeRoot';
+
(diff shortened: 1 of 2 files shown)


appservice/src/tree/DeploymentTreeItem.ts
```diff
@@ -55,14 +56,36 @@ export class DeploymentTreeItem extends AzureTreeItem<ISiteTreeRoot> {
...
+        }
+    }
+
     public async redeployDeployment(): Promise<void> {
         const redeploying: string = `Redeploying commit "${this.id}" to "${this.root.client.fullName}"`;
         const deployed: string = `Commit "${this.id}" has been redeployed to "${this.root.client.fullName}".`;
         window.withProgress({ location: ProgressLocation.Notification, title: redeploying }, async (): Promise<void> => {
             const kuduClient: KuduClient = await getKuduClient(this.root.client);
             ext.outputChannel.appendLine(redeploying);
+            const refreshingInteveral: NodeJS.Timer = setInterval(async () => { await this.refresh(); }, 1000); /* the status of the label changes during deployment so poll for that*/
```
EricJizbaMSFT: Wrap this in a try/finally so that `clearInterval` is called even if `await kuduClient.deployment.deploy(this.id);` throws an exception

PR #322, opened 2018-11-15
Add some telemetry to web app creation

nturinski:
Fixes https://github.com/Microsoft/vscode-azureappservice/issues/696
Sorry, I'm not sure if there's a preferred way we add telemetry so I just kind of threw it in where I thought it made sense.

### DIFF
--- a/appservice/src/createAppService/createAppService.ts
+++ b/appservice/src/createAppService/createAppService.ts
@@ -71,8 +71,6 @@ export async function createAppService(
             break;
         case AppKind.app:
             if (createOptions.advancedCreation) {
-                actionContext.properties.advancedCreation = 'true';
-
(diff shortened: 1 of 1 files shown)


appservice/src/createAppService/createAppService.ts
```diff
@@ -71,6 +71,8 @@ export async function createAppService(
             break;
         case AppKind.app:
             if (createOptions.advancedCreation) {
+                actionContext.properties.advancedCreation = 'true';
```
StephenWeatherford: Looks fine, although consider how you will distinguish between advancedCreation being undefined because it occurred before this checkin vs it wasn't advanced creation. Personally I'd consider 'false'.

PR #325, opened 2018-11-16
Add optional debounce property in registerCommand

nturinski:
I originally tried to have the context hold the lastClicked time, but of course, each individual event has its own context soooo.... I'm up for a more elegant solution though 😄


StephenWeatherford:
You know, I'm actually thinking we should ask vscode to do this.  The issue is really that people are double-clicking a tree item not meant to support it.  It seems questionable that they should allow that.  Or else that they should not be tolerant of the scenario.  Thoughts?
cc @EricJizbaMSFT

PR #326, opened 2018-11-19
Use RegExp to check for line breaks and trunc

nturinski:
My commit message only contained `\n` characters, but this RegExp should check for `\r`, `\n`, and `\r\n`.   Found the RegExp here:  https://stackoverflow.com/questions/10805125/how-to-remove-all-line-breaks-from-a-string

Fixes https://github.com/Microsoft/vscode-azureappservice/issues/713


appservice/src/tree/DeploymentTreeItem.ts
```diff
@@ -132,4 +133,14 @@ export class DeploymentTreeItem extends AzureTreeItem<ISiteTreeRoot> {
             return '';
         }
     }
+
+    private removeLineBreaks(message: string): string {
```
EricJizbaMSFT: nit: `removeLineBreaks` makes it sound like you're just flattening the message into one line. Suggest renaming to something like `getFirstLine`

PR #333, opened 2018-11-21
Show progress bar while loading GitHub Repos

nturinski:
I couldn't think of a more elegant solution than having a global array to concatenate the repositories.  I wasn't sure how to pass a Promise object and then resolve it and continue concatenate to it.


EricJizbaMSFT:
See https://github.com/Microsoft/vscode-azuretools/pull/334

nturinski:
Closing in favor #334

PR #336, opened 2018-11-21
Fix (recently used) for "Load more" quick pick

EricJizbaMSFT:
So I ran into quite the interesting problem when testing out this PR: https://github.com/Microsoft/vscode-azuretools/pull/334. If you keep clicking load more...you will get a quick pick like this 😂:
screen shot 2018-11-21 at 11 33 19 am

@fiveisprime has also complained in the past that it looks weird when "recently used" is the "detail" option like this:
screen shot 2018-11-21 at 12 09 09 pmscreen shot 2018-11-21 at 12 09 03 pm

So I changed it to look like this:
screen shot 2018-11-21 at 12 06 57 pmscreen shot 2018-11-21 at 12 07 11 pm
Thoughts? I think this looks better.

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -1,7 +1,7 @@
 {
     "name": "vscode-azureextensionui",
     "author": "Microsoft Corporation",
-    "version": "0.19.3",
+    "version": "0.19.4",
     "description": "Common UI tools for developing Azure extensions for VS Code",
     "tags": [
         "azure",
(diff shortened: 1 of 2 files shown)


nturinski:
For longer path files, it might look really odd to have it all the way at the end like that.  But I'm not super concerned about it and that's a pretty gross bug.

PR #338, opened 2018-11-26
Add the repo name to the connecting message

nturinski:
Fixes https://github.com/Microsoft/vscode-azureappservice/issues/722

The repoUrl is the `https://github.com/Microsoft/vscode-azuretools/`.  By removing `https://github.com/`, it should leave the repo name.

### DIFF
--- a/appservice/src/connectToGitHub.ts
+++ b/appservice/src/connectToGitHub.ts
@@ -16,8 +16,8 @@ import { signRequest } from './signRequest';
 import { SiteClient } from './SiteClient';
 import { nonNullProp } from './utils/nonNull';
 
-type gitHubOrgData = { repos_url?: string };
-type gitHubReposData = { repos_url?: string, url?: string, html_url?: string };
+type gitHubOrgData = { login?: string, repos_url?: string };
+type gitHubReposData = { name?: string, repos_url?: string, url?: string, html_url?: string };
 type gitHubLink = { prev?: string, next?: string, last?: string, first?: string };
 // tslint:disable-next-line:no-reserved-keywords
 type gitHubWebResource = WebResource & { resolveWithFullResponse?: boolean, nextLink?: string, type?: string };
@@ -40,7 +40,7 @@ export async function connectToGitHub(node: AzureTreeItem, client: SiteClient, c
     const gitHubUser: Object[] = await getJsonRequest(requestOptions, node, client, context);
     requestOptions.url = 'https://api.github.com/user/orgs';
     const gitHubOrgs: Object[] = await getJsonRequest(requestOptions, node, client, context);
-    const orgQuickPicks: IAzureQuickPickItem<{}>[] = createQuickPickFromJsons([gitHubUser], 'login', undefined, ['repos_url']).concat(createQuickPickFromJsons(gitHubOrgs, 'login', undefined, ['repos_url']));
+    const orgQuickPicks: IAzureQuickPickItem<{}>[] = createQuickPickFromJsons([gitHubUser], 'login', undefined, ['repos_url', 'login']).concat(createQuickPickFromJsons(gitHubOrgs, 'login', undefined, ['repos_url', 'login']));
     const orgQuickPick: gitHubOrgData = (await ext.ui.showQuickPick(orgQuickPicks, { placeHolder: 'Choose your organization.' })).data;
     requestOptions.url = nonNullProp(orgQuickPick, 'repos_url');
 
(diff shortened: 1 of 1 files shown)


appservice/src/connectToGitHub.ts
```diff
@@ -63,9 +63,12 @@ export async function connectToGitHub(node: AzureTreeItem, client: SiteClient, c
         isMercurial: false
     };
 
+    // tslint:disable-next-line:no-non-null-assertion
+    const repoName: string = siteSourceControl.repoUrl!.substring('https://github.com/'.length);
```
EricJizbaMSFT: How do you get the org name when showing the org quick pick? There has to be a simpler (and more reliable) way to do this...

appservice/src/connectToGitHub.ts
```diff
@@ -63,9 +63,11 @@ export async function connectToGitHub(node: AzureTreeItem, client: SiteClient, c
         isMercurial: false
     };
 
+    const repoName: string | undefined = orgQuickPick.login && repoQuickPick.name ? `${orgQuickPick.login}/${repoQuickPick.name}` : undefined;
+
     try {
-        const connectingToGithub: string = localize('ConnectingToGithub', '"{0}" is being connected to the GitHub repo. This may take several minutes...', client.fullName);
-        const connectedToGithub: string = localize('ConnectedToGithub', '"{0}" has been connected to the GitHub repo.', client.fullName);
+        const connectingToGithub: string = repoName ? localize('ConnectingToGithub', '"{0}" is being connected to repo "{1}". This may take several minutes...', client.fullName, repoName) : localize('ConnectingToGithub', '"{0}" is being connected to the GitHub repo. This may take several minutes...', client.fullName);
```
EricJizbaMSFT: Do we really need two separate messages here? Seems unlikely that repoName would be undefined. You're not checking for undefined when you show it in the quick pick are you?

appservice/src/connectToGitHub.ts
```diff
@@ -199,7 +201,7 @@ async function getGitHubReposQuickPicks(cache: ICachedQuickPicks<gitHubReposData
         }
     } while (requestOptions.nextLink && startTime + timeoutMs > Date.now());
 
-    cache.picks = cache.picks.concat(createQuickPickFromJsons(gitHubRepos, 'name', undefined, ['url', 'html_url']));
+    cache.picks = cache.picks.concat(createQuickPickFromJsons(gitHubRepos, 'name', undefined, ['url', 'html_url', 'name']));
```
EricJizbaMSFT: Can you remove `['url', 'html_url', 'name']` as a parameter? It should just be returning the original object in `data`...

The signature should be:
```
function createQuickPickFromJsons<T extends Object>(jsons: T[], label: string, description?: string): IAzureQuickPickItem<T>[] {
```

appservice/src/connectToGitHub.ts
nturinski: Yeah, I agree it seems low-risk, was just being thorough.  I am checking for undefined in the quick pick though: https://github.com/Microsoft/vscode-azuretools/blob/1fd5e3b0b1e36f90f514831a2d872776f4e3e8d3/appservice/src/connectToGitHub.ts#L148.

If there's no label property, it just skips the entry entirely.

nturinski: ... which means that it's guaranteed to have that data then

appservice/src/connectToGitHub.ts
```diff
@@ -54,21 +55,23 @@ export async function connectToGitHub(node: AzureTreeItem, client: SiteClient, c
...
         isMercurial: false
     };
 
+    const repoName: string = `${orgQuickPick.login}/${repoQuickPick.name}`;
+
     try {
-        const connectingToGithub: string = localize('ConnectingToGithub', '"{0}" is being connected to the GitHub repo. Check output channel for status', client.fullName);
-        const connectedToGithub: string = localize('ConnectedToGithub', 'The GitHub repo is now connected and deployed to "{0}" .', client.fullName);
+        const connectingToGithub: string = localize('ConnectingToGithub', '"{0}" is being connected to repo "{1}". Check output channel for status.', client.fullName, repoName);
+        const connectedToGithub: string = localize('ConnectedToGithub', '"{0}" is connected and deployed to "{1}".', repoName, client.fullName);
```
EricJizbaMSFT: nit: I prefer `Repo "{0}"` to be consistent with above message

PR #339, opened 2018-11-26
GitHub connecting and redeploy leverage waitForDeployment

nturinski:
Not awaiting the promises is giving me a tslint error, but I thought `.then()` would satisfy the no-floating-promise rule.  Any idea?

### DIFF
--- a/appservice/src/connectToGitHub.ts
+++ b/appservice/src/connectToGitHub.ts
@@ -10,7 +10,10 @@ import { Response } from 'request';
 import * as request from 'request-promise';
 import * as vscode from 'vscode';
 import { AzureTreeItem, DialogResponses, IActionContext, IAzureQuickPickItem, IParsedError, parseError } from 'vscode-azureextensionui';
+import KuduClient from 'vscode-azurekudu';
+import { waitForDeploymentToComplete } from './deploy/waitForDeploymentToComplete';
 import { ext } from './extensionVariables';
+import { getKuduClient } from './getKuduClient';
 import { localize } from './localize';
 import { signRequest } from './signRequest';
 import { SiteClient } from './SiteClient';
@@ -69,11 +72,15 @@ export async function connectToGitHub(node: AzureTreeItem, client: SiteClient, c
     try {
         const connectingToGithub: string = localize('ConnectingToGithub', '"{0}" is being connected to repo "{1}". This may take several minutes...', client.fullName, repoName);
         const connectedToGithub: string = localize('ConnectedToGithub', '"{0}" has been connected to repo "{1}".', client.fullName, repoName);
+        const kuduClient: KuduClient = await getKuduClient(client);
         await vscode.window.withProgress({ location: vscode.ProgressLocation.Notification, title: connectingToGithub }, async (): Promise<void> => {
             ext.outputChannel.appendLine(connectingToGithub);
-            await client.updateSourceControl(siteSourceControl);
-            vscode.window.showInformationMessage(connectedToGithub);
-            ext.outputChannel.appendLine(connectedToGithub);
+            client.updateSourceControl(siteSourceControl).then(() => {
+                vscode.window.showInformationMessage(connectedToGithub);
+                ext.outputChannel.appendLine(connectedToGithub);
+            });
+            await waitForDeploymentToComplete(client, kuduClient);
+
         });
     } catch (err) {
         try {
--- a/appservice/src/connectToGitHub.ts
+++ b/appservice/src/connectToGitHub.ts
@@ -63,9 +63,12 @@ export async function connectToGitHub(node: AzureTreeItem, client: SiteClient, c
         isMercurial: false
     };
 
+    // tslint:disable-next-line:no-non-null-assertion
+    const repoName: string = siteSourceControl.repoUrl!.substring('https://github.com/'.length);
+
     try {
-        const connectingToGithub: string = localize('ConnectingToGithub', '"{0}" is being connected to the GitHub repo. This may take several minutes...', client.fullName);
-        const connectedToGithub: string = localize('ConnectedToGithub', '"{0}" has been connected to the GitHub repo.', client.fullName);
+        const connectingToGithub: string = localize('ConnectingToGithub', '"{0}" is being connected to repo "{1}". This may take several minutes...', client.fullName, repoName);
+        const connectedToGithub: string = localize('ConnectedToGithub', '"{0}" has been connected to repo "{1}".', client.fullName, repoName);
         await vscode.window.withProgress({ location: vscode.ProgressLocation.Notification, title: connectingToGithub }, async (): Promise<void> => {
             ext.outputChannel.appendLine(connectingToGithub);
             await client.updateSourceControl(siteSourceControl);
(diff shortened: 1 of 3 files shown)


nturinski:
The children don't seem to be refreshing correctly.  Rather than refreshing just the DeploymentsTreeItem, I have to refresh the WebAppTreeItem to see the changes reflected in the children.  Think it's a bug from linking the packages?

appservice/src/connectToGitHub.ts
```diff
@@ -64,13 +67,18 @@ export async function connectToGitHub(node: AzureTreeItem, client: SiteClient, c
...
-            ext.outputChannel.appendLine(connectedToGithub);
+        const deploying: string = localize('deploying', 'Deploying to "{0}"... Check output window for status.', client.fullName);
+        const deploymentComplete: string = localize('deploymentComplete', 'Deployment to "{0}" completed.', client.fullName);
+        const kuduClient: KuduClient = await getKuduClient(client);
+        await vscode.window.withProgress({ location: vscode.ProgressLocation.Notification, title: deploying }, async (): Promise<void> => {
+            ext.outputChannel.appendLine(deploying);
+            // tslint:disable-next-line:no-floating-promises
+            client.updateSourceControl(siteSourceControl).then(() => {
+                vscode.window.showInformationMessage(deploymentComplete);
+                ext.outputChannel.appendLine(deploymentComplete);
```
EricJizbaMSFT: Your messages are not showing at the right time. Deploy doesn't finish until waitForDeploymentToComplete

appservice/src/connectToGitHub.ts
```diff
@@ -64,13 +67,18 @@ export async function connectToGitHub(node: AzureTreeItem, client: SiteClient, c
...
-        const connectedToGithub: string = localize('ConnectedToGithub', '"{0}" has been connected to the GitHub repo.', client.fullName);
-        await vscode.window.withProgress({ location: vscode.ProgressLocation.Notification, title: connectingToGithub }, async (): Promise<void> => {
-            ext.outputChannel.appendLine(connectingToGithub);
-            await client.updateSourceControl(siteSourceControl);
-            vscode.window.showInformationMessage(connectedToGithub);
-            ext.outputChannel.appendLine(connectedToGithub);
+        const deploying: string = localize('deploying', 'Deploying to "{0}"... Check output window for status.', client.fullName);
+        const deploymentComplete: string = localize('deploymentComplete', 'Deployment to "{0}" completed.', client.fullName);
+        const kuduClient: KuduClient = await getKuduClient(client);
+        await vscode.window.withProgress({ location: vscode.ProgressLocation.Notification, title: deploying }, async (): Promise<void> => {
```
EricJizbaMSFT: The title should not be "deploying" until after you change the source control

nturinski: In this case, changing the source control and deploying are the same command.  So I'm not really sure how to distinguish between the two actions.

appservice/src/connectToGitHub.ts
nturinski: This is the order that I'm seeing:

image

This makes sense to me.  It shows the kudu logs while the build is happening and then says the Deployment is completed when that is finished.

EricJizbaMSFT: A couple things:
1. Not sure we need to print a message that says "Check output window" to the output window 😂
1. Okay makes sense about the order, but I'd still prefer if you did it after `waitForDeploymentToComplete` to make it more clear in the code (rather than using `then`)

appservice/src/tree/DeploymentTreeItem.ts
```diff
@@ -78,17 +79,20 @@ export class DeploymentTreeItem extends AzureTreeItem<ISiteTreeRoot> {
...
             ext.outputChannel.appendLine(redeploying);
             const refreshingInteveral: NodeJS.Timer = setInterval(async () => { await this.refresh(); }, 1000); /* the status of the label changes during deployment so poll for that*/
             try {
-                await kuduClient.deployment.deploy(this.id);
-                await this.parent.refresh(); /* refresh entire node because active statuses has changed */
-                window.showInformationMessage(deployed);
-                ext.outputChannel.appendLine(deployed);
+                // tslint:disable-next-line:no-floating-promises
+                kuduClient.deployment.deploy(this.id).then(async () => {
+                    await this.parent.refresh(); /* refresh entire node because active statuses has changed */
```
EricJizbaMSFT: Yeah I say get rid of `then` and just move this code after `waitForDeploymentToComplete`

appservice/src/connectToGitHub.ts
EricJizbaMSFT: Ah - well then I prefer the "connecting to github" message over the "deploy" message. But it should still say "Check output window for status." instead of " This may take several minutes..."

nturinski:
Regarding option 2, that will give us the same "bug" where the deployment log is outputted at the very end in a giant chunk rather than at real-time since we are awaiting `updateSourceControl` which won't return until it is complete.

EricJizbaMSFT:
I'm not saying add the `await` back...I'm just saying get rid of the `then`...

appservice/src/connectToGitHub.ts
```diff
@@ -64,11 +67,13 @@ export async function connectToGitHub(node: AzureTreeItem, client: SiteClient, c
     };
 
     try {
-        const connectingToGithub: string = localize('ConnectingToGithub', '"{0}" is being connected to the GitHub repo. This may take several minutes...', client.fullName);
-        const connectedToGithub: string = localize('ConnectedToGithub', '"{0}" has been connected to the GitHub repo.', client.fullName);
+        const connectingToGithub: string = localize('ConnectingToGithub', '"{0}" is being connected to the GitHub repo... Check output channel for status', client.fullName);
+        const connectedToGithub: string = localize('ConnectedToGithub', 'The GitHub repo is now connected and is being deployed to "{0}" .', client.fullName);
```
EricJizbaMSFT: Why "is being deployed"? Isn't the deploy already finished?

appservice/src/tree/DeploymentTreeItem.ts
```diff
@@ -78,17 +79,18 @@ export class DeploymentTreeItem extends AzureTreeItem<ISiteTreeRoot> {
     }
 
     public async redeployDeployment(): Promise<void> {
-        const redeploying: string = `Redeploying commit "${this.id}" to "${this.root.client.fullName}"`;
-        const deployed: string = `Commit "${this.id}" has been redeployed to "${this.root.client.fullName}".`;
+        const redeploying: string = localize('redeploying', 'Redeploying commit "{0}" to "{1}"... Check output window for status.', this.id, this.root.client.fullName);
```
EricJizbaMSFT: The "..." in the middle of these messages has always bothered me. Why isn't it just a single period?

nturinski: I thought we did it because it was an ongoing message.  I don't really have an opinion on it though.

appservice/src/connectToGitHub.ts
nturinski: Well, it's connected and deployed, but still building according to the status codes. The deployment doesn't become active until after it's finished building so I thought it'd be misleading to say that it's deployed.

PR #340, opened 2018-11-28
Improve path doesn't exist error

EricJizbaMSFT:
Partial fix for https://github.com/Microsoft/vscode-azurefunctions/issues/772

### DIFF
--- a/appservice/src/deploy/deployZip.ts
+++ b/appservice/src/deploy/deployZip.ts
@@ -18,6 +18,10 @@ import { formatDeployLog } from './formatDeployLog';
 import { waitForDeploymentToComplete } from './waitForDeploymentToComplete';
 
 export async function deployZip(client: SiteClient, fsPath: string, configurationSectionName: string, aspPromise: Promise<AppServicePlan | undefined>): Promise<void> {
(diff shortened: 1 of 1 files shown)


appservice/src/deploy/deployZip.ts
```diff
@@ -18,6 +18,10 @@ import { formatDeployLog } from './formatDeployLog';
 import { waitForDeploymentToComplete } from './waitForDeploymentToComplete';
 
 export async function deployZip(client: SiteClient, fsPath: string, configurationSectionName: string, aspPromise: Promise<AppServicePlan | undefined>): Promise<void> {
+    if (!(await fse.pathExists(fsPath))) {
+        throw new Error(localize('pathNotExist', 'Failed to deploy path that does not exist: {0}', fsPath));
+    }
+
```
PrashanthCorp: I don't get why throwing the error is being moved to another function.

EricJizbaMSFT: deployZip calls getZipFileToDeploy a few lines down. If i move the error, then it will also be thrown for the case when `if (FileUtilities.getFileExtension(fsPath) === 'zip') {`

PR #343, opened 2018-11-29
Make Redeploy and logs more reliable

nturinski:
This PR does two things to make redeploy more reliable:
1. We wait to make sure that the "latest" deploy is the same as the commit being redeployed before listening to the logs.  Because of the delay in the request being sent vs. being processed, Kudu would see the latest result as the currently active deployment.  This resulted in a false positive.
1. Store `startTime` rather than `initialReceivedTime`.  `initialReceivedTime` doesn't work for redeploys because it is only updated on the _first_ deploy.  `startTime` seems to update when a deployment has been initialized.  This was causing an odd bug at the end that would display the logs of the  last commit of the repo because its `initialReceivedTime` is always the most recent.  I included an image of the object to illustrate.  I tested this with zipdeploy and local git to confirm that these still worked as intended.

image

I also added the repo url to the disconnect message.  We can't be too specific here due to the fact users can disconnect local git repositories.
Github:
image
Local git:
image

### DIFF
--- a/appservice/src/tree/DeploymentTreeItem.ts
+++ b/appservice/src/tree/DeploymentTreeItem.ts
@@ -81,7 +81,6 @@ export class DeploymentTreeItem extends AzureTreeItem<ISiteTreeRoot> {
     public async redeployDeployment(): Promise<void> {
         const redeploying: string = localize('redeploying', 'Redeploying commit "{0}" to "{1}". Check output window for status.', this.id, this.root.client.fullName);
         const redeployed: string = localize('redeployed', 'Commit "{0}" has been redeployed to "{1}".', this.id, this.root.client.fullName);
-        const timeout: string = localize('redeployTimeout', 'Redeploying commit "{0}" was unable to resolve and has timed out.', this.id);
         await window.withProgress({ location: ProgressLocation.Notification, title: redeploying }, async (): Promise<void> => {
             const kuduClient: KuduClient = await getKuduClient(this.root.client);
             const refreshingInteveral: NodeJS.Timer = setInterval(async () => { await this.refresh(); }, 1000); /* the status of the label changes during deployment so poll for that*/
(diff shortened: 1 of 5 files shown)


appservice/src/tree/DeploymentTreeItem.ts
```diff
@@ -81,18 +81,34 @@ export class DeploymentTreeItem extends AzureTreeItem<ISiteTreeRoot> {
...
+                    getResultInterval = setInterval(
+                        async () => {
+                            const deployResult: DeployResult | undefined = <DeployResult | undefined>await kuduClient.deployment.getResult('latest');
+                            if (deployResult && deployResult.id === this.id) {
+                                resolve();
+                            }
+                        },
+                        1000
+                    );
+                    setTimeout(() => reject(new Error(timeout)), 5000);
```
EricJizbaMSFT: 5 seconds seems kinda low. I'd probably increase this to about 20. Remember Microsoft internet is faster than average so you have to take your testing with a grain of salt

appservice/src/tree/DeploymentTreeItem.ts
```diff
@@ -81,18 +81,34 @@ export class DeploymentTreeItem extends AzureTreeItem<ISiteTreeRoot> {
...
+                await new Promise((resolve: () => void, reject: (error: Error) => void): void => {
+                    kuduClient.deployment.deploy(this.id).catch(reject);
+                    getResultInterval = setInterval(
+                        async () => {
+                            const deployResult: DeployResult | undefined = <DeployResult | undefined>await kuduClient.deployment.getResult('latest');
+                            if (deployResult && deployResult.id === this.id) {
+                                resolve();
+                            }
+                        },
+                        1000
```
EricJizbaMSFT: I'd also increase this interval a bit - maybe 2 or 3 seconds?

appservice/src/tree/DeploymentTreeItem.ts
```diff
@@ -81,18 +81,34 @@ export class DeploymentTreeItem extends AzureTreeItem<ISiteTreeRoot> {
     public async redeployDeployment(): Promise<void> {
         const redeploying: string = localize('redeploying', 'Redeploying commit "{0}" to "{1}". Check output window for status.', this.id, this.root.client.fullName);
         const redeployed: string = localize('redeployed', 'Commit "{0}" has been redeployed to "{1}".', this.id, this.root.client.fullName);
-        window.withProgress({ location: ProgressLocation.Notification, title: redeploying }, async (): Promise<void> => {
+        const timeout: string = localize('redeployTimeout', 'Redeploying commit "{0}" was unable to resolve and has timed out.', this.id);
```
EricJizbaMSFT: nit: can you move this error closer to where it's actually being used

PR #351, opened 2018-12-04
Update to gulp 4.0 in shared packages

EricJizbaMSFT:
Fixes some warnings from `npm audit`

### DIFF
--- a/kudu/gulpfile.js
+++ b/kudu/gulpfile.js
@@ -3,12 +3,11 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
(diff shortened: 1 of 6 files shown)


kudu/gulpfile.js
```diff
@@ -8,7 +8,7 @@ const autorest = require('autorest');
 const path = require('path');
 const fse = require('fs-extra');
 
-gulp.task('build', async () => {
+gulp.task('build', async (callback) => {
```
StephenWeatherford: Doesn't seem like you should mix models - either return a promise or use a callback, not both...

StephenWeatherford: on a timer??

PR #352, opened 2018-12-05
Additional check for Python runtime

nturinski:
Fixes https://github.com/Microsoft/vscode-azureappservice/issues/767
To help resolve this issue: https://github.com/Microsoft/vscode-azureappservice/issues/642

The only way to have a Python runtime is to create a Linux app, so it makes sense to default the web app to Linux and Python when we detect a `requirements.txt`

### DIFF
--- a/appservice/src/createAppService/IAppServiceWizardContext.ts
+++ b/appservice/src/createAppService/IAppServiceWizardContext.ts
@@ -61,15 +61,8 @@ export interface IAppServiceWizardContext extends IResourceGroupWizardContext, I
 
     /**
      * The runtime to put to the top of the QuickPick list to recommend to the user.
-     * This should be set in `getWizardRecommendations`
+     * This should be set in `setWizardContextDefaults`
      */
 
     recommendedSiteRuntime?: LinuxRuntimes;
-
-    /**
-     * The OS that we detect that the user should use.  Will be set automatically for Basic Creation.
-     * This should be set in `getWizardRecommendations`
-     */
-
-    recommendedWebsiteOS?: WebsiteOS;
 }
--- a/appservice/src/createAppService/IAppServiceWizardContext.ts
+++ b/appservice/src/createAppService/IAppServiceWizardContext.ts
@@ -5,7 +5,7 @@
 
 import { AppServicePlan, Site, SkuDescription } from 'azure-arm-website/lib/models';
 import { IResourceGroupWizardContext, IStorageAccountWizardContext } from 'vscode-azureextensionui';
-import { AppKind, WebsiteOS } from './AppKind';
+import { AppKind, LinuxRuntimes, WebsiteOS } from './AppKind';
 
 export interface IAppServiceWizardContext extends IResourceGroupWizardContext, IStorageAccountWizardContext {
     newSiteKind: AppKind;
@@ -61,8 +61,15 @@ export interface IAppServiceWizardContext extends IResourceGroupWizardContext, I
 
     /**
      * The runtime to put to the top of the QuickPick list to recommend to the user.
-     * This should be set in `setWizardContextDefaults`
+     * This should be set in `getWizardRecommendations`
      */
 
-     detectedSiteRuntime?: string;
+    recommendedSiteRuntime?: LinuxRuntimes;
+
+    /**
+     * The OS that we detect that the user should use.  Will be set automatically for Basic Creation.
+     * This should be set in `getWizardRecommendations`
+     */
+
+    recommendedWebsiteOS?: WebsiteOS;
 }
--- a/appservice/src/createAppService/IAppServiceWizardContext.ts
+++ b/appservice/src/createAppService/IAppServiceWizardContext.ts
@@ -58,4 +58,11 @@ export interface IAppServiceWizardContext extends IResourceGroupWizardContext, I
      * By specifying this in the context, we can ensure that Azure is only queried once for the entire wizard
      */
     plansTask?: Promise<AppServicePlan[]>;
+
+    /**
+     * The runtime to put to the top of the QuickPick list to recommend to the user.
+     * This should be set in `setWizardContextDefaults`
+     */
+
+     detectedSiteRuntime?: string;
 }
--- a/appservice/src/createAppService/SiteRuntimeStep.ts
+++ b/appservice/src/createAppService/SiteRuntimeStep.ts
@@ -6,7 +6,6 @@
 import { AzureWizardPromptStep, IAzureQuickPickItem } from 'vscode-azureextensionui';
 import { ext } from '../extensionVariables';
 import { AppKind, WebsiteOS } from './AppKind';
-import { pythonRuntime } from './createWebApp';
 import { IAppServiceWizardContext } from './IAppServiceWizardContext';
 
 interface ILinuxRuntimeStack {
@@ -33,13 +32,9 @@ export class SiteRuntimeStep extends AzureWizardPromptStep<IAppServiceWizardCont
                         data: rt
                     };
                 });
-                if (wizardContext.detectedSiteRuntime) {
-                    switch (wizardContext.detectedSiteRuntime) {
-                        case pythonRuntime:
-                            runtimeItems = this.sortQuickPicksByRuntime(runtimeItems, pythonRuntime);
-                        default:
-                        // do nothing if we do not handle that detectedSiteRuntime
-                    }
+                // tslint:disable-next-line:strict-boolean-expressions
+                if (wizardContext.recommendedSiteRuntime) {
+                    runtimeItems = this.sortQuickPicksByRuntime(runtimeItems, wizardContext.recommendedSiteRuntime);
                 }
                 wizardContext.newSiteRuntime = (await ext.ui.showQuickPick(runtimeItems, { placeHolder: 'Select a runtime for your new Linux app.' })).data.name;
             }
--- a/appservice/src/createAppService/SiteRuntimeStep.ts
+++ b/appservice/src/createAppService/SiteRuntimeStep.ts
@@ -36,7 +36,7 @@ export class SiteRuntimeStep extends AzureWizardPromptStep<IAppServiceWizardCont
                 if (wizardContext.detectedSiteRuntime) {
                     switch (wizardContext.detectedSiteRuntime) {
                         case pythonRuntime:
-                            runtimeItems = this.unshiftRuntimeQuickPickItem(pythonRuntime, runtimeItems);
+                            runtimeItems = this.sortQuickPicksByRuntime(runtimeItems, pythonRuntime);
                         default:
                         // do nothing if we do not handle that detectedSiteRuntime
                     }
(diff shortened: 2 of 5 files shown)


appservice/src/createAppService/createWebApp.ts
```diff
@@ -34,6 +34,14 @@ export async function setWizardContextDefaults(wizardContext: IAppServiceWizardC
                 wizardContext.newSiteOS = WebsiteOS.windows;
             }
         });
+
+        // requirements.txt are used to pip install so a good way to determine it's a Python app
+        await workspace.findFiles('requirements.txt').then((files: Uri[]) => {
+            if (files.length > 0) {
+                wizardContext.newSiteOS = WebsiteOS.linux;
+                wizardContext.newSiteRuntime = 'python|3.7';
```
EricJizbaMSFT: Makes sense to default to Linux, but I'm not so sure about defaulting to a specific version of python. We had this discussion when choosing the default for node and decided to prompt the user for runtime

appservice/src/createAppService/createWebApp.ts
```diff
@@ -34,6 +34,14 @@ export async function setWizardContextDefaults(wizardContext: IAppServiceWizardC
                 wizardContext.newSiteOS = WebsiteOS.windows;
             }
         });
+
+        // requirements.txt are used to pip install so a good way to determine it's a Python app
+        await workspace.findFiles('requirements.txt').then((files: Uri[]) => {
```
EricJizbaMSFT: nit: `fse.pathExists` would probably be faster than iterating through all files

appservice/src/createAppService/SiteRuntimeStep.ts
```diff
@@ -153,4 +161,16 @@ export class SiteRuntimeStep extends AzureWizardPromptStep<IAppServiceWizardCont
...
         ];
     }
+
+    private unshiftRuntimeQuickPickItem(runtime: string, runtimeItems: IAzureQuickPickItem<ILinuxRuntimeStack>[]): IAzureQuickPickItem<ILinuxRuntimeStack>[] {
+        const runtimeQuickPickItemIndex: number = runtimeItems.findIndex((qpItem: IAzureQuickPickItem<ILinuxRuntimeStack>) => {
+            return qpItem.data.name === runtime;
+        });
+        if (runtimeQuickPickItemIndex >= 0) {
+            const runtimeQuickPickItem: IAzureQuickPickItem<ILinuxRuntimeStack> = runtimeItems.splice(runtimeQuickPickItemIndex)[0];
+            runtimeItems.unshift(runtimeQuickPickItem);
```
EricJizbaMSFT: So I don't think we want to shift a single runtime. I think we want to move _all_ python runtimes to the top or _all_ node runtimes to the top if it's node. (I realize there's only one python runtime at the moment, but I expect there to be more in the future)

I would use `sort` rather than `unshift` in that case

appservice/src/createAppService/createWebApp.ts
```diff
@@ -23,17 +27,15 @@ export async function setWizardContextDefaults(wizardContext: IAppServiceWizardC
...
-
-        await workspace.findFiles('*.csproj').then((files: Uri[]) => {
-            if (files.length > 0) {
-                wizardContext.newSiteOS = WebsiteOS.windows;
-            }
-        });
+        const fsPath: string = workspace.workspaceFolders[0].uri.fsPath;
+        if (await fse.pathExists(path.join(fsPath, 'package.json'))) {
+            wizardContext.newSiteOS = WebsiteOS.linux;
+        } else if (await fse.pathExists(path.join(fsPath, '*.csproj'))) {
```
EricJizbaMSFT: `await fse.pathExists(path.join(fsPath, '*.csproj')` <-- I don't think this works with glob patterns

appservice/src/createAppService/createWebApp.ts
```diff
@@ -23,17 +27,15 @@ export async function setWizardContextDefaults(wizardContext: IAppServiceWizardC
     await LocationListStep.setLocation(wizardContext, 'centralus');
```
EricJizbaMSFT: I'm beginning to think `setWizardContextDefaults` should be directly in the appservice repo rather than in the shared package. (And then you'd pass that stuff in as a part of `IAppCreateOptions`). Thoughts? Doesn't have to be done in this PR.

nturinski: Yeah, I agree.  I think it's starting to get a little bit messy and seems like there'll be some options for Functions that will just make it a lot of branching logic.

appservice/src/createAppService/SiteRuntimeStep.ts
```diff
@@ -24,15 +25,22 @@ export class SiteRuntimeStep extends AzureWizardPromptStep<IAppServiceWizardCont
...
                     return {
                         id: rt.name,
                         label: rt.displayName,
                         description: '',
                         data: rt
                     };
                 });
-
+                if (wizardContext.detectedSiteRuntime) {
+                    switch (wizardContext.detectedSiteRuntime) {
```
EricJizbaMSFT: You should be able to just sort by runtime (if it's defined) without using a switch at all

appservice/src/createAppService/createWebApp.ts
```diff
@@ -23,17 +27,19 @@ export async function setWizardContextDefaults(wizardContext: IAppServiceWizardC
...
-        });
-
-        await workspace.findFiles('*.csproj').then((files: Uri[]) => {
-            if (files.length > 0) {
-                wizardContext.newSiteOS = WebsiteOS.windows;
-            }
-        });
+        const fsPath: string = workspace.workspaceFolders[0].uri.fsPath;
+        if (await fse.pathExists(path.join(fsPath, 'package.json'))) {
+            wizardContext.newSiteOS = WebsiteOS.linux;
```
EricJizbaMSFT: `detectedSiteRuntime` should be `node` in this case, correct?

appservice/src/createAppService/createWebApp.ts
```diff
@@ -23,17 +27,19 @@ export async function setWizardContextDefaults(wizardContext: IAppServiceWizardC
...
+        } else if (await fse.pathExists(path.join(fsPath, 'requirements.txt'))) {
+            // requirements.txt are used to pip install so a good way to determine it's a Python app
+            wizardContext.newSiteOS = WebsiteOS.linux;
+            wizardContext.detectedSiteRuntime = pythonRuntime;
```
EricJizbaMSFT: Currently this is only set if `advancedCreation` is set to false. Since this just changes the ordering of the quick pick - shouldn't it be set even if `advancedCreation` is true?

appservice/src/createAppService/IAppServiceWizardContext.ts
```diff
@@ -58,4 +58,11 @@ export interface IAppServiceWizardContext extends IResourceGroupWizardContext, I
      * By specifying this in the context, we can ensure that Azure is only queried once for the entire wizard
      */
     plansTask?: Promise<AppServicePlan[]>;
+
+    /**
+     * The runtime to put to the top of the QuickPick list to recommend to the user.
+     * This should be set in `setWizardContextDefaults`
+     */
+
+     detectedSiteRuntime?: string;
```
EricJizbaMSFT: Maybe `recommendedSiteRuntime`? I feel like that more closely matches your comment above. "detected" is more about "how you picked this value" and I think the name should be more about "how you're going to use this value"

appservice/src/createAppService/SiteRuntimeStep.ts
nturinski: What do you think about adding a little description saying (recommended) to anything that we sort? Something to indicate that we shifted the runtimes to the top?

appservice/src/createAppService/createWebApp.ts
```diff
@@ -23,23 +23,40 @@ export async function createWebApp(
     return await createAppService(AppKind.app, actionContext, subscriptionContext, createOptions, showCreatingTreeItem);
 }
 
-export async function setWizardContextDefaults(wizardContext: IAppServiceWizardContext): Promise<void> {
-    await LocationListStep.setLocation(wizardContext, 'centralus');
-    // defaults that for if one workspace is opened
+export async function getWizardRecommendations(wizardContext: IAppServiceWizardContext, actionContext: IActionContext): Promise<void> {
+    actionContext.properties.recommendedSiteRuntime = falseStr;
+    actionContext.properties.recommendedWebsiteOS = falseStr;
```
EricJizbaMSFT: Just default to undefined instead of using `falseStr`, which is a much more common pattern

appservice/src/createAppService/IAppServiceWizardContext.ts
```diff
@@ -61,8 +61,15 @@ export interface IAppServiceWizardContext extends IResourceGroupWizardContext, I
...
 
-     detectedSiteRuntime?: string;
+    recommendedSiteRuntime?: LinuxRuntimes;
+
+    /**
+     * The OS that we detect that the user should use.  Will be set automatically for Basic Creation.
+     * This should be set in `getWizardRecommendations`
+     */
+
+    recommendedWebsiteOS?: WebsiteOS;
```
EricJizbaMSFT: This has the same naming as `recommendedSiteRuntime` but it's used in a different way (aka you don't use it to sort the quick pick). It's not even used in the wizard, which is the whole point of the "WizardContext"

appservice/src/createAppService/createWebApp.ts
```diff
@@ -23,23 +23,40 @@ export async function createWebApp(
     return await createAppService(AppKind.app, actionContext, subscriptionContext, createOptions, showCreatingTreeItem);
 }
 
-export async function setWizardContextDefaults(wizardContext: IAppServiceWizardContext): Promise<void> {
-    await LocationListStep.setLocation(wizardContext, 'centralus');
-    // defaults that for if one workspace is opened
+export async function getWizardRecommendations(wizardContext: IAppServiceWizardContext, actionContext: IActionContext): Promise<void> {
```
EricJizbaMSFT: I suggest you keep this as one function (`setWizardContextDefaults`) and just pass in `advancedCreation`

appservice/src/createAppService/createWebApp.ts
```diff
@@ -19,21 +21,30 @@ export async function createWebApp(
...
-                wizardContext.newSiteOS = WebsiteOS.windows;
+        if (!advancedCreation) {
+            await LocationListStep.setLocation(wizardContext, 'centralus');
+            // we only set the OS for the non-advanced creation scenario
+            // tslint:disable-next-line:strict-boolean-expressions
+            if (wizardContext.recommendedSiteRuntime) {
+                wizardContext.newSiteOS = WebsiteOS.linux;
             }
-        });
+            await workspace.findFiles('*.csproj').then((files: Uri[]) => {
```
EricJizbaMSFT: Wrap this in an else case. No need to search files if we already know it's linux

PR #353, opened 2018-12-06
Add an optional setting to obfuscate settings

nturinski:
Fixes https://github.com/Microsoft/vscode-azureappservice/issues/640

Uses a similar structure as this PR: https://github.com/Microsoft/vscode-azuretools/pull/349

Because App Settings can contain sensitive information (such as password), we want to hide the values by default.

### DIFF
--- a/appservice/src/tree/AppSettingTreeItem.ts
+++ b/appservice/src/tree/AppSettingTreeItem.ts
@@ -5,6 +5,7 @@
 
 import { StringDictionary } from 'azure-arm-website/lib/models';
 import * as path from 'path';
+import { workspace } from 'vscode';
 import { AzureTreeItem, DialogResponses } from 'vscode-azureextensionui';
 import { ext } from '../extensionVariables';
 import { AppSettingsTreeItem, validateAppSettingKey } from './AppSettingsTreeItem';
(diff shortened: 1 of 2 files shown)


EricJizbaMSFT:
I like the portal's experience better than a VS Code setting:
screen shot 2018-12-06 at 1 49 42 pm

Not saying we have to do `Click to edit`, but maybe `<Hidden value. Click to view>`

nturinski:
@fiveisprime Thoughts?

nturinski:
Now the default behavior is to hide all app setting values.  A click will reveal.  If you click again, it will hide the setting again.

image

appservice/src/tree/AppSettingTreeItem.ts
```diff
@@ -14,22 +14,26 @@ export class AppSettingTreeItem extends AzureTreeItem<ISiteTreeRoot> {
...
+    private _value: string;
+    private _hideValue: boolean;
 
     constructor(parent: AppSettingsTreeItem, key: string, value: string) {
         super(parent);
-        this.key = key;
-        this.value = value;
+        this._key = key;
+        this._value = value;
+        this.commandId = 'appService.toggleValueVisability';
```
EricJizbaMSFT: You'll have to pass this command Id in. Just like the problem we had with deployments node commandId

appservice/src/tree/AppSettingTreeItem.ts
```diff
@@ -41,32 +45,36 @@ export class AppSettingTreeItem extends AzureTreeItem<ISiteTreeRoot> {
...
     }
 
     public async deleteTreeItemImpl(): Promise<void> {
-        await ext.ui.showWarningMessage(`Are you sure you want to delete setting "${this.key}"?`, { modal: true }, DialogResponses.deleteResponse, DialogResponses.cancel);
-        await this.parent.deleteSettingItem(this.key);
+        await ext.ui.showWarningMessage(`Are you sure you want to delete setting "${this._key}"?`, { modal: true }, DialogResponses.deleteResponse, DialogResponses.cancel);
+        await this.parent.deleteSettingItem(this._key);
+    }
+
+    public toggleValueVisability(): void {
```
EricJizbaMSFT: Visability -> Visibility

PR #357, opened 2018-12-13
Refactor confirmOverwriteSettings to AppSettingsTreeItem

nturinski:
At the risk of making this class too large, this seemed like the right spot for it since only the AppSettingsTreeItem uses this function.


EricJizbaMSFT:
> this seemed like the right spot for it since only the AppSettingsTreeItem uses this function

When does AppSettingsTreeItem use this function? Also the function never calls `this.` for anything. I don't think this function is related to the tree and I think it should be on its own

nturinski:
Well, I guess the way I refactored it, AppSettingsTreeItem is what is calling the function now.  I don't mind putting it in its own file, but this function is only used when uploading/downloading app settings which the only entry point is the AppSettingsTreeItem so it seemed appropriate to me to be in there.  If it's in its own file, should I just throw it in `src`?  I felt like it was getting kinda cluttered.

PR #358, opened 2018-12-18
Remove commandId so clicking on treeItem does not reveal

nturinski:
@fiveisprime didn't like having the click-on-reveal due to the fact users may unintentionally reveal settings.   He says the eye is a common enough UI.

### DIFF
--- a/appservice/package.json
+++ b/appservice/package.json
@@ -1,7 +1,7 @@
 {
     "name": "vscode-azureappservice",
     "author": "Microsoft Corporation",
-    "version": "0.27.0",
+    "version": "0.27.1",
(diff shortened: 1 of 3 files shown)


EricJizbaMSFT:
I completely agree that the eye makes sense, but I don't think that's why we kept "click to view". The problem is that the eye doesn't show up unless you hover over the node in the tree

nturinski:
I'll let @fiveisprime speak for himself on this matter. 

I personally think that it's not that hidden as a user will most likely hover over the setting to interact with it in the first place.  We also have other commands that behave this way such as filter subscriptions and view deployment logs.

EricJizbaMSFT:
@nturinski What's the status on this? Can it be closed?

PR #360, opened 2019-01-02
Parse html errors

EricJizbaMSFT:
See test cases for example errors and what they look like now.

Fixes https://github.com/Microsoft/vscode-azurefunctions/issues/841
Related https://github.com/Microsoft/vscode-azurefunctions/issues/878

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -34,6 +34,7 @@
         "azure-arm-resource": "^3.0.0-preview",
         "azure-arm-storage": "^3.1.0",
         "fs-extra": "^4.0.3",
+        "html-to-text": "^4.0.0",
         "ms-rest": "^2.2.2",
         "ms-rest-azure": "^2.4.4",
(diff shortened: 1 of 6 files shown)


ui/package.json
```diff
@@ -34,6 +34,7 @@
         "azure-arm-resource": "^3.0.0-preview",
         "azure-arm-storage": "^3.1.0",
         "fs-extra": "^4.0.3",
+        "html-to-text": "^4.0.0",
```
PrashanthCorp: I feel this should be locked down, without the ^. There's quite some test code that essentially tests this package's functionality (rather than _how we use it_). A package update making our code fail (if the `parseIfHtml` behavior changes subtly) is more overhead, especially for this upstream repo which can break the downstream extensions.

EricJizbaMSFT: That's what the package-lock.json is for

PR #362, opened 2019-01-05
Add 10 second delay to basic asp app's first deployment

nturinski:
Related to this issue and PR:
https://github.com/Microsoft/vscode-azureappservice/issues/644
https://github.com/Microsoft/vscode-azureappservice/pull/798

This will check that the ASP is a basic tier.  If this is the first deployment (because this happens after deployment completes, the first deployment is now 1), then we will delay 10 seconds.

The output channel looks like this:
image

NOTE: The last line would have a logtime of 5:10:47.

I think that this implies that it will take additional time to restart.  That said, it says "begin restart within 10 seconds" so that's only guaranteeing that the restart will start, not that the app will be up.

### DIFF
--- a/appservice/src/deploy/deployZip.ts
+++ b/appservice/src/deploy/deployZip.ts
@@ -44,6 +44,16 @@ export async function deployZip(client: SiteClient, fsPath: string, configuratio
             const kuduClient: KuduClient = await getKuduClient(client);
             await kuduClient.pushDeployment.zipPushDeploy(fs.createReadStream(zipFilePath), { isAsync: true });
             await waitForDeploymentToComplete(client, kuduClient);
+            if (asp && asp.sku && asp.sku.tier && asp.sku.tier.toLowerCase() === 'basic') {
+                const deployments: number = (await kuduClient.deployment.getDeployResults()).length;
+                // if this is the first deployment, implement a 10 second delay for apps in a basic plan due to long start times
+                if (deployments === 1) {
+                    const delay: (delayMs: number) => Promise<void> = async (delayMs: number): Promise<void> => {
+                        await new Promise<void>((resolve: () => void): void => { setTimeout(resolve, delayMs); });
+                    };
+                    await delay(10000);
+                }
+            }
         }
     } finally {
         if (createdZip) {
--- a/appservice/src/deploy/deployZip.ts
+++ b/appservice/src/deploy/deployZip.ts
@@ -67,8 +67,11 @@ async function getZipFileToDeploy(fsPath: string, configurationSectionName?: str
 }
 
 async function delayFirstWebAppDeploy(client: SiteClient, asp: AppServicePlan | undefined, kuduClient: KuduClient): Promise<void> {
-    // this delay is only valid for Linux web apps, so return for anything else
+    // this delay is only valid for Linux web apps on a basic asp, so return for anything else
     if (client.isFunctionApp) {
+        return;
+    }
+    if (!asp || !asp.sku || !asp.sku.tier || asp.sku.tier.toLowerCase() !== 'basic') {
         return;
     }
     const siteConfigResource: SiteConfigResource = await client.getSiteConfig();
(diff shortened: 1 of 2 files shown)


appservice/src/deploy/deployZip.ts
```diff
@@ -44,6 +44,16 @@ export async function deployZip(client: SiteClient, fsPath: string, configuratio
             const kuduClient: KuduClient = await getKuduClient(client);
             await kuduClient.pushDeployment.zipPushDeploy(fs.createReadStream(zipFilePath), { isAsync: true });
             await waitForDeploymentToComplete(client, kuduClient);
+            if (asp && asp.sku && asp.sku.tier && asp.sku.tier.toLowerCase() === 'basic') {
+                const deployments: number = (await kuduClient.deployment.getDeployResults()).length;
+                // if this is the first deployment, implement a 10 second delay for apps in a basic plan due to long start times
+                if (deployments === 1) {
+                    const delay: (delayMs: number) => Promise<void> = async (delayMs: number): Promise<void> => {
```
StephenWeatherford: Would be more readable as a nested or (better) file-level function

StephenWeatherford: Also, I thought there was supposed to be a message?

StephenWeatherford:
Have you considered waiting for a GET to respond successfully (with a max time of 10s or whatever), so that it can exit early if it comes up earlier?

appservice/src/deploy/deployZip.ts
```diff
@@ -44,6 +44,16 @@ export async function deployZip(client: SiteClient, fsPath: string, configuratio
             const kuduClient: KuduClient = await getKuduClient(client);
             await kuduClient.pushDeployment.zipPushDeploy(fs.createReadStream(zipFilePath), { isAsync: true });
             await waitForDeploymentToComplete(client, kuduClient);
+            if (asp && asp.sku && asp.sku.tier && asp.sku.tier.toLowerCase() === 'basic') {
```
EricJizbaMSFT: I thought Matt said we only needed to do this on Linux - not windows?

Also I don't think we need this on Function Apps?

appservice/src/deploy/deployZip.ts
```diff
@@ -44,6 +44,16 @@ export async function deployZip(client: SiteClient, fsPath: string, configuratio
...
+                const deployments: number = (await kuduClient.deployment.getDeployResults()).length;
+                // if this is the first deployment, implement a 10 second delay for apps in a basic plan due to long start times
```
EricJizbaMSFT: Can you add a link to the issue for this since it's only a workaround and should be removed eventually?

appservice/src/deploy/deployZip.ts
```diff
@@ -44,6 +44,16 @@ export async function deployZip(client: SiteClient, fsPath: string, configuratio
...
+                if (deployments === 1) {
```
EricJizbaMSFT: `if (deployments === 1) {` works reliably? I guess I figured there may be other "temp" deployments or something causing this number to be higher even for the first one

appservice/src/deploy/deployZip.ts
EricJizbaMSFT: I think Stephen is talking about "delay", but honestly I'd like all of this new code in it's own function. Just so that it looks cleaner and so that it'll be easier to remove once this workaround isn't necessary

appservice/src/deploy/deployZip.ts
nturinski: I only tested it with a few apps, but new apps only had 1 deployment after it successfully completed.  

Since @fiveisprime seems to mostly be worried about tutorial followers, I don't think we're expecting it to fail so I think it'd be okay?

nturinski:
Since this is intended to be a very temporary stopgap and it's only for the first deployment on basic Linux web apps, I'm okay with just waiting the 10 seconds.

appservice/src/deploy/deployZip.ts
```diff
@@ -62,3 +66,30 @@ async function getZipFileToDeploy(fsPath: string, configurationSectionName?: str
...
+                }
+                if (!asp || !asp.sku || !asp.sku.tier || asp.sku.tier.toLowerCase() !== 'basic') {
+                    resolve();
+                }
+                if (!client.isLinux) {
+                    resolve();
+                }
+
+                const deployments: number = (await kuduClient.deployment.getDeployResults()).length;
+                if (deployments > 1) {
```
EricJizbaMSFT: Ya got some weird indents goin on

PR #363, opened 2019-01-08
Enable Deployments Node for Push deploy scenarios

nturinski:
Deployments works for even when the app is ScmType.None.  The deployment logs work and there is a history of deployments.  Redeploy does NOT work, but throws an (unhelpful) error that we catch and give the user something a little more friendly :)

Due to the limited support, I still have the `Connect to GitHub repo...` tree node since I think we want to guide users towards that solution still.

### DIFF
--- a/appservice/src/tree/DeploymentTreeItem.ts
+++ b/appservice/src/tree/DeploymentTreeItem.ts
@@ -6,7 +6,7 @@
 import * as os from 'os';
 import * as path from 'path';
 import { ProgressLocation, TextDocument, window, workspace } from 'vscode';
-import { AzureTreeItem, IParsedError, parseError } from 'vscode-azureextensionui';
+import { AzureTreeItem } from 'vscode-azureextensionui';
 import KuduClient from 'vscode-azurekudu';
 import { DeployResult, LogEntry } from 'vscode-azurekudu/lib/models';
 import { waitForDeploymentToComplete } from '../deploy/waitForDeploymentToComplete';
--- a/appservice/src/tree/DeploymentTreeItem.ts
+++ b/appservice/src/tree/DeploymentTreeItem.ts
@@ -6,7 +6,7 @@
 import * as os from 'os';
 import * as path from 'path';
 import { ProgressLocation, TextDocument, window, workspace } from 'vscode';
-import { AzureTreeItem } from 'vscode-azureextensionui';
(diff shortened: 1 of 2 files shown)


appservice/src/tree/DeploymentTreeItem.ts
```diff
@@ -105,6 +105,13 @@ export class DeploymentTreeItem extends AzureTreeItem<ISiteTreeRoot> {
                 await this.parent.refresh(); /* refresh entire node because active statuses has changed */
                 window.showInformationMessage(redeployed);
                 ext.outputChannel.appendLine(redeployed);
+            } catch (error) {
+                const parsedError: IParsedError = parseError(error);
+                if (parsedError.message.includes('Repository could not be found')) {
```
EricJizbaMSFT: This would cover most cases, but I don't like checking for specific text in an error message because it won't work if/when the error gets translated.

Can you do a smarter check before you even try to redeploy? Is there a property on the deployment that says how it was created? Would it make sense to just do `siteConfig.scmType === ScmType.None` again?

nturinski: Nah, a user can redeploy to a git commit even if their ScmType is None.  I'll investigate the DeployResult objects.

nturinski: Think we could use this more reliably?

EricJizbaMSFT: isReadonly sounds good as long as you've tested it

nturinski: Pssssht, of course, I tested it...more than once...

appservice/src/tree/DeploymentTreeItem.ts
```diff
@@ -79,6 +79,9 @@ export class DeploymentTreeItem extends AzureTreeItem<ISiteTreeRoot> {
     }
 
     public async redeployDeployment(): Promise<void> {
+        if (this._deployResult.isReadonly) {
+            throw new Error(localize('redeployNotSupported', 'Redeploy does not support non-git deployments.'));
```
EricJizbaMSFT: nit: I like your previous wording better. Specifically, `Redeploy is not supported for` instead of `Redeploy does not support`

EricJizbaMSFT: No strong preference

PR #365, opened 2019-01-09
Prompt for function app runtime even on windows

ejizba:
Modeled behavior on the portal. This will only pop up in the "advanced creation" scenario


nturinski:
Relies on this PR https://github.com/Microsoft/vscode-azurefunctions/pull/898/files

PR #367, opened 2019-01-09
Append line to output that redeployment is happening

nturinski:
Fixes https://github.com/Microsoft/vscode-azureappservice/issues/756

### DIFF
--- a/appservice/src/tree/DeploymentTreeItem.ts
+++ b/appservice/src/tree/DeploymentTreeItem.ts
@@ -9,6 +9,7 @@ import { ProgressLocation, TextDocument, window, workspace } from 'vscode';
 import { AzureTreeItem } from 'vscode-azureextensionui';
 import KuduClient from 'vscode-azurekudu';
 import { DeployResult, LogEntry } from 'vscode-azurekudu/lib/models';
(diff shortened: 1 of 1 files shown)


appservice/src/tree/DeploymentTreeItem.ts
```diff
@@ -82,6 +83,7 @@ export class DeploymentTreeItem extends AzureTreeItem<ISiteTreeRoot> {
         const redeploying: string = localize('redeploying', 'Redeploying commit "{0}" to "{1}". Check output window for status.', this.id, this.root.client.fullName);
         const redeployed: string = localize('redeployed', 'Commit "{0}" has been redeployed to "{1}".', this.id, this.root.client.fullName);
         await window.withProgress({ location: ProgressLocation.Notification, title: redeploying }, async (): Promise<void> => {
+            ext.outputChannel.appendLine(formatDeployLog(this.root.client, localize('reployingOutput', 'Redeploying commit "{0}" to "{1}".', this.id, this.root.client.fullName)));
```
EricJizbaMSFT: nit: I prefer to have "..." at the end of messages that are ongoing

PR #372, opened 2019-01-11
Retry syncing triggers for linux consumption deployment

EricJizbaMSFT:
Fixes https://github.com/Microsoft/vscode-azurefunctions/issues/804

Output looks like this:
screen shot 2019-01-10 at 4 18 49 pm


EricJizbaMSFT:
The last error is thrown just like normal if all retries fail. See the screenshot where it lists `Error: test` (It also prompted like normal, but I didn't screenshot that)

PR #373, opened 2019-01-11
Local Git deployments return logs in real-time

nturinski:
Fixes https://github.com/Microsoft/vscode-azureappservice/issues/754

Rather than relying on simpleGit for pushing to the remote, we are using child_process instead.  The cmd returns all of the same output that we could get from Kudu so we can pipe it directly.

The cpUtils are copied mostly from the azure-functions repo, but I added a new parameter called CommandOptions.

The issue was that the command contained the user's credentials and would output it to the channel.  I added an option to obfuscate a value.  The output looks like this:

image


appservice/src/utils/cpUtils.ts
```diff
@@ -85,12 +83,17 @@ export namespace cpUtils {
         });
     }
 
-    function obfuscateValue(data: string, value: string): string {
-        let astericks: string = '';
-        for (const _i of value) {
-            astericks += '*';
+    function obfuscateValue(data: string, commandOptions?: ICommandOptions): string {
```
EricJizbaMSFT: nit: `commandOptions?: ICommandOptions` should be `commandOptions: ICommandOptions | undefined`

The first means you can forget to pass `commandOptions`. The second forces you to pass it in every time.

appservice/src/deploy/localGitDeploy.ts
```diff
@@ -8,52 +8,47 @@ import * as opn from 'opn';
...
-        } else if (err.message.indexOf('error: failed to push') >= 0) {
+        } else if (result.cmdOutputIncludingStderr.indexOf('error: failed to push') >= 0) {
             const forcePush: vscode.MessageItem = { title: localize('forcePush', 'Force Push') };
             const pushReject: string = localize('localGitPush', 'Push rejected due to Git history diverging.');
             await ext.ui.showWarningMessage(pushReject, forcePush, DialogResponses.cancel);
-            await localGit.push(remote, 'HEAD:master', { '-f': true });
+            await cpUtils.executeCommand(`git push -f ${remote} HEAD:master`, commandOptions);
         } else {
-            throw err;
+            throw result.cmdOutputIncludingStderr;
```
EricJizbaMSFT: throw new Error(result.cmdOutputIncludingStderr);

appservice/src/utils/cpUtils.ts
```diff
@@ -0,0 +1,111 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+import * as cp from 'child_process';
+import * as os from 'os';
+import * as vscode from 'vscode';
+import { localize } from '../localize';
+
+export namespace cpUtils {
+    export async function executeCommand(command: string, commandOptions?: ICommandOptions, ...args: string[]): Promise<string> {
+        const outputChannel: vscode.OutputChannel | undefined = commandOptions && commandOptions.outputChannel ? commandOptions.outputChannel : undefined;
```
EricJizbaMSFT: nit: `commandOptions ? commandOptions.outputChannel : undefined;`

appservice/src/utils/cpUtils.ts
```diff
@@ -0,0 +1,111 @@
...
+        const result: ICommandResult = await tryExecuteCommand(command, commandOptions, ...args);
+        // command is only being used to output at this point so we can alter it
+        command = obfuscateValue(command, commandOptions);
+        if (result.code !== 0) {
+            // We want to make sure the full error message is displayed to the user, not just the error code.
+            // If outputChannel is defined, then we simply call 'outputChannel.show()' and throw a generic error telling the user to check the output window
+            // If outputChannel is _not_ defined, then we include the command's output in the error itself and rely on AzureActionHandler to display it properly
+            if (outputChannel) {
+                outputChannel.show();
+                throw new Error(localize('commandErrorWithOutput', 'Failed to run "{0}" command. Check output window for more details.', command));
```
EricJizbaMSFT: This error illustrates at least one reason why it's better to pass in "git" as the command and the args separately. The message would be 

> Failed to run "git" command

vs.

> Failed to run "git push -f ...... HEAD:master" command

...which probably has too much text for the notification

appservice/src/utils/cpUtils.ts
```diff
@@ -0,0 +1,111 @@
...
+import * as os from 'os';
+import * as vscode from 'vscode';
+import { localize } from '../localize';
+
+export namespace cpUtils {
+    export async function executeCommand(command: string, commandOptions?: ICommandOptions, ...args: string[]): Promise<string> {
+        const outputChannel: vscode.OutputChannel | undefined = commandOptions && commandOptions.outputChannel ? commandOptions.outputChannel : undefined;
+        const result: ICommandResult = await tryExecuteCommand(command, commandOptions, ...args);
+        // command is only being used to output at this point so we can alter it
+        command = obfuscateValue(command, commandOptions);
```
EricJizbaMSFT: You also need to obfuscate the args

appservice/src/utils/cpUtils.ts
nturinski: Do we want to enforce that they have to pass in commandOptions?  That means you'd always have to pass in undefined explicitly?

EricJizbaMSFT: This is a private function only called within this file. When would you pass `undefined`?

nturinski:
This isn't exactly done, but wanted some guidance on some of my patterns/naming conventions :)

EricJizbaMSFT:
Can you merge this with master? Was gonna review, but the diff is pretty gnarly

appservice/src/utils/cpUtils.ts
```diff
@@ -0,0 +1,83 @@
...
+        const command: string = commandOptions.command;
+        const result: ICommandResult = await tryExecuteCommand(commandOptions);
+        // command is only being used to output at this point so we can alter it
+        if (result.code !== 0) {
+            // We want to make sure the full error message is displayed to the user, not just the error code.
+            // If outputChannel is defined, then we simply call 'outputChannel.show()' and throw a generic error telling the user to check the output window
+            // If outputChannel is _not_ defined, then we include the command's output in the error itself and rely on AzureActionHandler to display it properly
+
+            if (commandOptions.showOutputChannel()) {
+                throw new Error(localize('commandErrorWithOutput', 'Failed to run "{0}" command. Check output window for more details.', command));
```
EricJizbaMSFT: `command` could have a value that needs to be obfuscated

nturinski: Well, I refactored to have command only have the command and args to have values that should be obfuscated.

nturinski:
Closing this in favor of https://github.com/Microsoft/vscode-azuretools/pull/402

PR #375, opened 2019-01-17
Provide shared webpack support

StephenWeatherford:


dev/reporter-options.json
```diff
@@ -0,0 +1,6 @@
+{
+    "reporterEnabled": "spec, mocha-junit-reporter",
+    "mochaJunitReporterReporterOptions": {
+        "mochaFile": "test-results.xml"
```
EricJizbaMSFT: I remember having problems on certain OS's if I didn't specify the full path. Does this report results on all three OS's?

dev/src/webpack/getDefaultWebpackConfig.ts
```diff
@@ -0,0 +1,288 @@
...
+                                    return '[path][name].[ext]';
+                                }
+                            }
+                        }
+                    ]
+                },
+
+                // Note: If you use`vscode-nls` to localize your extension than you likely also use`vscode-nls-dev` to create language bundles at build time.
+                // To support webpack, a loader has been added to vscode-nls-dev .Add the section below to the`modules/rules` configuration.
+                // {
```
EricJizbaMSFT: Can we just add this by default instead of commenting it out? All of our extensions should use vscode-nls eventually anyways

dev/package.json
```diff
@@ -27,18 +27,32 @@
...
         "compile": "tsc -watch -p ./",
         "lint": "tslint --project tsconfig.json -e src/*.d.ts -t verbose",
-        "test": "mocha out/test/**/*.js --ui tdd --reporter mocha-junit-reporter",
+        "test": "mocha out/test/**/*.js --ui tdd --reporter mocha-multi-reporters --reporter-options configFile=reporter-options.json",
         "prepare": "node ./node_modules/vscode/bin/install"
     },
     "dependencies": {},
     "devDependencies": {
+        "@types/clean-webpack-plugin": "^0.1.2",
+        "@types/copy-webpack-plugin": "^4.4.2",
```
EricJizbaMSFT: I still think almost all of these should be `dependencies` instead of `devDependencies`. That could be why you needed to turn off `no-implicit-dependencies`, which is a rule I actually really like. I get that it's counterintuitive, but this is already a dev package and it relies on these packages for it's main purpose. It won't be a problem in the extensions as long as they list `vscode-azureextensiondev` as _their_ devDependency

dev/src/webpack/getDefaultWebpackConfig.ts
StephenWeatherford: As none are using it, it's not tested.

dev/package.json
StephenWeatherford: I remember we talked about this before but couldn't remember what we'd decided, but everything right now was dev dependencies so I kept it that way (was thinking you had argued for that, couldn't remember why, maybe not).  We can either keep them dev dependencies and force users to also add as dev dependencies, or do them as regular dependencies and add the third party notices work (even though the actual using product won't ship with them). Don't have a strong opinion. Would it be easier for me to do a third party notice manually (have to do before publishing), or for you to set it up in the pipeline?

dev/src/webpack/getDefaultWebpackConfig.ts
StephenWeatherford: And another dependence that wouldn't actually be used, so I'm hesitant to bake it in.

dev/package.json
EricJizbaMSFT: > everything right now was dev dependencies so I kept it that way

Well yes you've added everything so far :P  (except maybe one)

> was thinking you had argued for that

Nope https://github.com/Microsoft/vscode-azuretools/pull/277#discussion_r222777914. I argued _for_ listing them in dependencies and you said "I think I'll leave it as is for now". I think it's wrong to list these as devDependencies and feel more strongly the more you add. It results in duplicate dependencies listed in the extensions and IMO leads to more bugs. For example if you remove or add a devDependency here that means you have to remember to do the same in every single extension repo. You're losing the whole benefit of npm's dependency structure that would handle this for us.

> Would it be easier for me to do a third party notice manually (have to do before publishing), or for you to set it up in the pipeline?

Easier for you to do one manually. I haven't done a whole lot of TPN stuff in this repo since we don't plan to webpack the packages directly.

dev/src/webpack/getDefaultWebpackConfig.ts
EricJizbaMSFT: functions and event grid use it :) I test it by selecting "Configure Display Language" and setting it to "pseudo". If you don't want to do this then I'm fine doing it myself since none of "your" extensions use it yet haha

StephenWeatherford: I'll enable it when we get to those, then.

dev/package.json
StephenWeatherford: Moved to dev deps.

dev/index.d.ts
```diff
@@ -37,3 +40,60 @@ export class TestOutputChannel implements OutputChannel {
...
+export declare function getDefaultWebpackConfig(options: DefaultWebpackOptions): webpack.Configuration;
+
+/**
+ * Installs the azure account extension before running tests (otherwise our extension would fail to activate)
+ * NOTE: The version isn't super important since we don't actually use the account extension in tests
+ */
+export declare function gulp_installAzureAccount(): Promise<void> | Stream;
+
+/**
+ * Spawns a child process
```
EricJizbaMSFT: I would write a better description than this. Honestly I think a normal `cp.spawn` will work in 99% of cases and webpack is just weird (as far as I know webpack is the only case that actually needs this)

dev/tslint.json
```diff
@@ -57,6 +57,8 @@
                 "keyIndex": 0,
                 "messageIndex": 1
             }
-        ]
+        ],
+        "no-implicit-dependencies": false,
```
EricJizbaMSFT: I actually really like this rule. We had a situation in the app service extension where `azure-storage` was listed as a dev dependency on accident which resulted in an unusable vsix (we didn't release the vsix - but we did send it to people to test haha). This rule would've caught that. Also you can exclude specific modules like this:
```
"no-implicit-dependencies": [
    true,
    [
        "vscode"
    ]
],
```

dev/src/webpack/getDefaultWebpackConfig.ts
```diff
@@ -0,0 +1,288 @@
...
+            __dirname: false
+        },
+
+        entry: {
+            // Note: Each entry is a completely separate Node.js application that cannot interact with any
+            // of the others, and that individually includes all dependencies necessary (i.e. common
+            // dependencies will have a copy in each entry file, no sharing).
+
+            // The entrypoint bundle for this extension, see https://webpack.js.org/configuration/entry-context/
+            extension: './extension.ts',
```
EricJizbaMSFT: Would prefer to finalize this conversation about names before other repos start using this, but otherwise PR looks good
https://github.com/Microsoft/vscode-azureappservice/pull/828#discussion_r249152159

dev/index.d.ts
StephenWeatherford: I think the original was mainly to deal with listening for process errors. Gulp no longer needs that, and I'm not sure what needed the .cmd/exe fix, doesn't seem to repro now, so I've removed.

dev/src/webpack/getDefaultWebpackConfig.ts
StephenWeatherford: Settled.

dev/index.d.ts
StephenWeatherford: Actually, still an issue (due to Windows shell behavior vs abc/abc.cmd/abc.exe/etc), but moved to only webpack code.

PR #376, opened 2019-01-17
Change "deployToStorageAccount" SAS to match func cli

EricJizbaMSFT:
deployToStorageAccount is a temporary measure while Linux Consumption is in preview and these numbers are all kinda arbitrary anyways, but I think it makes sense to match the func cli.

Got this from here:
https://github.com/Azure/azure-functions-core-tools/blob/master/src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs#L445


StephenWeatherford:
whoops

PR #379, opened 2019-01-18
Fix location logic for a few app service cases

EricJizbaMSFT:
1. Since function app supports advanced creation now, users can select an OS. Thus I have to set this _after_ that happens. Fixes https://github.com/Microsoft/vscode-azurefunctions/issues/917
1. Use `linuxWorkersEnabled` in the web app case. Not as noticeable, but it reduces the number of locations from 29 to 26. I verified the portal does the same thing.

### DIFF
--- a/appservice/src/createAppService/SiteOSStep.ts
+++ b/appservice/src/createAppService/SiteOSStep.ts
@@ -3,10 +3,11 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { AzureWizardPromptStep, IAzureQuickPickItem } from 'vscode-azureextensionui';
+import { WebSiteManagementClient } from 'azure-arm-website';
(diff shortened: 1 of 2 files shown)


appservice/src/createAppService/SiteOSStep.ts
```diff
@@ -20,6 +21,18 @@ export class SiteOSStep extends AzureWizardPromptStep<IAppServiceWizardContext>
...
+        if (wizardContext.newSiteOS === 'linux') {
+            if (wizardContext.newSiteKind === AppKind.functionapp) {
+                options = { linuxDynamicWorkersEnabled: true };
+            } else {
+                options = { linuxWorkersEnabled: true };
+            }
+        }
+        // Overwrite the generic 'locationsTask' with a list of locations specific to provider "Microsoft.Web"
+        const client: WebSiteManagementClient = createAzureClient(wizardContext, WebSiteManagementClient);
+        wizardContext.locationsTask = client.listGeoRegions(options);
```
nturinski: It's kinda weird to me that we're setting the locations in the OS code (though I get that it's dependant on it being Linux)

PR #385, opened 2019-01-23
Move the debounce call to prevent it getting logged in telemetry ever…

nturinski:
Move the debounce call to prevent it getting logged in telemetry every click


StephenWeatherford:
Thx.

PR #387, opened 2019-01-24
Saw/copy exclude

StephenWeatherford:
Node modules which were excluded from webpack weren't getting copied properly to dist.  Partial fix for https://github.com/Microsoft/vscode-cosmosdb/issues/1000.

Will publish as part of 0.1.3.

### DIFF
--- a/dev/src/webpack/excludeNodeModulesAndDependencies.ts
+++ b/dev/src/webpack/excludeNodeModulesAndDependencies.ts
@@ -43,7 +43,7 @@ export function excludeNodeModulesAndDependencies(
     //   so they can be found through normal require calls.
     // tslint:disable-next-line: strict-boolean-expressions
     webpackConfig.plugins = webpackConfig.plugins || [];
-    Object.assign(webpackConfig.plugins, new CopyWebpackPlugin(copyEntries));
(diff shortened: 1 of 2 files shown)


dev/src/webpack/excludeNodeModulesAndDependencies.ts
```diff
@@ -43,7 +43,7 @@ export function excludeNodeModulesAndDependencies(
     //   so they can be found through normal require calls.
     // tslint:disable-next-line: strict-boolean-expressions
     webpackConfig.plugins = webpackConfig.plugins || [];
-    Object.assign(webpackConfig.plugins, new CopyWebpackPlugin(copyEntries));
+    webpackConfig.plugins.push(new CopyWebpackPlugin(copyEntries));
```
EricJizbaMSFT: Does it work if there are two CopyWebpackPlugins with different copyEntries? Since you've got this here and one recently added in this cosmos PR... https://github.com/Microsoft/vscode-cosmosdb/pull/1008/files

StephenWeatherford: Not a problem, thanks for asking.

PR #389, opened 2019-01-24
Allow RegExp for TestUserInput

nturinski:
The point of this is to let users just write a regular expression and pick the input on the first match.  For example, /LTS/g will find the first runtime that has LTS in its label or description.


ui/src/TestUserInput.ts
```diff
@@ -65,7 +79,7 @@ export class TestUserInput implements IAzureUserInput {
     // tslint:disable-next-line:no-any
     public async showWarningMessage<T extends MessageItem>(message: string, ...args: any[]): Promise<T> {
         if (this._inputs.length > 0) {
-            const result: string | undefined = this._inputs.shift();
+            const result: string | RegExp | undefined = this._inputs.shift();
```
EricJizbaMSFT: Seems like this case should also throw an error or use the RegExp to match item.title. Don't really care which

PR #394, opened 2019-02-02
Slot Settings

nturinski:
Adds the ability to toggle a setting as a Slot Setting (which means it does NOT change after a slot swap).  I figured it could just be a context menu action and toggle it. 

This is true on the portal as well, but there is no distinction between which slot (production or testing) the slot setting exists on.  So if you have a setting named `setting` and mark it as a slot setting, `setting` is now a slot setting on every single slot in that web app.  Which is why there isn't a slot equivalence for the SDK calls.

image

"The web app"
image

"Deployment slot"
image

### DIFF
--- a/appservice/package-lock.json
+++ b/appservice/package-lock.json
@@ -1,6 +1,6 @@
 {
     "name": "vscode-azureappservice",
-    "version": "0.29.0",
+    "version": "0.29.1",
     "lockfileVersion": 1,
     "requires": true,
     "dependencies": {
--- a/appservice/package-lock.json
+++ b/appservice/package-lock.json
@@ -10,7 +10,7 @@
             "integrity": "sha512-UGcGgeMoGpFh97pQkzR+cgSrJDjVmO+CZ2N+WDI7i0QCOJE+PocpfHEEtePhlttULdbfOrzNi0yexg66E52Prw==",
             "dev": true,
             "requires": {
-                "@types/glob": "7.1.1"
+                "@types/glob": "*"
             }
         },
         "@types/caseless": {
@@ -31,7 +31,7 @@
             "integrity": "sha512-JAMFhOaHIciYVh8fb5/83nmuO/AHwmto+Hq7a9y8FzLDcC1KCU344XDOMEmahnrTFlHjgh4L0WJFczNIX2GxnQ==",
             "dev": true,
             "requires": {
-                "@types/node": "8.10.38"
+                "@types/node": "*"
             }
         },
         "@types/fs-extra": {
@@ -40,7 +40,7 @@
             "integrity": "sha512-Z5nu9Pbxj9yNeXIK3UwGlRdJth4cZ5sCq05nI7FaI6B0oz28nxkOtp6Lsz0ZnmLHJGvOJfB/VHxSTbVq/i6ujA==",
             "dev": true,
             "requires": {
-                "@types/node": "8.10.38"
+                "@types/node": "*"
             }
         },
         "@types/glob": {
@@ -49,9 +49,9 @@
             "integrity": "sha512-1Bh06cbWJUHMC97acuD6UMG29nMt0Aqz1vF3guLfG+kHHJhy3AyohZFFxYk2f7Q1SQIrNwvncxAE0N/9s70F2w==",
             "dev": true,
             "requires": {
-                "@types/events": "1.2.0",
-                "@types/minimatch": "3.0.3",
-                "@types/node": "8.10.38"
+                "@types/events": "*",
+                "@types/minimatch": "*",
+                "@types/node": "*"
             }
         },
         "@types/minimatch": {
@@ -71,7 +71,7 @@
             "integrity": "sha512-UQnV/R3AWC7ZGZQjAjr+SAwkWHiDYZxzFHqFFW36K17f8q4frldjFH4DO1WHMVngJ8f1OfKSpgmcGnWZakVbyw==",
             "dev": true,
             "requires": {
-                "@types/retry": "0.12.0"
+                "@types/retry": "*"
             }
         },
         "@types/request": {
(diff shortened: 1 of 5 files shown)


nturinski:
Also would like to point out this is another case where'd I'd like to use `initAndRefreshImpl()`

appservice/package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "vscode-azureappservice",
-    "version": "0.29.0",
+    "version": "0.29.1",
```
EricJizbaMSFT: This is out of date with master. You'll have to do 0.29.2 if you want to publish this

appservice/src/tree/AppSettingTreeItem.ts
```diff
@@ -77,4 +77,28 @@ export class AppSettingTreeItem extends AzureTreeItem<ISiteTreeRoot> {
         this._hideValue = !this._hideValue;
         await this.refresh();
     }
+
+    public async toggleSlotSetting(): Promise<void> {
+        const slotSettings: SlotConfigNamesResource = await this.root.client.listSlotConfigurationNames();
+        if (slotSettings.appSettingNames) {
```
EricJizbaMSFT: I feel like I should still be able to toggle slot settings even if `slotSettings.appSettingNames` isn't defined

appservice/src/tree/AppSettingTreeItem.ts
```diff
@@ -77,4 +77,28 @@ export class AppSettingTreeItem extends AzureTreeItem<ISiteTreeRoot> {
...
+            const slotSettingIndex: number = slotSettings.appSettingNames.findIndex((value: string) => { return value === this._key; });
```
EricJizbaMSFT: How does case sensitivity work for this? Can I have two settings with different casing?

appservice/src/tree/AppSettingTreeItem.ts
```diff
@@ -77,4 +77,28 @@ export class AppSettingTreeItem extends AzureTreeItem<ISiteTreeRoot> {
...
+
+            if (slotSettingIndex > 0) {
```
EricJizbaMSFT: Shouldn't this be `>=`?

appservice/src/tree/AppSettingTreeItem.ts
```diff
@@ -77,4 +77,28 @@ export class AppSettingTreeItem extends AzureTreeItem<ISiteTreeRoot> {
...
+            }
+            await this.root.client.updateSlotConfigurationNames(slotSettings);
+            await this.refresh();
+        }
+    }
+
+    public async refreshImpl(): Promise<void> {
+        const slotSettings: SlotConfigNamesResource = await this.root.client.listSlotConfigurationNames();
+        if (slotSettings.appSettingNames && slotSettings.appSettingNames.find((value: string) => { return value === this._key; })) {
+            this.description = 'Slot Setting';
```
EricJizbaMSFT: localize

appservice/src/tree/AppSettingsTreeItem.ts
```diff
@@ -57,11 +57,18 @@ export class AppSettingsTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
...
     public async loadMoreChildrenImpl(_clearCache: boolean): Promise<AzureTreeItem<ISiteTreeRoot>[]> {
         this._settings = await this.root.client.listApplicationSettings();
+        const slotConfigurationSettings: SlotConfigNamesResource = await this.root.client.listSlotConfigurationNames();
         const treeItems: AppSettingTreeItem[] = [];
         // tslint:disable-next-line:strict-boolean-expressions
         const properties: { [name: string]: string } = this._settings.properties || {};
         Object.keys(properties).forEach((key: string) => {
-            treeItems.push(new AppSettingTreeItem(this, key, properties[key], this._commandId));
+            const appSettingTreeItem: AppSettingTreeItem = new AppSettingTreeItem(this, key, properties[key], this._commandId);
+            if (slotConfigurationSettings.appSettingNames && slotConfigurationSettings.appSettingNames.find((value: string) => {
```
EricJizbaMSFT: This code is a duplicate of `refreshImpl`. I prefer this pattern:
https://github.com/Microsoft/vscode-azurestorage/blob/master/src/azureStorageExplorer/storageAccounts/storageAccountNode.ts#L63

Basically you make a static `createAppSettingTreeItem` and make the constructor private to ensure that `refreshImpl` is always called when the tree item is created.

appservice/src/tree/AppSettingTreeItem.ts
nturinski: The settings are case sensitive across the slots (including production).  So SETTING and setting are considered to be different slot settings.  However, settings aren't case sensitive within one slot, so you wouldn't be able to have setting and SETTING in the same slot.

appservice/src/tree/AppSettingsTreeItem.ts
nturinski: Hmm, one issue with doing it like that is I'm going to have to call `const slotSettings: SlotConfigNamesResource = await this.root.client.listSlotConfigurationNames();` each time a child is created vs. the one time (and then once per refresh).  Should I store that on the parent and refresh it whenever it's updated?

appservice/src/tree/AppSettingTreeItem.ts
```diff
@@ -77,4 +85,30 @@ export class AppSettingTreeItem extends AzureTreeItem<ISiteTreeRoot> {
...
+        if (!slotSettings.appSettingNames) {
+            slotSettings.appSettingNames = [];
+        }
+        const slotSettingIndex: number = slotSettings.appSettingNames.findIndex((value: string) => { return value === this._key; });
+
+        if (slotSettingIndex >= 0) {
+                slotSettings.appSettingNames.splice(slotSettingIndex, 1);
+            } else {
+                slotSettings.appSettingNames.push(this._key);
+            }
```
EricJizbaMSFT: got some weird formatting goin on here

appservice/src/tree/AppSettingsTreeItem.ts
EricJizbaMSFT: If you store it in the parent you just have to make sure you update it when the child is refreshed.

Or you could just make `slotSettings` an optional parameter to `refreshImpl`.

Or you could just leave it as is - idc

nturinski:
Confirmed it works with Functions.  I'll wait to merge and see what @fiveisprime thinks about the UI experience tomorrow.

PR #397, opened 2019-02-05
Fix finding of deep links

StephenWeatherford:
In the particular case I looked at we were picking up the wrong version of a module in determining dependencies, thus ending up with missing modules copied to node_modules in cosmos. Now look at subdependencies when listed, and not just "requires".

https://github.com/Microsoft/vscode-cosmosdb/issues/1020

### DIFF
--- a/dev/package-lock.json
+++ b/dev/package-lock.json
@@ -1,6 +1,6 @@
 {
     "name": "vscode-azureextensiondev",
-    "version": "0.1.5",
+    "version": "0.1.6",
     "lockfileVersion": 1,
     "requires": true,
     "dependencies": {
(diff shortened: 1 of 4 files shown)


dev/package-lock.json
```diff
@@ -16,8 +16,8 @@
             "integrity": "sha512-PDkSRY7KltW3M60hSBlerxI8SFPXsO3AL/aRVsO4Kh9IHRW74Ih75gUuTd/aE4LSSFqypb10UIX3QzOJwBQMGQ==",
             "dev": true,
             "requires": {
-                "@types/events": "*",
-                "@types/node": "*"
+                "@types/events": "1.2.0",
```
EricJizbaMSFT: Did you update this package lock before or after realizing you were on npm 5.6?

StephenWeatherford:
> Have you every looked at `npm ls --json`? It seems like a much easier tree to parse. Here's what I get for "engine.io". I may be wrong, but I think you just follow "dependencies" all the way down to get the whole list.

I did. Didn't look like it provided any better info, although I could be wrong.

dev/package-lock.json
StephenWeatherford: Had thought it was after but obviously not.  Fixed.

StephenWeatherford: I think I did an npm update as well, hence still a lot of changes but not the same kind.

StephenWeatherford:
> > Have you every looked at `npm ls --json`? It seems like a much easier tree to parse. Here's what I get for "engine.io". I may be wrong, but I think you just follow "dependencies" all the way down to get the whole list.
> 
> I did. Didn't look like it provided any better info, although I could be wrong.

Looks like it might warrant another look, but I'm not worry about it right now.

PR #396, opened 2019-02-05
Let functions extension handle exitCode for preDeployTask

EricJizbaMSFT:
See https://github.com/Microsoft/vscode-azurefunctions/issues/826. I need to check the exit code and handle it in the "func pack" case

This actually simplifies the code for the app service extension as well since I split it up into runPreDeployTask and tryRunPreDeployTask

PR #400, opened 2019-02-06
Remove RunFromPackage setting for Github deployment

nturinski:
Fixes https://github.com/Microsoft/vscode-azurefunctions/issues/888

I thought `verifyNoRunFromPackageSetting` made sense in preDeployTask since it should be verified prior to deployment, but I can move it to its own file if you don't like it there.

### DIFF
--- a/appservice/package.json
+++ b/appservice/package.json
@@ -1,7 +1,11 @@
 {
     "name": "vscode-azureappservice",
     "author": "Microsoft Corporation",
+<<<<<<< HEAD
     "version": "0.29.3",
+=======
+    "version": "0.30.0",
+>>>>>>> 32e2625cc1df1c9e4f231c07bc95edb27a304c40
     "description": "Common tools for developing Azure App Service extensions for VS Code",
     "tags": [
         "azure",
--- a/appservice/package.json
+++ b/appservice/package.json
@@ -1,7 +1,7 @@
 {
     "name": "vscode-azureappservice",
     "author": "Microsoft Corporation",
-    "version": "0.29.2",
+    "version": "0.29.3",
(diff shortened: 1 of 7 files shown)


appservice/package.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "vscode-azureappservice",
     "author": "Microsoft Corporation",
-    "version": "0.29.2",
+    "version": "0.29.3",
```
EricJizbaMSFT: I already bumped the version in https://github.com/Microsoft/vscode-azuretools/pull/396 and I can just include it in that

appservice/src/deploy/runPreDeployTask.ts
```diff
@@ -85,3 +88,20 @@ async function waitForPreDeployTask(preDeployTask: vscode.Task, actionContext: I
         }
     }
 }
+
+// prior to git deploying, these settings must be deleted or it will fail
+export async function verifyNoRunFromPackageSetting(client: SiteClient): Promise<void> {
```
EricJizbaMSFT: Yeah I don't like it in this file

appservice/src/connectToGitHub.ts
```diff
@@ -36,6 +37,8 @@ export async function connectToGitHub(node: AzureTreeItem, client: SiteClient, c
         const noToken: string = localize('noToken', 'No oAuth2 Token.');
         throw new Error(noToken);
     }
+    await verifyNoRunFromPackageSetting(client);
```
EricJizbaMSFT: I would to this after all user input, aka at the last chance possible. If they cancel at any point, we don't want to change their app settings

PR #401, opened 2019-02-06
View Commit in Github

nturinski:
Fixes https://github.com/Microsoft/vscode-azureappservice/issues/726

There are some limitations to this:
1. User must be connected to a GitHub repo
1. Commit must be from the currently connected GitHub repo because repo url is gotten from source level, not deployResult level

If the `request` fails, an error is thrown.

### DIFF
--- a/appservice/src/createAppService/AppServicePlanListStep.ts
+++ b/appservice/src/createAppService/AppServicePlanListStep.ts
@@ -66,15 +66,13 @@ export class AppServicePlanListStep extends AzureWizardPromptStep<IAppServiceWiz
         for (const plan of plans) {
             const isNewSiteLinux: boolean = wizardContext.newSiteOS === WebsiteOS.linux;
             const isPlanLinux: boolean = nonNullProp(plan, 'kind').toLowerCase().includes(WebsiteOS.linux);
-            const client: WebSiteManagementClient = createAzureClient(wizardContext, WebSiteManagementClient);
-            const fullAspDetails: AppServicePlan = await client.appServicePlans.get(nonNullProp(plan, 'resourceGroup'), nonNullProp(plan, 'name'));
             // plan.kind will contain "linux" for Linux plans, but will _not_ contain "windows" for Windows plans. Thus we check "isLinux" for both cases
             if (isNewSiteLinux === isPlanLinux) {
                 picks.push({
                     id: plan.id,
                     label: nonNullProp(plan, 'name'),
                     description: `${nonNullProp(plan, 'sku').name} (${plan.geoRegion})`,
-                    detail: `${plan.resourceGroup} (${fullAspDetails.numberOfSites})`,
+                    detail: plan.resourceGroup,
                     data: plan
                 });
             }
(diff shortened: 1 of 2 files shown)


appservice/src/tree/DeploymentTreeItem.ts
```diff
@@ -144,6 +148,30 @@ export class DeploymentTreeItem extends AzureTreeItem<ISiteTreeRoot> {
         });
     }
 
+    public async viewCommitInGitHub(): Promise<void> {
+        // if it's not connected to a .git repo, this won't work
```
EricJizbaMSFT: So you're doing quite a few checks here after the user clicks "View Commit in GitHub", but I think this is a case where I would rather just hide the context menu preemptively (if `ScmType !== Github`). I think we've gone to far down the rabbit hole of "never hide a context menu". I think it's okay in situations where the UI makes clear that nodes are different types or in different states. For example, the Cosmos DB extension has "Mongo" and "SQL" nodes at the same level with different context menus, but it's fine since we know they're different types. Just make clear in the UI if I'm connected to Github, Local Git, Zip deploy, etc.

appservice/src/tree/DeploymentTreeItem.ts
```diff
@@ -144,6 +148,30 @@ export class DeploymentTreeItem extends AzureTreeItem<ISiteTreeRoot> {
...
+        if (this.parent.contextValue === DeploymentsTreeItem.contextValueConnected) {
+            const sourceControl: SiteSourceControl = await this.root.client.getSourceControl();
+            const githubUrl: string = 'https://github.com';
+
+            if (sourceControl.repoUrl && sourceControl.repoUrl.match(githubUrl)) {
+                try {
+                    const githubCommitUrl: string = `${sourceControl.repoUrl}/commit/${this._deployResult.id}`;
+                    // tslint:disable-next-line:no-unsafe-any
+                    const githubPage: Response = await request({url: githubCommitUrl, resolveWithFullResponse: true}).promise();
+                    if (githubPage.statusCode === 200) {
```
EricJizbaMSFT: I'm hesitant to check the statusCode here. There's a risk that you'll block the user from viewing a commit in github that actually would've worked and I don't think the reward is worth it. If you don't check this they might get an error page on github saying "can't find this commit" but that seems fine to me

appservice/src/createAppService/AppServicePlanListStep.ts
```diff
@@ -66,13 +66,15 @@ export class AppServicePlanListStep extends AzureWizardPromptStep<IAppServiceWiz
...
+            const client: WebSiteManagementClient = createAzureClient(wizardContext, WebSiteManagementClient);
+            const fullAspDetails: AppServicePlan = await client.appServicePlans.get(nonNullProp(plan, 'resourceGroup'), nonNullProp(plan, 'name'));
             // plan.kind will contain "linux" for Linux plans, but will _not_ contain "windows" for Windows plans. Thus we check "isLinux" for both cases
             if (isNewSiteLinux === isPlanLinux) {
                 picks.push({
                     id: plan.id,
                     label: nonNullProp(plan, 'name'),
                     description: `${nonNullProp(plan, 'sku').name} (${plan.geoRegion})`,
-                    detail: plan.resourceGroup,
+                    detail: `${plan.resourceGroup} (${fullAspDetails.numberOfSites})`,
```
EricJizbaMSFT: How'd this stuff sneak into this PR? :P

PR #402, opened 2019-02-06
Wait for expected id to be latest before looking for logs

nturinski:
In favor of this PR: https://github.com/Microsoft/vscode-azuretools/pull/373

Fixes https://github.com/Microsoft/vscode-azurefunctions/issues/758
Fixes https://github.com/Microsoft/vscode-azureappservice/issues/754

Tested with redeploy and local git deployments.  This seems to work better since it will wait for the build to be triggered with the anticipated id before outputting the logs.  Before, the latest (if we didn't await the push or redeploy) would be the previous deployment's logs.

### DIFF
--- a/appservice/src/deploy/waitForDeploymentToComplete.ts
+++ b/appservice/src/deploy/waitForDeploymentToComplete.ts
@@ -3,6 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import { IParsedError, parseError } from 'vscode-azureextensionui';
 import KuduClient from 'vscode-azurekudu';
 import { DeployResult, LogEntry } from 'vscode-azurekudu/lib/models';
 import { ext } from '../extensionVariables';
@@ -37,11 +38,11 @@ export async function waitForDeploymentToComplete(client: SiteClient, kuduClient
         try {
             logEntries = <LogEntry[]>await kuduClient.deployment.getLogEntry(deployment.id);
         } catch (error) {
+            const parsedError: IParsedError = parseError(error);
             // Swallow 404 errors for a deployment while its still in the "temp" phase
             // (We can't reliably get logs until the deployment has shifted to the "permanent" phase)
-            // tslint:disable-next-line:no-unsafe-any
-            if (!deployment.isTemp || !error || error.statusCode !== 404) {
-                throw error;
+            if (!deployment.isTemp || parsedError.errorType !== '404') {
+                throw parsedError;
             }
         }
 
@@ -96,10 +97,10 @@ async function tryGetLatestDeployment(kuduClient: KuduClient, permanentId: strin
                 const latestDeployment: DeployResult = await kuduClient.deployment.getResult('latest');
                 [deployment, permanentId] = latestDeployment.id === expectedId ? [latestDeployment, latestDeployment.id] : [undefined, undefined];
             } catch (error) {
+                const parsedError: IParsedError = parseError(error);
                 // swallow 404 error since "latest" might not exist on the first deployment
-                if (error.statusCode !== 404) {
-                    // tslint:disable-next-line:no-unsafe-any
-                    throw error;
+                if (parsedError.errorType !== '404') {
+                    throw parsedError;
                 }
             }
         } else if (initialStartTime) {
@@ -118,10 +119,10 @@ async function tryGetLatestDeployment(kuduClient: KuduClient, permanentId: strin
             try {
                 deployment = <DeployResult | undefined>await kuduClient.deployment.getResult('latest');
             } catch (error) {
+                const parsedError: IParsedError = parseError(error);
                 // swallow 404 error since "latest" might not exist on the first deployment
-                if (error.statusCode !== 404) {
-                    // tslint:disable-next-line:no-unsafe-any
-                    throw error;
+                if (parsedError.errorType !== '404') {
+                    throw parsedError;
                 }
             }
             if (deployment && deployment.startTime) {
--- a/appservice/src/deploy/waitForDeploymentToComplete.ts
+++ b/appservice/src/deploy/waitForDeploymentToComplete.ts
@@ -18,7 +18,7 @@ export async function waitForDeploymentToComplete(client: SiteClient, kuduClient
     let initialStartTime: Date | undefined;
     let deployment: DeployResult | undefined;
     let permanentId: string | undefined;
-    // a 20 second timeout period to let Kudu initialize the deployment
+    // a 30 second timeout period to let Kudu initialize the deployment
     const maxTimeToWaitForPermanentId: number = Date.now() + 30 * 1000;
 
     // tslint:disable-next-line:no-constant-condition
--- a/appservice/src/deploy/waitForDeploymentToComplete.ts
+++ b/appservice/src/deploy/waitForDeploymentToComplete.ts
@@ -27,7 +27,7 @@ export async function waitForDeploymentToComplete(client: SiteClient, kuduClient
         [deployment, permanentId, initialStartTime] = await tryGetLatestDeployment(kuduClient, permanentId, initialStartTime, expectedId);
         if ((deployment === undefined || !deployment.id)) {
             if (expectedId && Date.now() < maxTimeToWaitForExpectedId) {
-                ext.outputChannel.appendLine(formatDeployLog(client, localize('waitingForBuild', 'Waiting for build to trigger...')));
+                ext.outputChannel.appendLine(formatDeployLog(client, localize('waitingForBuild', 'Starting deployment...')));
                 await delay(pollingInterval);
                 continue;
             }
--- a/appservice/src/deploy/waitForDeploymentToComplete.ts
+++ b/appservice/src/deploy/waitForDeploymentToComplete.ts
@@ -8,29 +8,30 @@ import { DeployResult, LogEntry } from 'vscode-azurekudu/lib/models';
 import { ext } from '../extensionVariables';
 import { localize } from '../localize';
 import { SiteClient } from '../SiteClient';
+import { delay } from '../utils/delay';
 import { nonNullProp } from '../utils/nonNull';
 import { formatDeployLog } from './formatDeployLog';
 
-export async function waitForDeploymentToComplete(client: SiteClient, kuduClient: KuduClient, permanentId?: string, pollingInterval: number = 5000): Promise<void> {
+export async function waitForDeploymentToComplete(client: SiteClient, kuduClient: KuduClient, expectedId?: string, pollingInterval: number = 5000): Promise<void> {
     const alreadyDisplayedLogs: string[] = [];
     let nextTimeToDisplayWaitingLog: number = Date.now();
     let initialStartTime: Date | undefined;
     let deployment: DeployResult | undefined;
+    let permanentId: string | undefined;
     // a 20 second timeout period to let Kudu initialize the deployment
-    const maxTimeToWaitForPermanentId: number = Date.now() + 20 * 1000;
+    const maxTimeToWaitForPermanentId: number = Date.now() + 30 * 1000;
 
     // tslint:disable-next-line:no-constant-condition
     while (true) {
-        [deployment, permanentId, initialStartTime] = await getLatestDeployment(kuduClient, permanentId, initialStartTime);
+        [deployment, permanentId, initialStartTime] = await tryGetLatestDeployment(kuduClient, permanentId, initialStartTime, expectedId);
         if ((deployment === undefined || !deployment.id)) {
-            if (permanentId && Date.now() < maxTimeToWaitForPermanentId) {
-                await new Promise((resolve: () => void): void => { setTimeout(resolve, pollingInterval); });
+            if (expectedId && Date.now() < maxTimeToWaitForPermanentId) {
+                await delay(pollingInterval);
                 continue;
             }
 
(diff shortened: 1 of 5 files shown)


appservice/src/deploy/waitForDeploymentToComplete.ts
```diff
@@ -105,3 +109,28 @@ async function getLatestDeployment(kuduClient: KuduClient, permanentId: string |
 
     return [deployment, permanentId, initialStartTime];
 }
+
+async function waitForDeploymentToStart(kuduClient: KuduClient, id: string): Promise<void> {
```
EricJizbaMSFT: I feel like all of this logic can be more tightly integrated with the existing code rather than being it's own function. For example, `expectedId` seems very similar to `permanentId` and you've got two loops with two polling intervals when you could just have one

Biggest thing you'd have to change is this, but I'm sure you could refactor somehow (since `deployment` would probably be undefined until it actually starts):
```
if (deployment === undefined || !deployment.id) {
        throw new Error(localize('failedToFindDeployment', 'Failed to get status of deployment.'));
}
```

appservice/src/deploy/waitForDeploymentToComplete.ts
```diff
@@ -11,13 +11,17 @@ import { SiteClient } from '../SiteClient';
 import { nonNullProp } from '../utils/nonNull';
 import { formatDeployLog } from './formatDeployLog';
 
-export async function waitForDeploymentToComplete(client: SiteClient, kuduClient: KuduClient, pollingInterval: number = 5000): Promise<void> {
+export async function waitForDeploymentToComplete(client: SiteClient, kuduClient: KuduClient, pollingInterval: number = 5000, expectedId?: string): Promise<void> {
```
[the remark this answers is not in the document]
nturinski: Okaaaay, but that does make it a breaking change technically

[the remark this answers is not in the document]
EricJizbaMSFT: Breaking changes only matter for stuff listed in index.ts: https://github.com/Microsoft/vscode-azuretools/blob/master/appservice/src/index.ts

waitForDeploymentToComplete is an internal function and we can change the signature however we want without worrying about the version

nturinski:
Bumped to new minor number as this is a breaking change.

appservice/src/deploy/waitForDeploymentToComplete.ts
```diff
@@ -11,18 +11,23 @@ import { SiteClient } from '../SiteClient';
...
 
-export async function waitForDeploymentToComplete(client: SiteClient, kuduClient: KuduClient, pollingInterval: number = 5000): Promise<void> {
+export async function waitForDeploymentToComplete(client: SiteClient, kuduClient: KuduClient, permanentId?: string, pollingInterval: number = 5000): Promise<void> {
     const alreadyDisplayedLogs: string[] = [];
     let nextTimeToDisplayWaitingLog: number = Date.now();
-    let permanentId: string | undefined;
     let initialStartTime: Date | undefined;
     let deployment: DeployResult | undefined;
+    // a 20 second timeout period to let Kudu initialize the deployment
+    const maxTimeToWaitForPermanentId: number = Date.now() + 20 * 1000;
```
EricJizbaMSFT: Any ideas why I ran into this error before?: https://github.com/Microsoft/vscode-azurefunctions/issues/982

Do we just need to increase the timeout?

appservice/src/deploy/waitForDeploymentToComplete.ts
```diff
@@ -81,7 +86,9 @@ async function getLatestDeployment(kuduClient: KuduClient, permanentId: string |
     let deployment: DeployResult | undefined;
     if (permanentId) {
         // Use "permanentId" to find the deployment during its "permanent" phase
-        deployment = await kuduClient.deployment.getResult(permanentId);
+        // if we have a "permanentId" we know which deployment we are looking for, so wait until latest id reflects that
+        const latestDeployment: DeployResult = await kuduClient.deployment.getResult('latest');
+        deployment = latestDeployment.id === permanentId ? latestDeployment : undefined;
     } else if (initialStartTime) {
         // Use "initialReceivedTime" to find the deployment during its "temp" phase
         deployment = (await kuduClient.deployment.getDeployResults())
```
EricJizbaMSFT: Can you refresh my memory on why the existing `latest` and `initialStartTime` logic doesn't work? Your change is definitely an improvement and I like it, but we're still going to miss the logs during the "temp" phase (although I don't remember what happens in the temp phase or how long it takes)

appservice/src/deploy/waitForDeploymentToComplete.ts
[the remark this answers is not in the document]
nturinski: Doh, I totally missed this comment.  You right doe.

appservice/src/deploy/waitForDeploymentToComplete.ts
nturinski: Hmm, I kind of doubt that it's the timeout.  We're just waiting for the kudu build to initialize so that we can find the deployment id.  I get the feeling that for some reason, the build never initializes.  Not sure if that'd be something on our end or if Kudu never picks up our redeploy request?

appservice/src/deploy/waitForDeploymentToComplete.ts
nturinski: I think this is what you're asking:

So the reason `Redeploy` and `Local Git` deployment don't work with old code is due to a timing issue.  Both methods can't be `await`ed in order for the logs to stream in real-time because the API call doesn't resolve _until_ after the deployment has completed.  But due to the Kudu taking some time to trigger the new build, our old code would grab the "latest" (which would be the previous deployment) and just return all of those logs immediately.

EricJizbaMSFT: Okay yep I remember now

appservice/src/deploy/waitForDeploymentToComplete.ts
EricJizbaMSFT: The only reason I could see it being the timeout is because there is some sort of "temp" phase for deployments that happens after it triggers and before `deployment.id` is the permanent id. I just don't remember what happens in that temp phase or how long it takes

EricJizbaMSFT: Also when I filed that bug - it did successfully redeploy even though I got the timeout error.

nturinski: I'm not 100% sure about this, but I think the "temp" phase is only for new deployments.  However, redeploy already has a `DeployResult` associated with it, so rather than creating a new one, Kudu reuses it to redeploy that deployment. (the game is to see how many times I can say deploy in one sentence)

EricJizbaMSFT: Okay well maybe just try redeploy a few times and time it somehow? (some console.logs maybe?). If it's remotely close to 20 then bump it otherwise I'm fine to leave it as is

appservice/src/deploy/waitForDeploymentToComplete.ts
```diff
@@ -8,24 +8,30 @@ import { DeployResult, LogEntry } from 'vscode-azurekudu/lib/models';
...
-export async function waitForDeploymentToComplete(client: SiteClient, kuduClient: KuduClient, pollingInterval: number = 5000): Promise<void> {
+export async function waitForDeploymentToComplete(client: SiteClient, kuduClient: KuduClient, expectedId?: string, pollingInterval: number = 5000): Promise<void> {
     const alreadyDisplayedLogs: string[] = [];
     let nextTimeToDisplayWaitingLog: number = Date.now();
-    let permanentId: string | undefined;
     let initialStartTime: Date | undefined;
     let deployment: DeployResult | undefined;
+    let permanentId: string | undefined;
+    // a 30 second timeout period to let Kudu initialize the deployment
+    const maxTimeToWaitForPermanentId: number = Date.now() + 30 * 1000;
```
EricJizbaMSFT: nit: maxTimeToWaitForExpectedId

appservice/src/deploy/waitForDeploymentToComplete.ts
```diff
@@ -72,35 +78,43 @@ export async function waitForDeploymentToComplete(client: SiteClient, kuduClient
...
+            // Use "latest" to get the deployment before we know the "initialReceivedTime" or "permanentId"
+            deployment = <DeployResult | undefined>await kuduClient.deployment.getResult('latest');
+            if (deployment && deployment.startTime) {
+                // Make note of the startTime because that is when kudu has began the deployment process,
+                // so that we can use that to find the deployment going forward
+                initialStartTime = deployment.startTime;
+            }
         }
+    } catch (error) {
+        // swallow error since "latest" might not exist yet
```
EricJizbaMSFT: Can we minimize the scope of this try catch? Like just put it around the call to get latest? And just swallow the 404 error? Or whatever it is...

appservice/src/deploy/waitForDeploymentToComplete.ts
```diff
@@ -72,35 +78,43 @@ export async function waitForDeploymentToComplete(client: SiteClient, kuduClient
...
-            // so that we can use that to find the deployment going forward
-            initialStartTime = deployment.startTime;
+    try {
+        if (permanentId) {
+            // Use "permanentId" to find the deployment during its "permanent" phase
+            deployment = await kuduClient.deployment.getResult(permanentId);
+        } else if (expectedId) {
+            // if we have a "expectedId" we know which deployment we are looking for, so wait until latest id reflects that
+            const latestDeployment: DeployResult = await kuduClient.deployment.getResult('latest');
+            permanentId = latestDeployment.id === expectedId ? latestDeployment.id : undefined;
```
EricJizbaMSFT: I don't see you setting `deployment` in this case

nturinski: I didn't intentionally since it will be set the next time the command is called (due to having a `permanentId` now).  I guess this could cause a 5 second delay for output though so I'll change that.

appservice/src/deploy/waitForDeploymentToComplete.ts
```diff
@@ -8,24 +8,31 @@ import { DeployResult, LogEntry } from 'vscode-azurekudu/lib/models';
...
+    // a 30 second timeout period to let Kudu initialize the deployment
+    const maxTimeToWaitForExpectedId: number = Date.now() + 30 * 1000;
 
     // tslint:disable-next-line:no-constant-condition
     while (true) {
-        [deployment, permanentId, initialStartTime] = await getLatestDeployment(kuduClient, permanentId, initialStartTime);
+        [deployment, permanentId, initialStartTime] = await tryGetLatestDeployment(kuduClient, permanentId, initialStartTime, expectedId);
+        if ((deployment === undefined || !deployment.id)) {
+            if (expectedId && Date.now() < maxTimeToWaitForExpectedId) {
+                ext.outputChannel.appendLine(formatDeployLog(client, localize('waitingForBuild', 'Waiting for build to trigger...')));
```
nturinski: On the first deploy, there is a considerable amount of time where nothing is displayed in the output channel, so I thought this might be useful.  Thoughts?

EricJizbaMSFT: A message is fine, but I would prefer if it was consistent with zipDeploy:
https://github.com/Microsoft/vscode-azuretools/blob/master/appservice/src/deploy/deployZip.ts#L36

appservice/src/deploy/waitForDeploymentToComplete.ts
```diff
@@ -72,36 +79,55 @@ export async function waitForDeploymentToComplete(client: SiteClient, kuduClient
...
+            // Use "permanentId" to find the deployment during its "permanent" phase
+            deployment = await kuduClient.deployment.getResult(permanentId);
+        } else if (expectedId) {
+            // if we have a "expectedId" we know which deployment we are looking for, so wait until latest id reflects that
+            try {
+                const latestDeployment: DeployResult = await kuduClient.deployment.getResult('latest');
+                [deployment, permanentId] = latestDeployment.id === expectedId ? [latestDeployment, latestDeployment.id] : [undefined, undefined];
+            } catch (error) {
+                // swallow 404 error since "latest" might not exist on the first deployment
+                if (error.statusCode !== 404) {
```
EricJizbaMSFT: `parseError(error).errorType !== 404`

appservice/src/deploy/waitForDeploymentToComplete.ts
```diff
@@ -27,7 +27,7 @@ export async function waitForDeploymentToComplete(client: SiteClient, kuduClient
         [deployment, permanentId, initialStartTime] = await tryGetLatestDeployment(kuduClient, permanentId, initialStartTime, expectedId);
         if ((deployment === undefined || !deployment.id)) {
             if (expectedId && Date.now() < maxTimeToWaitForExpectedId) {
-                ext.outputChannel.appendLine(formatDeployLog(client, localize('waitingForBuild', 'Waiting for build to trigger...')));
+                ext.outputChannel.appendLine(formatDeployLog(client, localize('waitingForBuild', 'Starting deployment...')));
```
EricJizbaMSFT: So I didn't think about the fact that this message is logged multiple times. In that case, "Waiting" makes more sense, but I don't think I like the "Waiting" messages.

I guess I'd prefer a message that was _truly_ more consistent with zipDeploy in the sense that it's only displayed once

PR #403, opened 2019-02-08
Fix opn and clipboardy on Linux

StephenWeatherford:
Two issues:
1) opn has a dynamically-loaded binary and needs to always be excluded (also added clipboardy)
2) copy-webpack-plugin doesn't preserve the executable bit when copying node_modules, switch to filemanager-webpack-plugin

### DIFF
--- a/dev/src/webpack/getDefaultWebpackConfig.ts
+++ b/dev/src/webpack/getDefaultWebpackConfig.ts
@@ -6,7 +6,7 @@
 // tslint:disable: no-unsafe-any // Lots of plugin functions use any
 
 import * as CleanWebpackPlugin from 'clean-webpack-plugin';
-import * as CopyWebpackPlugin from 'copy-webpack-plugin';
+import * as FileManagerPlugin from 'filemanager-webpack-plugin';
 import * as fse from 'fs-extra';
 import * as path from 'path';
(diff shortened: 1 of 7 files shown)


dev/src/webpack/getDefaultWebpackConfig.ts
```diff
@@ -109,10 +114,17 @@ export function getDefaultWebpackConfig(options: DefaultWebpackOptions): webpack
...
-            new CopyWebpackPlugin([
-                // Test files -> dist/test (these files are ignored during packaging)
-                { from: './out/test', to: 'test/' }
-            ]),
+            new FileManagerPlugin({
+                onEnd: {
+                    copy: [
+                        // Test files -> dist/test (these files are ignored during packaging)
+                        {
+                            source: path.join(options.projectRoot, 'out/test'),
```
EricJizbaMSFT: nit:
```
path.join(options.projectRoot, 'out', 'test'),
```

nturinski:
@StephenWeatherford Nice.  This will fix the regression in App Service? https://github.com/Microsoft/vscode-azureappservice/issues/857

StephenWeatherford:
> @StephenWeatherford Nice. This will fix the regression in App Service? [Microsoft/vscode-azureappservice#857](https://github.com/Microsoft/vscode-azureappservice/issues/857)

Yes.

PR #404, opened 2019-02-09
Fix usage of cwd

StephenWeatherford:
cwd needs to remain the root path. Not sure why this issue only affects docker, but I tested the change with functions and it's fine.

### DIFF
--- a/dev/src/gulp/gulp_webpack.ts
+++ b/dev/src/gulp/gulp_webpack.ts
@@ -17,7 +17,9 @@ export function gulp_webpack(mode: string): cp.ChildProcess {
 
     return cp.spawn(
         path.join(
-            './node_modules/.bin/',
+            '.',
+            'node_modules',
+            '.bin',
             // https://github.com/nodejs/node-v0.x-archive/issues/2318#issuecomment-249355505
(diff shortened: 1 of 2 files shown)


EricJizbaMSFT:
I think this change is already in master?
https://github.com/Microsoft/vscode-azuretools/blob/master/dev/src/gulp/gulp_webpack.ts

EricJizbaMSFT:
Never mind disregard comment

dev/src/gulp/gulp_webpack.ts
```diff
@@ -15,14 +16,15 @@ export function gulp_webpack(mode: string): cp.ChildProcess {
     env.NODE_OPTIONS = '--max-old-space-size=8192';
 
     return cp.spawn(
-        // https://github.com/nodejs/node-v0.x-archive/issues/2318#issuecomment-249355505
-        process.platform === 'win32' ? 'webpack.cmd' : 'webpack',
+        path.join(
+            './node_modules/.bin/',
```
EricJizbaMSFT: nit: split up the node_modules and bin in path.join

StephenWeatherford:
Both published.

PR #405, opened 2019-02-09
Fix defaultExternalNodeModules being filtered out

EricJizbaMSFT:
I tried @StephenWeatherford's fix for opn and noticed it wasn't quite working

Honestly I think there's still an issue here if opn is a dependency of a dependency instead of a direct dependency (right?), but I'd rather get this in just for next week's release

### DIFF
--- a/dev/package-lock.json
+++ b/dev/package-lock.json
@@ -1,6 +1,6 @@
 {
     "name": "vscode-azureextensiondev",
-    "version": "0.1.7",
+    "version": "0.1.8",
     "lockfileVersion": 1,
     "requires": true,
     "dependencies": {
--- a/dev/package.json
+++ b/dev/package.json
@@ -1,7 +1,7 @@
 {
     "name": "vscode-azureextensiondev",
     "author": "Microsoft Corporation",
-    "version": "0.1.7",
+    "version": "0.1.8",
     "description": "Common dev dependency tools for developing Azure extensions for VS Code",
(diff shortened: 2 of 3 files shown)


EricJizbaMSFT:
The issue is we need to check _dependencies_ within packageLockJson. I don't think any packageLock just has packages at the root of the json object

StephenWeatherford:
I don't understand the issue from the comments, but I'll go ahead and put it through.

EricJizbaMSFT:
@StephenWeatherford the packagelock file looks like this:
```json
{
    "name": "vscode-azurefunctions",
    "version": "0.14.0",
    "lockfileVersion": 1,
    "requires": true,
    "dependencies": {
        "opn": {
            "version": "5.4.0",
            "resolved": "https://registry.npmjs.org/opn/-/opn-5.4.0.tgz",
            "integrity": "sha512-YF9MNdVy/0qvJvDtunAOzFw9iasOQHpVthTCvGzxt61Il64AYSGdK+rYwld7NAfk9qJ7dt+hymBNSc9LNYS+Sw==",
            "requires": {
                "is-wsl": "^1.1.0"
            }
        },
```

The code to "Only use defaultExternalNodeModules entries that are actually in package-lock.json" was always returning an empty array because it was looking for `!!packageLockJson['opn']` instead of `!!packageLockJson.dependencies['opn']`

PR #407, opened 2019-02-12
Create new ASP for users in basic creation

nturinski:
Fixes https://github.com/Microsoft/vscode-azureappservice/issues/795 (kind of)

For some reason, the method we call to list all the asp's today does not retrieve the property `numOfSites`.  In order to get that information, I had to retrieve each one with its own API call and it delayed the quickPick prompt considerably.  So I'm thinking it might not be worth it to display that in the UI.

If a user is in basic creation, I feel like rather than prompting them about performance issues, we should just create the new ASP for them when they're creating their fourth app.  I don't really see a downfall for doing that for them.

### DIFF
--- a/appservice/src/createAppService/AppServicePlanCreateStep.ts
+++ b/appservice/src/createAppService/AppServicePlanCreateStep.ts
@@ -10,7 +10,9 @@ import { ext } from '../extensionVariables';
 import { localize } from '../localize';
 import { nonNullProp, nonNullValueAndProp } from '../utils/nonNull';
 import { getAppServicePlanModelKind, WebsiteOS } from './AppKind';
+import { appServicePlanNamingRules } from './AppServicePlanNameStep';
 import { IAppServiceWizardContext } from './IAppServiceWizardContext';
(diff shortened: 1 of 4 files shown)


nturinski:
I was hoping you weren't going to mention that...

nturinski:
As I've been refactoring, I've realized that the asp and rg names are OS reliant, and we can't guarantee we will have that information until the user has been prompted for it (since we don't pick a runtime for the user if we don't see the appropriate files) so I think this PR will have to remain in the appservice package.

EricJizbaMSFT:
Take a look at createFunctionAppSettings https://github.com/Microsoft/vscode-azuretools/blob/master/appservice/src/createAppService/IAppCreateOptions.ts#L13

Maybe we could add a property to creationOptions like:
```
afterPrompt?(context: WizardContext): Promise<WizardContext>;
```

I'm not saying you _have_ to do this refactoring now, but we've had several PRs affecting this code recently and it would be a lot easier (IMO) if it was in the app service directly

EricJizbaMSFT:
Closing this due to lack of activity

PR #410, opened 2019-02-12
Add the support for WildFly image

### LINKED ISSUE #409: Add the WildFly support
To add the support about the new Java image named `WildFly` in the Linux Web App.

jdneo:

The usage of WildFly image is very likely to Tomcat: simply deploy the war file to the Web App...

Resolve #409

### DIFF
--- a/appservice/src/utils/javaUtils.ts
+++ b/appservice/src/utils/javaUtils.ts
@@ -12,8 +12,17 @@ export namespace javaUtils {
     const DEFAULT_PORT: string = '8080';
     const PORT_KEY: string = 'PORT';
 
-    export function isJavaTomcatRuntime(runtime: string | undefined): boolean {
-        return !!runtime && runtime.toLowerCase().startsWith('tomcat');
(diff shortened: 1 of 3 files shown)


appservice/src/utils/javaUtils.ts
```diff
@@ -12,8 +12,17 @@ export namespace javaUtils {
     const DEFAULT_PORT: string = '8080';
     const PORT_KEY: string = 'PORT';
 
-    export function isJavaTomcatRuntime(runtime: string | undefined): boolean {
-        return !!runtime && runtime.toLowerCase().startsWith('tomcat');
+    export function needDeployWarFile(runtime: string | undefined): boolean {
+        if (!runtime) {
```
EricJizbaMSFT: This can be done in one line with the use of a RegExp:
```
return !!runtime && /^(tomcat|wildfly)/i.test(runtime);
```

If you're not familiar with RegExp, you can use a site like this to get explanations and test it out: https://regex101.com/r/1oyXnJ/1

PR #411, opened 2019-02-12
Enable smart default for Java project when creating a new Web App

jdneo:

Resolve https://github.com/Microsoft/vscode-azureappservice/issues/848

### DIFF
--- a/appservice/src/createAppService/createWebApp.ts
+++ b/appservice/src/createAppService/createWebApp.ts
@@ -45,8 +45,6 @@ export async function setWizardContextDefaults(wizardContext: IAppServiceWizardC
         } else if (await javaUtils.isJavaProject()) {
             wizardContext.newSiteOS = WebsiteOS.linux;
             wizardContext.newPlanSku = { name: 'P1v2', tier: 'PremiumV2', size: 'P1v2', family: 'P', capacity: 1 };
-            // To avoid 'Requested features are not supported in region' error
-            await LocationListStep.setLocation(wizardContext, 'westeurope');
         } else {
             await workspace.findFiles('*.csproj').then((files: Uri[]) => {
                 if (files.length > 0) {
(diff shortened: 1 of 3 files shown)


appservice/src/createAppService/createWebApp.ts
```diff
@@ -30,6 +31,9 @@ export async function setWizardContextDefaults(wizardContext: IAppServiceWizardC
         } else if (await fse.pathExists(path.join(fsPath, 'requirements.txt'))) {
             // requirements.txt are used to pip install so a good way to determine it's a Python app
             wizardContext.recommendedSiteRuntime = LinuxRuntimes.python;
+        } else if (await javaUtils.isJavaProject(fsPath)) {
```
EricJizbaMSFT: You're setting things that result in fewer prompts, which should only be done `if (!advancedCreation)` (like below). `recommendedSiteRuntime` is unique because it moves runtimes to the top of the quick pick, but _still prompts the user_. Thus it can ignore `advancedCreation`.

appservice/src/createAppService/createWebApp.ts
```diff
@@ -30,6 +31,9 @@ export async function setWizardContextDefaults(wizardContext: IAppServiceWizardC
...
+            wizardContext.newSiteOS = WebsiteOS.linux;
+            wizardContext.newPlanSku = { name: 'P1v2', tier: 'PremiumV2', size: 'P1v2', family: 'P', capacity: 1 };
```
EricJizbaMSFT: On a related note - do you want to leverage `recommendedSiteRuntime`? You'd have to refactor a bit since I assume you want to recommend multiple runtimes (java, tomcat, and wildfly), but that could be done. Just a question - not required for this PR.

jdneo: Yes, I was intended to address it in another PR.

jdneo:
Some explanation about the new commit:
- We only think the workspace is a Java workspace **only** if **all** of the workspace folder root contains `pom.xml` or `build.gradle`.
- If it's a Java project, the default location will be changed to West Europe. Cuz I'm keeping receiving `Requested features are not supported in region` in Central US. Might be the P1V2 resource exceeds the limit in this region? Not quite sure for that.

PR #413, opened 2019-02-14
Recomend the Java related runtime in list

jdneo:

### DIFF
--- a/appservice/src/createAppService/SiteRuntimeStep.ts
+++ b/appservice/src/createAppService/SiteRuntimeStep.ts
@@ -5,7 +5,7 @@
 
 import { AzureWizardPromptStep, IAzureQuickPickItem } from 'vscode-azureextensionui';
 import { ext } from '../extensionVariables';
-import { AppKind, LinuxRuntimes, WebsiteOS } from './AppKind';
+import { AppKind, WebsiteOS } from './AppKind';
 import { IAppServiceWizardContext } from './IAppServiceWizardContext';
 
 interface ILinuxRuntimeStack {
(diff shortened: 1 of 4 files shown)


appservice/src/createAppService/createWebApp.ts
```diff
@@ -23,28 +23,38 @@ export async function createWebApp(
 }
 
 export async function setWizardContextDefaults(wizardContext: IAppServiceWizardContext, actionContext: IActionContext, advancedCreation?: boolean): Promise<void> {
+    const isJavaProject: boolean = await javaUtils.isJavaProject();
+    if (isJavaProject) {
```
jdneo: @EricJizbaMSFT Do you think we need to enable the multi-root for `js` & `python` here?

The current implementation is separated.

appservice/src/createAppService/createWebApp.ts
```diff
@@ -23,28 +23,38 @@ export async function createWebApp(
...
+        actionContext.properties.recommendedSiteRuntime = wizardContext.recommendedSiteRuntime ? [...wizardContext.recommendedSiteRuntime.keys()].join(',') : undefined;
     }
 
     if (!advancedCreation) {
         await LocationListStep.setLocation(wizardContext, 'centralus');
+        if (isJavaProject) {
+            // considering high resource requirement for Java applications, a higher plan sku is set here
+            wizardContext.newPlanSku = { name: 'P1v2', tier: 'PremiumV2', size: 'P1v2', family: 'P', capacity: 1 };
+            // to avoid 'Requested features are not supported in region' error
+            await LocationListStep.setLocation(wizardContext, 'westeurope');
```
jdneo: I found the `P1v2` is not always available in `Central US`. I've met this problem, but not stable repro.

So the location is set to `West Europe` here.

jdneo:
Hi @EricJizbaMSFT,

Would you mind to take a look at this PR?

For the issue that P1v2 may fail to provision, I've sent emails to the service team and waiting for the response.

appservice/src/createAppService/createWebApp.ts
EricJizbaMSFT: I'm not worried about multi-root. According to VS Code folks it has pretty low usage

appservice/src/createAppService/SiteRuntimeStep.ts
```diff
@@ -167,15 +167,19 @@ export class SiteRuntimeStep extends AzureWizardPromptStep<IAppServiceWizardCont
         ];
     }
 
-    private sortQuickPicksByRuntime(runtimeItems: IAzureQuickPickItem<ILinuxRuntimeStack>[], runtime: string): IAzureQuickPickItem<ILinuxRuntimeStack>[] {
+    private sortQuickPicksByRuntime(runtimeItems: IAzureQuickPickItem<ILinuxRuntimeStack>[], recomendedRuntimes: Map<LinuxRuntimes, number>): IAzureQuickPickItem<ILinuxRuntimeStack>[] {
```
EricJizbaMSFT: The map seems over-complex. What if it was just a sorted array of recommendedRuntimes. For example, this function could be implemented like this:
```typescript
private sortQuickPicksByRuntime(runtimeItems: IAzureQuickPickItem<ILinuxRuntimeStack>[], recommendedRuntimes: string[]): IAzureQuickPickItem<ILinuxRuntimeStack>[] {
    function getPriority(item: IAzureQuickPickItem<ILinuxRuntimeStack>): number {
        const index: number = recommendedRuntimes.findIndex(runtime => item.data.name.includes(runtime));
        return index === -1 ? recommendedRuntimes.length : index;
    }
    return runtimeItems.sort((a, b) => getPriority(a) - getPriority(b));
}
```

appservice/src/createAppService/createWebApp.ts
EricJizbaMSFT: Sounds fine to me. I don't have a strong preference on default location.

appservice/src/createAppService/SiteRuntimeStep.ts
jdneo: Cool, using `array` is much better!

appservice/src/createAppService/createWebApp.ts
jdneo: Ok. I'll revert the code for switching location once the issue has been addressed.

appservice/src/createAppService/createWebApp.ts
```diff
@@ -23,28 +23,38 @@ export async function createWebApp(
...
         if (await fse.pathExists(path.join(fsPath, 'package.json'))) {
-            wizardContext.recommendedSiteRuntime = LinuxRuntimes.node;
+            wizardContext.recommendedSiteRuntime = [ LinuxRuntimes.node ];
         } else if (await fse.pathExists(path.join(fsPath, 'requirements.txt'))) {
             // requirements.txt are used to pip install so a good way to determine it's a Python app
-            wizardContext.recommendedSiteRuntime = LinuxRuntimes.python;
+            wizardContext.recommendedSiteRuntime = [ LinuxRuntimes.python ];
         }
-        actionContext.properties.recommendedSiteRuntime = wizardContext.recommendedSiteRuntime;
+        actionContext.properties.recommendedSiteRuntime = wizardContext.recommendedSiteRuntime ? [...wizardContext.recommendedSiteRuntime.keys()].join(',') : undefined;
```
EricJizbaMSFT: Simplify this line to work for an array instead of a map, but otherwise looks good

jdneo: Oh sorry, I missed this part. Updated...

PR #414, opened 2019-02-14
Do not zip .vscode when deploying

nturinski:
Fixes https://github.com/Microsoft/vscode-azureappservice/issues/861

I know that we don't like to black box logic like this, but it does seem like this would never be used for a deployment.


nturinski:
I was already mentally prepared for this. 😅

PR #415, opened 2019-02-14
Add support for .funcignore

EricJizbaMSFT:
.funcignore is already supported by the func cli and uses .gitignore patterns instead of glob (which are better IMO)

I'm going to remove support for the previous settings in functions, so this is a breaking change. Also I just cleaned up the deploy signature because I felt like it.

Related: https://github.com/Microsoft/vscode-azurefunctions/issues/418

### DIFF
--- a/appservice/src/FileUtilities.ts
+++ b/appservice/src/FileUtilities.ts
@@ -4,7 +4,8 @@
  *--------------------------------------------------------------------------------------------*/
 
 import * as archiver from 'archiver';
-import * as fs from 'fs';
+import * as fse from 'fs-extra';
+import { glob as globGitignore } from 'glob-gitignore';
 import * as os from 'os';
 import * as path from 'path';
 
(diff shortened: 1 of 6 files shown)


appservice/src/FileUtilities.ts
```diff
@@ -40,30 +27,67 @@ export async function zipFile(filePath: string): Promise<string> {
...
+
+async function addFilesByGlob(zipper: archiver.Archiver, folderPath: string, globPattern: string, ignorePattern: string | string[] | undefined): Promise<void> {
+    zipper.glob(globPattern, {
+        cwd: folderPath,
+        dot: true,
+        ignore: ignorePattern
+    });
+}
+
+async function addFilesByGitignore(zipper: archiver.Archiver, folderPath: string, gitignoreName: string): Promise<void> {
```
nturinski: Should we ignore files in the `.gitignore`? I think we discussed this before, but couldn't some of these files be needed for deployment?

Though I see you're using it for `.funcignore` but that seems different to me.

appservice/src/FileUtilities.ts
```diff
@@ -40,30 +27,67 @@ export async function zipFile(filePath: string): Promise<string> {
...
+export async function zipDirectoryGlob(folderPath: string, globPattern: string = '**/*', ignorePattern?: string | string[]): Promise<string> {
+    return await zipDirectoryInternal(folderPath, async (z) => {
+        await addFilesByGlob(z, folderPath, globPattern, ignorePattern);
+    });
+}
+
+/**
+ * Zips directory using gitignore filtering
+ */
+export async function zipDirectoryGitignore(folderPath: string, gitignoreName: string): Promise<string> {
```
nturinski: Just a question, but is it okay to not have an option to respect both the userSettings and the `.funcignore`?  Should it always be one or the other?

appservice/src/FileUtilities.ts
EricJizbaMSFT: It's not actually using the gitignore - it's just using the _filtering rules_ of a gitignore. gitignore has it's own patterns that are unique and not exactly the same as glob: https://git-scm.com/docs/gitignore#_pattern_format

appservice/src/FileUtilities.ts
EricJizbaMSFT: It's definitely an option to support both, but I decided it wasn't worth it (willing to discuss if you disagree). For the following reasons:
1. There would be confusion between which takes priority
1. I'd prefer to have one consistent way to do this and the func cli supports ".funcignore"
1. .gitignore patterns are much easier than glob patterns IMO
1. In functions we never prompt the user for these settings (like you do for app service) so I'm guessing they have low usage anyways (users would've had to discover them themselves)

EricJizbaMSFT: Also both you and Matt seemed to imply you preferred just one consistent option in your comments here: https://github.com/Microsoft/vscode-azurefunctions/issues/418

appservice/src/FileUtilities.ts
nturinski: No, I know, I was questioning the naming convention of your function 😋

EricJizbaMSFT: You're welcome to suggest a name that makes more sense to you :)

PR #419, opened 2019-02-20
Update "Report an Issue" template

EricJizbaMSFT:
We got feedback that characters like `&lt;` made it _harder_ to read the comments they surround (Fixes https://github.com/Microsoft/vscode-azureappservice/issues/782) and @StephenWeatherford has mentioned he wants the call stack several times. Here's my reasoning for other changes:
1. Added 'IMPORTANT' callout as suggested in above issue
1. Made the comments "hidden" once the issue is posted with `<!--`. This matches the VS Code issue template and most other templates I could find. I think the comments just make the issue ugly
1. Add "1." and "2." to maybe (?) make it more obvious that there's a bunch of whitespace that needs to be addressed. Also matches VS Code issue template.

Example:
screen shot 2019-02-20 at 9 25 29 am

screen shot 2019-02-20 at 9 25 20 am

NOTE: I decided not to add call stack to unit tests. The stack is dependent on the machine and the compiled code, so it just didn't seem worth it.

Fixes https://github.com/Microsoft/vscode-azuretools/issues/152

### DIFF
--- a/ui/src/parseError.ts
+++ b/ui/src/parseError.ts
@@ -12,11 +12,14 @@ import { localize } from './localize';
 export function parseError(error: any): IParsedError {
     let errorType: string = '';
     let message: string = '';
+    let stack: string | undefined;
 
     if (typeof (error) === 'object' && error !== null) {
         if (error.constructor !== Object) {
             errorType = error.constructor.name;
         }
+
+        stack = error.stack;
 
         // See https://github.com/Microsoft/vscode-azureappservice/issues/419 for an example error that requires these 'unpack's
         error = unpackErrorFromField(error, 'value');
@@ -66,6 +69,7 @@ export function parseError(error: any): IParsedError {
     return {
         errorType: errorType,
         message: message,
+        stack: stack,
         // NOTE: Intentionally not using 'error instanceof UserCancelledError' because that doesn't work if multiple versions of the UI package are used in one extension
         // See https://github.com/Microsoft/vscode-azuretools/issues/51 for more info
         isUserCancelledError: errorType === 'UserCancelledError'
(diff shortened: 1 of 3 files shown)


StephenWeatherford:
Great, I just added callstacks, too. :-(

ui/src/parseError.ts
```diff
@@ -12,12 +12,15 @@ import { localize } from './localize';
...
     let errorType: string = '';
     let message: string = '';
+    let stack: string | undefined;
 
     if (typeof (error) === 'object' && error !== null) {
         if (error.constructor !== Object) {
             errorType = error.constructor.name;
         }
 
+        stack = error.stack;
```
StephenWeatherford: My PR removes the first part of the path that may contain username. How about I just add that functionality after you merge?

ui/src/reportAnIssue.ts
```diff
@@ -14,19 +14,35 @@ import { getPackageInfo } from "./getPackageInfo";
 export function reportAnIssue(actionId: string, parsedError: IParsedError): void {
     const { extensionName, extensionVersion } = getPackageInfo();
 
-    const body: string = `
-&lt;Please be sure to remove any private information before submitting.&gt;
+    let body: string = `
```
StephenWeatherford: In mine, I added the request to fill out repro steps right to the top to hopefully encourage them more and give them less excuse. :-)  I'll let you decide.

StephenWeatherford: Added #423

ui/src/parseError.ts
EricJizbaMSFT: Sounds good

ui/src/reportAnIssue.ts
EricJizbaMSFT: Not entirely sure what you mean and I can't find your old commit at this point so I'll merge as-is and you can edit if you want

PR #420, opened 2019-02-20
Use correct issues URL for ARM

StephenWeatherford:
Code incorrect assumes that the issues URL is related to the official extension ID - not so with ARM, use package.json instead when available.

### DIFF
--- a/ui/src/getPackageInfo.ts
+++ b/ui/src/getPackageInfo.ts
@@ -39,7 +39,9 @@ export function getPackageInfo(ctx?: ExtensionContext): { extensionName: string,
     const extensionVersion: string | undefined = packageJson.version;
     const aiKey: string | undefined = packageJson.aiKey;
     const publisher: string | undefined = packageJson.publisher;
-    const bugsUrl: string | undefined = packageJson.bugs && packageJson.bugs.url;
+    const bugsUrl: string | undefined = !packageJson.bugs ? undefined :
+        typeof packageJson.bugs === 'string' ? packageJson.bugs :
+        packageJson.bugs.url;
 
     if (!aiKey) {
         throw new Error('Extension\'s package.json is missing aiKey');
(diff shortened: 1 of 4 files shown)


EricJizbaMSFT:
Uh literally submitted this PR 5 hours ago: https://github.com/Microsoft/vscode-azuretools/pull/419

ui/src/reportAnIssue.ts
```diff
@@ -12,10 +12,10 @@ import { getPackageInfo } from "./getPackageInfo";
  * Used to open the browser to the "New Issue" page on GitHub with relevant context pre-filled in the issue body
  */
 export function reportAnIssue(actionId: string, parsedError: IParsedError): void {
-    const { extensionName, extensionVersion } = getPackageInfo();
+    const { extensionName, extensionVersion, bugsUrl } = getPackageInfo();
 
     const body: string = `
-&lt;Please be sure to remove any private information before submitting.&gt;
+&lt;Please fill out the repro steps to help us better understand your issue, and be sure to remove any private information before submitting.&gt;
```
PrashanthCorp: I'm assuming  `+&lt` should be `<!--- `. 
Either ways, I don't know which PR ought to be reviewed. :P

StephenWeatherford: I'll remove this part from mine.

StephenWeatherford:
Now this only deals with the issues URL for ARM.

ui/src/getPackageInfo.ts
```diff
@@ -55,12 +56,15 @@ export function getPackageInfo(ctx?: ExtensionContext): { extensionName: string,
...
-    return { extensionName, extensionVersion, aiKey, extensionId };
+    return { extensionName, extensionVersion, aiKey, extensionId, bugsUrl };
 }
 
 interface IPackageJson {
     version?: string;
     name?: string;
     publisher?: string;
     aiKey?: string;
+    bugs?: {
```
EricJizbaMSFT: [Docs](https://docs.npmjs.com/files/package.json#bugs) say the following about bugs:
> If you want to provide only a url, you can specify the value for “bugs” as a simple string instead of an object.

I don't _think_ any of our repos specify "bugs" as a string, but probably worth being safe

PR #427, opened 2019-02-22
Make storage connections more sovereign friendly for functions

EricJizbaMSFT:
Similar to https://github.com/Microsoft/vscode-azurefunctions/pull/1010, I made this more friendly to sovereign, but didn't actually test on a sovereign so don't make any claims that it'll work. I obviously still tested on global Azure so theoretically it will.

Did the following:
1. Pass connection string as-is instead of trying to parse out just key and name
1. Add EndpointSuffix to connection strings by default. This follows same pattern as the portal

Last piece to fix https://github.com/Microsoft/vscode-azurefunctions/issues/962

### DIFF
--- a/appservice/src/createAppService/SiteCreateStep.ts
+++ b/appservice/src/createAppService/SiteCreateStep.ts
@@ -76,9 +76,15 @@ export class SiteCreateStep extends AzureWizardExecuteStep<IAppServiceWizardCont
                 fileShareName = `${fileShareName.slice(0, maxFileShareNameLength - randomLetters - 1)}-${randomUtils.getRandomHexString(randomLetters)}`;
             }
 
+            let endpointSuffix: string = wizardContext.environment.storageEndpointSuffix;
+            // https://github.com/Azure/azure-sdk-for-node/issues/4706
+            if (endpointSuffix.startsWith('.')) {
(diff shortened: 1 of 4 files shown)


appservice/src/createAppService/SiteCreateStep.ts
```diff
@@ -76,9 +76,15 @@ export class SiteCreateStep extends AzureWizardExecuteStep<IAppServiceWizardCont
                 fileShareName = `${fileShareName.slice(0, maxFileShareNameLength - randomLetters - 1)}-${randomUtils.getRandomHexString(randomLetters)}`;
             }
 
+            let endpointSuffix: string = wizardContext.environment.storageEndpointSuffix;
+            // https://github.com/Azure/azure-sdk-for-node/issues/4706
```
StephenWeatherford: nit: you could do something like:
endpointSuffix = endpointSuffix.replace(/^\./, '');

EricJizbaMSFT: Good thing I tested this because markdown apparently removed a slash 😬😬😬
Preview tab:
screen shot 2019-02-22 at 11 55 10 am

Write tab:
screen shot 2019-02-22 at 11 55 15 am

PR #433, opened 2019-03-05
Add node lts to the runtime quickpick

nturinski:
image

So there's a new runtime in the portal that is just called LTS.  I'm not sure what version this is running, but I'd assume that the service team will be keeping it up to date as possible?  Either way, this makes us not have to update the LTS manually anymore.

Also, P.S. but why is Ruby above Node in the Portal's dropdown o.O?

### DIFF
--- a/appservice/src/createAppService/SiteRuntimeStep.ts
+++ b/appservice/src/createAppService/SiteRuntimeStep.ts
@@ -53,8 +53,8 @@ export class SiteRuntimeStep extends AzureWizardPromptStep<IAppServiceWizardCont
     private getLinuxRuntimeStack(): ILinuxRuntimeStack[] {
         return [
             {
-                name: 'node|10.10',
-                displayName: 'Node.js 10.10 (LTS - Recommended for new apps)'
+                name: 'node|lts',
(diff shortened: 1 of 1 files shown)


appservice/src/createAppService/SiteRuntimeStep.ts
```diff
@@ -53,8 +53,8 @@ export class SiteRuntimeStep extends AzureWizardPromptStep<IAppServiceWizardCont
     private getLinuxRuntimeStack(): ILinuxRuntimeStack[] {
         return [
             {
-                name: 'node|10.10',
-                displayName: 'Node.js 10.10 (LTS - Recommended for new apps)'
+                name: 'node|lts',
+                displayName: 'Node.js LTS (Recommended for new apps)'
```
EricJizbaMSFT: I'm not sure I'm ready to make this "Recommended for new apps" until we know exactly how it works. For example, what happens when Node updates LTS to a new major version? Do people's apps get updated automatically? If so, that doesn't sound great

@fiveisprime thoughts?

EricJizbaMSFT:
Closing this due to lack of activity

PR #436, opened 2019-03-07
WaitForDeploymentToComplete accepts cancel token

nturinski:
Fixes https://github.com/Microsoft/vscode-azureappservice/issues/883

Changing from Kudu to KuduLite broke LocalGit deployment for Python apps and Node.js v10 apps.  The remote url should be accepted now.

The error in the issue brought to my attention that we were parsing a generic error message so I've changed it to be specific to diverging history.  There are also credentials in that error message, so I've masked those when we parse.

The push and wait for deploy have been wrapped in a Promise because both need to finish before we can proceed.  Previously, since we did not wait on the push, the error was not being handled properly.  Because we need to deploy again in this scenario, we needed the ability to cancel the `waitForDeploymentToComplete` command

I increased the timeout to 60 seconds because on a fresh deployment, 30 seconds was often not enough and we would report the deployment as a failure even if it wasn't.  If we pick up a temp id, then we know it has at least started so I set the deployment as the temp so it will not check for time-out anymore.

### DIFF
--- a/appservice/src/deploy/localGitDeploy.ts
+++ b/appservice/src/deploy/localGitDeploy.ts
@@ -39,6 +39,8 @@ export async function localGitDeploy(client: SiteClient, fsPath: string): Promis
             ext.outputChannel.appendLine(formatDeployLog(client, (localize('localGitDeploy', `Deploying Local Git repository to "${client.fullName}"...`))));
             localGit.push(remote, 'HEAD:master').catch(async (error) => {
                 tokenSource.cancel();
+                // due to timing issues, output here when we know we need to cancel
+                ext.outputChannel.appendLine(formatDeployLog(client, localize('cancelledDeployment', 'Cancelled deployment.')));
                 // tslint:disable-next-line:no-unsafe-any
                 const parsedError: IParsedError = parseAndRemoveCredentialsFromError(error);
 
--- a/appservice/src/deploy/localGitDeploy.ts
+++ b/appservice/src/deploy/localGitDeploy.ts
@@ -7,7 +7,7 @@ import { User } from 'azure-arm-website/lib/models';
 import * as opn from 'opn';
 import * as git from 'simple-git/promise';
 import * as vscode from 'vscode';
-import { DialogResponses } from 'vscode-azureextensionui';
+import { DialogResponses, IParsedError, parseError } from 'vscode-azureextensionui';
 import KuduClient from 'vscode-azurekudu';
 import { ext } from '../extensionVariables';
 import { getKuduClient } from '../getKuduClient';
@@ -21,9 +21,10 @@ import { waitForDeploymentToComplete } from './waitForDeploymentToComplete';
 export async function localGitDeploy(client: SiteClient, fsPath: string): Promise<void> {
     const kuduClient: KuduClient = await getKuduClient(client);
     const publishCredentials: User = await client.getWebAppPublishCredential();
-    const remote: string = nonNullProp(publishCredentials, 'scmUri');
+    const remote: string = `https://${nonNullProp(publishCredentials, 'publishingUserName')}:${nonNullProp(publishCredentials, 'publishingPassword')}@${client.gitUrl}`;
     const localGit: git.SimpleGit = git(fsPath);
     const commitId: string = (await localGit.log()).latest.hash;
+    const tokenSource: vscode.CancellationTokenSource = new vscode.CancellationTokenSource();
     try {
         const status: git.StatusResult = await localGit.status();
         if (status.files.length > 0) {
@@ -32,11 +33,36 @@ export async function localGitDeploy(client: SiteClient, fsPath: string): Promis
             await ext.ui.showWarningMessage(message, { modal: true }, deployAnyway, DialogResponses.cancel);
         }
         await verifyNoRunFromPackageSetting(client);
-        // tslint:disable-next-line:no-floating-promises
-        localGit.push(remote, 'HEAD:master');
+
+        // wrap two async actions in a promise because both need to run and finish simaltaneously
+        await new Promise((resolve: () => void, reject: (err: IParsedError) => void): void => {
+            ext.outputChannel.appendLine(formatDeployLog(client, (localize('localGitDeploy', `Deploying Local Git repository to "${client.fullName}"...`))));
+            localGit.push(remote, 'HEAD:master').catch(async (error) => {
+                tokenSource.cancel();
+                // tslint:disable-next-line:no-unsafe-any
+                const parsedError: IParsedError = parseAndRemoveCredentialsFromError(error);
+
+                // if the push fails, check if it's due to diverging history to offer force push
+                if (parsedError.message.indexOf('Updates were rejected because the remote contains work that you do') >= 0) {
+                    await forcePush();
+                    resolve();
+                } else {
+                    reject(parsedError);
+                }
+            });
+
+            waitForDeploymentToComplete(client, kuduClient, commitId, tokenSource.token).then(resolve).catch((error: Error) => {
+                const parsedError: IParsedError = parseAndRemoveCredentialsFromError(error);
+                // waitForDeploymentToComplete throws a UserCancelledError when cancelled so that we don't mistake it as "completing"
+                if (!parsedError.isUserCancelledError) {
+                    reject(parsedError);
+                }
+            });
+        });
     } catch (err) {
         // tslint:disable-next-line:no-unsafe-any
-        if (err.message.indexOf('spawn git ENOENT') >= 0) {
+        const parsedError: IParsedError = parseAndRemoveCredentialsFromError(err);
+        if (parsedError.message.indexOf('spawn git ENOENT') >= 0) {
             const installString: string = localize('Install', 'Install');
             const input: string | undefined = await vscode.window.showErrorMessage(localize('GitRequired', 'Git must be installed to use Local Git Deploy.'), installString);
             if (input === installString) {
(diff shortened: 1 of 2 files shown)


appservice/src/deploy/localGitDeploy.ts
```diff
@@ -32,30 +33,63 @@ export async function localGitDeploy(client: SiteClient, fsPath: string): Promis
...
+        // wrap two async actions in a promise because both need to run and finish simaltaneously
+        await new Promise((resolve: () => void, reject: (err: IParsedError) => void): void => {
+            ext.outputChannel.appendLine(formatDeployLog(client, (localize('localGitDeploy', `Deploying Local Git repository to "${client.fullName}"...`))));
+            localGit.push(remote, 'HEAD:master').catch(async (error) => {
+                tokenSource.cancel();
+                // tslint:disable-next-line:no-unsafe-any
+                const parsedError: IParsedError = parseAndRemoveCredentialsFromError(error);
+
+                // if the push fails, check if it's due to diverging history to offer force push
+                if (parsedError.message.indexOf('Updates were rejected because the remote contains work that you do') >= 0) {
```
EricJizbaMSFT: I don't trust this error message to stay the same - or even to stay in english. I think we should be checking a code or something else. If there's no code, maybe we should be checking for diverging history before a git push

appservice/src/deploy/localGitDeploy.ts
```diff
@@ -32,30 +33,63 @@ export async function localGitDeploy(client: SiteClient, fsPath: string): Promis
...
+                    await forcePush();
+                    resolve();
+                } else {
+                    reject(parsedError);
+                }
+            });
+
+            waitForDeploymentToComplete(client, kuduClient, commitId, tokenSource.token).then(resolve).catch((error: Error) => {
+                const parsedError: IParsedError = parseAndRemoveCredentialsFromError(error);
```
EricJizbaMSFT: > There are also credentials in that error message, so I've masked those when we parse.

We've been over this before, but I think this needs further discussion. We need to make extra sure there's no credentials in error messages/telemetry and I'm not convinced this handles all cases.

As an aside - why didn't we have this problem before your changes?

appservice/src/deploy/localGitDeploy.ts
```diff
@@ -21,9 +21,10 @@ import { waitForDeploymentToComplete } from './waitForDeploymentToComplete';
 export async function localGitDeploy(client: SiteClient, fsPath: string): Promise<void> {
     const kuduClient: KuduClient = await getKuduClient(client);
     const publishCredentials: User = await client.getWebAppPublishCredential();
-    const remote: string = nonNullProp(publishCredentials, 'scmUri');
+    const remote: string = `https://${nonNullProp(publishCredentials, 'publishingUserName')}:${nonNullProp(publishCredentials, 'publishingPassword')}@${client.gitUrl}`;
```
EricJizbaMSFT: Have we been in contact with the App Service team to figure out why this changed in kudu lite and if it was intentional? Maybe we can get them to fix on their side?

appservice/src/deploy/localGitDeploy.ts
```diff
@@ -32,30 +33,65 @@ export async function localGitDeploy(client: SiteClient, fsPath: string): Promis
...
+            localGit.push(remote, 'HEAD:master').catch(async (error) => {
+                tokenSource.cancel();
+                // due to timing issues, output here when we know we need to cancel
+                ext.outputChannel.appendLine(formatDeployLog(client, localize('cancelledDeployment', 'Cancelled deployment.')));
+                // tslint:disable-next-line:no-unsafe-any
+                const parsedError: IParsedError = parseAndRemoveCredentialsFromError(error);
+
+                // if the push fails, check if it's due to diverging history to offer force push
+                if (parsedError.message.indexOf('Updates were rejected because the remote contains work that you do') >= 0) {
+                    await forcePush();
```
EricJizbaMSFT: I think errors from forcePush are being ignored right now

appservice/src/deploy/waitForDeploymentToComplete.ts
```diff
@@ -13,21 +14,29 @@ import { delay } from '../utils/delay';
...
+        if (token && token.isCancellationRequested) {
+            throw new UserCancelledError();
+        }
+
         [deployment, permanentId, initialStartTime] = await tryGetLatestDeployment(kuduClient, permanentId, initialStartTime, expectedId);
         if ((deployment === undefined || !deployment.id)) {
             if (expectedId && Date.now() < maxTimeToWaitForExpectedId) {
-                ext.outputChannel.appendLine(formatDeployLog(client, localize('waitingForBuild', 'Starting deployment...')));
+                if (!startDeploymentOutput) {
+                    ext.outputChannel.appendLine(formatDeployLog(client, localize('waitingForBuild', 'Starting deployment...')));
```
EricJizbaMSFT: Can we just move this message outside of `waitForDeploymentToComplete`? It should be displayed before you even do the first git push

appservice/src/deploy/waitForDeploymentToComplete.ts
```diff
@@ -87,50 +96,51 @@ export async function waitForDeploymentToComplete(client: SiteClient, kuduClient
...
-                    throw parsedError;
-                }
+        // Use "permanentId" to find the deployment during its "permanent" phase
+        deployment = await kuduClient.deployment.getResult(permanentId);
+    } else if (expectedId) {
+        // if we have a "expectedId" we know which deployment we are looking for, so wait until latest id reflects that
+        try {
+            const latestDeployment: DeployResult = await kuduClient.deployment.getResult('latest');
+            // if the latest deployment is temp, then a deployment has triggered so we should watch it even if it doesn't match the expectedId
+            deployment = latestDeployment.isTemp ? latestDeployment : undefined;
```
EricJizbaMSFT: Isn't `deployment` ovewritten the next line?

EricJizbaMSFT:
Also I know this touches a lot of the same code, but if you could split this up into smaller PRs that would be great. Seems like you fixed 4-5 bugs here and we might be able to fit 2-3 in the release without waiting for all of them

appservice/src/deploy/localGitDeploy.ts
nturinski:
```
"To https://naturins-22-error-03.scm.azurewebsites.net:443/naturins-22-error-03.git
 ! [rejected]        HEAD -> master (fetch first)
error: failed to push some refs to 'https://$naturins-22-error-03:***@naturins-22-error-03.scm.azurewebsites.net:443/naturins-22-error-03.git'
hint: Updates were rejected because the remote contains work that you do
hint: not have locally. This is usually caused by another repository pushing
hint: to the same ref. You may want to first integrate the remote changes
hint: (e.g., 'git pull ...') before pushing again.
hint: See the 'Note about fast-forwards' in 'git push --help' for details.
```

This is the error that comes back.  I looked into trying to figure out diverging history prior to a push, but it involves fetching the entire remote and then comparing it locally.  It wasn't that complicated, but it took a bit longer to do since we'd do it with every push.

I honestly don't know how often people use Force Push (should have gathered telemetry, stupid past Nathan 😞) in the first place or if we should even encourage them to.  Seems kind of like overriding their old deployment might not be something we would want to allow since it's kind of a "one button deploy" scenario.

appservice/src/deploy/localGitDeploy.ts
nturinski: We did for any error message that wasn't a result from "failed to push" errors.  But I think most error messages weren't actually being output because we were handling it customly.

appservice/src/deploy/localGitDeploy.ts
nturinski: Check yo email soooonnnn

appservice/src/deploy/waitForDeploymentToComplete.ts
nturinski: Only if there is a match.  The reason for this is if the latest is a temp, even if it doesn't match our current ID, we should track it because that means something has triggered.  Once it does match, we know that we can save it as our permanentId.

EricJizbaMSFT: > Only if there is a match

But if there's not a match, doesn't the next line set it to undefined?

I feel like you want to do something like this instead:
```
if (latestDeployment.isTemp) {
    deployment = latestDeployment;
} else if (latestDeployment.id === expectedId) {
    deployment = latestDeployment;
    permanentId = latestDeployment.id;
}
```

appservice/src/deploy/localGitDeploy.ts
EricJizbaMSFT: Okay well I would prefer if we had a more broad try/catch encompassing most of `localGitDeploy` and then you only need to call `parseAndRemoveCredentialsFromError` once

Also I would like to see some unit tests for `parseAndRemoveCredentialsFromError`. For example, I think `replace` only replaces the first occurrence and not all occurrences?

appservice/src/deploy/localGitDeploy.ts
EricJizbaMSFT: I think I'm fine with this - looks like `client.gitUrl` will handle most of the weird stuff like custom domains, etc. that I mentioned offline.

I am curious if there is a way to connect with a temporary key instead of always using the main credentials. But if the old scmUri used these credentials, I'm fine keeping the behavior as-is for this release.

appservice/src/deploy/localGitDeploy.ts
nturinski: Does it not throw the error, and then get parsed in the catch? Or do you mean errors from the force push itself? (since I don't await that)

EricJizbaMSFT: You're inside of `await new Promise(...` so the only errors that are tracked from any of this code are the ones passed in to `reject`

EricJizbaMSFT:
Closing this for my own sanity and because it's covered by several other PRs

PR #438, opened 2019-03-07
Move the starting deployment output from waitForDeploymentToComplete to localGitDeploy

nturinski:


nturinski:
Going to remove the line in question for a better deployment starting output in separate PR.

PR #439, opened 2019-03-07
parse errors and mask creds

nturinski:


appservice/src/deploy/localGitDeploy.ts
```diff
@@ -35,6 +35,8 @@ export async function localGitDeploy(client: SiteClient, fsPath: string): Promis
         // tslint:disable-next-line:no-floating-promises
         localGit.push(remote, 'HEAD:master');
     } catch (err) {
+        const parsedError: IParsedError = parseError(err);
+        parsedError.message.replace(new RegExp(nonNullProp(publishCredentials, 'publishingPassword'), 'g'), '***');
```
StephenWeatherford: I think string.replace is non-destructive...?

nturinski: You are absolutely correct.  I need to re-assign 😃

EricJizbaMSFT: UNIT TESTS

appservice/src/deploy/localGitDeploy.ts
```diff
@@ -35,6 +36,8 @@ export async function localGitDeploy(client: SiteClient, fsPath: string): Promis
         // tslint:disable-next-line:no-floating-promises
         localGit.push(remote, 'HEAD:master');
     } catch (err) {
+        const parsedError: IParsedError = parseError(err);
+        parsedError.message = maskValue(parsedError.message, nonNullProp(publishCredentials, 'publishingPassword'));
```
EricJizbaMSFT: You're still not masking all errors if you do it here since the force push code could throw an error. I suggest something like this to ensure that all of localGitDeploy is covered:
```typescript
export async function localGitDeploy(client: SiteClient, fsPath: string): Promise<void> {
    try {
        await localGitDeployCore(client, fsPath, publishCredentials);
    } catch (error) {
        // mask publishCredentials.password
    }
}

export async function localGitDeployCore(client: SiteClient, fsPath: string, publishCredentials: User): Promise<void> {
...
```

NOTE: This would look obviously look different if you took my other suggestion below that has `callWithMaskHandling`

appservice/src/deploy/localGitDeploy.ts
```diff
@@ -45,14 +48,14 @@ export async function localGitDeploy(client: SiteClient, fsPath: string): Promis
...
-        } else if (err.message.indexOf('error: failed to push') >= 0) {
+        } else if (parsedError.message.indexOf('error: failed to push') >= 0) {
             const forcePush: vscode.MessageItem = { title: localize('forcePush', 'Force Push') };
             const pushReject: string = localize('localGitPush', 'Push rejected due to Git history diverging.');
             await ext.ui.showWarningMessage(pushReject, forcePush, DialogResponses.cancel);
             // tslint:disable-next-line:no-floating-promises
             localGit.push(remote, 'HEAD:master', { '-f': true });
         } else {
-            throw err;
+            throw parsedError;
```
EricJizbaMSFT: Seems weird to throw parsedError. I would do `throw new Error(parsedError.message);`

appservice/test/maskValue.test.ts
```diff
@@ -0,0 +1,46 @@
...
+            let errorMessage: string = `To https://naturins-22-error-03.scm.azurewebsites.net:443/naturins-22-error-03.git
+            ! [rejected]        HEAD -> master (fetch first)
+           error: failed to push some refs to 'https://$naturins-22-error-03:scHQERrAlXSmlCeN1mrhDzsHWeDz2XZt5R343HgCNmxS0xlswcaA2Cowflda@naturins-22-error-03.scm.azurewebsites.net:443/naturins-22-error-03.git'
+           hint: Updates were rejected because the remote contains work that you do
+           hint: not have locally. This is usually caused by another repository pushing
+           hint: to the same ref. You may want to first integrate the remote changes
+           hint: (e.g., 'git pull ...') before pushing again.
+           hint: See the 'Note about fast-forwards' in 'git push --help' for details.`;
+
+            errorMessage = maskValue(errorMessage, credentials);
```
EricJizbaMSFT: These unit tests are a good start, but they don't quite cover the scenario Stephen brought up (where you forgot to reassign the result after masking the value). Instead of just `maskValue`, I would create a function like this so that the tests are more valuable
```
export async function callWithMaskHandling<T>(valueToMask: string, callback: () => Promise<T>): Promise<T> {
    try {
        return await callback();
    } catch (error) {
        // mask logic
    }
}
```

appservice/src/deploy/localGitDeploy.ts
```diff
@@ -34,7 +36,7 @@ export async function localGitDeploy(client: SiteClient, fsPath: string): Promis
         }
         await verifyNoRunFromPackageSetting(client);
         ext.outputChannel.appendLine(formatDeployLog(client, (localize('localGitDeploy', `Deploying Local Git repository to "${client.fullName}"...`))));
-        await tryPushAndWaitForDeploymentToComplete();
+        await callWithMaskHandling(async (): Promise<void> => { await tryPushAndWaitForDeploymentToComplete(); }, publishingPassword);
```
EricJizbaMSFT: I still want this around all of localGitDeploy just to be safe. I don't like picking and choosing the lines that we mask

appservice/test/callWithMaskHandling.test.ts
```diff
@@ -0,0 +1,91 @@
...
+           hint: to the same ref. You may want to first integrate the remote changes
+           hint: (e.g., 'git pull ...') before pushing again.
+           hint: See the 'Note about fast-forwards' in 'git push --help' for details.`;
+
+            await assertThrowsAsync(async (): Promise<void> => {
+                await callWithMaskHandling(async () => {
+                    throw new Error(errorMessage);
+                }, credentials);
+            }, (err) => {
+                if (err.message.indexOf(credentials) === -1) {
```
EricJizbaMSFT: I would do something like this:
```
(err) => {
                return JSON.stringify(parseError(err)).includes(credentials);
            }
```

Better to validate the whole parsedError that we actually end up using. Plus that logic would handle non-standard errors (aka if you had done `throw errorMessage` instead of `throw new Error(errorMessage);` - actually you might want to add a test case for non-standard errors

EricJizbaMSFT: Also might be worth creating a `validateError` function rather than duplicating this logic for all tests

PR #441, opened 2019-03-07
Add cancel token, wrap push and wait in a promise

nturinski:


appservice/src/deploy/localGitDeploy.ts
```diff
@@ -45,17 +65,29 @@ export async function localGitDeploy(client: SiteClient, fsPath: string): Promis
...
+
+            localGit.push(remote, 'HEAD:master', pushOptions).catch(async (error) => {
+                tokenSource.cancel();
+                // tslint:disable-next-line:no-unsafe-any
+                reject(error);
+            });
+
+            waitForDeploymentToComplete(client, kuduClient, commitId, token).then(resolve).catch((error: Error) => {
+                // waitForDeploymentToComplete throws a UserCancelledError when cancelled so that we don't mistake it as "completing"
+                if (!parseError(error).isUserCancelledError) {
```
EricJizbaMSFT: I think it's a little confusing for `waitForDeploymentToComplete` to throw a _User_CancelledError. What if it just returns nothing and you do something like this:
```
waitForDeploymentToComplete(client, kuduClient, commitId, token).then(() => {
    if (!tokenSource.canceled) {
        resolve();
    }
}).catch(reject);
```

appservice/src/deploy/localGitDeploy.ts
```diff
@@ -45,17 +65,29 @@ export async function localGitDeploy(client: SiteClient, fsPath: string): Promis
             }
             return undefined;
             // tslint:disable-next-line:no-unsafe-any
-        } else if (err.message.indexOf('error: failed to push') >= 0) {
```
EricJizbaMSFT: Discussed offline. I don't think you need to move this

appservice/src/deploy/localGitDeploy.ts
EricJizbaMSFT: Actually if you take my above suggestion to return nothing and if you switch these lines:
```
// tslint:disable-next-line:no-unsafe-any
reject(error);
tokenSource.cancel();                
```

You can just do this:
```
waitForDeploymentToComplete(client, kuduClient, commitId, token).then(resolve).catch(reject);
```

Since `reject` is called before `cancel`, you don't have to worry about the result of `waitForDeploymentToComplete` since the promise was guaranteed to already have finished.

appservice/src/deploy/localGitDeploy.ts
```diff
@@ -45,17 +65,29 @@ export async function localGitDeploy(client: SiteClient, fsPath: string): Promis
...
-            localGit.push(remote, 'HEAD:master', { '-f': true });
         } else {
             throw err;
         }
     }
 
-    ext.outputChannel.appendLine(formatDeployLog(client, (localize('localGitDeploy', `Deploying Local Git repository to "${client.fullName}"...`))));
-    await waitForDeploymentToComplete(client, kuduClient, commitId);
+    async function pushAndWaitForDeploymentToComplete(token?: vscode.CancellationToken, forcePush: boolean = false): Promise<void> {
+        await new Promise((resolve: () => void, reject: (error: Error) => void): void => {
```
EricJizbaMSFT: I would create a new cancellation token source inside `pushAndWaitForDeploymentToComplete` every time. I don't think it's optional because either way `localGit.push` could throw an error

appservice/src/deploy/localGitDeploy.ts
nturinski: Doesn't this mean that tokenSource will never be cancelled? Since it would reject with the error before it can cancel the token.

EricJizbaMSFT: You're thinking of `throw error;`, which cancels the current code path. However, `reject(error);` is different and should continue

appservice/src/deploy/localGitDeploy.ts
```diff
@@ -46,16 +48,40 @@ export async function localGitDeploy(client: SiteClient, fsPath: string): Promis
...
+            });
+
+            waitForDeploymentToComplete(client, kuduClient, commitId, token).then(() => {
+                if (!token.isCancellationRequested) {
+                    resolve();
+                }
+            }).catch(reject);
+        });
+
+        tokenSource.dispose();
```
EricJizbaMSFT: Almost any time you do `dispose` it should be in a try/finally block so that it happens even on error

appservice/src/deploy/localGitDeploy.ts
```diff
@@ -46,16 +48,43 @@ export async function localGitDeploy(client: SiteClient, fsPath: string): Promis
...
+                    reject(error);
+                });
+
+                waitForDeploymentToComplete(client, kuduClient, commitId, token).then(() => {
+                    if (!token.isCancellationRequested) {
+                        resolve();
+                    }
+                }).catch(reject);
+            });
+        } catch (error) {
```
EricJizbaMSFT: No need for try/catch/finally - just try/finally will do the same thing.

appservice/src/deploy/localGitDeploy.ts
EricJizbaMSFT: ping

nturinski: Ahhh, okay, that makes a lot more sense. 👍

PR #445, opened 2019-03-13
Update to new wizard with back button

EricJizbaMSFT:


appservice/src/createAppService/AppServicePlanListStep.ts
```diff
@@ -35,24 +35,23 @@ export class AppServicePlanListStep extends AzureWizardPromptStep<IAppServiceWiz
         );
     }
 
-    public async prompt(wizardContext: IAppServiceWizardContext): Promise<IAppServiceWizardContext> {
-        if (!wizardContext.plan && !wizardContext.newPlanName) {
```
StephenWeatherford: consider reusing shouldPrompt

StephenWeatherford: ditto on others

appservice/src/createAppService/SiteRuntimeStep.ts
```diff
@@ -14,39 +14,39 @@ interface ILinuxRuntimeStack {
...
 
 export class SiteRuntimeStep extends AzureWizardPromptStep<IAppServiceWizardContext> {
-    public async prompt(wizardContext: IAppServiceWizardContext): Promise<IAppServiceWizardContext> {
-        if (!wizardContext.newSiteRuntime) {
-            if (wizardContext.newSiteKind === AppKind.functionapp) {
-                const runtimeItems: IAzureQuickPickItem<string>[] = [
-                    { label: 'JavaScript', data: 'node' },
-                    { label: '.NET', data: 'dotnet' }
-                ];
+    public async prompt(wizardContext: IAppServiceWizardContext): Promise<void> {
```
StephenWeatherford: So why does this one not have the canprompt check?  I was wondering if the extra check was really necessary.  Or maybe an assertion instead (preferably reusing shouldPrompt)

appservice/src/createAppService/createAppService.ts
```diff
@@ -92,9 +89,15 @@ export async function createAppService(
...
     }
     executeSteps.push(new SiteCreateStep(createOptions.createFunctionAppSettings));
-    const wizard: AzureWizard<IAppServiceWizardContext> = new AzureWizard(promptSteps, executeSteps, wizardContext);
 
-    wizardContext = await wizard.prompt(actionContext);
+    if (wizardContext.newSiteOS !== undefined) {
+        SiteOSStep.setLocationsTask(wizardContext);
+    }
+
+    const title: string = localize('creatingTitle', 'Create new {0}', getAppKindDisplayName(appKind));
```
StephenWeatherford: This can cause trouble with localization.  E.g. in German the spelling of "new" here will depend on the specific app kind's gender.  It would be better to pass in the entire creatingTitle as a phrase rather than a piece of a sentence.

appservice/src/createAppService/AppServicePlanListStep.ts
EricJizbaMSFT: Reusing how? I _deleted_ `if (!wizardContext.plan && !wizardContext.newPlanName) {`

StephenWeatherford: Huh...  Wonder what my mental state was this morning.

PR #446, opened 2019-03-13
Fix back button through filtered step

EricJizbaMSFT:
[SiteRuntimeStep](https://github.com/Microsoft/vscode-azuretools/blob/master/appservice/src/createAppService/SiteRuntimeStep.ts) is a unique step because it filters itself based on the previous step (don't need to pick runtime if you're creating a windows app). Most other steps filter themselves on defaults set before the wizard was even created. I added a unit test to cover this scenario and fixed the bug (back button wasn't actually going back).

Also, `instanceof` notoriously doesn't work well when linking packages so I got rid of that when setting `rootUserInput`

### DIFF
--- a/ui/src/wizard/AzureWizard.ts
+++ b/ui/src/wizard/AzureWizard.ts
@@ -6,9 +6,8 @@
 import { isNullOrUndefined } from 'util';
 import * as vscode from 'vscode';
 import * as types from '../../index';
-import { AzureUserInput, IRootUserInput } from '../AzureUserInput';
 import { GoBackError } from '../errors';
-import { ext } from '../extensionVariables';
+import { ext, IRootUserInput } from '../extensionVariables';
 import { AzureWizardExecuteStep } from './AzureWizardExecuteStep';
 import { AzureWizardPromptStep } from './AzureWizardPromptStep';
 import { AzureWizardUserInput, IInternalAzureWizard } from './AzureWizardUserInput';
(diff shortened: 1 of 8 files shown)


ui/src/wizard/AzureWizard.ts
```diff
@@ -32,51 +31,49 @@ export class AzureWizard<T> implements types.AzureWizard<T>, IInternalAzureWizar
     }
 
     public get currentStep(): number {
-        return this._finishedPromptSteps.length + 1;
+        return this._finishedPromptSteps.filter(s => s.prompted).length + 1;
     }
 
     public get totalSteps(): number {
-        return this._finishedPromptSteps.length + this._promptSteps.filter(s => s.shouldPrompt(this._wizardContext)).length + 1;
+        return this._finishedPromptSteps.filter(s => s.prompted).length + this._promptSteps.filter(s => s.shouldPrompt(this._wizardContext)).length + 1;
```
StephenWeatherford: Why the +1?

ui/src/wizard/AzureWizard.ts
```diff
@@ -32,51 +31,49 @@ export class AzureWizard<T> implements types.AzureWizard<T>, IInternalAzureWizar
...
                         } else {
                             throw err;
                         }
                     }
                 }
 
+                this._finishedPromptSteps.push(step);
                 step = this._promptSteps.pop();
             }
         } finally {
```
StephenWeatherford: Why is this setting ext.ui.rootUserInput to vscode.window when it was undefined before?

ui/src/wizard/AzureWizard.ts
EricJizbaMSFT: +1 is for the current step. Aka the step that has already been popped off of `this._promptSteps`, but won't be pushed to `this._finishedPromptSteps` until after the UI is shown

ui/src/wizard/AzureWizard.ts
EricJizbaMSFT: Realistically it didn't matter since `vscode.window` was already the default, but I agree the logic was weird. Fixed

ui/src/wizard/AzureWizard.ts
StephenWeatherford: Ah.

PR #447, opened 2019-03-13
Add descriptions to Deployments

nturinski:
Adds descriptions to the Deployments node.

Also passes the scmType to the children to change their context value.  `View Commit in Github` should definitely be hidden unless it's a Github repository which now is easy for the user to identify since it will be in the description.

Fixes https://github.com/Microsoft/vscode-azureappservice/issues/753


appservice/src/tree/DeploymentTreeItem.ts
```diff
@@ -30,15 +30,16 @@ enum DeployStatus {
 }
 
 export class DeploymentTreeItem extends AzureTreeItem<ISiteTreeRoot> {
-    public static contextValue: string = 'deployment';
+    public static contextValue: string;
```
EricJizbaMSFT: This change breaks `showTreeItemPicker` (aka code [like this](https://github.com/Microsoft/vscode-azureappservice/blob/dc917610ae6968a6d2864a4e0b3ce4c81f904066/src/commands/deployments/redeployDeployment.ts#L11)). The static value can't be undefined

The simplest fix would be to define a static `contextValues` array with all the possible values, but IMO the better fix would be to modify `showTreeItemPicker` to work with a RegExp instead of a string (kind of like what you did for TestUserInput [here](https://github.com/Microsoft/vscode-azuretools/pull/389)).

appservice/src/tree/DeploymentsTreeItem.ts
```diff
@@ -92,10 +93,27 @@ export class DeploymentsTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
...
         if (siteConfig.scmType === ScmType.GitHub || siteConfig.scmType === ScmType.LocalGit) {
             this.contextValue = DeploymentsTreeItem.contextValueConnected;
         } else {
             this.contextValue = DeploymentsTreeItem.contextValueUnconnected;
         }
+
+        this.description = this.getDescriptionFromScmType(siteConfig, sourceControl);
+    }
+
+    private getDescriptionFromScmType(siteConfig: SiteConfig, sourceControl: SiteSourceControl): string {
```
EricJizbaMSFT: I would prefer to use getters for `description` and `contextValue`. Something like this where you reference private variables like `_scmType` and `_repoUrl` that are updated in `refreshImpl`:
```typescript
private _scmType: string;
private _repoUrl?: string;

public get description(): string {
    switch (this._scmType) {
        case ScmType.LocalGit:
            return localize('localGit', 'Local Git');
        case ScmType.GitHub:
            // remove github from the repoUrl which leaves only the org/repo names
            return this._repoUrl ? this._repoUrl.substring('https://github.com/'.length) : localize('github', 'Github');
        case ScmType.None:
        default:
            return '';
    }
}
```

appservice/src/tree/DeploymentsTreeItem.ts
```diff
@@ -92,10 +93,27 @@ export class DeploymentsTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
 
     public async refreshImpl(): Promise<void> {
         const siteConfig: SiteConfig = await this.root.client.getSiteConfig();
+        const sourceControl: SiteSourceControl = await this.root.client.getSourceControl();
         if (siteConfig.scmType === ScmType.GitHub || siteConfig.scmType === ScmType.LocalGit) {
```
EricJizbaMSFT: I know it's not completely related to the change in this PR, but I just realized your logic for setting contextValue here is different than the logic in the constructor. Which logic is correct? I would prefer to consolidate the logic into one getter as described in the other comment.

appservice/src/tree/DeploymentsTreeItem.ts
```diff
@@ -92,10 +93,27 @@ export class DeploymentsTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
...
         }
+
+        this.description = this.getDescriptionFromScmType(siteConfig, sourceControl);
+    }
+
+    private getDescriptionFromScmType(siteConfig: SiteConfig, sourceControl: SiteSourceControl): string {
+        switch (siteConfig.scmType) {
+            case ScmType.LocalGit:
+                return localize('localGit', 'Local Git');
+                break;
```
EricJizbaMSFT: nit: `break` after a `return` is unreachable

appservice/src/tree/DeploymentTreeItem.ts
```diff
@@ -138,7 +139,7 @@ export class DeploymentTreeItem extends AzureTreeItem<ISiteTreeRoot> {
             opn(githubCommitUrl);
             return;
         } else {
-            throw new Error(localize('noRepoUrl', 'There is no repo url associated with deployment "{0}".', this._deployResult.id));
+            throw new Error(localize('noRepoUrl', 'There is no Github repo url associated with deployment "{0}".', this._deployResult.id));
```
EricJizbaMSFT: nit: "GitHub" should have a capital "H" in all of your strings here

appservice/src/tree/DeploymentsTreeItem.ts
```diff
@@ -27,10 +27,11 @@ export class DeploymentsTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
 
     private readonly _connectToGitHubCommandId: string;
 
-    public constructor(parent: AzureParentTreeItem<ISiteTreeRoot>, siteConfig: SiteConfig, connectToGitHubCommandId: string) {
+    public constructor(parent: AzureParentTreeItem<ISiteTreeRoot>, siteConfig: SiteConfig, sourceControl: SiteSourceControl, connectToGitHubCommandId: string) {
         super(parent);
         this.contextValue = siteConfig.scmType === ScmType.None ? DeploymentsTreeItem.contextValueUnconnected : DeploymentsTreeItem.contextValueConnected;
```
EricJizbaMSFT: What do you think about adding things like 'repoUrl' and 'scmType' (or even `siteConfig` and `sourceControl`) to `ISiteRoot` or `SiteClient`? It seems like multiple different tree items are using those types of properties and I feel like it's inefficient to get them so many times. We'd probably have to add a 'refresh' function to ISiteRoot, though

nturinski: Yeah, I agree.  We tend to grab it a lot for deployment-related things.  I think I prefer having it as a property of `SiteClient` and having refresh just create a new, updated `client`.  Do you have a preference?

EricJizbaMSFT: I don't want to add anything to SiteClient that involves getting the plan due to https://github.com/Microsoft/vscode-azureappservice/issues/769, but site config and source control should be fine

nturinski: So I think it only makes sense to add `SiteConfig` to `SiteClient`.  `SourceControl` is only used three times and it's mostly related to Git deployment.  Furthermore, `SiteConfig` is included in the `Site` object, but `SourceControl` requires a separate API call.

nturinski: Never mind about `SiteConfig` being included in the `Site` object.  It's an optional parameter and isn't returned with whatever API call we used to get the `Site`.

appservice/src/tree/DeploymentTreeItem.ts
```diff
@@ -28,15 +28,16 @@ enum DeployStatus {
...
-    public readonly contextValue: string = DeploymentTreeItem.contextValue;
+    public static contextValue: RegExp = new RegExp('deployment\/.*');
+    public readonly contextValue: string;
     public label: string;
     public receivedTime: Date;
     public parent: DeploymentsTreeItem;
     private _deployResult: DeployResult;
 
-    constructor(parent: DeploymentsTreeItem, deployResult: DeployResult) {
+    constructor(parent: DeploymentsTreeItem, deployResult: DeployResult, scmType?: string) {
```
EricJizbaMSFT: Is scmType really optional? I'm guessing this should be `scmType: string | undefined` instead. We've been over this before, but there's a difference between the two. My suggestion forces people to pass _something_ in every time they call `new DeploymentTreeItem()`. With the current implementation, it's easy to forget to pass in scmType and TypeScript won't complain

appservice/src/tree/DeploymentTreeItem.ts
```diff
@@ -79,6 +80,10 @@ export class DeploymentTreeItem extends AzureTreeItem<ISiteTreeRoot> {
         }
     }
 
+    public isAncestorOfImpl(contextValue: string): boolean {
```
EricJizbaMSFT: nit: `contextValue: string | RegExp`

PR #448, opened 2019-03-14
Clean up log stream output

EricJizbaMSFT:
1. Use `append` instead of `appendLine` since the log stream data already has line breaks
1. Fixes https://github.com/Microsoft/vscode-azurefunctions/issues/1089 where the functions log stream sometimes displays duplicate information. As far as I can tell the 'data' comes in on a per-line basis with a timestamp, so I'm pretty confident this logic only removes duplicate lines. Y'all think I need to be more careful in case the data comes in smaller chunks?

### DIFF
--- a/appservice/src/startStreamingLogs.ts
+++ b/appservice/src/startStreamingLogs.ts
@@ -62,10 +62,9 @@ export async function startStreamingLogs(client: SiteClient, verifyLoggingEnable
 
                 await new Promise((onLogStreamEnded: () => void, reject: (err: Error) => void): void => {
                     let recentData: string = '';
-                    let newLogStream: ILogStream;
                     const logsRequest: request.Request = requestApi(`${client.kuduUrl}/api/logstream/${logsPath}`);
                     const recentDataTimer: NodeJS.Timer = setInterval(() => { recentData = ''; }, 2 * 1000);
-                    newLogStream = {
+                    const newLogStream: ILogStream = {
                         dispose: (): void => {
                             logsRequest.removeAllListeners();
                             logsRequest.destroy();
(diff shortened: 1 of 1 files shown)


appservice/src/startStreamingLogs.ts
```diff
@@ -61,8 +61,10 @@ export async function startStreamingLogs(client: SiteClient, verifyLoggingEnable
                 }
 
                 await new Promise((onLogStreamEnded: () => void, reject: (err: Error) => void): void => {
+                    let recentData: string = '';
                     let newLogStream: ILogStream;
                     const logsRequest: request.Request = requestApi(`${client.kuduUrl}/api/logstream/${logsPath}`);
+                    const recentDataTimer: NodeJS.Timer = setInterval(() => { recentData = ''; }, 2 * 1000);
                     newLogStream = {
```
StephenWeatherford: Where is newLogStream used?

appservice/src/startStreamingLogs.ts
```diff
@@ -79,8 +82,13 @@ export async function startStreamingLogs(client: SiteClient, verifyLoggingEnable
                         outputChannel: outputChannel
                     };
 
-                    logsRequest.on('data', (chunk: Buffer | string) => {
-                        outputChannel.appendLine(chunk.toString());
+                    logsRequest.on('data', (data: Buffer | string) => {
```
StephenWeatherford: What exactly is causing the bug?  Needs more commehnts.

EricJizbaMSFT: Added more comments

appservice/src/startStreamingLogs.ts
EricJizbaMSFT: Uh a few places in this function. It's actually the return value of this root `startStreamingLogs` function. 

`startStreamingLogs` is admittedly a confusing mess of promises which could possibly be improved, but part of that is just the nature of a long-running action like log streaming

appservice/src/startStreamingLogs.ts
StephenWeatherford: So how likely is it that data will come in right before the tick and the duplicated data right after the tick?

EricJizbaMSFT:
Closing this one for now. Duplicate data is annoying, but not as bad as accidentally hiding non-duplicate data IMO. I want to be more careful about this

PR #451, opened 2019-03-18
Provide subscription prompt step from tree

EricJizbaMSFT:
wizard everything lol

### DIFF
--- a/ui/src/treeDataProvider/AzureTreeDataProvider.ts
+++ b/ui/src/treeDataProvider/AzureTreeDataProvider.ts
@@ -15,6 +15,7 @@ import { ext } from '../extensionVariables';
 import { localize } from '../localize';
 import { parseError } from '../parseError';
 import { TestAzureAccount } from '../TestAzureAccount';
+import { AzureWizardPromptStep } from '../wizard/AzureWizardPromptStep';
 import { AzureParentTreeItem } from './AzureParentTreeItem';
 import { AzureTreeItem } from './AzureTreeItem';
 import { GenericTreeItem } from './GenericTreeItem';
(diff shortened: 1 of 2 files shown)


ui/src/treeDataProvider/AzureTreeDataProvider.ts
```diff
@@ -220,6 +221,25 @@ export class AzureTreeDataProvider<TRoot = ISubscriptionRoot> implements IAzureT
         return element.parent;
     }
 
+    public async getSubscriptionPromptStep(wizardContext: Partial<types.ISubscriptionWizardContext>): Promise<types.AzureWizardPromptStep<types.ISubscriptionWizardContext> | undefined> {
+        const subscriptions: AzureTreeItem<ISubscriptionRoot>[] = await this.ensureRootTreeItems();
+        if (subscriptions.length === 1) {
+            Object.assign(wizardContext, subscriptions[0].root);
+            return undefined;
```
StephenWeatherford: There are more properties in ISubscriptionRoot than ISubscriptionWizardContext (tenantId/userId).  Is this assignment a good idea?  Should ISubscriptionWizardContext extend ISubscriptionRoot?

EricJizbaMSFT: > Should ISubscriptionWizardContext extend ISubscriptionRoot?

I'm hesitant to do that for this PR because it would affect other repos (If they create a wizardContext, they'll now be required to specify those extra props). For now, I made sure to only copy necessary props

ui/src/treeDataProvider/AzureTreeDataProvider.ts
```diff
@@ -220,6 +221,25 @@ export class AzureTreeDataProvider<TRoot = ISubscriptionRoot> implements IAzureT
         return element.parent;
     }
 
+    public async getSubscriptionPromptStep(wizardContext: Partial<types.ISubscriptionWizardContext>): Promise<types.AzureWizardPromptStep<types.ISubscriptionWizardContext> | undefined> {
```
nturinski: Why aren't you creating a .ts file that is the SubscriptionPromptStep rather than implementing it in the `AzureTreeDataProvider.ts`?

ui/src/treeDataProvider/AzureTreeDataProvider.ts
```diff
@@ -220,6 +221,25 @@ export class AzureTreeDataProvider<TRoot = ISubscriptionRoot> implements IAzureT
...
+        if (subscriptions.length === 1) {
+            assignRootToWizardContext(wizardContext, subscriptions[0].root);
+            return undefined;
+        } else {
+            // tslint:disable-next-line: no-var-self
+            const tree: AzureTreeDataProvider<TRoot> = this;
+            class SubscriptionPromptStep extends AzureWizardPromptStep<types.ISubscriptionWizardContext> {
+                public async prompt(): Promise<void> {
+                    const ti: AzureTreeItem<TRoot | ISubscriptionRoot> = await tree.promptForRootTreeItem(SubscriptionTreeItem.contextValue);
+                    assignRootToWizardContext(wizardContext, <types.ISubscriptionRoot>ti.root);
```
nturinski: What if the user cancels the prompt and `ti` is undefined?  Does `promptForRootTreeItem` use our `UserCancelledError` handling?

ui/src/treeDataProvider/AzureTreeDataProvider.ts
EricJizbaMSFT: Just a stylistic preference. It's not that many lines of code and it should only be used within this function anyways

ui/src/treeDataProvider/AzureTreeDataProvider.ts
EricJizbaMSFT: Yeah it'll throw a `promptForRootTreeItem`, not return undefined.

PR #452, opened 2019-03-18
Add regexp to accepted contextValue

nturinski:
The gif shows a kind of weird "bug" that I'm not sure exactly how to address.  But basically, because the node has children, it displays them, but after being selected, the context value does not match what is expected, so it throws the error.  The second attempt is a context value we are expecting and the command executes as expected.

vscode_regexp_ui

### DIFF
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -6,7 +6,7 @@
     "dependencies": {
         "@types/fs-extra": {
             "version": "4.0.8",
-            "resolved": "http://registry.npmjs.org/@types/fs-extra/-/fs-extra-4.0.8.tgz",
+            "resolved": "https://registry.npmjs.org/@types/fs-extra/-/fs-extra-4.0.8.tgz",
             "integrity": "sha512-Z5nu9Pbxj9yNeXIK3UwGlRdJth4cZ5sCq05nI7FaI6B0oz28nxkOtp6Lsz0ZnmLHJGvOJfB/VHxSTbVq/i6ujA==",
             "dev": true,
             "requires": {
@@ -933,7 +933,7 @@
         },
         "deep-assign": {
             "version": "1.0.0",
-            "resolved": "http://registry.npmjs.org/deep-assign/-/deep-assign-1.0.0.tgz",
+            "resolved": "https://registry.npmjs.org/deep-assign/-/deep-assign-1.0.0.tgz",
             "integrity": "sha1-sJJ0O+hCfcYh6gBnzex+cN0Z83s=",
             "dev": true,
             "requires": {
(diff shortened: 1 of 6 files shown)


ui/package-lock.json
```diff
@@ -1,12 +1,12 @@
...
     "name": "vscode-azureextensionui",
-    "version": "0.21.1",
+    "version": "0.21.2",
     "lockfileVersion": 1,
     "requires": true,
     "dependencies": {
         "@types/fs-extra": {
             "version": "4.0.8",
-            "resolved": "https://registry.npmjs.org/@types/fs-extra/-/fs-extra-4.0.8.tgz",
+            "resolved": "http://registry.npmjs.org/@types/fs-extra/-/fs-extra-4.0.8.tgz",
```
EricJizbaMSFT: Can you do all that stuff to clean out your npm cache? So we don't get these http changes?

ui/src/treeDataProvider/AzureTreeDataProvider.ts
```diff
@@ -175,16 +175,17 @@ export class AzureTreeDataProvider<TRoot = ISubscriptionRoot> implements IAzureT
...
             expectedContextValues = [expectedContextValues];
         }
 
         // tslint:disable-next-line:strict-boolean-expressions
         let treeItem: AzureTreeItem<TRoot | ISubscriptionRoot> = startingTreeItem || await this.promptForRootTreeItem(expectedContextValues);
-        while (!expectedContextValues.some((val: string) => treeItem.contextValue === val)) {
-            if (treeItem instanceof AzureParentTreeItem) {
-                treeItem = await treeItem.pickChildTreeItem(expectedContextValues);
+
+        while (!expectedContextValues.some((val: string | RegExp) => (val instanceof RegExp && treeItem.contextValue.match(val) !== null) || treeItem.contextValue === val)) {
```
EricJizbaMSFT: Use `val.test(treeItem.contextValue)` instead of `treeItem.contextValue.match(val) !== null`

ui/src/treeDataProvider/AzureTreeDataProvider.ts
```diff
@@ -175,16 +175,17 @@ export class AzureTreeDataProvider<TRoot = ISubscriptionRoot> implements IAzureT
...
+            if ((<AzureParentTreeItem>treeItem).hasMoreChildrenImpl) {
```
EricJizbaMSFT: I'm fine with this change, but add a comment saying why we're not using `instanceof`

EricJizbaMSFT: Also - I think I'd prefer to check `if ((<AzureParentTreeItem>treeItem).loadMoreChildrenImpl) {` just because `hasMoreChildrenImpl` already returns a boolean and that seems potentially more confusing

PR #458, opened 2019-03-25
Allow hiding step count

EricJizbaMSFT:
Sometimes the step count isn't useful because of the structure of sub wizards. This lets people hide the count. Used in this PR: https://github.com/Microsoft/vscode-azurefunctions/pull/1119

Also only show step count when executing if there's more than one step.

### DIFF
--- a/ui/src/wizard/AzureWizard.ts
+++ b/ui/src/wizard/AzureWizard.ts
@@ -14,6 +14,7 @@ import { AzureWizardUserInput, IInternalAzureWizard } from './AzureWizardUserInp
 import { getExecuteSteps, IWizardNode } from './IWizardNode';
 
 export class AzureWizard<T> implements types.AzureWizard<T>, IInternalAzureWizard {
+    public hideStepCount: boolean;
     private _title: string | undefined;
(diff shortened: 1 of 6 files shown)


ui/src/wizard/AzureWizard.ts
```diff
@@ -52,6 +53,7 @@ export class AzureWizard<T> implements types.AzureWizard<T>, IInternalAzureWizar
 
                 actionContext.properties.lastStepAttempted = `prompt-${step.constructor.name}`;
                 this._title = step.wizardNode.effectiveTitle;
+                this.hideStepCount = step.hideStepCount;
```
StephenWeatherford: So this is a per-step setting?  Not sure that makes sense to me.

EricJizbaMSFT: I ran into this when users were creating a project or function. In both wizards, I first prompt for something that determines most of the rest of the wizard (project language or function template). The step count was showing as `1/1` even though it should be `1/?`. The only other option I see is to hide on a per-wizard basis, but that's not what I wanted in this scenario. It's still helpful for users to see how many steps there are _after_ they select a language/template. We could always add the per-wizard option later

PR #460, opened 2019-03-26
Add powershell to list of runtimes under feature flag

EricJizbaMSFT:
And remove use of opn

### DIFF
--- a/appservice/src/connectToGitHub.ts
+++ b/appservice/src/connectToGitHub.ts
@@ -5,7 +5,6 @@
 
 import { SiteSourceControl } from 'azure-arm-website/lib/models';
 import { TokenCredentials, WebResource } from 'ms-rest';
-import * as opn from 'opn';
 import { Response } from 'request';
 import * as request from 'request-promise';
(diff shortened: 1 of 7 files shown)


appservice/src/connectToGitHub.ts
```diff
@@ -103,14 +103,13 @@ async function showGitHubAuthPrompt(node: AzureTreeItem, client: SiteClient, con
...
-            // tslint:disable-next-line:no-unsafe-any
-            opn('https://aka.ms/B7g6sw');
+            await openUrl('https://aka.ms/B7g6sw');
         }
     }
 
     if (input === goToPortal) {
         context.properties.githubGoToPortal = 'true';
-        node.openInPortal(`${client.id}/vstscd`);
+        await node.openInPortal(`${client.id}/vstscd`);
```
nturinski: Wait, is this an async method?

EricJizbaMSFT: It is after my recent PR in UI :)

But it honestly doesn't matter if you await opening a url. Only lint really cares

PR #463, opened 2019-03-29
Simplify ordering of execute steps

EricJizbaMSFT:
The current ordering of execute steps is overly complex and doesn't even work all the time as demonstrated by https://github.com/Microsoft/vscode-azurefunctions/issues/1137. I changed it so that the order is based on a simple `priority` property that must be defined in each execute step. I definitely wish I had done it this way in the first place because it feels so much more intuitive, flexible, and simple.


ui/src/wizard/AzureWizard.ts
```diff
@@ -91,7 +88,7 @@ export class AzureWizard<T> implements types.AzureWizard<T>, IInternalAzureWizar
         await vscode.window.withProgress({ location: vscode.ProgressLocation.Notification }, async progress => {
             let currentStep: number = 1;
 
-            const steps: AzureWizardExecuteStep<T>[] = getExecuteSteps(this._wizardNode);
+            const steps: AzureWizardExecuteStep<T>[] = this._executeSteps.sort((a, b) => b.priority - a.priority);
```
StephenWeatherford: Should you warn about dup priorities?

EricJizbaMSFT: Not sure what you mean by "warn":
- I don't want to do `vscode.window.showWarningMessage` because I'm guessing most cases will work just fine even if there's duplicate priorities and this feels like it would "block" those cases
- I'm not opposed to `console.warn`, but I wouldn't add it myself. Just a personal preference to avoid `console.*` in production code

StephenWeatherford:
I was referring to an assert/exception, although console.warn would be fine.  It’s up to you if you think it makes sense or not.

PR #466, opened 2019-04-03
Allow start proxy to listen to other ports

nturinski:
The changed startProxy to listen to specific ports rather than just NOT 2222.  2222 is used by SSH, remoteDebugging uses 49494 (as far as I can tell).  This doesn't seem to be set by us.


appservice/src/TunnelProxy.ts
```diff
@@ -155,8 +155,8 @@ export class TunnelProxy {
         this._openSockets = [];
     }
 
-    public async startProxy(): Promise<void> {
-        await this.checkTunnelStatusWithRetry();
+    public async startProxy(portNumber: number): Promise<void> {
```
EricJizbaMSFT: `portNumber` and `this._port` are too similar in naming. I suggest you pass both ports in to the constructor and give them more unique names with documentation

PR #467, opened 2019-04-05
Fix Azure Data Studio

StephenWeatherford:
Azure Data Studio is currently using vscode 1.30 with Chrome 61.  Chrome 61 doesn't support the regex /s flag and so our extensions fail to load.

Azure Data Studio
Version: 1.5.2 (user setup)
Commit: f74080c96310354bc92211f1826c0fdce78ca2f6
Date: 2019-03-22T06:25:26.627Z
VS Code 1.30.1
Electron: 2.0.12
Chrome: 61.0.3163.100
Node.js: 8.9.3
V8: 6.1.534.41
OS: Windows_NT x64 10.0.17763

From https://github.com/Microsoft/vscode-docker/issues/881

### DIFF
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -1,6 +1,6 @@
 {
     "name": "vscode-azureextensionui",
-    "version": "0.23.0",
+    "version": "0.23.1",
     "lockfileVersion": 1,
     "requires": true,
     "dependencies": {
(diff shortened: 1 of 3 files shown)


StephenWeatherford:
BTW, this also fixes our extension in older versions of vscode.

PR #470, opened 2019-04-09
Refactor how webpack handles icons in shared packages

EricJizbaMSFT:
Mainly I was trying to fix icons when linking packages, but there's the added side-effect of simplifying the default webpack config.

I bumped the minor version of both packages as this is a breaking change. Extensions have to create their own `StringReplacePlugin` (if using it) and they need to specify the following in their `.vscodeignore` file:
```
!node_modules/vscode-*/resources/**/*.svg
```


ui/src/treeDataProvider/AzureParentTreeItem.ts
```diff
@@ -188,7 +187,7 @@ export abstract class AzureParentTreeItem<TRoot = ISubscriptionRoot> extends Azu
 
         if (this.hasMoreChildrenImpl()) {
             picks.push({
-                label: `$(sync) ${loadMoreLabel}`,
+                label: localize('LoadMore', '$(sync) Load more...'),
```
StephenWeatherford: nit: Would be safer to remove the $(sync) from the localizable portion

StephenWeatherford:
I'd like to think about this one.  I don't like that you have to know to set it up, and that we're shipping something out of dist and normal locations like resources/images.  Maybe we can discuss.

EricJizbaMSFT:
If there's any debate, I say this isn't worth the time right now. I'll close this PR and re-visit after build conference is over

PR #471, opened 2019-04-11
Change tunnelProxy to only listen to either 2222 or not 2222

nturinski:
There are only 2 states being considered there. If the port is set to default (2222), then the tunnel is ready for SSH. If it is different from the default, then the tunnel is ready for remote debugging.

### DIFF
--- a/appservice/package.json
+++ b/appservice/package.json
@@ -1,7 +1,7 @@
 {
     "name": "vscode-azureappservice",
     "author": "Microsoft Corporation",
-    "version": "0.36.1",
+    "version": "0.37.0",
     "description": "Common tools for developing Azure App Service extensions for VS Code",
     "tags": [
         "azure",
--- a/appservice/package.json
+++ b/appservice/package.json
@@ -1,7 +1,7 @@
 {
     "name": "vscode-azureappservice",
     "author": "Microsoft Corporation",
-    "version": "0.36.0",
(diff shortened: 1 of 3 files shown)


appservice/src/TunnelProxy.ts
```diff
@@ -141,20 +141,20 @@ class RetryableTunnelStatusError extends Error { }
...
     private _client: SiteClient;
     private _publishCredential: User;
     private _server: Server;
     private _openSockets: TunnelSocket[];
+    private _useDefaultPort: boolean;
 
-    constructor(localPort: number, remotePort: number, client: SiteClient, publishCredential: User) {
-        this._localPort = localPort;
-        this._remotePort = remotePort;
+    constructor(port: number, client: SiteClient, publishCredential: User, useDefaultPort: boolean = false) {
```
EricJizbaMSFT: I think `sshPort` is more accurate than `defaultPort`. It's also not really "using" that port - it's just validating that port. I'd prefer a variable name like `isSsh`.

appservice/package.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "vscode-azureappservice",
     "author": "Microsoft Corporation",
-    "version": "0.36.0",
+    "version": "0.36.1",
```
EricJizbaMSFT: Mmm still a breaking change.

PR #472, opened 2019-04-11
Use arm rest api instead of kudu for webjobs

nturinski:
Related to https://github.com/Microsoft/vscode-azureappservice/issues/891 but does not fix the issue

### DIFF
--- a/appservice/src/SiteClient.ts
+++ b/appservice/src/SiteClient.ts
@@ -260,7 +260,7 @@ export class SiteClient {
         return await this._client.getPublishingUser({});
     }
 
-    public async getWebJobs(): Promise<WebJobCollection> {
+    public async listWebJobs(): Promise<WebJobCollection> {
         return this.slotName ?
             await this._client.webApps.listWebJobsSlot(this.resourceGroup, this.siteName, this.slotName) :
             await this._client.webApps.listWebJobs(this.resourceGroup, this.siteName);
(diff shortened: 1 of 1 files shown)


appservice/src/SiteClient.ts
```diff
@@ -259,4 +259,10 @@ export class SiteClient {
     public async getPublishingUser(): Promise<User> {
         return await this._client.getPublishingUser({});
     }
+
+    public async getWebJobs(): Promise<WebJobCollection> {
```
EricJizbaMSFT: nit: Prefer to keep the naming consistent with the below call (listWebJobs)

appservice/src/SiteClient.ts
```diff
@@ -259,4 +259,10 @@ export class SiteClient {
...
+        return this.slotName ?
+            await this._client.webApps.listWebJobsSlot(this.resourceGroup, this.siteName, this.slotName) :
+            await this._client.webApps.listWebJobs(this.resourceGroup, this.siteName);
```
EricJizbaMSFT: Do you need to add a listWebJobsNext? Does that exist?

nturinski:
It seems to work the same as `kudu.listJobs` did, but the source of the problem is that WebJobs isn't supported on Linux and returns 404 instead.

https://github.com/Microsoft/vscode-azureappservice/pull/934

EricJizbaMSFT:
Okay cool as long as it works on windows

appservice/src/SiteClient.ts
nturinski: Does not exist

nturinski: Ugh, sorry this is incorrect.  It wasn't showing up before, but I see it in the API.

PR #473, opened 2019-04-12
Add telemetry to localGit and add modal for force push

nturinski:
Fixes https://github.com/Microsoft/vscode-azureappservice/issues/896
Fices https://github.com/Microsoft/vscode-azureappservice/issues/897

### DIFF
--- a/appservice/src/deploy/localGitDeploy.ts
+++ b/appservice/src/deploy/localGitDeploy.ts
@@ -6,7 +6,7 @@
 import { User } from 'azure-arm-website/lib/models';
 import * as git from 'simple-git/promise';
 import * as vscode from 'vscode';
-import { DialogResponses, UserCancelledError } from 'vscode-azureextensionui';
+import { DialogResponses, TelemetryProperties, UserCancelledError } from 'vscode-azureextensionui';
 import { ext } from '../extensionVariables';
 import { localize } from '../localize';
 import { SiteClient } from '../SiteClient';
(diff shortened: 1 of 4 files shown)


appservice/src/deploy/localGitDeploy.ts
```diff
@@ -30,10 +30,13 @@ export async function localGitDeploy(client: SiteClient, fsPath: string): Promis
             try {
                 const status: git.StatusResult = await localGit.status();
                 if (status.files.length > 0) {
+                    telemetryProperties.pushWithUncommitChanges = 'false';
```
EricJizbaMSFT: I'd suggest using `cancelStep`. It's a somewhat common pattern that we already use and it's already picked up in a few queries I think. Here's an example: https://github.com/Microsoft/vscode-azureappservice/blob/80fa5b5317389e5549b86a2bc038808666cf806d/src/commands/deploy.ts#L136

I would do this before the dialog:
```
telemetryProperties.cancelStep = 'pushWithUncommitChanges';
```

And this after the dialog:
```
telemetryProperties.cancelStep = undefined;
telemetryProperties.pushWithUncommitChanges = 'true';
```

appservice/src/deploy/localGitDeploy.ts
```diff
@@ -46,15 +49,18 @@ export async function localGitDeploy(client: SiteClient, fsPath: string): Promis
...
                 } else if (err.message.indexOf('error: failed to push') >= 0) {
                     const forcePushMessage: vscode.MessageItem = { title: localize('forcePush', 'Force Push') };
                     const pushReject: string = localize('localGitPush', 'Push rejected due to Git history diverging.');
 
-                    if (await ext.ui.showWarningMessage(pushReject, forcePushMessage, DialogResponses.cancel) === forcePushMessage) {
+                    if (await ext.ui.showWarningMessage(pushReject, { modal: true }, forcePushMessage, DialogResponses.cancel) === forcePushMessage) {
+                        telemetryProperties.forcePush = 'true';
                         await tryPushAndWaitForDeploymentToComplete(true);
                     } else {
+                        telemetryProperties.forcePush = 'false';
```
EricJizbaMSFT: telemetryProperties.cancelStep = 'forcePush'

PR #475, opened 2019-04-15
Function to filter out Linux runtimes with a RegExp

nturinski:
Fixes https://github.com/Microsoft/vscode-azureappservice/issues/917

### DIFF
--- a/appservice/src/createAppService/SiteRuntimeStep.ts
+++ b/appservice/src/createAppService/SiteRuntimeStep.ts
@@ -45,7 +45,7 @@ export class SiteRuntimeStep extends AzureWizardPromptStep<IAppServiceWizardCont
             });
 
             // filters out Node 4.x and 6.x as they are EOL
-            runtimeItems = this.filterRuntimes(runtimeItems, /node\|(4|6)\./);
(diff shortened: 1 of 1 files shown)


appservice/src/createAppService/SiteRuntimeStep.ts
```diff
@@ -43,6 +43,9 @@ export class SiteRuntimeStep extends AzureWizardPromptStep<IAppServiceWizardCont
                     data: rt
                 };
             });
+
+            // filters out Node 4.x and 6.x as they are EOL
+            runtimeItems = this.filterRuntimes(runtimeItems, /node\|(4|6)\./);
```
EricJizbaMSFT: nit: I almost always throw an `i` flag at the end of a regexp to handle weird casing

EricJizbaMSFT: The `filterRuntimes` function seems unnecessary. It can already be a one-liner:
```
runtimeItems = runtimeItems.filter(qp => !/node\|(4|6)\./i.test(qp.data.name));
```

(Remember I turned off typedef for arrow-parameter. Looks much cleaner without the typedef IMO)

PR #479, opened 2019-04-16
Sort LinuxRuntimes alphabetically and push recommended to the top

nturinski:
Fixes https://github.com/Microsoft/vscode-azureappservice/issues/880

The reason Python is at the top is due to having a Python project opened.  Otherwise, it'd default to alphabetical ordering (which leads to weird ordering for Node 10 being at the top).  I can fix this if it's worth it.  

@fiveisprime Also if we don't want to do alphabetically and instead order it by specific runtimes (like Node, Python, and then .NET for example) I can do that.

image

### DIFF
--- a/appservice/src/createAppService/SiteRuntimeStep.ts
+++ b/appservice/src/createAppService/SiteRuntimeStep.ts
@@ -197,7 +197,7 @@ export class SiteRuntimeStep extends AzureWizardPromptStep<IAppServiceWizardCont
     }
 
     private sortQuickPicksByRuntime(runtimeQuickPicks: IAzureQuickPickItem<ILinuxRuntimeStack>[], runtimeRecommendations: string[]): IAzureQuickPickItem<ILinuxRuntimeStack>[] {
-        function getPriority(ti1: IAzureQuickPickItem<ILinuxRuntimeStack>, ti2: IAzureQuickPickItem<ILinuxRuntimeStack>): number {
+        return runtimeQuickPicks.sort((ti1: IAzureQuickPickItem<ILinuxRuntimeStack>, ti2: IAzureQuickPickItem<ILinuxRuntimeStack>) => {
             const index1: number = runtimeRecommendations.findIndex((runtime: string) => ti1.data.name.includes(runtime.toLocaleLowerCase()));
             const index2: number = runtimeRecommendations.findIndex((runtime: string) => ti2.data.name.includes(runtime.toLocaleLowerCase()));
 
@@ -208,8 +208,6 @@ export class SiteRuntimeStep extends AzureWizardPromptStep<IAppServiceWizardCont
             } else {
                 return ti1.data.name.localeCompare(ti2.data.name);
             }
-        }
-
-        return runtimeQuickPicks.sort((a: IAzureQuickPickItem<ILinuxRuntimeStack>, b: IAzureQuickPickItem<ILinuxRuntimeStack>) => getPriority(a, b));
+        });
     }
 }
--- a/appservice/src/createAppService/SiteRuntimeStep.ts
+++ b/appservice/src/createAppService/SiteRuntimeStep.ts
@@ -89,7 +89,7 @@ export class SiteRuntimeStep extends AzureWizardPromptStep<IAppServiceWizardCont
     private sortQuickPicksByRuntime(runtimeQuickPicks: IAzureQuickPickItem<ApplicationStack>[], runtimeRecommendations: string[]): IAzureQuickPickItem<ApplicationStack>[] {
         const recommendedQuickPicks: IAzureQuickPickItem<ApplicationStack>[] = [];
         // go backwards so that when runtime is unshifted to front, the order is unaffected
-        for (let i = runtimeQuickPicks.length - 1; i >= 0; i--) {
+        for (let i: number = runtimeQuickPicks.length - 1; i >= 0; i = i - 1) {
             for (const rt of runtimeRecommendations) {
                 if (nonNullProp(runtimeQuickPicks[i].data, 'name').toLocaleLowerCase().includes(rt.toLocaleLowerCase())) {
                     recommendedQuickPicks.unshift(runtimeQuickPicks.splice(i, 1)[0]);
--- a/appservice/src/createAppService/SiteRuntimeStep.ts
+++ b/appservice/src/createAppService/SiteRuntimeStep.ts
@@ -196,11 +196,20 @@ export class SiteRuntimeStep extends AzureWizardPromptStep<IAppServiceWizardCont
         ];
     }
 
-    private sortQuickPicksByRuntime(runtimeItems: IAzureQuickPickItem<ILinuxRuntimeStack>[], recommendedRuntimes: string[]): IAzureQuickPickItem<ILinuxRuntimeStack>[] {
-        function getPriority(item: IAzureQuickPickItem<ILinuxRuntimeStack>): number {
-            const index: number = recommendedRuntimes.findIndex((runtime: string) => item.data.name.includes(runtime));
-            return index === -1 ? recommendedRuntimes.length : index;
+    private sortQuickPicksByRuntime(runtimeQuickPicks: IAzureQuickPickItem<ILinuxRuntimeStack>[], runtimeRecommendations: string[]): IAzureQuickPickItem<ILinuxRuntimeStack>[] {
+        function getPriority(ti1: IAzureQuickPickItem<ILinuxRuntimeStack>, ti2: IAzureQuickPickItem<ILinuxRuntimeStack>): number {
+            const index1: number = runtimeRecommendations.findIndex((runtime: string) => ti1.data.name.includes(runtime.toLocaleLowerCase()));
+            const index2: number = runtimeRecommendations.findIndex((runtime: string) => ti2.data.name.includes(runtime.toLocaleLowerCase()));
+
+            if (index1 > -1) {
+                return -1;
+            } else if (index2 > -1) {
+                return 1;
+            } else {
+                return ti1.data.name.localeCompare(ti2.data.name);
+            }
         }
-        return runtimeItems.sort((a: IAzureQuickPickItem<ILinuxRuntimeStack>, b: IAzureQuickPickItem<ILinuxRuntimeStack>) => getPriority(a) - getPriority(b));
+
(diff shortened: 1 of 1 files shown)


appservice/src/createAppService/SiteRuntimeStep.ts
```diff
@@ -196,11 +196,18 @@ export class SiteRuntimeStep extends AzureWizardPromptStep<IAppServiceWizardCont
...
+        return runtimeQuickPicks.sort((ti1: IAzureQuickPickItem<ILinuxRuntimeStack>, ti2: IAzureQuickPickItem<ILinuxRuntimeStack>) => {
+            const index1: number = runtimeRecommendations.findIndex((runtime: string) => ti1.data.name.includes(runtime.toLocaleLowerCase()));
+            const index2: number = runtimeRecommendations.findIndex((runtime: string) => ti2.data.name.includes(runtime.toLocaleLowerCase()));
+
+            if (index1 > -1) {
+                return -1;
+            } else if (index2 > -1) {
+                return 1;
+            } else {
+                return ti1.data.name.localeCompare(ti2.data.name);
```
EricJizbaMSFT: I don't think this will sort the runtimes for the recommended case. Try your changes with a node app open - I'm guessing the node runtimes won't be sorted. You need to check if `ti1` and `ti2` are the same runtime (aka both "node" or both "python") and if so sort by localCompare _first_. If they're not the same, that's when you use the `getPriority` function.

appservice/src/createAppService/SiteRuntimeStep.ts
```diff
@@ -196,11 +196,18 @@ export class SiteRuntimeStep extends AzureWizardPromptStep<IAppServiceWizardCont
         ];
     }
 
-    private sortQuickPicksByRuntime(runtimeItems: IAzureQuickPickItem<ILinuxRuntimeStack>[], recommendedRuntimes: string[]): IAzureQuickPickItem<ILinuxRuntimeStack>[] {
-        function getPriority(item: IAzureQuickPickItem<ILinuxRuntimeStack>): number {
```
EricJizbaMSFT: I don't understand why you got rid of the `getPriority` function. It reduces the amount of copied code

nturinski: Because of tslint. It said "code doesn't need a wrapper, just call it directly"

appservice/src/createAppService/SiteRuntimeStep.ts
nturinski: Eh, it kinda sorts it?  But if we're not sorting it alphabetically, I'll probably do it your way

image

appservice/src/createAppService/SiteRuntimeStep.ts
```diff
@@ -86,12 +86,17 @@ export class SiteRuntimeStep extends AzureWizardPromptStep<IAppServiceWizardCont
...
-
-            const index: number = recommendedRuntimes.findIndex((runtime: string) => nonNullProp(item.data, 'name').includes(runtime));
-            return index === -1 ? recommendedRuntimes.length : index;
+    private sortQuickPicksByRuntime(runtimeQuickPicks: IAzureQuickPickItem<ApplicationStack>[], runtimeRecommendations: string[]): IAzureQuickPickItem<ApplicationStack>[] {
+        const recommendedQuickPicks: IAzureQuickPickItem<ApplicationStack>[] = [];
+        // go backwards so that when runtime is unshifted to front, the order is unaffected
+        for (let i: number = runtimeQuickPicks.length - 1; i >= 0; i = i - 1) {
+            for (const rt of runtimeRecommendations) {
+                if (nonNullProp(runtimeQuickPicks[i].data, 'name').toLocaleLowerCase().includes(rt.toLocaleLowerCase())) {
+                    recommendedQuickPicks.unshift(runtimeQuickPicks.splice(i, 1)[0]);
```
EricJizbaMSFT: I don't understand - is this sorting the list? It looks like you're moving `runtimeRecommendations` to the top of the list, but not actually sorting by version. Isn't that the same as what we had before your changes?

nturinski: Yep, it's only moving the recommended to the top.

And no, before that's what I _wanted_ it to do, but the bug came from the fact that the runtimes that were moved up to the top were getting swapped with whatever it replaced in order.  Which is why you'd sometimes see .NET core in the middle of the node runtimes or some of the other odd behaviors.

I thought we didn't really care about sorting, so I just left the default Portal ordering.

nturinski:
So the new sorting method will sort in ascending order EXCEPT if it has been listed as the default.

I'm okay with not doing this but wondered what you thought about taking advantage of that flag.  It might also be worth adding a description indicating that the qp is the default, but could be confusing since there will technically be like 5 "defaults".

image

appservice/src/createAppService/SiteRuntimeStep.ts
```diff
@@ -79,19 +79,36 @@ export class SiteRuntimeStep extends AzureWizardPromptStep<IAppServiceWizardCont
         const runtimesParsed: ApplicationStackJsonResponse = <ApplicationStackJsonResponse>JSON.parse(runtimes);
 
         return runtimesParsed.value.map((runtime) => {
-            return nonNullProp(runtime.properties, 'majorVersions').map((majorVersion) => {
+            return nonNullProp(runtime.properties, 'majorVersions').sort((rt1, rt2) => {
+                if (rt1.isDefault) {
+                    return -1;
+                } else if (rt2.isDefault) {
```
EricJizbaMSFT: eh I'm not sure I trust their default. I think I'd prefer to just sort by latest on top

appservice/src/createAppService/SiteRuntimeStep.ts
```diff
@@ -79,19 +79,36 @@ export class SiteRuntimeStep extends AzureWizardPromptStep<IAppServiceWizardCont
...
+                    return 1;
+                } else {
+                    // runtimVersion comes in the format RUNTIME|X.X
+                    const runtimeRegExp: RegExp = /.*(?=\|)+./g;
+                    const v1: number = parseFloat(nonNullProp(rt1, 'runtimeVersion').replace(runtimeRegExp, ''));
+                    const v2: number = parseFloat(nonNullProp(rt2, 'runtimeVersion').replace(runtimeRegExp, ''));
+                    return v2 - v1;
```
EricJizbaMSFT: I think this logic treats "1.1" and "1.10" as the same thing, but they're different. You could either use the 'semver' node module or just check the major and minor version separately

appservice/src/createAppService/SiteRuntimeStep.ts
```diff
@@ -79,19 +79,36 @@ export class SiteRuntimeStep extends AzureWizardPromptStep<IAppServiceWizardCont
...
 
-    private sortQuickPicksByRuntime(runtimeItems: IAzureQuickPickItem<ApplicationStack>[], recommendedRuntimes: string[]): IAzureQuickPickItem<ApplicationStack>[] {
-        function getPriority(item: IAzureQuickPickItem<ApplicationStack>): number {
-
-            const index: number = recommendedRuntimes.findIndex((runtime: string) => nonNullProp(item.data, 'name').includes(runtime));
-            return index === -1 ? recommendedRuntimes.length : index;
+    private sortQuickPicksByRuntime(runtimeQuickPicks: IAzureQuickPickItem<ApplicationStack>[], runtimeRecommendations: string[]): IAzureQuickPickItem<ApplicationStack>[] {
+        const recommendedQuickPicks: IAzureQuickPickItem<ApplicationStack>[] = [];
+        // go backwards so that when runtime is unshifted to front, the order is unaffected
+        for (let i: number = runtimeQuickPicks.length - 1; i >= 0; i = i - 1) {
```
EricJizbaMSFT: So this method is called `sortQuickPicksByRuntime`. I guess I'd prefer if all the sorting logic was in this one function rather than having half of it in `getLinuxRuntimeStack`. Also I prefer using the `.sort` method instead of this for/unshift/concat logic because it makes me nervous about an off-by-one error

nturinski: I think I'd rather rename this to `shiftRecommendedToTop` as there is no actual sorting going on here.

I don't want to do actual sorting here because it's a lot easier to sort by runtime before I've flattened the array (when I make the API call).  

Since we're traversing the array backward, I think even if we splice out an element, we shouldn't have any indexing errors, but I can find another method.

Maybe I wasn't understanding the `sort` method correctly, but when it was moving recommended runtimes to the top, it was swapping the recommended runtime's index with whatever was previous at the top which is why sometimes the versions would be out of order or would get sandwiched in between other runtimes.

I was checking if either item was the recommended runtime and returned -1 or 1 appropriately, but if not, returned 0.  That should have caused it to keep its position, but since it's getting swapped with the recommended runtime, that position was incorrect.

EricJizbaMSFT: I'm more okay with this if you change the name. However in an ideal world this all has unit-tests and that would be easier to do if it was one function.

You can definitely do all of this inside one sort function. I don't think the behavior is guaranteed when you return 0, which is why it would have weird results. I think this sort logic would work:
```
const aInfo = getRuntimeInfo(a);
const bInfo = getRuntimeInfo(b);
if (aInfo.name !== bInfo.name) {
    return getPriority(a) - getPriority(b);
} else if (aInfo.major !== bInfo.major) {
    return aInfo.major - bInfo.major;
} else {
    return aInfo.minor - bInfo.minor;
}

function getRuntimeInfo() {
// use a regexp to get name, major, and minor from node|8.0
}
```

appservice/src/createAppService/SiteRuntimeStep.ts
nturinski: **sigh**, I just noticed, but Java SE is listed twice because the display name for Java 8 and Java 11 are the same from the API.

I plan to add a description, either `java8` or `java11` to the runtimes that this is relevant too.  Is that an alright solution?

nturinski: image

EricJizbaMSFT: I think ideally we only show the description when necessary (seems like java runtimes are the only ones that need it). But I'm fine with that screenshot

nturinski:
Closing in favor of #486

PR #482, opened 2019-04-18
Remove all usage of /m flag

nturinski:
This is a hotfix for these issues: 

https://github.com/Microsoft/vscode-azureappservice/issues/942
https://github.com/Microsoft/vscode-azurearmtools/issues/206

I wanted to leave them commented in so it'd be easy to reimplement later, but I'm opened to just removing the code.

It's based off this release https://github.com/Microsoft/vscode-azuretools/commit/e727f1662f325b2065813486f668d693e1bef256 which both Azure App Service and ARM are using.

### DIFF
--- a/ui/src/parseError.ts
+++ b/ui/src/parseError.ts
@@ -3,8 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import * as htmlToText from 'html-to-text';
-import * as os from 'os';
+import * as htmlToText from 'html-to-text'
 import { IParsedError } from '../index';
 import { localize } from './localize';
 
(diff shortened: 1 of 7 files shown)


ui/src/parseError.ts
```diff
@@ -155,33 +157,34 @@ function unpackErrorFromField(error: any, prop: string): any {
     return error;
 }
 
-function getCallstack(error: { stack?: string }): string | undefined {
-    // tslint:disable-next-line: strict-boolean-expressions
-    const stack: string = error.stack || '';
+// temporarily disable for extension activation hotfix
```
StephenWeatherford: nit: It's better to use /* */ comments so you get fewer changes

nturinski:
As per PR advice, closing this in favor of rel branch https://github.com/Microsoft/vscode-azuretools/tree/rel/0.20.5

PR #486, opened 2019-04-23
Sort linux runtimes by version

EricJizbaMSFT:
A few other benefits:
1. Java picks display stack as description since label is the same
1. Pass promise to `ext.ui.showQuickPick` so that we get that loading thing
1. Added tests

(In this example, I had a python project open so that's why Python is on top)
Screen Shot 2019-04-22 at 7 52 52 PM

Supersedes https://github.com/Microsoft/vscode-azuretools/pull/479 - offered to take over from Nathan because he was sick

### DIFF
--- a/appservice/src/createAppService/SiteRuntimeStep.ts
+++ b/appservice/src/createAppService/SiteRuntimeStep.ts
@@ -12,7 +12,7 @@ import { ext } from '../extensionVariables';
 import { localize } from '../localize';
 import { signRequest } from '../signRequest';
 import { nonNullProp } from '../utils/nonNull';
-import { AppKind, WebsiteOS } from './AppKind';
+import { AppKind, LinuxRuntimes, WebsiteOS } from './AppKind';
 import { IAppServiceWizardContext } from './IAppServiceWizardContext';
 
 type ApplicationStackJsonResponse = {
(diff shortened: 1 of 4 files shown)


appservice/src/createAppService/SiteRuntimeStep.ts
```diff
@@ -75,23 +63,90 @@ export class SiteRuntimeStep extends AzureWizardPromptStep<IAppServiceWizardCont
...
-            return index === -1 ? recommendedRuntimes.length : index;
+function convertToNumber(data: string | undefined): number {
+    if (data) {
+        const result: number = parseInt(data);
+        if (!isNaN(result)) {
+            return result;
         }
-        return runtimeItems.sort((a: IAzureQuickPickItem<ApplicationStack>, b: IAzureQuickPickItem<ApplicationStack>) => getPriority(a) - getPriority(b));
     }
+    return 0;
```
nturinski: Rather than return 0, I feel like we should return a higher number so that LTS gets displayed at the top.  I think it's what the Portal is trying to lead people to at least.

appservice/src/createAppService/SiteRuntimeStep.ts
```diff
@@ -75,23 +63,90 @@ export class SiteRuntimeStep extends AzureWizardPromptStep<IAppServiceWizardCont
...
+        })
+        // convert to quick pick
+        .map(mv => {
+            return {
+                id: mv.runtimeVersion,
+                label: mv.displayVersion,
+                data: mv.runtimeVersion,
+                // include stack as description if it has a version
+                // tslint:disable-next-line: strict-boolean-expressions
+                description: /[0-9]/.test(mv.stackDisplay) ? mv.stackDisplay : undefined
```
nturinski: Clever :) I like this solution (I was just gonna special case Java)

appservice/src/createAppService/SiteRuntimeStep.ts
EricJizbaMSFT: Fine with me

PR #488, opened 2019-04-25
Fix report an issue URLs

StephenWeatherford:
Two issues:
1) URL can be too long for the browser to handle if stack is long
2) Temporarily use opn because vscode.open removes newlines and other characters (Microsoft/vscode#25852)

### DIFF
--- a/ui/src/reportAnIssue.ts
+++ b/ui/src/reportAnIssue.ts
@@ -3,26 +3,58 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import * as assert from 'assert';
 import * as vscode from 'vscode';
 import { IParsedError } from '../index';
 import { getPackageInfo } from "./getPackageInfo";
-import { limitLines, numberOfLines } from './utils/limitLines';
+import { parseError } from './parseError';
 import { openUrl } from './utils/openUrl';
+import { countLines, limitLines } from './utils/textStrings';
+
+// Some browsers don't have very long URLs
+// 2000 seems a reasonable number for most browsers,
+// see https://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers
+export const maxUrlLength: number = 2000;
 
 /**
  * Used to open the browser to the "New Issue" page on GitHub with relevant context pre-filled in the issue body
  */
 export async function reportAnIssue(actionId: string, parsedError: IParsedError): Promise<void> {
+    const link: string = getReportAnIssueLink(actionId, parsedError);
+    await openUrl(link);
+}
+
+export function getReportAnIssueLink(actionId: string, parsedError: IParsedError): string {
     const { extensionName, extensionVersion, bugsUrl } = getPackageInfo();
 
-    // Some browsers don't have very long URLs
-    // tslint:disable-next-line: typedef
-    const maxUrlLength = 2000;
-
     // tslint:disable-next-line: strict-boolean-expressions
-    const stack: string = parsedError.stack || '';
+    const stack: string = (parsedError.stack || '').replace(/\r\n/g, '\n');
 
-    const body: string = `
+    // Try repeatedly with fewer lines of stack until we have a URL under the limit
+    for (let stackLines: number = countLines(stack); stackLines >= 0; stackLines -= 1) {
+        const url: string = createLink(stackLines, parsedError.message);
+        if (url.length <= maxUrlLength) {
+            return url;
+        }
+    }
+
+    // If it's still too long, shorten the message;
+    const urlWithNoStack: string = createLink(0, parsedError.message);
+    const ellipses: string = '...';
+    const reduceByChars: number = urlWithNoStack.length - maxUrlLength + ellipses.length;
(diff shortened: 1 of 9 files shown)


ui/src/reportAnIssue.ts
```diff
@@ -33,22 +41,30 @@ Product: ${vscode.env.appName}
 Product Version: ${vscode.version}
 Language: ${vscode.env.language}`;
 
-    if (parsedError.stack) {
-        body = body.concat(`
+    // Try repeatedly with fewer lines of stack until we have a URL of a reasonable length
+    // tslint:disable-next-line: no-increment-decrement
+    for (let stackLines: number = numberOfLines(stack); stackLines >= 0; --stackLines) {
+        let bodyWithStack: string = body;
+        if (stack) {
```
EricJizbaMSFT:
```
if (stack && stackLines > 0) {
```

No need to add extra characters if we can avoid it

ui/src/reportAnIssue.ts
```diff
@@ -33,22 +41,30 @@ Product: ${vscode.env.appName}
 Product Version: ${vscode.version}
 Language: ${vscode.env.language}`;
 
-    if (parsedError.stack) {
-        body = body.concat(`
+    // Try repeatedly with fewer lines of stack until we have a URL of a reasonable length
+    // tslint:disable-next-line: no-increment-decrement
```
EricJizbaMSFT: Rather than disable the rule, the [suggested fix](https://palantir.github.io/tslint/rules/increment-decrement/) is super simple: `stackLines -= 1`

ui/src/reportAnIssue.ts
```diff
@@ -14,7 +15,14 @@ import { openUrl } from './utils/openUrl';
 export async function reportAnIssue(actionId: string, parsedError: IParsedError): Promise<void> {
     const { extensionName, extensionVersion, bugsUrl } = getPackageInfo();
 
-    let body: string = `
+    // Some browsers don't have very long URLs
+    // tslint:disable-next-line: typedef
```
EricJizbaMSFT: Why disable this rule? Just add `: number`

ui/src/reportAnIssue.ts
```diff
@@ -14,7 +15,14 @@ import { openUrl } from './utils/openUrl';
 export async function reportAnIssue(actionId: string, parsedError: IParsedError): Promise<void> {
     const { extensionName, extensionVersion, bugsUrl } = getPackageInfo();
 
-    let body: string = `
+    // Some browsers don't have very long URLs
```
EricJizbaMSFT: Not sure if you randomly picked 2000, but seems like the actual limit in several browsers. Suggest adding a link like this in the comment:

https://blogs.msdn.microsoft.com/ieinternals/2014/08/13/url-length-limits/

or

https://support.microsoft.com/help/208427/maximum-url-length-is-2-083-characters-in-internet-explorer

ui/src/reportAnIssue.ts
```diff
@@ -33,22 +41,30 @@ Product: ${vscode.env.appName}
...
-    }
 
-    const baseUrl: string = bugsUrl || `https://github.com/Microsoft/${extensionName}/issues`;
-    const url: string = `${baseUrl}/new?body=${encodeURIComponent(body)}`;
+            const baseUrl: string = bugsUrl || `https://github.com/Microsoft/${extensionName}/issues`;
+            const url: string = `${baseUrl}/new?body=${encodeURIComponent(bodyWithStack)}`;
 
-    // tslint:disable-next-line:no-floating-promises
-    await openUrl(url);
+            if (url.length < maxUrlLength || stackLines === 0) {
```
EricJizbaMSFT: Not gonna lie - makes me nervous having this kind of code in something so common as "Report an Issue" without unit tests. For example, the risks include opening a bunch of links if we forget a `return` or opening zero links if `url.length < maxUrlLength || stackLines === 0` has any problems.

Perhaps you move `await openUrl(url);` outside of the loop and create a function called `addStackToBody` with a few tests?

ui/src/reportAnIssue.ts
nturinski: Maybe from this?

https://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers

ui/src/reportAnIssue.ts
StephenWeatherford: Ah, I had intended to write unit tests, just forgot.

ui/src/reportAnIssue.ts
StephenWeatherford: Nathan's was the one, but it all seemed to vague I hesitated to include a link.  will do so.

ui/src/reportAnIssue.ts
StephenWeatherford: Stupid rules.

ui/src/reportAnIssue.ts
StephenWeatherford: if stack = "", stack is falsy

EricJizbaMSFT: No I'm talking about the case where the body by itself is close to 2000 characters. Aka the stack has contents and you've already tried a bunch of times to add some of it's lines

PR #489, opened 2019-04-25
Remove setWizardDefaults and move to Azure App Extension

nturinski:
Related to https://github.com/Microsoft/vscode-azureappservice/issues/780

There was a lot of custom logic in creating a web app that was specific to app service.  We should leverage the createOptions object and base our defaults on that


EricJizbaMSFT:
I actually think all of `createWebApp` should be moved to the app service repo. Everything except for the steps themselves.

Also - are we completely done with testing for the build release of app service? I'm not too worried about functions because it doesn't use this code, but I want to be absolutely sure we don't plan any more changes for app service before we take a refactor like this.

nturinski:
Yeah, I think the only thing I want to include that hasn't been tested is the sort [Linux runtimes PR](https://github.com/Microsoft/vscode-azuretools/pull/486) that you handled (thanks again😃).

I think it makes sense to release a 0.37.2 for `appservice` that includes your host site plan and the runtimes sorting and bump `App Service Extension` to that package, but not pick up these changes.

appservice/src/createAppService/IAppCreateOptions.ts
```diff
@@ -1,11 +1,19 @@
...
- *--------------------------------------------------------------------------------------------*/
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.txt in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import { SkuDescription } from "azure-arm-website/lib/models";
+import { IActionContext } from "vscode-azureextensionui";
+import { LinuxRuntimes } from "./AppKind";
 
 export interface IAppCreateOptions {
```
EricJizbaMSFT: remove this file since it's no longer used in this package

PR #491, opened 2019-04-25
Minify call stack even more

EricJizbaMSFT:
Somewhat tangential to Stephen's work here where the stack is too long: https://github.com/Microsoft/vscode-azuretools/pull/488. I figured another improvement would be just to collect and post less information from the stack.

Check the test cases to see what the call stack would look like. It contains all the info I ever found useful, but of course let me know if it's missing something major. I kind of feel better saying what we _do_ want to include rather than trying to remove things we _don't_ want to include in the stack.

### DIFF
--- a/ui/src/parseError.ts
+++ b/ui/src/parseError.ts
@@ -164,7 +164,10 @@ function getCallstack(error: { stack?: string }): string | undefined {
             let result: string = '';
             // Get just the file name, line number and column number
             const fileMatch: RegExpMatchArray | null = l.match(/[^\/\\\(\s]+\.(t|j)s:[0-9]+:[0-9]+/i);
+
+            // Ignore any lines without a file match (e.g. "at Generator.next (<anonymous>)")
             if (fileMatch) {
+                // Get the function name
                 const functionMatch: RegExpMatchArray | null = l.match(/^[\s]*at ([^\(\\\/]+(?:\\|\/)?)+/i);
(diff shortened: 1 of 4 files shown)


EricJizbaMSFT:
ping @StephenWeatherford updated thoughts on this? Should have the function name now...

ui/src/parseError.ts
```diff
@@ -159,29 +158,40 @@ function getCallstack(error: { stack?: string }): string | undefined {
...
-
-    // Trim each line, including getting rid of 'at'
-    result = result.replace(/^\s*(at\s)?\s*/mg, '');
-    result = result.replace(/\s+$/mg, '');
+    const minifiedLines: (string | undefined)[] = stack
+        .split(/(\r\n|\n)/g) // split by line ending
+        .map(l => {
+            let result: string = '';
+            // Get just the file name, line number and column number
+            const fileMatch: RegExpMatchArray | null = l.match(/[^\/\\\(\s]+\.(t|j)s:[0-9]+:[0-9]+/i);
```
StephenWeatherford: It would be very helpful if you put an example line into the comments

EricJizbaMSFT: Isn't that what unit tests are for? 😂

StephenWeatherford: No.  :-)

StephenWeatherford:
> Apheasement accepted :-)

PR #494, opened 2019-04-30
Try catch around add app setting

nturinski:
Fixes https://github.com/Microsoft/vscode-azureappservice/issues/943


appservice/src/tree/AppSettingsTreeItem.ts
```diff
@@ -109,7 +109,15 @@ export class AppSettingsTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
 
         showCreatingTreeItem(newKey);
         settings.properties[newKey] = newValue;
-        await this.root.client.updateApplicationSettings(settings);
+
+        try {
+            await this.root.client.updateApplicationSettings(settings);
+        } catch (err) {
+            // if this is rejected, we need to clear the property otherwise it will cause every subsequent add to fail
+            delete settings.properties[newKey];
```
EricJizbaMSFT: I'm guessing `editSettingItem` (which is used for "edit" and "rename" of an already existing setting) will have a similar problem

nturinski: Good point

nturinski:
Just a note, I'm not very worried about edit value failing since this passes:

image

Also I'm calling `editSettingsValue` first now because if it failed out, the name value would be wrong until refresh (because the reassign happens before the error is thrown)

appservice/src/tree/AppSettingsTreeItem.ts
```diff
@@ -71,14 +71,24 @@ export class AppSettingsTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
     public async editSettingItem(oldKey: string, newKey: string, value: string): Promise<void> {
         const settings: StringDictionary = await this.ensureSettings();
```
EricJizbaMSFT:
```
// clone cached settings into new object
const settings: StringDictionary = { ...await this.ensureSettings() };
```

appservice/src/tree/AppSettingsTreeItem.ts
```diff
@@ -71,14 +71,24 @@ export class AppSettingsTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
...
+
+        try {
+            await this.root.client.updateApplicationSettings(settings);
+        } catch (err) {
+            // if this was a rename, delete the invalid key and readd the old key
+            renameHelper(newKey, oldKey);
+            throw err;
         }
 
-        await this.root.client.updateApplicationSettings(settings);
```
EricJizbaMSFT: this._settings = await this.root.client.updateApplicationSettings(settings);

appservice/src/tree/AppSettingsTreeItem.ts
nturinski: This is a great idea, but apparently only makes a shallow copy so anytime properties were referenced, it was still referring to the `this._settings.properties`.  I used the JSON method to make a deep-copy

Ref: https://scotch.io/bar-talk/copying-objects-in-javascript

appservice/src/tree/AppSettingsTreeItem.ts
```diff
@@ -69,16 +69,17 @@ export class AppSettingsTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
...
                 delete settings.properties[oldKey];
             }
             settings.properties[newKey] = value;
         }
 
-        await this.root.client.updateApplicationSettings(settings);
+        this._settings = await this.root.client.updateApplicationSettings(settings);
     }
 
     public async deleteSettingItem(key: string): Promise<void> {
```
EricJizbaMSFT: Seems like you might as well do the same thing for delete, too

PR #497, opened 2019-05-07
Refactor javaUtils into Azure App Service extension

nturinski:
Related to https://github.com/Microsoft/vscode-azureappservice/issues/979


appservice/package.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "vscode-azureappservice",
     "author": "Microsoft Corporation",
-    "version": "0.38.0",
+    "version": "0.38.1",
```
EricJizbaMSFT: Removing javaUtils is a breaking change

PR #498, opened 2019-05-10
Allow adding additional telemetry info

StephenWeatherford:
My use case is this in ARM:

    ext.reporter = createTelemetryReporter(
        context,
        {
            autoDetectJsonTemplates: String(vscode.workspace.getConfiguration('armTools').get<boolean>(configKeys.autoDetectJsonTemplates))
        });

The given config key is useful to know, but it affects whether we activate functionality on opening JSON files, so if it's false, no telemetry would normally be sent, and we don't want to send an event on every JSON file opened.


StephenWeatherford:
Yep, that makes more sense

PR #499, opened 2019-05-10
 Make the tree more friendly to non-Azure scenarios

EricJizbaMSFT:
This refactor was done to enable two main scenarios - both of which display non-Azure tree items:
1. The local project tree view for the functions extension (https://github.com/microsoft/vscode-azurefunctions/issues/1163)
1. Switch Docker over to the shared tree code

The main changes were:
1. Move all subscription related code out of the TreeDataProvider and into it's own tree item (`AzureAccountTreeItem`). In most cases, the `AzureAccountTreeItem` will be passed in to the tree as the root tree item, but in Docker's case it will be a child tree item of the `Registries` node.
1. Provide a more generic version of `AzureTreeItem` that doesn't have any Azure-specific requirements. I thought of a lot of different names (none of which seemed great), but ended up with `AzExtTreeItem`. The intention was to imply that the tree item is for an **Az**ure **Ext**ension, but not necessarily actually an Azure resource. This also seemed consistent with the name of the npm package.

In terms of review, I suggest focusing on `index.d.ts` and `README.md` (there should be no real functional change with this PR). You can also look as [this commit](https://github.com/microsoft/vscode-azuretools/commit/c800cf7541cb472529854efbcf54cbb04a103b24) in particular since the first two commits were just renames and copying.


ui/README.md
```diff
@@ -14,7 +14,7 @@ This package provides common Azure UI elements for VS Code extensions:
 Use `registerCommand`, `registerEvent`, or ` callWithTelemetryAndErrorHandling` to consistently display error messages and track commands with telemetry. You must call `registerUIExtensionVariables` first in your extension's `activate()` method. The simplest example is to register a command (in this case, refreshing a node):
 ```typescript
 registerUIExtensionVariables(...);
-registerCommand('yourExtension.Refresh', (node: AzureTreeItem) => { node.refresh(); });
+registerCommand('yourExtension.Refresh', (node: AzExtTreeItem) => { node.refresh(); });
```
StephenWeatherford: I guess another name possibility would be UiTreeItem, but I'm fine with this.

ui/src/treeDataProvider/AzureAccountTreeItem.ts
```diff
@@ -0,0 +1,171 @@
...
+    public readonly label: string = 'Azure';
+    public readonly childTypeLabel: string = localize('subscription', 'subscription');
+    public autoSelectInTreeItemPicker: boolean = true;
+    public disposables: Disposable[] = [];
+
+    private _subscriptionTreeItemType: SubscriptionTreeItemType;
+    private _azureAccount: AzureAccount;
+    private _subscriptionTreeItems: SubscriptionTreeItem[] | undefined;
+
+    constructor(parent: AzExtParentTreeItem | undefined, subscriptionTreeItemType: SubscriptionTreeItemType, testAccount?: TestAzureAccount) {
```
StephenWeatherford: In the readme you refer to passing in a "subscriptionTreeItem" (here the param is named subscriptionTreeItemType)

ui/README.md
```diff
@@ -88,25 +88,35 @@ Follow these steps to create your basic Azure Tree:
         }
     }
     ```
-1. Instantiate a new instance of `AzureTreeDataProvider` in your extension's `activate()` method, passing the `SubscriptionTreeItem` type and `loadMoreCommandId`. The `loadMoreCommandId` maps the 'Load More...' node to the command registered by your extension.
+1. Instantiate an `AzureAccountTreeItem` in your extension's `activate()` method. In most cases this will be the root of your tree, so you should pass in `undefined` as the parent and your `SubscriptionTreeItem` type to provide children.
```
StephenWeatherford: I find the SubscriptionTreeItem/SubscriptionTreeItem naming confusing.  I'd call it a constructor function (http://www.javascriptkit.com/javatutors/oopjs2.shtml).

ui/index.d.ts
```diff
@@ -198,42 +180,49 @@ export interface IGenericTreeItemOptions {
     iconPath?: string | Uri | { light: string | Uri; dark: string | Uri };
     commandId?: string;
     contextValue: string;
+
+    /**
+     * If true, the tree item picker will execute `commandId`, refresh the tree, and re-prompt at the same level of the tree.
+     * If `commandId` is not defined, it will throw an error.
+     */
+    includeInTreeItemPicker?: boolean;
```
StephenWeatherford: I don't understand the comments here.  In what scenario would you execute a command or refresh?

EricJizbaMSFT: So far this is used for "Sign in to Azure...", "Create an Account...", "Select Subscriptions...", and "Attach a Database Account..."

All of those actions will basically "populate" this level of the tree, which is why we execute the command and then refresh that level of the tree.

EricJizbaMSFT: Added a note

EricJizbaMSFT:
Refactored the naming a bit based on your confusion. Instead of each extension doing something like this:
```
export class WebAppProvider extends SubscriptionTreeItem {
```
They can now do this:
```
export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
```

The account tree item will follow the same "Base" pattern. Instead of passing `subscriptionTreeItemType` in to the constructor, they just have to implement `createSubscriptionTreeItem` (which is also better because it allows for async code if necessary).

ui/index.d.ts
```diff
@@ -243,48 +233,123 @@ export declare abstract class AzureParentTreeItem<TRoot = ISubscriptionRoot> ext
...
+    createChild<T extends AzExtTreeItem>(userOptions?: any): Promise<T>;
+
+    /**
+     * Get the currently cached children for this tree item. This will load the first batch if they have not been loaded yet.
+     */
+    getCachedChildren(): Promise<AzExtTreeItem[]>;
+}
+
+/**
+ * A tree time for an Azure Account, which will display subscriptions. In most cases, this will be the root of an extension's tree.
```
StephenWeatherford: suggest: For Azure-centered extensions, this will be at the root of the tree.

ui/src/treeDataProvider/AzExtParentTreeItem.ts
```diff
@@ -0,0 +1,324 @@
...
+
+    public get creatingTreeItems(): AzExtTreeItem[] {
+        return this._creatingTreeItems;
+    }
+
+    //#region Methods implemented by base class
+    public abstract loadMoreChildrenImpl(clearCache: boolean): Promise<AzExtTreeItem[]>;
+    public abstract hasMoreChildrenImpl(): boolean;
+    // tslint:disable-next-line:no-any
+    public createChildImpl?(showCreatingTreeItem: (label: string) => void, userOptions?: any): Promise<AzExtTreeItem>;
```
StephenWeatherford: Can we switch to unknown?

ui/src/treeDataProvider/AzExtParentTreeItem.ts
```diff
@@ -0,0 +1,324 @@
...
+                            iconPath: loadingIconPath
+                        });
+                        this._creatingTreeItems.push(creatingTreeItem);
+                        this.treeDataProvider.refreshUIOnly(this);
+                    },
+                    userOptions);
+
+                this.addChildToCache(newTreeItem);
+                this.treeDataProvider._onTreeItemCreateEmitter.fire(newTreeItem);
+                // tslint:disable-next-line: no-any
```
StephenWeatherford: unknown

ui/src/treeDataProvider/AzExtParentTreeItem.ts
```diff
@@ -0,0 +1,324 @@
...
+class InvalidTreeItem extends AzExtParentTreeItem {
+    public readonly contextValue: string;
+    public readonly label: string;
+    public readonly description: string;
+
+    // tslint:disable-next-line:no-any
+    private _error: any;
+
+    // tslint:disable-next-line:no-any
+    constructor(parent: AzExtParentTreeItem, label: string, error: any, contextValue: string, description: string = localize('invalid', 'Invalid')) {
```
StephenWeatherford: unknown

StephenWeatherford: etc

ui/src/treeDataProvider/AzExtParentTreeItem.ts
EricJizbaMSFT: That would require updating the version of typescript to at least 3.0.0. I can do that in a separate PR

PR #501, opened 2019-05-14
Update some dev dependencies and start using "unknown"

EricJizbaMSFT:
Per Mr. Stephen (https://github.com/microsoft/vscode-azuretools/pull/499#discussion_r283599551) we can start using `unknown` instead of `any` as long as we update to TypeScript 3.0+


nturinski:
I'm scared of the unknown 😨

PR #503, opened 2019-05-15
Go all-in on `IActionContext`

### LINKED ISSUE #120: showNodePicker needs to pass in userOptions to createChild
When people implement createChild, they can optionally specify an additional 'userOptions' [argument](https://github.com/Microsoft/vscode-azuretools/blob/master/ui/index.d.ts#L115). However, the nodePicker never passes anything to createChild. We have to figure out some way to pass along the userOptions

EricJizbaMSFT:
This work was inspired because I wanted to fix #120 (which has been around for _a while_). IMO the best fix was to pass a "context" object all the way down through the tree item picker, but this made me realize I'd have to pass that object in _a lot_ of places. Instead of using `userOptions` (like we had in `createChildImpl` before), I decided to go all-in on actionContext and make that the default "context" object that we pass around in this entire package. Now wizards, the tree item picker, and any command/event will leverage actionContext as their base "context" object (which can obviously be extended as desired).

Since I was using `actionContext` so much, I decided to make it the first parameter for any command, rather than using `this`. The only reason I originally used `this` was because people could essentially ignore the `this` parameter if they didn't want to use actionContext. However, that was always confusing and led to things like `const me = this;`.

### DIFF
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -147,7 +147,7 @@ export declare abstract class AzExtTreeItem {
     /**
      * Implement this to support the 'delete' action in the tree. Should not be called directly
      */
-    public deleteTreeItemImpl?(): Promise<void>;
+    public deleteTreeItemImpl?(context: IActionContext): Promise<void>;
 
     /**
      * Implement this to execute any async code when this node is refreshed. Should not be called directly
@@ -169,7 +169,7 @@ export declare abstract class AzExtTreeItem {
     /**
      * This class wraps deleteTreeItemImpl and ensures the tree is updated correctly when an item is deleted
      */
-    public deleteTreeItem(): Promise<void>;
+    public deleteTreeItem(context: IActionContext): Promise<void>;
 
     /**
      * Displays a 'Loading...' icon and temporarily changes the item's description while `callback` is being run
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -406,7 +406,7 @@ export declare function registerCommand(commandId: string, callback: (context: I
 
 /**
  * Used to register VSCode events. It wraps your callback with consistent error and telemetry handling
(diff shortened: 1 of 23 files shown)


ui/index.d.ts
```diff
@@ -666,18 +685,18 @@ export interface IWizardOptions<T> {
...
     public constructor(wizardContext: T, options: IWizardOptions<T>);
 
-    public prompt(actionContext: IActionContext): Promise<void>;
-    public execute(actionContext: IActionContext): Promise<void>;
+    public prompt(): Promise<void>;
+    public execute(): Promise<void>;
 }
 
-export declare abstract class AzureWizardExecuteStep<T extends {}> {
+export declare abstract class AzureWizardExecuteStep<T extends IActionContext> {
```
StephenWeatherford: Why extend action context rather than having it as a member?

StephenWeatherford:
me like

ui/index.d.ts
EricJizbaMSFT: It just felt simpler to settle on one context object as the base for all other context objects. Perhaps you made this comment because something like this wouldn't work?:
```typescript
const wizardContext = { ...actionContext, wizardProp = 'exampleProp' };

// This will work since `properties` is a reference to the `properties` object on `actionContext`
wizardContext.properties.exampleProp = 'exampleValue';

// this will be ignored since `suppressErrorDisplay` is not changed on the original `actionContext`
// If we had used `Object.assign` above, it would've worked. But I think people prefer spread operator over Object.assign
wizardContext.suppressErrorDisplay = true;
```

I just refactored `IActionContext` to hopefully alleviate that concern. This is what it'll look like now:
```typescript
const wizardContext = { ...actionContext };
wizardContext.telemetry.properties.exampleProp = 'exampleValue';
wizardContext.errorHandling.suppressDisplay = true;
```
(I also just like `.telemetry.properties` because it makes it more obvious what those properties are for)

StephenWeatherford: More wordy, but agreed better structured/behaved.

PR #506, opened 2019-05-17
Remove free tier and add it to common

nturinski:
Related to https://github.com/microsoft/vscode-azureappservice/issues/992

Verified it works! 🎉 
image


nturinski:
This seems to indicate that it'll with almost every region except for some government accounts and South Africa.

PR #508, opened 2019-05-21
Add accountStatus back to telemetry

EricJizbaMSFT:
`accountStatus` was removed from telemetry in https://github.com/microsoft/vscode-azuretools/pull/499 because I didn't have access to `actionContext` in the new `AzureAccountTreeItemBase.loadMoreChildrenImpl`, but that's now possible as of https://github.com/microsoft/vscode-azuretools/pull/503.

Also removed some default empty strings in telemetry. They're already marked as optional in `index.d.ts` and it makes things just a touch cleaner.

Old stuff:
> ** TELEMETRY("vscode-cosmosdb/AzureTreeDataProvider.getChildren", 0.10.2) properties={"isActivationEvent":"true","cancelStep":"","result":"Succeeded","stack":"","error":"","errorMessage":"","contextValue":"azureextensionui.azureAccount","accountStatus":"LoggedIn","hasMoreChildren":"false"}, measures={"duration":0.004,"childCount":3}

New stuff:
> ** TELEMETRY("vscode-cosmosdb/AzureTreeDataProvider.getChildren", 0.10.2) properties={"result":"Succeeded","isActivationEvent":"true","contextValue":"azureextensionui.azureAccount","accountStatus":"LoggedIn","hasMoreChildren":"false"}, measures={"childCount":3,"duration":0.005}

### DIFF
--- a/ui/src/treeDataProvider/AzureAccountTreeItemBase.ts
+++ b/ui/src/treeDataProvider/AzureAccountTreeItemBase.ts
@@ -69,7 +69,8 @@ export abstract class AzureAccountTreeItemBase extends AzExtParentTreeItem imple
         return false;
     }
 
-    public async loadMoreChildrenImpl(_clearCache: boolean): Promise<AzExtTreeItem[]> {
+    public async loadMoreChildrenImpl(_clearCache: boolean, context: types.IActionContext): Promise<AzExtTreeItem[]> {
+        context.telemetry.properties.accountStatus = this._azureAccount.status;
         const existingSubscriptions: SubscriptionTreeItemBase[] = this._subscriptionTreeItems ? this._subscriptionTreeItems : [];
         this._subscriptionTreeItems = [];
 


ui/src/callWithTelemetryAndErrorHandling.ts
```diff
@@ -20,15 +20,9 @@ function initContext(): [number, IActionContext] {
     const context: IActionContext = {
         telemetry: {
             properties: {
-                isActivationEvent: 'false',
-                cancelStep: '',
-                result: 'Succeeded',
-                stack: '',
-                error: '',
-                errorMessage: ''
+                result: 'Succeeded'
```
StephenWeatherford: Removing this does have one disadvantage.  If we don't send this by default, we won't be able to register them until they're used.  I.e., we won't get actual stacks, error messages or cancel steps coming through our telemetry for each event until a) a user actually hits an error, and b) we categorize it after noticing it's missing and then wait for another hit.

Maybe we should add a new env var that's turned on when running tests which causes these to be set to empty strings only then.  I've never liked sending these in all events, but I don't like not registering everything up front, either.

EricJizbaMSFT: > I've never liked sending these in all events, but I don't like not registering everything up front, either.

I guess I don't have a strong preference. Want me to revert this part of the PR for now?

EricJizbaMSFT:
@StephenWeatherford reduced this PR to just adding back accountStatus to telemetry. The other stuff as discussed offline was taking me too long, but I might get back to it eventually

PR #509, opened 2019-05-22
Suppress error display for cancels

nturinski:
Related https://github.com/microsoft/vscode-azureappservice/issues/898

### DIFF
--- a/appservice/src/deploy/localGitDeploy.ts
+++ b/appservice/src/deploy/localGitDeploy.ts
@@ -62,6 +62,7 @@ export async function localGitDeploy(client: SiteClient, fsPath: string, context
                         await tryPushAndWaitForDeploymentToComplete(true);
                     } else {
                         context.telemetry.properties.cancelStep = 'forcePush';
+                        context.errorHandling.suppressDisplay = true;
                         throw new UserCancelledError();
                     }
                 } else {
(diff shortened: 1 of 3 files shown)


appservice/src/deploy/localGitDeploy.ts
```diff
@@ -62,6 +62,7 @@ export async function localGitDeploy(client: SiteClient, fsPath: string, context
                         await tryPushAndWaitForDeploymentToComplete(true);
                     } else {
                         context.telemetry.properties.cancelStep = 'forcePush';
+                        context.errorHandling.suppressDisplay = true;
                         throw new UserCancelledError();
```
StephenWeatherford: I thought we automatically suppressed display for any UserCancelledError…?

EricJizbaMSFT: Yeah Stephen's right - I think the bug is here: https://github.com/microsoft/vscode-azuretools/blob/68d1dc9591575be239d9f0fecbada370408b4a98/appservice/src/utils/callWithMaskHandling.ts#L13

nturinski: What would be the actual solution here?  I was thinking about calling callWithTelemetryandError around callWithMaskHandling, but it'd duplicate the event in telemetry.

EricJizbaMSFT: Inside of `callWithMaskHandling` you need to check if the error is a UserCancelledError and then just throw that original error instead of trying to mask the message.

The problem is you're converting a UserCancelledError into a normal Error, which we don't handle the same way.

Issue #510, opened 2019-05-23
Switch over to azure-sdk-for-js


PrashanthCorp:
- [ ] UI (Prashanth)
- [ ]  appservice
- [ ]  kudu

formulahendry:
+1
Want to know the plan of switching over to azure-sdk-to-js

EricJizbaMSFT:
Currently blocked on https://github.com/microsoft/vscode-azure-account/issues/140

bwateratmsft:
This should now be unblocked; https://github.com/microsoft/vscode-azure-account/issues/140 has been "fixed" (it now surfaces both the old and new SDKs at least until people have enough time to move off the old).

EricJizbaMSFT:
@bwateratmsft yeah we really want to move to the new sdks, too 🙂 . Was planning on working on this in the next few weeks

bwateratmsft:
@EricJizbaMSFT That would be awesome. I'm going to try to convince the powers-that-be to let me work on this for Docker as well; I may be able to help here too (no promises!)

PR #511, opened 2019-05-23
ManagementPlane API switch

PrashanthCorp:
Addresses #510  for the UI part of the code base.

### DIFF
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -3,11 +3,12 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { ResourceGroup } from 'azure-arm-resource/lib/resource/models';
-import { Location } from 'azure-arm-resource/lib/subscription/models';
-import { StorageAccount } from 'azure-arm-storage/lib/models';
-import { ServiceClientCredentials } from 'ms-rest';
-import { AzureEnvironment, AzureServiceClientOptions } from 'ms-rest-azure';
+import { ResourceGroup } from '@azure/arm-resources/esm/models';
+import { Location } from '@azure/arm-subscriptions/esm/models';
+import { StorageAccount } from '@azure/arm-storage/esm/models';
+import { ServiceClientCredentials } from '@azure/ms-rest-js';
+import { Environment, EnvironmentParameters  } from "@azure/ms-rest-azure-env";
+import { AzureServiceClientOptions } from '@azure/ms-rest-azure-js/lib/azureServiceClient';
 import { Disposable, Event, ExtensionContext, InputBoxOptions, Memento, MessageItem, MessageOptions, OpenDialogOptions, OutputChannel, QuickPickItem, QuickPickOptions, TextDocument, TreeDataProvider, TreeItem, Uri, QuickPick, InputBox, Progress } from 'vscode';
 import { AzureExtensionApi, AzureExtensionApiProvider } from './api';
 
@@ -98,7 +99,7 @@ export interface ISubscriptionRoot {
     subscriptionPath: string;
     tenantId: string;
     userId: string;
-    environment: AzureEnvironment;
+    environment: Environment;
 }
 
 /**
@@ -721,7 +722,7 @@ export declare abstract class AzureWizardPromptStep<T extends {}> {
 export interface ISubscriptionWizardContext {
     credentials: ServiceClientCredentials;
     subscriptionId: string;
-    environment: AzureEnvironment;
+    environment: Environment;
     subscriptionDisplayName: string;
 }
 
@@ -983,10 +984,6 @@ export interface UIExtensionVariables {
     reporter: ITelemetryReporter;
 }
 
-export interface IAddUserAgent {
-    addUserAgentInfo(additionalUserAgentInfo: any): void;
-}
-
 /**
  * Retrieves a user agent string specific to the VS Code extension, of the form `${extensionName}/${extensionVersion}`,
  * and appends it to the given user agent string, if it isn't already in the string. Passing in no existingUserAgent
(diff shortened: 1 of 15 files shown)


ui/index.d.ts
```diff
@@ -3,11 +3,12 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { ResourceGroup } from 'azure-arm-resource/lib/resource/models';
-import { Location } from 'azure-arm-resource/lib/subscription/models';
-import { StorageAccount } from 'azure-arm-storage/lib/models';
-import { ServiceClientCredentials } from 'ms-rest';
-import { AzureEnvironment, AzureServiceClientOptions } from 'ms-rest-azure';
+import { ResourceGroup } from '@azure/arm-resources/esm/models';
```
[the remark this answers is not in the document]
PrashanthCorp: This was what I asked. I felt it weird that I could reference a specific file.

ui/index.d.ts
```diff
@@ -983,39 +984,30 @@ export interface UIExtensionVariables {
...
  * Retrieves a user agent string specific to the VS Code extension, of the form `${extensionName}/${extensionVersion}`,
  * and appends it to the given user agent string, if it isn't already in the string. Passing in no existingUserAgent
  * will return just the extension portion to use in a user agent.
  */
 export declare function appendExtensionUserAgent(existingUserAgent?: string): string;
 
-/**
- * Adds the extension user agent to the given ServiceClient or other object support AddUserAgentInfo
- */
-export declare function addExtensionUserAgent(client: IAddUserAgent): void;
```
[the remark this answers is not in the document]
PrashanthCorp: In the old API, we would create a management client, then add user agent info. 
In the new API, we can provide a method that adds user agent info, which is then invoked during client construction. I don't see method tied to any management client that allows us to add user agent info post-construction. 


As an aside, the method `addUserAgentInfo` has become `addUserAgentHeader`, but I don't think it can be invoked by the client directly.

ui/src/TestAzureAccount.ts
```diff
@@ -2,10 +2,12 @@
...
  *--------------------------------------------------------------------------------------------*/
-import { SubscriptionClient } from 'azure-arm-resource';
-import { SubscriptionListResult, TenantListResult } from 'azure-arm-resource/lib/subscription/models';
-import { ServiceClientCredentials } from 'ms-rest';
-import { ApplicationTokenCredentials, AzureEnvironment, loginWithServicePrincipalSecret } from 'ms-rest-azure';
+import { SubscriptionClient } from '@azure/arm-subscriptions';
+import { SubscriptionListResult, TenantListResult } from '@azure/arm-subscriptions/lib/models';
+import { Environment } from '@azure/ms-rest-azure-env';
+import { ServiceClientCredentials } from '@azure/ms-rest-js';
+import { ApplicationTokenCredentials } from '@azure/ms-rest-nodeauth';
```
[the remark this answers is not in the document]
PrashanthCorp: AzureAccountTreeItem takes a dependency on this file, so merely moving this file to `azureextensiondev`  doesn't really help.

ui/package-lock.json
```diff
@@ -94,7 +178,7 @@
         },
         "ansi-colors": {
             "version": "1.1.0",
-            "resolved": "https://registry.npmjs.org/ansi-colors/-/ansi-colors-1.1.0.tgz",
+            "resolved": "http://registry.npmjs.org/ansi-colors/-/ansi-colors-1.1.0.tgz",
```
EricJizbaMSFT: https/http - you know the drill

ui/src/TestAzureAccount.ts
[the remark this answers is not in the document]
EricJizbaMSFT: I mean yeah you can't literally copy/paste and be done - you'd have to refactor a bit. Never mind - don't worry about it for this PR

ui/index.d.ts
```diff
@@ -1017,39 +1018,30 @@ export interface UIExtensionVariables {
...
 /**
  * Creates an Azure client, ensuring best practices are followed. For example:
  * 1. Adds extension-specific user agent
  * 2. Uses resourceManagerEndpointUrl to support sovereigns
  */
-export function createAzureClient<T extends IAddUserAgent>(
-    clientInfo: { credentials: ServiceClientCredentials; subscriptionId: string; environment: AzureEnvironment; },
-    clientType: new (credentials: ServiceClientCredentials, subscriptionId: string, baseUri?: string, options?: AzureServiceClientOptions) => T): T;
+export function createAzureClient<T>(
+    clientInfo: { credentials: ServiceClientCredentials; subscriptionId: string; environment: EnvironmentParameters; },
```
EricJizbaMSFT: What's the difference between `EnvironmentParameters` and `Environment`? Why doesn't this match the type on `ISubscriptionContext`?

ui/src/azure-account.api.d.ts
```diff
@@ -4,9 +4,9 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { Event } from 'vscode';
-import { ServiceClientCredentials } from 'ms-rest';
-import { AzureEnvironment } from 'ms-rest-azure';
-import { SubscriptionModels } from 'azure-arm-resource';
+import { ServiceClientCredentials } from '@azure/ms-rest-js';
+import { Environment } from '@azure/ms-rest-azure-env/lib/azureEnvironment'
```
EricJizbaMSFT: So this file is copied from here: https://github.com/microsoft/vscode-azure-account/blob/master/src/azure-account.api.d.ts

Are the old types (`AzureEnvironment` and `ServiceClientCredentials`) compatible with the new types (`Environment ` and `ServiceClientCredentials`)? Have we talked to Christof about this at all?

ui/src/wizard/ResourceGroupCreateStep.ts
```diff
@@ -20,7 +20,7 @@ export class ResourceGroupCreateStep<T extends types.IResourceGroupWizardContext
         // tslint:disable-next-line:no-non-null-assertion
         const newLocation: string = wizardContext.location!.name!;
         const resourceClient: ResourceManagementClient = createAzureClient(wizardContext, ResourceManagementClient);
-        const rgExists: boolean = await resourceClient.resourceGroups.checkExistence(newName);
+        const rgExists: boolean = (await resourceClient.resourceGroups.checkExistence(newName)).body;
```
EricJizbaMSFT: What's with the `.body` here?

ui/src/wizard/StorageAccountCreateStep.ts
```diff
@@ -4,7 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 // tslint:disable-next-line:no-require-imports
```
EricJizbaMSFT: nit: remove the tslint disable

ui/src/wizard/ResourceGroupCreateStep.ts
PrashanthCorp: The previous API call returned a Promise<boolean>, the new one returns a Promise<ResourceGroupsCheckExistenceResponse>, which has a body of boolean.

EricJizbaMSFT: That feels like a bug in the sdk

ui/index.d.ts
PrashanthCorp: `EnvironmentParameters` is an interface for all the fields required to invoke the constructor for the `Environment` class. 

The `EnvironmentParameters` interface has just readonly primitive fields - one boolean, everything else string. 
The `Environment` class has methods like add and get, and a few more fields. 

I've aimed to use just the parameters wherever just the type is required, and Environment otherwise.

ui/src/azure-account.api.d.ts
PrashanthCorp: I haven't talked to him. Will talk to him. 

Brings me to the question: Why was this file copied? This adds problems with versioning and documentation (there wasn't any warning in the file for instance)

EricJizbaMSFT: I agree versioning/docs would be ideal, but it's just not a pressing issue for the following reasons:
1. It's a `.d.ts` file, so it's only there for the types. It never gets compiled or run from this repo
1. It hasn't changed since 2017 when it was originally added to this repo
1. It's supposed to be backwards compatible

Also, we intentionally put this file in the shared ui package so that each individual extension would not have to deal with it.

ui/index.d.ts
EricJizbaMSFT: If you want to go even further down that path, you could just use this as the type:
```
    clientInfo: { credentials: ServiceClientCredentials; subscriptionId: string; environment: { resourceManagerEndpointUrl: string }; }
```

ui/src/azure-account.api.d.ts
PrashanthCorp: I had a look at ui/package.json, and there's no reference to the azure-account library. This makes sense to me now. 
As for backwards compatibility it looks like we might have to break it for this change.

PR #513, opened 2019-05-31
Proper fix for the error parsing for handleWithMasking

nturinski:


appservice/src/utils/callWithMaskHandling.ts
```diff
@@ -10,6 +10,11 @@ export async function callWithMaskHandling<T>(callback: () => Promise<T>, valueT
         return await callback();
     } catch (error) {
         const parsedError: IParsedError = parseError(error);
+
+        if (parsedError.isUserCancelledError) {
+            throw parsedError;
```
StephenWeatherford: throw error

PR #515, opened 2019-06-03
Add timestamp to outputChannel with AzureOutputChannel wrapper

nturinski:
I need to use this PR to leverage it in the appservice package that uses `formatDeployLog` before I can add this to the App Service and Functions extension.


ui/index.d.ts
```diff
@@ -317,6 +309,13 @@ export declare abstract class AzureAccountTreeItemBase extends AzExtParentTreeIt
      */
     public constructor(parent?: AzExtParentTreeItem, testAccount?: TestAzureAccount);
 
+    //#region Methods implemented by base class
```
EricJizbaMSFT: I'm find with most of your changes to this file, but this one screwed up the regions

ui/index.d.ts
```diff
@@ -1056,3 +1055,24 @@ export function createAzureSubscriptionClient<T extends IAddUserAgent>(
  * Multiple APIs with different versions can be supplied, but ideally a single backwards-compatible API is all that's necessary.
  */
 export function createApiProvider(azExts: AzureExtensionApi[]): AzureExtensionApiProvider;
+
+// tslint:disable-next-line:max-classes-per-file
```
EricJizbaMSFT: I thought we weren't linting this file?

ui/index.d.ts
```diff
@@ -1056,3 +1055,24 @@ export function createAzureSubscriptionClient<T extends IAddUserAgent>(
...
+
+/**
+ * Wraps the vscode.OutputChannel to add timestamps to all messages by default.
+ * appendLine accepts optional parameters to override the date and to prepend the message with the resourceName
+ * Use append if you don't want timestamps for that output message
+ */
+export declare class AzureOutputChannel implements OutputChannel {
```
EricJizbaMSFT: Don't you need to use this in the ui package, too? For the resource group step at least? I'm just curious how you plan to make this work with registerUIExtensionVariables

ui/index.d.ts
```diff
@@ -1056,3 +1055,24 @@ export function createAzureSubscriptionClient<T extends IAddUserAgent>(
  * Multiple APIs with different versions can be supplied, but ideally a single backwards-compatible API is all that's necessary.
  */
 export function createApiProvider(azExts: AzureExtensionApi[]): AzureExtensionApiProvider;
+
+// tslint:disable-next-line:max-classes-per-file
+
+/**
+ * Wraps the vscode.OutputChannel to add timestamps to all messages by default.
+ * appendLine accepts optional parameters to override the date and to prepend the message with the resourceName
+ * Use append if you don't want timestamps for that output message
```
EricJizbaMSFT: I don't like overwriting appendLine. I'd prefer if you added another method like `appendLog` or something

ui/src/AzureOutputChannel.ts
```diff
@@ -0,0 +1,44 @@
...
+        this._outputChannel.show();
+    }
+
+    public hide(): void {
+        this._outputChannel.hide();
+    }
+
+    public dispose(): void {
+        this._outputChannel.dispose();
+        this.dispose();
```
EricJizbaMSFT: Why is `this.dispose()` here?

ui/index.d.ts
nturinski: Sorry, I think it was just the automatic format on save for tslint

ui/src/callWithTelemetryAndErrorHandling.ts
```diff
@@ -84,7 +84,7 @@ function handleError(context: IActionContext, callbackId: string, error: unknown
 
     if (!context.errorHandling.suppressDisplay) {
         // Always append the error to the output channel, but only 'show' the output channel for multiline errors
-        ext.outputChannel.appendLine(localize('outputError', 'Error: {0}', errorData.message));
+        ext.outputChannel.appendLog(localize('outputError', 'Error: {0}', errorData.message));
```
EricJizbaMSFT: I think `appendLog` makes complete sense for long-running operations like the wizard steps, but I'm not sure we want to do `appendLog` for _everything_. If you want to do that, I'd like to see some screenshots and get Matt's opinion first.

ui/src/AzureOutputChannel.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    public append(value: string): void {
+        this._outputChannel.append(value);
+    }
+
+    public appendLine(value: string): void {
+        this._outputChannel.appendLine(value);
+    }
+
+    public appendLog(value: string, resourceName?: string, date?: Date): void {
```
EricJizbaMSFT: Can you refactor `resourceName` and `date` into an `options` object? There's several examples in `index.d.ts`.

I want to avoid things like this:
```
appendLog(value, undefined, 'today')
```

And instead do something like this:
```
appendLog(value, { date: 'today' })
```

It's not too bad with only two parameters, but you can imagine something ugly like this if you ever add support for more options:
```
appendLog(value, undefined, undefined, undefined, 'otherProp')
```

ui/index.d.ts
```diff
@@ -10,6 +10,7 @@ import { ServiceClientCredentials } from 'ms-rest';
 import { AzureEnvironment, AzureServiceClientOptions } from 'ms-rest-azure';
 import { Disposable, Event, ExtensionContext, InputBoxOptions, Memento, MessageItem, MessageOptions, OpenDialogOptions, OutputChannel, QuickPickItem, QuickPickOptions, TextDocument, TreeDataProvider, TreeItem, Uri, QuickPick, InputBox, Progress } from 'vscode';
 import { AzureExtensionApi, AzureExtensionApiProvider } from './api';
+import { AzureOutputChannel } from './src';
```
EricJizbaMSFT: ...this is not how `index.d.ts` works...

You have to do the `export declare class` thing, and you should add docs for the class you're adding

ui/src/AzureOutputChannel.ts
```diff
@@ -0,0 +1,50 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.txt in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import { OutputChannel, ViewColumn } from "vscode";
+
+export class AzureOutputChannel implements OutputChannel {
```
EricJizbaMSFT: Can we name this `AzExtOutputChannel`? I want to start following the naming pattern I used for `AzExtTreeItem` vs. `AzureTreeItem`. "AzExt" should be used for anything that could apply outside of Azure (e.g. an attached mongo database) and "Azure" should be reserved for things that _only_ work in Azure.

ui/src/callWithTelemetryAndErrorHandling.ts
nturinski: image

Note: `appservice`  package isn't updated yet thus no timestamps

ui/index.d.ts
```diff
@@ -10,6 +10,7 @@ import { ServiceClientCredentials } from 'ms-rest';
 import { AzureEnvironment, AzureServiceClientOptions } from 'ms-rest-azure';
 import { Disposable, Event, ExtensionContext, InputBoxOptions, Memento, MessageItem, MessageOptions, OpenDialogOptions, OutputChannel, Progress, QuickPickItem, QuickPickOptions, TextDocument, TreeDataProvider, TreeItem, Uri } from 'vscode';
 import { AzureExtensionApi, AzureExtensionApiProvider } from './api';
+import { AzExtOutputChannel } from './src';
```
EricJizbaMSFT: Remove this line

ui/index.d.ts
```diff
@@ -1112,6 +1113,26 @@ export function createAzureSubscriptionClient<T extends IAddUserAgent>(
  */
 export function createApiProvider(azExts: AzureExtensionApi[]): AzureExtensionApiProvider;
 
+/**
+ * Wraps the vscode.OutputChannel to add timestamps to all messages by calling appendLog.
```
EricJizbaMSFT: I would keep the docs for the class more generic and move all of the comments specific to `appendLog` down onto `appendLog`

ui/src/AzExtOutputChannel.ts
```diff
@@ -0,0 +1,50 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.txt in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import { OutputChannel, ViewColumn } from "vscode";
+
+export class AzExtOutputChannel implements OutputChannel {
```
EricJizbaMSFT: Rather than `implements OutputChannel` do `implements types.AzExtOutputChannel`

with 
```
import * as types from '../../index';
```

That helps force you to keep the class itself up to date with the typing in index.d.ts

nturinski:
Honestly, I'm not sure either.  It's to satisfy this issue: https://github.com/microsoft/vscode-azureappservice/issues/788

But I think personally I kinda like having the time prepended to make all of the lines evenly padded.  It looks kinda off to me when there's some with a timestamp and some without.

nturinski:
@fiveisprime Any opinion on the screenshot? https://user-images.githubusercontent.com/5290572/59539162-0d007480-8eb1-11e9-893a-509b5bd18d10.png

EricJizbaMSFT:
Closing due to lack of activity

PR #516, opened 2019-06-03
Various fixes to support the docker tree

EricJizbaMSFT:
Included in this:
1. Support canPickMany in the treeItemPicker. This will only use a multi-select quick pick at the last stage
1. Handle docker-specific error formats in parseError
1. Allow hiding step count for an entire wizard

### DIFF
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -8,7 +8,7 @@ import { Location } from 'azure-arm-resource/lib/subscription/models';
 import { StorageAccount } from 'azure-arm-storage/lib/models';
 import { ServiceClientCredentials } from 'ms-rest';
 import { AzureEnvironment, AzureServiceClientOptions } from 'ms-rest-azure';
-import { Disposable, Event, ExtensionContext, InputBoxOptions, Memento, MessageItem, MessageOptions, OpenDialogOptions, OutputChannel, QuickPickItem, QuickPickOptions, TextDocument, TreeDataProvider, TreeItem, Uri, QuickPick, InputBox, Progress } from 'vscode';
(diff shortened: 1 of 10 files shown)


ui/src/treeDataProvider/AzExtTreeDataProvider.ts
```diff
@@ -109,28 +111,42 @@ export class AzExtTreeDataProvider implements IAzExtTreeDataProviderInternal, ty
         this.refreshUIOnly(treeItem);
     }
 
-    public refreshUIOnly(treeItem: AzExtTreeItem | undefined): void {
-        this._onDidChangeTreeDataEmitter.fire(treeItem === this._rootTreeItem ? undefined : treeItem);
+    public refreshUIOnly(_treeItem: AzExtTreeItem | undefined): void {
+        // Pass undefined as temporary workaround for https://github.com/microsoft/vscode/issues/71698
+        this._onDidChangeTreeDataEmitter.fire(undefined);
```
EricJizbaMSFT: @nturinski this was your recommended workaround for "Element with id is already registered", right? I was getting that error all the time in Docker and couldn't handle it anymore. Plus the VS Code folks bumped the milestone on that issue, so I'm not sure when they'll get to it.

Anyways - just wanted to say thanks for finding the workaround!

ui/index.d.ts
```diff
@@ -82,6 +83,19 @@ export declare class AzExtTreeDataProvider implements TreeDataProvider<AzExtTree
     public getParent(treeItem: AzExtTreeItem): Promise<AzExtTreeItem | undefined>;
 }
 
+export interface ITreeItemPickerContext extends IActionContext {
+    /**
+     * If set to true, the last (and _only_ the last) stage of the tree item picker will show a multi-select quick pick
+     */
+    canPickMany?: boolean;
```
nturinski: Why only the last?  Should this be something we specify on the step instead?

PR #517, opened 2019-06-04
Expose getRelated name for SiteNameStep to be used in App Service extension

nturinski:
I need this for the app service plan validation.  Also, added `_` to the list of allowed characters since I'm 100% sure it works as our default is `appsvc_asp_${wizardContext.newSiteOS}_${location.name}`

Also there's an RG reg exp we can use here: https://docs.microsoft.com/en-us/rest/api/appservice/appserviceplans/update

### DIFF
--- a/appservice/src/createAppService/AppServicePlanNameStep.ts
+++ b/appservice/src/createAppService/AppServicePlanNameStep.ts
@@ -13,7 +13,7 @@ import { IAppServiceWizardContext } from './IAppServiceWizardContext';
 export const appServicePlanNamingRules: IAzureNamingRules = {
     minLength: 1,
     maxLength: 40,
-    invalidCharsRegExp: /[^a-zA-Z0-9\-]/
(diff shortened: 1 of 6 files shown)


nturinski:
I personally don't really care about having an underscore prefixing the number

appservice/src/createAppService/AppServicePlanNameStep.ts
```diff
@@ -13,7 +13,7 @@ import { IAppServiceWizardContext } from './IAppServiceWizardContext';
 export const appServicePlanNamingRules: IAzureNamingRules = {
     minLength: 1,
     maxLength: 40,
-    invalidCharsRegExp: /[^a-zA-Z0-9\-]/
+    invalidCharsRegExp: /[^a-zA-Z0-9\-_]/
```
EricJizbaMSFT: I think you need `\_`

nturinski:
It comes out like this 
image which I'm okay with

PR #518, opened 2019-06-04
Refactor GitHub steps into an Azure Wizard

nturinski:
Fixes https://github.com/microsoft/vscode-azureappservice/issues/764

### DIFF
--- a/appservice/src/github/GitHubBranchListStep.ts
+++ b/appservice/src/github/GitHubBranchListStep.ts
@@ -6,7 +6,7 @@
 import { AzureWizardPromptStep, IAzureQuickPickItem } from 'vscode-azureextensionui';
 import { ext } from '../extensionVariables';
 import { nonNullProp } from '../utils/nonNull';
-import { getGitHubQuickPicksWithLoadMore, gitHubBranchData, gitHubWebResource, ICachedQuickPicks } from './connectToGitHub';
+import { createRequestOptions, getGitHubQuickPicksWithLoadMore, gitHubBranchData, gitHubWebResource, ICachedQuickPicks } from './connectToGitHub';
 import { IConnectToGitHubWizardContext } from './IConnectToGitHubWizardContext';
 
 export class GitHubBranchListStep extends AzureWizardPromptStep<IConnectToGitHubWizardContext> {
@@ -26,8 +26,7 @@ export class GitHubBranchListStep extends AzureWizardPromptStep<IConnectToGitHub
     }
 
     private async getBranches(context: IConnectToGitHubWizardContext, picksCache: ICachedQuickPicks<gitHubBranchData>): Promise<IAzureQuickPickItem<gitHubBranchData | undefined>[]> {
-        const requestOption: gitHubWebResource = nonNullProp(context, 'requestOption');
-        requestOption.url = `${nonNullProp(context, 'repoData').url}/branches`;
-        return await getGitHubQuickPicksWithLoadMore<gitHubBranchData>(picksCache, requestOption, 'name');
+        const requestOption: gitHubWebResource = await createRequestOptions(context, `${nonNullProp(context, 'repoData').url}/branches`);
+        return await getGitHubQuickPicksWithLoadMore<gitHubBranchData>(context, picksCache, requestOption, 'name');
     }
 }
--- a/appservice/src/github/GitHubBranchListStep.ts
+++ b/appservice/src/github/GitHubBranchListStep.ts
@@ -1,0 +1,33 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { AzureWizardPromptStep, IAzureQuickPickItem } from 'vscode-azureextensionui';
+import { ext } from '../extensionVariables';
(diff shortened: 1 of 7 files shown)


appservice/src/github/GitHubBranchListStep.ts
```diff
@@ -0,0 +1,33 @@
...
+        } while (!branchQuickPick);
+
+        context.branchData = branchQuickPick;
+    }
+
+    public shouldPrompt(context: IConnectToGitHubWizardContext): boolean {
+        return !context.branchData;
+    }
+
+    private async getBranches(context: IConnectToGitHubWizardContext, picksCache: ICachedQuickPicks<gitHubBranchData>): Promise<IAzureQuickPickItem<gitHubBranchData | undefined>[]> {
```
EricJizbaMSFT: Can we rename this to getBranchPicks?

appservice/src/github/GitHubBranchListStep.ts
```diff
@@ -0,0 +1,33 @@
...
+
+import { AzureWizardPromptStep, IAzureQuickPickItem } from 'vscode-azureextensionui';
+import { ext } from '../extensionVariables';
+import { nonNullProp } from '../utils/nonNull';
+import { getGitHubQuickPicksWithLoadMore, gitHubBranchData, gitHubWebResource, ICachedQuickPicks } from './connectToGitHub';
+import { IConnectToGitHubWizardContext } from './IConnectToGitHubWizardContext';
+
+export class GitHubBranchListStep extends AzureWizardPromptStep<IConnectToGitHubWizardContext> {
+    public async prompt(context: IConnectToGitHubWizardContext): Promise<void> {
+        const placeHolder: string = 'Choose branch';
```
EricJizbaMSFT: "branchQuickPick" seems unnecessary and misleading since it's the data, not the pick

```
do {
    context.branchData = (await ext.ui.showQuickPick(this.getBranches(context, picksCache), { placeHolder })).data;
} while (!context.branchData);
```

appservice/src/github/connectToGitHub.ts
```diff
@@ -7,29 +7,35 @@ import { SiteSourceControl } from 'azure-arm-website/lib/models';
...
 // tslint:disable-next-line:no-reserved-keywords
-type gitHubWebResource = WebResource & { resolveWithFullResponse?: boolean, nextLink?: string, type?: string };
+export type gitHubWebResource = WebResource & { resolveWithFullResponse?: boolean, nextLink?: string, type?: string };
 
-export async function connectToGitHub(node: AzureTreeItem, client: SiteClient, context: IActionContext): Promise<void> {
-    const requestOptions: gitHubWebResource = new WebResource();
-    requestOptions.resolveWithFullResponse = true;
-    requestOptions.headers = {
+export async function connectToGitHub(node: AzureTreeItem, client: SiteClient, context: IConnectToGitHubWizardContext): Promise<void> {
+    context.requestOption = new WebResource();
```
EricJizbaMSFT: Not a huge fan of re-using the same requestOption. Can you create a new `requestOption` inside `getGitHubJsonResponse` and take the url as a parameter instead?

Specifically worried about one of the requests setting a header or something and other requests inheriting that on accident

appservice/src/github/connectToGitHub.ts
```diff
@@ -38,35 +44,39 @@ export async function connectToGitHub(node: AzureTreeItem, client: SiteClient, c
...
+            new GitHubRepoListStep(),
+            new GitHubBranchListStep()
+        ]
+    });
+
+    try {
+        await wizard.prompt();
+    } catch (error) {
+        const parsedError: IParsedError = parseError(error);
+        if (parsedError.message.indexOf('Bad credentials') > -1) {
```
EricJizbaMSFT: Why did you move this logic outside getGitHubJsonResponse? If you ever change this to be a subwizard instead of it's own wizard - you'd have to move this logic back since you wouldn't call wizard.prompt() directly.

nturinski: Well, it was because it was the only reason each step required the node and client to be passed in which seemed excessive.

EricJizbaMSFT: Ah well I'm fine with the change in that case (although like I said you might have to refactor if you ever move this to a subwizard)

nturinski: Future Nathan's problem 😼

appservice/src/github/connectToGitHub.ts
```diff
@@ -98,7 +85,7 @@ async function showGitHubAuthPrompt(node: AzureTreeItem, client: SiteClient, con
     const goToPortal: vscode.MessageItem = { title: localize('goToPortal', 'Go to Portal') };
     let input: vscode.MessageItem | undefined = DialogResponses.learnMore;
     while (input === DialogResponses.learnMore) {
-        input = await vscode.window.showErrorMessage(invalidToken, goToPortal, DialogResponses.learnMore);
+        input = await vscode.window.showErrorMessage(invalidToken, { modal: true }, goToPortal, DialogResponses.learnMore);
```
nturinski: I made this a modal because the Organization quick pick menu looked like it was still loading when there's an error until you click the dialog which could be confusing.

nturinski:
I'm going to continue making changes to this since it'll be necessary for another PR. It shouldn't be anything huge, just passing in the name of org/repo like we talked about offline.

nturinski:
Never mind, Matt didn't like the automatically getting org/repo so not gonna add more to this PR.

PR #520, opened 2019-06-10
Add shared logic to open read only content

EricJizbaMSFT:
Docker had this before https://github.com/microsoft/vscode-docker/pull/966 - I'm just implementing it in our shared package. It's the difference between this:
Screen Shot 2019-06-10 at 10 51 30 AM

And this:
Screen Shot 2019-06-10 at 10 52 54 AM

The readonly view is nice for two reasons:
1. It has a better tab title name
1. It doesn't prompt to save each file when you close them

A few downsides:
1. These readonly files never persist across sessions
1. You can't even edit it unless you copy the content to a new file
1. You can't do "File -> Save As" (Honestly this feels like a bug in VS Code...I might follow up on this)

Even with those downsides, I think this is still a better experience (At least in certain scenarios), but open for debate.


ui/src/openReadOnlyContent.ts
```diff
@@ -0,0 +1,60 @@
...
+import { CancellationToken, Event, EventEmitter, TextDocumentContentProvider, Uri, window, workspace, WorkspaceConfiguration } from "vscode";
+import { ext } from "./extensionVariables";
+import { nonNullValue } from "./utils/nonNull";
+import { randomUtils } from "./utils/randomUtils";
+
+let contentProvider: ReadOnlyContentProvider | undefined;
+const scheme: string = 'azureextensionuiReadonly';
+
+export async function openReadOnlyJson(node: { label: string, fullId: string }, data: {}): Promise<void> {
+    let tab: string = '	';
```
StephenWeatherford: ' '.repeat(8)

ui/src/openReadOnlyContent.ts
```diff
@@ -0,0 +1,60 @@
...
+    const config: WorkspaceConfiguration = workspace.getConfiguration('editor');
+    const insertSpaces: boolean = !!config.get<boolean>('insertSpaces');
+    if (insertSpaces) {
+        let tabSize: number | undefined = config.get<number>('tabSize');
+        if (!isNumber(tabSize) || tabSize < 0) {
```
StephenWeatherford: should this be <= 0?  Not sure what tabsize=0 means

ui/src/openReadOnlyContent.ts
```diff
@@ -0,0 +1,60 @@
...
+    private _onDidChangeEmitter: EventEmitter<Uri> = new EventEmitter<Uri>();
+    private _contentMap: Map<string, string> = new Map<string, string>();
+
+    public get onDidChange(): Event<Uri> {
+        return this._onDidChangeEmitter.event;
+    }
+
+    public async openReadOnlyContent(node: { label: string, fullId: string }, content: string, fileExtension: string): Promise<void> {
+        const idHash: string = encodeURIComponent(randomUtils.getPseudononymousStringHash(node.fullId));
+        const uri: Uri = Uri.parse(`${scheme}:///${idHash}/${node.label}${fileExtension}`);
```
StephenWeatherford: is three ///'s correct?

EricJizbaMSFT: Yeah I'm just following typical patterns for `file` uris. It just means the "host" part of the uri is an empty string. For example this code:
```
const url = new URL('file:///Users/ericjizba/example.txt');
console.log("HOST: " + url.host);
console.log("PATHNAME: " + url.pathname);

const url2 = new URL('file://Users/ericjizba/example.txt');
console.log("HOST: " + url2.host);
console.log("PATHNAME: " + url2.pathname);
```

Prints this:
```
HOST: 
PATHNAME: /Users/ericjizba/example.txt
HOST: Users
PATHNAME: /ericjizba/example.txt
```

ui/src/openReadOnlyContent.ts
EricJizbaMSFT: This a tab character, not 8 spaces

ui/src/openReadOnlyContent.ts
EricJizbaMSFT: It just means there's no tabs. I mean...I would never want that, but ¯\\\_(ツ)\_/¯

PR #531, opened 2019-07-19
Filter locations based on subscription

EricJizbaMSFT:
The api to get locations specific for function apps returns a list like this:
Screen Shot 2019-07-19 at 10 41 51 AM

But not all regions work for my subscription:
Screen Shot 2019-07-19 at 10 42 07 AM

So I changed the logic to always start from the list of locations available to a subscription and filter from there. Now I see a list like this:
Screen Shot 2019-07-19 at 10 44 43 AM

Related to https://github.com/microsoft/vscode-azurefunctions/issues/1359

### DIFF
--- a/ui/src/wizard/LocationListStep.ts
+++ b/ui/src/wizard/LocationListStep.ts
@@ -17,8 +17,16 @@ function generalizeLocationName(name: string | undefined): string {
     return (name || '').toLowerCase().replace(/\s/g, '');
 }
 
-export class LocationListStep<T extends types.ILocationWizardContext> extends AzureWizardPromptStep<T> implements types.LocationListStep<T> {
-    public static async setLocation<T extends types.ILocationWizardContext>(wizardContext: T, name: string): Promise<void> {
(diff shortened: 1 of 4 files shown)


ui/src/wizard/LocationListStep.ts
```diff
@@ -17,22 +17,36 @@ function generalizeLocationName(name: string | undefined): string {
...
+            wizardContext._allLocationsTask = client.subscriptions.listLocations(wizardContext.subscriptionId);
         }
 
-        return await wizardContext.locationsTask;
+        const allLocations: Location[] = await wizardContext._allLocationsTask;
+        if (wizardContext.locationsTask === undefined) {
+            return allLocations;
+        } else {
+            const locationsSubset: { name: string }[] = await wizardContext.locationsTask;
+            return allLocations.filter(l1 => locationsSubset.find(l2 => generalizeLocationName(l1.name) === generalizeLocationName(l2.name)));
```
nturinski: Is the locationsTask array expected to be hardcoded?  Shouldn't we just have it return a Location[] so we don't need to do this filter?

PR #532, opened 2019-07-19
Don't show recently used for first item

EricJizbaMSFT:
No need to show `(recently used)` and move it to the top if it's already at the top. I think the quick pick looks a little cleaner this way.


ui/src/AzureUserInput.ts
```diff
@@ -111,7 +111,7 @@ export class AzureUserInput implements types.IAzureUserInput, types.AzureUserInp
             const previousValue: string | undefined = this._persistence.get(persistenceKey);
             if (previousValue) {
                 const index: number = items.findIndex((item: T) => getPersistenceValue(item) === previousValue);
-                if (index !== -1) {
+                if (index > 0) {
```
StephenWeatherford: How about a comment?

PR #535, opened 2019-07-19
Add icons when not signed in to Azure

EricJizbaMSFT:
I was adding "Create New Project" per https://github.com/microsoft/vscode-azurefunctions/issues/1322 and thought some icons might be nice:

Screen Shot 2019-07-19 at 3 43 12 PM
Screen Shot 2019-07-19 at 3 43 28 PM

Also just cleaned up the icon code.

### DIFF
--- a/ui/src/treeDataProvider/AzureAccountTreeItemBase.ts
+++ b/ui/src/treeDataProvider/AzureAccountTreeItemBase.ts
@@ -3,7 +3,6 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import * as path from 'path';
 import { commands, Disposable, Extension, extensions, MessageItem, ProgressLocation, window } from 'vscode';
 import * as types from '../../index';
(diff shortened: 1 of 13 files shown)


ui/src/treeDataProvider/AzureAccountTreeItemBase.ts
```diff
@@ -148,6 +149,14 @@ export abstract class AzureAccountTreeItemBase extends AzExtParentTreeItem imple
         return undefined;
     }
 
+    public compareChildrenImpl(item1: AzExtTreeItem, item2: AzExtTreeItem): number {
```
StephenWeatherford: Is this related?

EricJizbaMSFT: Yeah without it the items in the above screenshot would be sorted alphabetically:

- Create a Free Azure Account...
- Create New Local Project...
- Sign in to Azure...

I don't have a strong preference on the order _other than_ the two Azure ones should be next to each other. Which unfortunately is exactly what happens with alphabetical sorting.

PR #541, opened 2019-07-24
Fix new Test* classes in dev package to work with webpack

EricJizbaMSFT:
Because of how the dev package is structured, we need to delay loading the 'vscode' package until the tests are actually running. Tests run fine in non-webpack scenario, but not in webpack (default used by our CI).

I'll send out a PR for storage shortly to show how this will be used.


EricJizbaMSFT:
See https://github.com/microsoft/vscode-azurestorage/pull/406 for example using this new stuff.

PR #543, opened 2019-07-26
More validation in createTreeItemsWithErrorHandling

EricJizbaMSFT:
Related to https://github.com/microsoft/vscode-docker/issues/1155

In that case, `treeItem.label` was failing for one tree item, blocking display of all tree items, and the user couldn't disconnect the registry. This PR has two benefits:
1. If something like `treeItem.label` fails, it'll be handled as a normal invalid tree item
1. For invalid tree items, the `source` object will be there so I can perform actions like "Disconnect"

### DIFF
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -231,12 +231,18 @@ export interface IInvalidTreeItemOptions {
      * Defaults to "Invalid" if undefined
      */
     description?: string;
+
+    /**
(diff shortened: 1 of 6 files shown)


EricJizbaMSFT:
In the case of Docker, `source` is a basic object describing a connected registry (cached across sessions). Even if I don't have the full tree item (because the cached object was invalid), I can still let the user disconnect a registry based on the cached information alone

PR #547, opened 2019-08-01
Add listKey methods for functions

EricJizbaMSFT:
These keys are needed to get function urls for linux function apps.

And refactor to use a `requestUtils` file that ensures user agent, resourceManagerEndpoint, and fewer tslint disables.


EricJizbaMSFT:
Not `requestUtils`. Doesn't seem specific to app service

PR #549, opened 2019-08-02
Support server side build for linux function apps

EricJizbaMSFT:
Currently this will _not_ be turned on by default. Users have to set `azureFunctions.scmDoBuildDuringDeployment` to true to enable this.

Most of this logic is done per instructions from the Functions team. Let me know if anything is especially confusing, but hopefully a good chunk of this code is only temporary workarounds.

This is a breaking change because I had to make `SiteClient.kudu` async (now it's `SiteClient.getKuduClient`). This change actually helped quite a few cases, though. For example it fixes https://github.com/microsoft/vscode-azurefunctions/issues/1046

### DIFF
--- a/appservice/src/SiteClient.ts
+++ b/appservice/src/SiteClient.ts
@@ -5,8 +5,9 @@
 
 import { WebSiteManagementClient } from 'azure-arm-website';
 import { AppServicePlan, FunctionEnvelopeCollection, FunctionSecrets, HostNameSslState, Site, SiteConfigResource, SiteLogsConfig, SiteSourceControl, SlotConfigNamesResource, SourceControlCollection, StringDictionary, User, WebAppInstanceCollection, WebJobCollection } from 'azure-arm-website/lib/models';
+import { ServiceClientOptions, WebResource } from 'ms-rest';
 import { addExtensionUserAgent, createAzureClient, ISubscriptionContext, parseError } from 'vscode-azureextensionui';
-import KuduClient from 'vscode-azurekudu';
+import { KuduClient } from 'vscode-azurekudu';
 import { FunctionEnvelope } from 'vscode-azurekudu/lib/models';
 import { localize } from './localize';
 import { nonNullProp, nonNullValue } from './utils/nonNull';
@@ -49,6 +50,10 @@ export class SiteClient {
     private readonly _subscription: ISubscriptionContext;
     private _funcPreviewSlotError: Error = new Error(localize('functionsSlotPreview', 'This operation is not supported for slots, which are still in preview.'));
 
+    private _cachedPlan: AppServicePlan | undefined;
+    private _cachedSiteRestrictedToken: string | undefined;
+    private _siteRestrictedTokenExpireTime: number = Date.now();
+
     constructor(site: Site, subscription: ISubscriptionContext) {
         let matches: RegExpMatchArray | null = nonNullProp(site, 'serverFarmId').match(/\/subscriptions\/(.*)\/resourceGroups\/(.*)\/providers\/Microsoft.Web\/serverfarms\/(.*)/);
         matches = nonNullValue(matches, 'Invalid serverFarmId.');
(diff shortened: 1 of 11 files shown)


appservice/src/SiteClient.ts
```diff
@@ -288,6 +301,58 @@ export class SiteClient {
...
+        }
+        return result;
+    }
+
+    /**
+     * Linux consumption currently requires a site-restricted token for all kudu calls (in addition to existing credentials)
+     * This is only a temporary requirement and should be going away eventually
+     */
+    private async getKuduClientOptions(): Promise<ServiceClientOptions | undefined> {
+        if (this.isFunctionApp && this.isLinux) {
```
nturinski: Doesn't getIsConsumption already check if this isFunctionApp?

appservice/src/SiteClient.ts
```diff
@@ -288,6 +301,58 @@ export class SiteClient {
...
+
+        return undefined;
+    }
+
+    private async getSiteRestrictedToken(): Promise<string> {
+        if (!this._cachedSiteRestrictedToken || Date.now() > this._siteRestrictedTokenExpireTime) {
+            const urlPath: string = `${this.id}/hostruntime/admin/host/token?api-version=2015-08-01`;
+            const request: requestUtils.Request = await requestUtils.getDefaultAzureRequest(urlPath, this._subscription);
+            // token only lasts 5 minutes. Use 4 just to be safe
+            this._siteRestrictedTokenExpireTime = Date.now() + 4 * 60 * 1000;
```
nturinski: Azure tokens usually have a number property that specifies when it'll expire.  I think it's called `expiresIn`.  Could just add the exact number that's there rather than making an estimate.

appservice/src/deploy/deployZip.ts
```diff
@@ -41,19 +42,23 @@ export async function deployZip(client: SiteClient, fsPath: string, aspPromise:
...
-            // Linux consumption doesn't support kudu zipPushDeploy
-            await deployToStorageAccount(client, zipFilePath);
-        } else {
-            await client.kudu.pushDeployment.zipPushDeploy(fs.createReadStream(zipFilePath), { isAsync: true });
-            await waitForDeploymentToComplete(client);
-            // https://github.com/Microsoft/vscode-azureappservice/issues/644
-            // This delay is a temporary stopgap that should be resolved with the new pipelines
-            await delayFirstWebAppDeploy(client, asp);
-
+        if (client.isFunctionApp && client.isLinux) {
```
nturinski: Same here, though I guess it doesn't really matter

appservice/src/deploy/deployZip.ts
```diff
@@ -101,3 +107,52 @@ async function delayFirstWebAppDeploy(client: SiteClient, asp: AppServicePlan |
...
+    if (!isConsumption) {
+        const dedicatedBuildSettings: [string, string][] = [
+            ['ENABLE_ORYX_BUILD', 'true'],
+            ['SCM_DO_BUILD_DURING_DEPLOYMENT', '1'],
+            ['BUILD_FLAGS', 'UseExpressBuild'],
+            ['XDG_CACHE_HOME', '/tmp/.cache']
+        ];
+
+        for (const [key, value] of dedicatedBuildSettings) {
+            if (!doBuild) {
```
nturinski: This never changes right?  Couldn't you just do a concat and then break out of the for loop?

appservice/src/SiteClient.ts
EricJizbaMSFT: This is definitely _not_ your typical azure token. It's a workaround and is _in addition_ to the typical token used for authenticating. The functions team told me to refresh the token based on the 5 minute interval - I never would've known to do it myself.

appservice/src/deploy/deployZip.ts
EricJizbaMSFT: The other check I can get rid of, but this one I need. We validate app settings for _all_ linux function apps, regardless of if it's consumption or dedicated

appservice/src/deploy/deployZip.ts
EricJizbaMSFT: Yeah that would be the exact same - I don't think either of us really care between the two options so Imma leave it

PR #554, opened 2019-08-02
Add App Insights create step for basic creation

nturinski:
For now, this is only the create step.  Eventually will add support for List and Name as well for advanced create scenario.


appservice/src/createAppService/AppInsightsCreateStep.ts
```diff
@@ -0,0 +1,61 @@
...
+            ext.outputChannel.appendLine(appInsightsNotAvailable);
+        }
+    }
+
+    public shouldExecute(wizardContext: IAppServiceWizardContext): boolean {
+        return !wizardContext.applicationInsights;
+    }
+
+    private async appInsightsSupportedInLocation(wizardContext: IAppServiceWizardContext, location: Location): Promise<boolean> {
+        const aiRegionMappingUrl: string = 'providers/microsoft.insights?api-version=2014-04-01-preview';
```
EricJizbaMSFT: Try `ResourceManagementClient.providers.get` instead of making your own request

appservice/src/connectToAppInsights.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export async function connectToAppInsights(client: SiteClient, appInsightComponent: ApplicationInsightsComponent): Promise<void> {
+
+    const appSettings: StringDictionary = await client.listApplicationSettings();
+    // tslint:disable-next-line:strict-boolean-expressions
+    appSettings.properties = appSettings.properties || {};
+
+    const appInsightKey: string = 'APPINSIGHTS_INSTRUMENTATIONKEY';
+    if (appInsightComponent.instrumentationKey) {
+        appSettings.properties[appInsightKey] = appInsightComponent.instrumentationKey;
```
EricJizbaMSFT: I feel like this should be done as a part of this function: https://github.com/microsoft/vscode-azurefunctions/blob/master/src/tree/SubscriptionTreeItem.ts#L173

appservice/src/createAppService/AppInsightsCreateStep.ts
```diff
@@ -0,0 +1,61 @@
...
+        ext.outputChannel.appendLine(verifyingAppInsightsAvailable);
+        if (await this.appInsightsSupportedInLocation(wizardContext, location)) {
+
+            const creatingNewAppInsights: string = localize('creatingNewAppInsightsInsights', 'Creating new application insights component "{0}"...', wizardContext.newSiteName);
+            ext.outputChannel.appendLine(creatingNewAppInsights);
+
+            const client: ApplicationInsightsManagementClient = createAzureClient(wizardContext, ApplicationInsightsManagementClient);
+            wizardContext.applicationInsights = await client.components.createOrUpdate(
+                nonNullValue(wizardContext.newResourceGroupName),
+                nonNullValue(wizardContext.newSiteName),
```
EricJizbaMSFT: There should be a property for app insights name and you should set it here: https://github.com/microsoft/vscode-azurefunctions/blob/master/src/tree/SubscriptionTreeItem.ts#L138

appservice/src/index.ts
```diff
@@ -15,12 +16,12 @@ export * from './createAppService/SiteOSStep';
 export * from './createAppService/SiteRuntimeStep';
 export * from './remoteDebug/remoteDebugCommon';
 export * from './remoteDebug/startRemoteDebug';
+export * from './connectToAppInsights';
 export * from './createSlot';
 export * from './deploy/deploy';
 export * from './deploy/runPreDeployTask';
 export * from './deleteSite';
 export * from './editScmType';
-export * from './getFile';
```
EricJizbaMSFT: Did you mean to delete this?

EricJizbaMSFT:
Since we don't have a time crunch anymore I'm fine if you want to do basic/advanced create at the same time

nturinski:
https://github.com/microsoft/vscode-azuretools/issues/560 

https://media.tenor.com/images/34bc40aac47a08c58032dd5e1d4bbcc8/tenor.gif

appservice/src/createAppService/AppInsightsCreateStep.ts
```diff
@@ -0,0 +1,58 @@
...
+        if (await this.appInsightsSupportedInLocation(wizardContext, location)) {
+            const client: ApplicationInsightsManagementClient = createAzureClient(wizardContext, ApplicationInsightsManagementClient);
+            const rgName: string = nonNullValue(wizardContext.newResourceGroupName);
+            const aiName: string = nonNullValue(wizardContext.newApplicationInsightsName);
+            try {
+                wizardContext.applicationInsights = await client.components.get(rgName, aiName);
+                ext.outputChannel.appendLine(localize('existingNewAppInsights', 'Using existing application insights component "{0}".', aiName));
+            } catch (error) {
+                // Only expecting a resource not found error here
+                const creatingNewAppInsights: string = localize('creatingNewAppInsightsInsights', 'Creating new application insights component "{0}"...', wizardContext.newSiteName);
```
EricJizbaMSFT: You should leverage parseError to check the error type. There's plenty of other errors unrelated to "resource not found" that could happen

appservice/src/createAppService/AppInsightsCreateStep.ts
```diff
@@ -0,0 +1,58 @@
...
+            ext.outputChannel.appendLine(appInsightsNotAvailable);
+        }
+    }
+
+    public shouldExecute(wizardContext: IAppServiceWizardContext): boolean {
+        return !wizardContext.applicationInsights;
+    }
+
+    private async appInsightsSupportedInLocation(wizardContext: IAppServiceWizardContext, location: Location): Promise<boolean> {
+        const resourceClient: ResourceManagementClient = createAzureClient(wizardContext, ResourceManagementClient);
```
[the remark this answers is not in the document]
EricJizbaMSFT: ping on this

appservice/src/createAppService/IAppServiceWizardContext.ts
```diff
@@ -72,4 +73,16 @@ export interface IAppServiceWizardContext extends IResourceGroupWizardContext, I
      */
 
     recommendedSiteRuntime?: LinuxRuntimes[];
+
+    /**
+     * Application insights components are necessary for Function apps log streaming.  By default, we should instance
+     * one for the user if there is a data farm available within the same region as the web app
+     */
+    applicationInsights?: ApplicationInsightsComponent;
```
EricJizbaMSFT: How about you consistently use `appInsights` instead of `applicationIsights` (similar to what you did with `AppInsightsCreateStep`)

Also feels like this one should be named `appInsightsComponent`. Technically we could add `component` to all the names, but that seems like it would get excessive. I guess this is the only one that bothers me the most

appservice/src/createAppService/IAppServiceWizardContext.ts
```diff
@@ -72,4 +73,16 @@ export interface IAppServiceWizardContext extends IResourceGroupWizardContext, I
      */
 
     recommendedSiteRuntime?: LinuxRuntimes[];
+
+    /**
+     * Application insights components are necessary for Function apps log streaming.  By default, we should instance
```
EricJizbaMSFT: "we should instance one"

appservice/src/createAppService/AppInsightsCreateStep.ts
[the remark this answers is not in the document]
nturinski: Renamed to not be using the mapping.  I think we discussed offline that we would not use the mapping initially and add it in later (if there's a need for it)?

appservice/src/createAppService/IAppServiceWizardContext.ts
nturinski: I'm not sure what this comment means.  Should I move "one" up a line?

appservice/src/createAppService/AppInsightsCreateStep.ts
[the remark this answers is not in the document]
EricJizbaMSFT: I don't remember that discussion, but I'm okay if you want to do it separately. I think you should verify how many regions support function apps vs how many support app insights rather than waiting for user feedback. If there's a big gap there we should use the api ASAP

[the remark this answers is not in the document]
nturinski: image

So for Functions, it seems like there's a 1:1 mapping.  But when I implement it for web apps, I'll probably have to use the mapping.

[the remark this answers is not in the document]
EricJizbaMSFT: You should check Windows, though. I don't think linux is a good sample because it's still in preview and the regions will likely increase as they GA

PR #557, opened 2019-08-05
Remove usage of gulp-download

EricJizbaMSFT:
This gets rid of some npm alerts about `lodash.template`

The packages I added seem to be much more widely used, smaller, and kept up to date (unlike `gulp-download`)


dev/package.json
```diff
@@ -56,11 +59,13 @@
         "glob": "^7.1.3",
         "gulp": "^4.0.0",
         "gulp-decompress": "^2.0.2",
-        "gulp-download": "^0.0.1",
         "ms-rest": "^2.2.2",
         "ms-rest-azure": "^2.4.4",
+        "request": "^2.88.0",
         "terser-webpack-plugin": "^1.2.2",
         "ts-loader": "^5.3.3",
+        "vinyl-buffer": "^1.0.1",
```
nturinski: Does this need to be in normal dependencies?  This isn't a devDependencies?

PR #558, opened 2019-08-05
Adding suppressAll option in ITelemetryContext

monalgarg:


ui/src/callWithTelemetryAndErrorHandling.ts
```diff
@@ -108,7 +108,7 @@ function handleError(context: IActionContext, callbackId: string, error: unknown
 }
 
 function handleTelemetry(context: IActionContext, callbackId: string, start: number): void {
-    if (!(context.telemetry.suppressIfSuccessful && context.telemetry.properties.result === 'Succeeded')) {
+    if (!((context.telemetry.suppressIfSuccessful || context.telemetry.suppressAll) && context.telemetry.properties.result === 'Succeeded')) {
```
EricJizbaMSFT: Double check this logic. The `suppressIfSuccessful` and check against result should go together

PR #562, opened 2019-08-08
AI Region Mapping (Basic Create)

nturinski:
Leverage the RegionMapping JSON to provide a paired location if there is no exact match for an App Insights component.

### DIFF
--- a/appservice/src/createAppService/AppInsightsCreateStep.ts
+++ b/appservice/src/createAppService/AppInsightsCreateStep.ts
@@ -17,13 +17,14 @@ import { IAppServiceWizardContext } from './IAppServiceWizardContext';
 
 export class AppInsightsCreateStep extends AzureWizardExecuteStep<IAppServiceWizardContext> {
     public priority: number = 135;
-    private _location: string;
 
     public async execute(wizardContext: IAppServiceWizardContext, progress: Progress<{ message?: string; increment?: number }>): Promise<void> {
-        const location: Location = nonNullProp(wizardContext, 'location');
+        const resourceLocation: Location = nonNullProp(wizardContext, 'location');
         const verifyingAppInsightsAvailable: string = localize('verifyingAppInsightsAvailable', 'Verifying that application insights is available for this location...');
         ext.outputChannel.appendLine(verifyingAppInsightsAvailable);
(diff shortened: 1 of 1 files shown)


appservice/src/createAppService/AppInsightsCreateStep.ts
```diff
@@ -53,11 +55,43 @@ export class AppInsightsCreateStep extends AzureWizardExecuteStep<IAppServiceWiz
...
         const supportedRegions: Provider = await resourceClient.providers.get('microsoft.insights');
         const componentsResourceType: ProviderResourceType | undefined = supportedRegions.resourceTypes && supportedRegions.resourceTypes.find(aiRt => aiRt.resourceType === 'components');
 
-        return !!componentsResourceType && !!componentsResourceType.locations && componentsResourceType.locations.some((loc) => loc === location.displayName);
+        if (!!componentsResourceType && !!componentsResourceType.locations && componentsResourceType.locations.some((loc) => loc === location.displayName)) {
+            this._location = nonNullProp(location, 'name');
+            return true;
+        } else {
+            // If there is no exact match, then query the regionMapping.json
+            const request: requestUtils.Request = await requestUtils.getDefaultRequest('https://appinsights.azureedge.net/portal/regionMapping.json');
```
EricJizbaMSFT: If this fails, I feel like we should just return `false` (aka don't enable app insights) rather than show this error to the user. Same for if you fail to parse the response, etc.

appservice/src/createAppService/AppInsightsCreateStep.ts
```diff
@@ -53,11 +55,43 @@ export class AppInsightsCreateStep extends AzureWizardExecuteStep<IAppServiceWiz
         return !wizardContext.appInsightsComponent;
     }
 
-    private async appInsightsSupportedInLocation(wizardContext: IAppServiceWizardContext, location: Location): Promise<boolean> {
+    public async appInsightsSupportedInLocation(wizardContext: IAppServiceWizardContext, location: Location): Promise<boolean> {
```
EricJizbaMSFT: How about you change this to return `string | undefined` instead of having a private `_location` variable. Return the location name if it's supported, otherwise undefined

Issue #560, opened 2019-08-06
Add naming rules for App Insights in NameStep


nturinski:
Regexp to use: ^[-\w\._\(\)]+$

EricJizbaMSFT:
Lol if you're going to create an issue it should be in the functions or app service repo

We almost never look through or triage vscode-azuretools

nturinski:
I'll remember to do it, don't worry! 😅

PR #563, opened 2019-08-12
Advanced Creation for App Insights

nturinski:
Here are the naming rules from the portal:
image

These rules are pretty much identical to resource groups.

When creating the app insights in the portal, the location will default in the following way:
1. The same region as the function app
2. A region in the region map

The user does have the option to change the region to whatever they want though.

To emulate this, I added recommended to the quickPick list.

prompt_advancedai


appservice/src/createAppService/AppInsightsListStep.ts
```diff
@@ -0,0 +1,85 @@
...
+import { AppInsightsNameStep } from "./AppInsightsNameStep";
+import { IAppServiceWizardContext } from "./IAppServiceWizardContext";
+
+export const appInsightsNamingRules: IAzureNamingRules = {
+    minLength: 1,
+    maxLength: 255,
+    invalidCharsRegExp: /[^a-zA-Z0-9\.\_\-\(\)]/
+};
+
+const skipForNow: string = '$(circle-slash) Skip for now';
```
EricJizbaMSFT: In other places I use `$(clock)`
Screen Shot 2019-08-13 at 9 05 33 AM

I'd prefer to stay consistent unless you have a strong objection

appservice/src/createAppService/AppInsightsLocationStep.ts
```diff
@@ -0,0 +1,105 @@
...
+        return [];
+    }
+
+    private async getQuickPicks(wizardContext: IAppServiceWizardContext): Promise<IAzureQuickPickItem<string | undefined>[]> {
+        let picks: IAzureQuickPickItem<string | undefined>[] = [];
+        const locationName: string = nonNullProp(nonNullProp(wizardContext, 'location'), 'name');
+        let pairedRegions: string[] = [locationName];
+        pairedRegions = pairedRegions.concat(await this.getPairedRegions(locationName));
+
+        const locations: Location[] = await LocationListStep.getLocations(wizardContext);
```
EricJizbaMSFT: Are you displaying the locations that specifically support app insights or all locations that support function apps?

appservice/src/createAppService/AppInsightsListStep.ts
```diff
@@ -0,0 +1,85 @@
...
+            };
+        } else {
+            return undefined;
+        }
+    }
+
+    private async getQuickPicks(wizardContext: IAppServiceWizardContext): Promise<IAzureQuickPickItem<ApplicationInsightsComponent | string | undefined>[]> {
+
+        const picks: IAzureQuickPickItem<ApplicationInsightsComponent | string | undefined>[] = [{
+            label: localize('newApplicationInsight', '$(plus) Create new application insight'),
```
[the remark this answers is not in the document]
nturinski: The portal tends to use Application Insight and then resource name.  I used the term component because that's how the API refers to them, but I actually don't see that vernacular anywhere in the portal, unfortunately.  I'd guess we should probably call it an `Application Insight resource`

image

image

appservice/src/createAppService/AppInsightsLocationStep.ts
```diff
@@ -0,0 +1,128 @@
...
+            if (regionMappingJson.regions[locationName]) {
+                return regionMappingJson.regions[locationName].pairedRegions;
+            }
+        } catch (error) {
+            // ignore the error
+        }
+        return [];
+    }
+
+    private generalizeLocationName(name: string | undefined): string {
```
EricJizbaMSFT: Copying code I see?

Honestly I'm really not a fan of this entire file. I feel like you should be changing LocationListStep to fit your purposes, not pulling bits and pieces into an entirely new file that only helps in one scenario. For example, it seems like you created this file for a few reasons:
1. You wanted to prompt for location twice
1. You wanted to put recommended locations at the top of the list

Both of those are general things that could be supported for many scenarios outside of App Insights. You're essentially taking a shortcut and not helping those future scenarios.

appservice/src/createAppService/AppInsightsLocationStep.ts
```diff
@@ -0,0 +1,128 @@
...
+    public async prompt(wizardContext: IAppServiceWizardContext): Promise<void> {
+        const options: IAzureQuickPickOptions = { placeHolder: localize('locationAppInsights', 'Select a location for the new App Insights component.'), id: `AppInsightsLocationStep/${wizardContext.subscriptionId}` };
+        wizardContext.newAppInsightsLocation = (await ext.ui.showQuickPick(this.getQuickPicks(wizardContext), options)).label;
+    }
+
+    public shouldPrompt(wizardContext: IAppServiceWizardContext): boolean {
+        return !wizardContext.appInsightsComponent && !wizardContext.newAppInsightsLocation;
+    }
+
+    public async getSupportedLocation(wizardContext: IAppServiceWizardContext, location: Location): Promise<string | undefined> {
```
EricJizbaMSFT: I don't see `getSupportedLocation` called in this file, so it probably shouldn't be in here

appservice/src/createAppService/AppInsightsLocationStep.ts
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    private async getLocations(wizardContext: IAppServiceWizardContext): Promise<Location[] | undefined> {
+        const resourceClient: ResourceManagementClient = createAzureClient(wizardContext, ResourceManagementClient);
+        const supportedRegions: Provider = await resourceClient.providers.get('microsoft.insights');
+        const componentsResourceType: ProviderResourceType | undefined = supportedRegions.resourceTypes && supportedRegions.resourceTypes.find(aiRt => aiRt.resourceType === 'components');
+        if (!!componentsResourceType && !!componentsResourceType.locations) {
+            return componentsResourceType.locations.map((locationDisplayName: string) => {
+                return {
+                    name: this.generalizeLocationName(locationDisplayName),
```
EricJizbaMSFT: Creating your own `Location` object feels pretty hacky

appservice/src/createAppService/AppInsightsLocationStep.ts
nturinski: `getSupportedLocation` isn't called in this file, but it calls two other functions that are.  Of course, this might all be refactored anyway.

appservice/src/createAppService/IAppServiceWizardContext.ts
```diff
@@ -75,13 +75,25 @@ export interface IAppServiceWizardContext extends IResourceGroupWizardContext, I
...
     /**
-     * The name of the new application insights component
+     * The task used to get existing App Insights components.
+     * By specifying this in the context, we can ensure that Azure is only queried once for the entire wizard
+     */
+    appInsightsTask?: Promise<ApplicationInsightsComponentListResult>;
+
+    /**
+     * Boolean indicating that the user opted out of creating an Application inisghts component.
+     * WThis will be defined after `AppInsightsLocationStep.prompt` occurs.
```
EricJizbaMSFT: `WThis`

appservice/src/createAppService/AppInsightsCreateStep.ts
```diff
@@ -20,70 +20,89 @@ export class AppInsightsCreateStep extends AzureWizardExecuteStep<IAppServiceWiz
...
-        return !wizardContext.appInsightsComponent;
+        return !wizardContext.appInsightsComponent && !!wizardContext.newAppInsightsName;
+    }
+
+    // returns the supported location, a location in the region map, or undefined
+    private async getSupportedLocation(wizardContext: IAppServiceWizardContext, location: Location): Promise<string | undefined> {
+        // tslint:disable-next-line: strict-boolean-expressions
+        const locations: string[] = await this.getLocations(wizardContext) || [];
+        const locationName: string = nonNullProp(location, 'name');
+        // still need to do a name step
```
EricJizbaMSFT: What does `still need to do a name step` mean?

appservice/src/createAppService/AppInsightsListStep.ts
```diff
@@ -0,0 +1,82 @@
...
+        }
+    }
+
+    public shouldPrompt(wizardContext: IAppServiceWizardContext): boolean {
+        return !wizardContext.appInsightsComponent;
+    }
+
+    public async getSubWizard(wizardContext: IAppServiceWizardContext): Promise<IWizardOptions<IAppServiceWizardContext> | undefined> {
+        if (!wizardContext.appInsightsComponent && !wizardContext.appInsightsSkip) {
+            return {
```
EricJizbaMSFT: Technically this should probably do LocationListStep.addStep (although it won't make a functional difference with the current setup of our wizard)

appservice/src/createAppService/AppInsightsCreateStep.ts
nturinski: 😅 This was to make sure that I covered this issue https://github.com/microsoft/vscode-azuretools/issues/560

PR #566, opened 2019-08-14
Make LocationListStep private and singleton in Azure Wizard

nturinski:
Discussed with Eric offline, but location being inserted several times by the wizard could be troublesome since we want to have all resources to be within the same location.  

This will prevent steps from adding location steps in the subwizard when it requires a location if the user already had a location step added to the wizard.  If the user forgot, then it'll add a location step for them and not break them.

### DIFF
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -799,13 +799,15 @@ export interface ILocationWizardContext extends ISubscriptionWizardContext {
     location?: Location;
 
     /**
-  * Optional task to describe the subset of locations that should be displayed.
-  * If not specified, all locations supported by the user's subscription will be displayed.
-  */
+     * Optional task to describe the subset of locations that should be displayed.
+     * If not specified, all locations supported by the user's subscription will be displayed.
+     */
     locationsTask?: Promise<{ name?: string }[]>;
 }
 
 export declare class LocationListStep<T extends ILocationWizardContext> extends AzureWizardPromptStep<T> {
+    private constructor();
+
     /**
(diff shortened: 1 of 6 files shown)


ui/index.d.ts
```diff
@@ -799,13 +799,20 @@ export interface ILocationWizardContext extends ISubscriptionWizardContext {
...
     locationsTask?: Promise<{ name?: string }[]>;
 }
 
 export declare class LocationListStep<T extends ILocationWizardContext> extends AzureWizardPromptStep<T> {
+    /**
+     * Adds a LocationListStep to the wizard.  This function will ensure there is only one LocationListStep per wizard context.
+     * @param wizardContext The context of the wizard
+     * @param promptSteps The array of steps to include the LocationListStep to
+     */
+    public static addStep<T extends ILocationWizardContext>(wizardContext: T, promptSteps: AzureWizardPromptStep<T>[]): void;
```
EricJizbaMSFT: Also declare the constructor as private to enforce this

ui/index.d.ts
```diff
@@ -799,13 +799,20 @@ export interface ILocationWizardContext extends ISubscriptionWizardContext {
     location?: Location;
 
     /**
-     * Optional task to describe the subset of locations that should be displayed.
-     * If not specified, all locations supported by the user's subscription will be displayed.
-     */
+  * Optional task to describe the subset of locations that should be displayed.
```
EricJizbaMSFT: nit: change this spacing back

PR #573, opened 2019-08-22
Add getDeployFsPath

nturinski:
The implementation is a little different from the extensions.

I didn't think it made sense for the azuretools package to handle getting the project workspace, so this method relies on the extension passing the fsPath to getDeployFsPath (the old behavior would prompt the user to select a workspace folder).

The extension should now handle this instead pre-deployment.


appservice/src/deploy/getDeployFsPath.ts
```diff
@@ -0,0 +1,61 @@
...
+import * as vscode from 'vscode';
+import { DialogResponses } from 'vscode-azureextensionui';
+import { ext } from '../extensionVariables';
+import { localize } from '../localize';
+
+/**
+ * Appends the deploySubpath setting if the target path matches the root of a workspace folder
+ * If the targetPath is a sub folder instead of the root, leave the targetPath as-is and assume they want that exact folder used
+ */
+export async function getDeployFsPath(targetPath: string, extensionPrefix: string): Promise<string> {
```
EricJizbaMSFT: If your plan is to only copy this function, you need to at least name it accurately (as it used to be): 
https://github.com/microsoft/vscode-azurefunctions/blob/master/src/commands/deploy/getDeployFsPath.ts#L39

appservice/src/deploy/getDeployFsPath.ts
```diff
@@ -0,0 +1,61 @@
...
+                    return fsPathWithSetting;
+                }
+            }
+        }
+    }
+
+    return targetPath;
+}
+
+function isSubpath(expectedParent: string, expectedChild: string): boolean {
```
EricJizbaMSFT: These should be put in a utils file. That's one of the reasons we have "util" files - so that you can just copy the file over to a new repo when you need it

nturinski: Oh, I figured we didn't want to clutter the utils folder with something that is only being used in one file.  Though I guess if I move workspace related code over, then it makes more sense to transfer this to its own file as well.

nturinski:
Well, the main difference between the two is that App Service accepts the file extension as a parameter for Java deployments so it has to check for that extra parameter.  It felt a little weird to me to have vscode workspace related things in the shared package since I figured the extensions handle setting up the projects and whatnot, but I understand your point about the entry points being a hassle to test for

appservice/src/deploy/getDeployFsPath.ts
nturinski: I mean, it still gets the deployFsPath 😬

appservice/src/deploy/getDeployFsPath.ts
```diff
@@ -5,37 +5,65 @@
...
 import { ext } from '../extensionVariables';
 import { localize } from '../localize';
+import { isPathEqual, isSubpath } from '../utils/fs';
+import { getWorkspaceSetting, updateGlobalSetting } from '../utils/settings';
+import * as workspaceUtil from '../utils/workspace';
+
+const deploySubpathSetting: string = 'deploySubpath';
+/**
+ * Entry point can be the workspace folder, the fsPath, or the tree item being deployed to
+ * In App Service, users can deploy specific artificat files (such as .jar) which is handled by selectWorkspaceFile
```
EricJizbaMSFT: artificat

appservice/src/deploy/getDeployFsPath.ts
```diff
@@ -5,37 +5,65 @@
...
+    } else if (vscode.workspace.workspaceFolders && vscode.workspace.workspaceFolders.length === 1) {
+        // If there is only one workspace and it has 'deploySubPath' set - return that value without prompting
+        const folderPath: string = vscode.workspace.workspaceFolders[0].uri.fsPath;
+        const deploySubpath: string | undefined = getWorkspaceSetting(deploySubpathSetting, extensionPrefix, folderPath);
+        if (deploySubpath) {
+            return path.join(folderPath, deploySubpath);
+        }
+    }
+
+    const selectFile: string = localize('selectDeployFile', `Select the ${fileExtension} file to deploy`);
```
EricJizbaMSFT: Make it properly localizable

```
localize('selectDeployFile', `Select the {0} file to deploy`, fileExtension);
```

appservice/src/utils/workspace.ts
```diff
@@ -0,0 +1,73 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as path from 'path';
+import * as vscode from 'vscode';
+import { IAzureQuickPickItem, IAzureUserInput } from 'vscode-azureextensionui';
+import { localize } from '../localize';
+
+export async function selectWorkspaceFolder(ui: IAzureUserInput, placeHolder: string, getSubPath?: (f: vscode.WorkspaceFolder) => string | undefined | Promise<string | undefined>): Promise<string> {
```
EricJizbaMSFT: These should just use `ext.ui` directly instead of passing `ui` around

PR #575, opened 2019-08-29
Add gulp_installVSCodeExtension to dev package

EricJizbaMSFT:
I want to add more testing for the python case in functions, but turns out the tests need the Python extension installed. So I created a generic version of `gulp_installAzureAccount` that can be used for any arbitrary extension

### DIFF
--- a/dev/src/gulp/gulp_installVSCodeExtension.ts
+++ b/dev/src/gulp/gulp_installVSCodeExtension.ts
@@ -1,0 +1,40 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { File } from 'decompress';
+import * as glob from 'glob';
+import * as gulp from 'gulp';
+// tslint:disable-next-line: no-require-imports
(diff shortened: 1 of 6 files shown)


dev/src/gulp/gulp_installVSCodeExtension.ts
```diff
@@ -0,0 +1,40 @@
...
+import * as request from 'request';
+import { Stream } from 'stream';
+import * as buffer from 'vinyl-buffer';
+import * as source from 'vinyl-source-stream';
+
+export function gulp_installVSCodeExtension(version: string, publisherId: string, extensionName: string): Promise<void> | Stream {
+    const extensionId: string = `${publisherId}.${extensionName}`;
+    const extensionPath: string = path.join(os.homedir(), `.vscode/extensions/${extensionId}-${version}`);
+    const existingExtensions: string[] = glob.sync(extensionPath.replace(version, '*'));
+    if (existingExtensions.length === 0) {
```
StephenWeatherford: Why don't we use "code --install-extension"?

EricJizbaMSFT: Interesting idea... Did a quick test and `code` isn't installed on build machines by default. I think it gets installed later in the build process (after all - the tests have to run against VS Code), but I don't feel like spending more time investigating at the moment - this works just fine for tests

PR #576, opened 2019-09-05
Include createSiteAppSettings for web apps

nturinski:
Fixes https://github.com/microsoft/vscode-azureappservice/issues/840

Since it returns an empty NameValue pair if it's a Windows app, this fixes this issue as well.

Had to make some properties optional since they weren't required for web apps, but shouldn't affect overall behavior for Functions.

### DIFF
--- a/appservice/src/createAppService/SiteCreateStep.ts
+++ b/appservice/src/createAppService/SiteCreateStep.ts
@@ -18,21 +18,21 @@ import { AppKind, WebsiteOS } from './AppKind';
 import { IAppServiceWizardContext } from './IAppServiceWizardContext';
 
 export interface IAppSettingsContext {
-    storageConnectionString: string;
-    fileShareName: string;
+    storageConnectionString?: string;
+    fileShareName?: string;
     os: string;
-    runtime: string;
+    runtime?: string;
     aiInstrumentationKey?: string;
 }
 
 export class SiteCreateStep extends AzureWizardExecuteStep<IAppServiceWizardContext> {
     public priority: number = 140;
 
-    private createFunctionAppSettings: ((context: IAppSettingsContext) => Promise<NameValuePair[]>) | undefined;
+    private createSiteAppSettings: ((context: IAppSettingsContext) => Promise<NameValuePair[]>) | undefined;
 
-    public constructor(createFunctionAppSettings?: ((context: IAppSettingsContext) => Promise<NameValuePair[]>)) {
+    public constructor(createSiteAppSettings?: ((context: IAppSettingsContext) => Promise<NameValuePair[]>)) {
         super();
-        this.createFunctionAppSettings = createFunctionAppSettings;
+        this.createSiteAppSettings = createSiteAppSettings;
     }
 
     public async execute(wizardContext: IAppServiceWizardContext, progress: Progress<{ message?: string; increment?: number }>): Promise<void> {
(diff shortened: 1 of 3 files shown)


appservice/src/createAppService/SiteCreateStep.ts
```diff
@@ -87,17 +97,16 @@ export class SiteCreateStep extends AzureWizardExecuteStep<IAppServiceWizardCont
...
-                    storageConnectionString,
-                    fileShareName,
-                    // tslint:disable-next-line:no-non-null-assertion
-                    os: nonNullProp(wizardContext, 'newSiteOS'),
-                    // tslint:disable-next-line:no-non-null-assertion
-                    runtime: nonNullProp(wizardContext, 'newSiteRuntime'),
-                    // tslint:disable-next-line: strict-boolean-expressions
-                    aiInstrumentationKey: wizardContext.appInsightsComponent && wizardContext.appInsightsComponent ? wizardContext.appInsightsComponent.instrumentationKey : undefined
-                });
+            if (this.createSiteAppSettings) {
```
EricJizbaMSFT: Move this out of the if/else so that you have less duplicate code. Something like this:
```
let storageConnectionString: string | undefined;
let fileShareName: string | undefined;
if (wizardContext.newSiteKind === AppKind.app) {
    ...
} else {
    ...
}

if (this.createSiteAppSettings) {
    newSiteConfig.appSettings = await this.createSiteAppSettings(
    {
        ...
    }
}
```

appservice/src/createAppService/SiteCreateStep.ts
```diff
@@ -18,21 +18,21 @@ import { AppKind, WebsiteOS } from './AppKind';
 import { IAppServiceWizardContext } from './IAppServiceWizardContext';
 
 export interface IAppSettingsContext {
-    storageConnectionString: string;
-    fileShareName: string;
```
EricJizbaMSFT: This feels like a breaking change so update the version accordingly. It might not change functionally, but it'll at least cause a build break

appservice/src/createAppService/SiteCreateStep.ts
```diff
@@ -18,21 +18,21 @@ import { AppKind, WebsiteOS } from './AppKind';
...
 }
 
 export class SiteCreateStep extends AzureWizardExecuteStep<IAppServiceWizardContext> {
     public priority: number = 140;
 
-    private createFunctionAppSettings: ((context: IAppSettingsContext) => Promise<NameValuePair[]>) | undefined;
+    private createSiteAppSettings: ((context: IAppSettingsContext) => Promise<NameValuePair[]>) | undefined;
 
-    public constructor(createFunctionAppSettings?: ((context: IAppSettingsContext) => Promise<NameValuePair[]>)) {
+    public constructor(createSiteAppSettings?: ((context: IAppSettingsContext) => Promise<NameValuePair[]>)) {
```
EricJizbaMSFT: Is `createSiteAppSettings` still optional or can we make it required now? Specifically, the bug is only fixed if you pass in `createSiteAppSettings`, right? So maybe we should enforce that?

appservice/src/createAppService/SiteCreateStep.ts
nturinski: I'm purely asking just for edification since I'll have to bump up the version if createSiteAppSettings is enforced, but I thought that making properties optional would prevent changes from being broken.  How come that would break the build?

EricJizbaMSFT: Because `string | undefined` can't be assigned to `string`. I don't actually know if it'll break the build and the fix would just be to use `nonNullProp` if it does - but it's more of an "on principal" thing. `IAppSettingsContext` used to guarantee it would give `createSiteAppSettings` a `fileShareName` and `storageConnectionString`, and now it no longer makes that guarantee - that seems breaking.

PR #577, opened 2019-09-05
Make `LocationListStep.addStep` more permissive

EricJizbaMSFT:
We don't actually need a full `ILocationWizardContext` just to add the step. This is true in the Docker extension where the subscription step is a part of the wizard.


EricJizbaMSFT:
There's generally not a great way to enforce anything on the wizard context - which is why the wizard steps use `nonNullProp` so often. Definitely room for improvement in that area if you want to take a stab at it :)

PR #592, opened 2019-09-19
Validate new application setting keys

wwlorey:
Related to https://github.com/microsoft/vscode-azureappservice/issues/646

This copies what the portal does for app setting key validation.

Screen Shot 2019-09-19 at 11 32 51 AM

Screen Shot 2019-09-19 at 11 28 01 AM

### DIFF
--- a/appservice/src/tree/AppSettingsTreeItem.ts
+++ b/appservice/src/tree/AppSettingsTreeItem.ts
@@ -14,14 +14,15 @@ import { ISiteTreeRoot } from './ISiteTreeRoot';
 export function validateAppSettingKey(settings: StringDictionary, client: SiteClient, newKey?: string, oldKey?: string): string | undefined {
     newKey = newKey ? newKey : '';
 
-    if (client.isLinux && RegExp('[^\\w\\.]+').test(newKey)) {
+    if (client.isLinux && /[^\w\.]+/.test(newKey)) {
         return 'App setting names can only contain letters, numbers (0-9), periods ("."), and underscores ("_")';
-    }
-    if (newKey.length === 0) {
-        return 'App setting names must have at least one character.';
     }
 
     newKey = newKey.trim();
+    if (newKey.length === 0) {
+        return 'App setting names must have at least one non-whitespace character.';
+    }
+
     oldKey = oldKey ? oldKey.trim().toLowerCase() : oldKey;
     if (settings.properties && newKey.toLowerCase() !== oldKey) {
         for (const key of Object.keys(settings.properties)) {
--- a/appservice/src/tree/AppSettingsTreeItem.ts
+++ b/appservice/src/tree/AppSettingsTreeItem.ts
@@ -6,15 +6,15 @@
 import { StringDictionary } from 'azure-arm-website/lib/models';
 import { AzureParentTreeItem, AzureTreeItem, IActionContext, ICreateChildImplContext, TreeItemIconPath } from 'vscode-azureextensionui';
 import { ext } from '../extensionVariables';
+import { SiteClient } from "../SiteClient";
 import { AppSettingTreeItem } from './AppSettingTreeItem';
 import { getThemedIconPath } from './IconPath';
 import { ISiteTreeRoot } from './ISiteTreeRoot';
 
-export function validateAppSettingKey(settings: StringDictionary, newKey?: string, oldKey?: string, isLinux: boolean = false): string | undefined {
-    // Default to Windows webapp (isLinux = false) to let users access looser name validation, even if saving the key fails
+export function validateAppSettingKey(settings: StringDictionary, client: SiteClient, newKey?: string, oldKey?: string): string | undefined {
     newKey = newKey ? newKey : '';
 
-    if (isLinux && RegExp('[^\\w\\.]+').test(newKey)) {
+    if (client.isLinux && RegExp('[^\\w\\.]+').test(newKey)) {
         return 'App setting names can only contain letters, numbers (0-9), periods ("."), and underscores ("_")';
     }
     if (newKey.length === 0) {
@@ -104,7 +104,7 @@ export class AppSettingsTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
         const settings: StringDictionary = JSON.parse(JSON.stringify(await this.ensureSettings(context)));
         const newKey: string = await ext.ui.showInputBox({
             prompt: 'Enter new setting key',
-            validateInput: (v?: string): string | undefined => validateAppSettingKey(settings, v, undefined, this.root.client.isLinux)
+            validateInput: (v?: string): string | undefined => validateAppSettingKey(settings, this.root.client, v)
         });
 
         const newValue: string = await ext.ui.showInputBox({
--- a/appservice/src/tree/AppSettingsTreeItem.ts
+++ b/appservice/src/tree/AppSettingsTreeItem.ts
@@ -19,8 +19,7 @@ export function validateAppSettingKey(settings: StringDictionary, newKey?: strin
         return 'App setting names must have at least one character.';
     }
 
-    newKey = newKey.trim();
-    oldKey = oldKey ? oldKey.trim().toLowerCase() : oldKey;
+    oldKey = oldKey ? oldKey.toLowerCase() : oldKey;
     if (settings.properties && newKey.toLowerCase() !== oldKey) {
         for (const key of Object.keys(settings.properties)) {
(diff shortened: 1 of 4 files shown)


appservice/src/tree/AppSettingsTreeItem.ts
```diff
@@ -11,11 +11,16 @@ import { getThemedIconPath } from './IconPath';
...
+    newKey = newKey ? newKey : '';
+    if (RegExp('[^\\w\\.]+').test(newKey)) {
+        return 'App setting names can only contain letters, numbers (0-9), periods ("."), and underscores ("_")';
+    }
     if (newKey.length === 0) {
-        return 'Key must have at least one non-whitespace character.';
+        return 'App setting names must have at least one character.';
     }
+
+    newKey = newKey.trim();
```
wwlorey: I'm not sure if whitespace needed to be trimmed from `newKey` here. It seemed safer to keep it

appservice/src/tree/AppSettingsTreeItem.ts
```diff
@@ -10,12 +10,17 @@ import { AppSettingTreeItem } from './AppSettingTreeItem';
 import { getThemedIconPath } from './IconPath';
 import { ISiteTreeRoot } from './ISiteTreeRoot';
 
-export function validateAppSettingKey(settings: StringDictionary, newKey?: string, oldKey?: string): string | undefined {
-    newKey = newKey ? newKey.trim() : '';
-    oldKey = oldKey ? oldKey.trim().toLowerCase() : oldKey;
+export function validateAppSettingKey(settings: StringDictionary, newKey?: string, oldKey?: string, isLinux?: boolean): string | undefined {
```
nturinski: Set this as a default to false so it's explicit that we are assuming it's Windows.

`isLinux = false: boolean`

Add a clarifying comment as to WHY we'd rather assume it's Windows instead of Linux

appservice/src/tree/AppSettingsTreeItem.ts
```diff
@@ -10,12 +10,19 @@ import { AppSettingTreeItem } from './AppSettingTreeItem';
...
 import { ISiteTreeRoot } from './ISiteTreeRoot';
 
-export function validateAppSettingKey(settings: StringDictionary, newKey?: string, oldKey?: string): string | undefined {
-    newKey = newKey ? newKey.trim() : '';
-    oldKey = oldKey ? oldKey.trim().toLowerCase() : oldKey;
+export function validateAppSettingKey(settings: StringDictionary, newKey?: string, oldKey?: string, isLinux: boolean = false): string | undefined {
+    // Default to Windows webapp (isLinux = false) to let users access looser name validation, even if saving the key fails
+    newKey = newKey ? newKey : '';
+
+    if (isLinux && RegExp('[^\\w\\.]+').test(newKey)) {
```
nturinski: Could you link where you got this RegExp from?

appservice/src/tree/AppSettingsTreeItem.ts
```diff
@@ -10,12 +10,19 @@ import { AppSettingTreeItem } from './AppSettingTreeItem';
 import { getThemedIconPath } from './IconPath';
 import { ISiteTreeRoot } from './ISiteTreeRoot';
 
-export function validateAppSettingKey(settings: StringDictionary, newKey?: string, oldKey?: string): string | undefined {
-    newKey = newKey ? newKey.trim() : '';
-    oldKey = oldKey ? oldKey.trim().toLowerCase() : oldKey;
+export function validateAppSettingKey(settings: StringDictionary, newKey?: string, oldKey?: string, isLinux: boolean = false): string | undefined {
+    // Default to Windows webapp (isLinux = false) to let users access looser name validation, even if saving the key fails
```
nturinski: I'm not sure about the wording here.  How about?

`Default to Windows web app because there are less strict constraints on app setting names for Windows.  If the key is invalid, Azure will prevent adding the key with an error.`

EricJizbaMSFT: Why do y'all need a default? Isn't `isLinux` always defined on SiteClient?

appservice/src/tree/AppSettingsTreeItem.ts
```diff
@@ -10,12 +10,19 @@ import { AppSettingTreeItem } from './AppSettingTreeItem';
 import { getThemedIconPath } from './IconPath';
 import { ISiteTreeRoot } from './ISiteTreeRoot';
 
-export function validateAppSettingKey(settings: StringDictionary, newKey?: string, oldKey?: string): string | undefined {
-    newKey = newKey ? newKey.trim() : '';
-    oldKey = oldKey ? oldKey.trim().toLowerCase() : oldKey;
+export function validateAppSettingKey(settings: StringDictionary, newKey?: string, oldKey?: string, isLinux: boolean = false): string | undefined {
```
nturinski: As discussed offline, let's move this into AppSettingsTreeItem and have the extensions use the treeItem to validate.  We won't need the `isLinux` boolean anymore as it will have access to the root treeItem.

wwlorey:
**Offline Feedback**: Move `valideAppSettingKey()` into the `AppSettingsTreeItem` class so we don't have to make assumptions on whether or not the web app is Linux.

wwlorey:
**Offline Feeback**: When the cosmos tree item validates keys, it uses `validateAppSettingKey` and has no reference to `AppSettingsTreeItem`. So we'll go back to the old implementation of separating key validation from the tree item class. 

Instead of passing in `isLinux` to the key validation function, we'll pass in the site client (which gives us `isLinux`) and make it a required parameter so that changes here in the future won't be breaking.

appservice/src/tree/AppSettingsTreeItem.ts
```diff
@@ -6,16 +6,23 @@
...
 import { getThemedIconPath } from './IconPath';
 import { ISiteTreeRoot } from './ISiteTreeRoot';
 
-export function validateAppSettingKey(settings: StringDictionary, newKey?: string, oldKey?: string): string | undefined {
-    newKey = newKey ? newKey.trim() : '';
-    oldKey = oldKey ? oldKey.trim().toLowerCase() : oldKey;
+export function validateAppSettingKey(settings: StringDictionary, client: SiteClient, newKey?: string, oldKey?: string): string | undefined {
+    newKey = newKey ? newKey : '';
+
+    if (client.isLinux && RegExp('[^\\w\\.]+').test(newKey)) {
```
EricJizbaMSFT: nit: Would prefer the shorthand for `RegExp` which is to use slashes like this: `/'[^\\w\\.]+'/.test(...`

appservice/src/tree/AppSettingsTreeItem.ts
```diff
@@ -6,16 +6,23 @@
...
+        return 'App setting names can only contain letters, numbers (0-9), periods ("."), and underscores ("_")';
+    }
     if (newKey.length === 0) {
-        return 'Key must have at least one non-whitespace character.';
+        return 'App setting names must have at least one character.';
```
EricJizbaMSFT: How come you removed "non-whitespace"? Can I have a key that is just whitespace? If so - it will be trimmed to nothing below which seems contradictory

appservice/src/tree/AppSettingsTreeItem.ts
```diff
@@ -6,16 +6,23 @@
...
 import { ext } from '../extensionVariables';
+import { SiteClient } from "../SiteClient";
 import { AppSettingTreeItem } from './AppSettingTreeItem';
 import { getThemedIconPath } from './IconPath';
 import { ISiteTreeRoot } from './ISiteTreeRoot';
 
-export function validateAppSettingKey(settings: StringDictionary, newKey?: string, oldKey?: string): string | undefined {
-    newKey = newKey ? newKey.trim() : '';
-    oldKey = oldKey ? oldKey.trim().toLowerCase() : oldKey;
+export function validateAppSettingKey(settings: StringDictionary, client: SiteClient, newKey?: string, oldKey?: string): string | undefined {
```
EricJizbaMSFT: This is a breaking change to a function exported from this package, so you should update the minor version to 0.46.0

appservice/src/tree/AppSettingsTreeItem.ts
wwlorey: It is possible to have a whitespace only key in the portal. The trim below is for comparison with existing keys but somewhere down the line (outside of this function) we trim the keys anyway so I'll add the 'non-whitespace' bit back in

PR #593, opened 2019-09-25
Update some app service icons

EricJizbaMSFT:
Part of https://github.com/microsoft/vscode-azurefunctions/issues/1494 and https://github.com/microsoft/vscode-azureappservice/issues/1164

See https://github.com/microsoft/vscode-azurefunctions/pull/1513 for screenshots


EricJizbaMSFT:
You know how in typescript you can do `var`, `let`, or `const`? I think it's supposed to be a "constant" in that sense. In other words, a value you set that doesn't change

In any case, it's the name of the file in the vscode-icons repo and I figure it's easier to stay consistent in case they ever update their icons

PR #594, opened 2019-09-25
Maybe update subscription icon?

EricJizbaMSFT:
On one hand:
1. These icons are taken from the vscode-icons repo and are more consistent with the new guidelines
1. The light theme looks better (the yellow key never looked great on light theme)

On the other hand:
1. The new key just looks ugly/boring to me haha
1. The new sign-in icon looks more complicated/blurry

What do you guys think? cc @fiveisprime 

|Before|After|
|---|---|
|Screen Shot 2019-09-25 at 4 00 02 PM|Screen Shot 2019-09-25 at 3 59 32 PM|
|Screen Shot 2019-09-25 at 4 00 14 PM|Screen Shot 2019-09-25 at 3 59 22 PM|
|Screen Shot 2019-09-25 at 4 01 01 PM|Screen Shot 2019-09-25 at 4 00 31 PM|
|Screen Shot 2019-09-25 at 4 00 49 PM|Screen Shot 2019-09-25 at 4 00 41 PM|


nturinski:
I agree; also the new key being monochrome looks odd especially in contrast to the Function App icons.

The new door icon is more "fun" but it's less clear to me what it means as well. I prefer both of the older icons.

fiveisprime:
The blurry sign in arrow is real distracting and I dislike the contrast between the Functions and subscription icons. 

Would it make sense to get a monochrome functions icon or is that breaking some branding guidelines?

fiveisprime:
Thoughts on this, @misolori? I can't tell if I'm biased because I'm really used to seeing the colors..

misolori:
If these are part of the "Azure" icons, I have a hard time justifying changing it based purely on the brand association with it (this goes for any icons from Azure).

If these were generic icons that weren't tied to services, then I'd be fine changing them.

PR #595, opened 2019-09-27
Have getDeployPath return the deployPath and the subDeployPath

nturinski:
Local Git deploys shouldn't respect the subDeployPath setting since it doesn't run preDeployTasks.  This change will let the extensions pick whether or not to use the subDeployPath.

### DIFF
--- a/appservice/src/deploy/getDeployFsPath.ts
+++ b/appservice/src/deploy/getDeployFsPath.ts
@@ -86,6 +86,8 @@ async function appendDeploySubpathSetting(targetPath: string, extensionPrefix: s
 }
 
 export type IDeployPaths = {
+    // the deploy path that the user actually deployed via the extension
     originalDeployFsPath: string,
+    // the deploy path after the deploySubpath setting has been appended
     effectiveDeployFsPath: string
 };
--- a/appservice/src/deploy/getDeployFsPath.ts
+++ b/appservice/src/deploy/getDeployFsPath.ts
@@ -17,17 +17,18 @@ const deploySubpathSetting: string = 'deploySubpath';
  * Entry point can be the workspace folder, the fsPath, or the tree item being deployed to
  * In App Service, users can deploy specific artifact files (such as .jar) which is handled by selectWorkspaceFile
  */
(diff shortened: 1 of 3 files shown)


appservice/src/deploy/getDeployFsPath.ts
```diff
@@ -81,3 +84,8 @@ async function appendDeploySubpathSetting(targetPath: string, extensionPrefix: s
 
     return targetPath;
 }
+
+export type deployPaths = {
+    deployFsPath: string,
+    deployFsSubpath: string
```
EricJizbaMSFT: Based on your names, I would expect deployFsSubpath to be `bin/release/netcoreapp2.0`, but I believe it's the whole path, right? Can we change the names to be more accurate? How about something like this:

originalDeployFsPath
effectiveDeployFsPath

Either way, you should add comments saying what each path is

appservice/src/deploy/getDeployFsPath.ts
```diff
@@ -81,3 +84,8 @@ async function appendDeploySubpathSetting(targetPath: string, extensionPrefix: s
 
     return targetPath;
 }
+
+export type deployPaths = {
```
EricJizbaMSFT: nit: `export type deployPaths` just looks wrong to me and I'm not entirely sure why. A few thoughts:
1. Can we pick a pattern for interfaces vs. types? I guess I'd prefer interfaces since I feel like we use that more often, but I couldn't find any conclusive docs saying they're better so 🤷‍♂️
1. Should this start with a lowercase letter or uppercase? I feel like a type is more similar to an interface or class (which uses uppercase) than a const (which uses lowercase)

appservice/src/deploy/getDeployFsPath.ts
```diff
@@ -81,3 +84,8 @@ async function appendDeploySubpathSetting(targetPath: string, extensionPrefix: s
 
     return targetPath;
 }
+
+export type IDeployPaths = {
+    originalDeployFsPath: string,
+    effectiveDeployFsPath: string
```
EricJizbaMSFT: nit: Please add comments to each of these describing what they are. Even with the better names they're still a little confusing, plus I think comments are more important in our shared npm packages (that could theoretically be used by other people)

PR #596, opened 2019-10-03
Retry when getting deployment logs

EricJizbaMSFT:
I've already seen retry help nightly builds based on [this PR](https://github.com/microsoft/vscode-azurefunctions/pull/1536) and I wanted to do the same thing for listing deployment logs. I believe this will help because we're making a lot of calls to kudu and this should prevent a single intermittent error from causing the whole deployment to be reported as a failure. But I added telemetry to validate that belief and we can remove this logic if we find it doesn't help.

This is a breaking change because `DeploymentTreeItem.redeployDeployment` needs actionContext passed in

Related to https://github.com/microsoft/vscode-azurefunctions/issues/1502

### DIFF
--- a/appservice/src/deploy/waitForDeploymentToComplete.ts
+++ b/appservice/src/deploy/waitForDeploymentToComplete.ts
@@ -3,8 +3,9 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import * as retry from 'p-retry';
 import { CancellationToken } from 'vscode';
-import { IParsedError, parseError } from 'vscode-azureextensionui';
+import { IActionContext, IParsedError, parseError } from 'vscode-azureextensionui';
 import { KuduClient } from 'vscode-azurekudu';
 import { DeployResult, LogEntry } from 'vscode-azurekudu/lib/models';
 import { ext } from '../extensionVariables';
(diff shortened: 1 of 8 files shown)


appservice/src/deploy/waitForDeploymentToComplete.ts
```diff
@@ -24,7 +29,7 @@ export async function waitForDeploymentToComplete(client: SiteClient, expectedId
     const kuduClient: KuduClient = await client.getKuduClient();
 
     while (!token || !token.isCancellationRequested) {
-        [deployment, permanentId, initialStartTime] = await tryGetLatestDeployment(kuduClient, permanentId, initialStartTime, expectedId);
+        [deployment, permanentId, initialStartTime] = await tryGetLatestDeployment(context, kuduClient, permanentId, initialStartTime, expectedId);
```
nturinski: Would it work to just use the `await retry()` logic on this call instead of doing it separately for each case in `tryGetLatestDeployment`?

EricJizbaMSFT: Doing that has a slight risk of retrying something that shouldn't be retried (like a null ref exception). I went back and forth on whether the cleaner code is worth the slight risk, but Imma leave it as-is for now. Ultimately I want to check telemetry after this is released and refactor anyways:
1. If it's not helping, remove this logic
1. If it is helping, add generic retry logic for _all_ kudu calls

PR #598, opened 2019-10-03
Fix slot creation for function apps

EricJizbaMSFT:
Modeled after the experience in the portal

Fixes https://github.com/microsoft/vscode-azurefunctions/issues/1441


appservice/src/createAppService/SiteCreateStep.ts
```diff
@@ -105,6 +100,12 @@ export class SiteCreateStep extends AzureWizardExecuteStep<IAppServiceWizardCont
     }
 }
 
+export function getNewFileShareName(siteName: string): string {
+    const randomLetters: number = 6;
```
nturinski: Why 6?  Is this how the portal does it?

EricJizbaMSFT: Because 6 is my favorite number lol :) I think they might do 4

PR #599, opened 2019-10-04
For telemetry, take `-nightly` out of extension name

bwateratmsft:
We're planning to release a nightly version of the extension on the marketplace. This requires it to have a separate extension ID, but for purposes of telemetry we would not want to upload separately--would require all duplicate GDPR classification, and make queries much more difficult.

The version numbers will be sufficiently different to easily distinguish nightly vs. not when needed.


bwateratmsft:
> Interesting idea. Is this going to be visible in the marketplace (don't know if there's a way not to be).
> 
> Somewhat mixed on having this in the shared code. Another option would be to pass in an optional override argument for the extensionName. But I don't really see any harm in the current code. Eric, thoughts?

Yeah, the plan is to have it on the marketplace, which requires us to have a separate extension ID; but we don't want the telemetry reported under an entirely different extension.

bwateratmsft:
> As mentioned offline, I'm skeptical about the nightly release thing on a broader level, but ultimately @bwateratmsft not gonna block you guys if you want to do it
I'll hold off on submitting until we reach consensus. :)

bwateratmsft:
Closing per our change in plans.

PR #601, opened 2019-10-07
Fix listing functions for slots

EricJizbaMSFT:
Fixes https://github.com/microsoft/vscode-azurefunctions/issues/830

### DIFF
--- a/appservice/src/slotFunctionOperations.ts
+++ b/appservice/src/slotFunctionOperations.ts
@@ -1,0 +1,94 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { FunctionEnvelope, FunctionEnvelopeCollection } from 'azure-arm-website/lib/models';
+import { ISubscriptionContext } from 'vscode-azureextensionui';
+import { requestUtils } from './utils/requestUtils';
+
+/**
+ * Temporary workaround because this isn't in azure sdk yet
+ */
+export async function listFunctionsSlot(subscription: ISubscriptionContext, id: string): Promise<FunctionEnvelopeCollection> {
+    const request: requestUtils.Request = await requestUtils.getDefaultAzureRequest(getUrlPath(id), subscription);
+    const response: IRawFunctionEnvelopeCollection = <IRawFunctionEnvelopeCollection>JSON.parse(await requestUtils.sendRequest(request));
+    const result: {}[] = response.value.map(convertRawResource);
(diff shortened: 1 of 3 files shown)


appservice/src/slotFunctionOperations.ts
```diff
@@ -0,0 +1,94 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { FunctionEnvelope, FunctionEnvelopeCollection } from 'azure-arm-website/lib/models';
+import { ISubscriptionContext } from 'vscode-azureextensionui';
+import { requestUtils } from './utils/requestUtils';
+
+/**
+ * Temporary workaround because this isn't in azure sdk yet
```
nturinski: Issue links in Function repo?

appservice/src/slotFunctionOperations.ts
```diff
@@ -0,0 +1,94 @@
...
+/**
+ * Mimics what the azure sdk does under the covers to create the returned object
+ */
+function convertRawResource(resource: IRawAzureResource): { [key: string]: string | undefined } {
+    const result: { [key: string]: string | undefined } = {};
+    for (const key of Object.keys(resource.properties)) {
+        result[convertPropertyName(key)] = convertPropertyValue(resource.properties[key]);
+    }
+
+    result.name = resource.name;
```
nturinski: Is this not covered by the for loop above?

appservice/src/slotFunctionOperations.ts
```diff
@@ -0,0 +1,94 @@
...
+
+    return result;
+}
+
+/**
+ * Converts property name like "function_app_id" to "functionAppId"
+ */
+function convertPropertyName(name: string): string {
+    // tslint:disable-next-line: no-constant-condition
+    while (true) {
```
nturinski: I think you could use this https://www.designcise.com/web/tutorial/how-to-replace-all-occurrences-of-a-word-in-a-javascript-string instead of using a while loop.

appservice/src/slotFunctionOperations.ts
EricJizbaMSFT: You mean the azure sdk repo? I don't see the point in filing an issue when we're on the deprecated version of the sdk

nturinski: I meant having issues in the Functions repo so we remember to change these.  Though I guess we'll have to change most calls when we change APIs.

EricJizbaMSFT: Ah - added a comment: https://github.com/microsoft/vscode-azurefunctions/issues/1083#issuecomment-539223138

PR #602, opened 2019-10-07
Fix parseError for Azure StatusCodeError

EricJizbaMSFT:
Previously the error would just be `Failed with code "401".`, which isn't very helpful. These type of errors happen for me when we create our own Azure request instead of using the Azure SDK (e.g. https://github.com/microsoft/vscode-azuretools/pull/601)

### DIFF
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -1,6 +1,6 @@
 {
     "name": "vscode-azureextensionui",
-    "version": "0.28.1",
+    "version": "0.28.2",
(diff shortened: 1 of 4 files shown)


EricJizbaMSFT:
We've been pretty good at keeping the unit tests up to date for `parseError`, so I'm pretty sure this logic is fine (aka doesn't break any known error formats). If we every run into an error like that, we'll just have to fix it at that time and add another unit test :)

PR #605, opened 2019-10-08
Make elastic premium plan more discoverable

EricJizbaMSFT:
Rather than display elastic premium as a part of "App Service Plan", it will be another option when picking between "Consumption" and "App Service Plan"
Screen Shot 2019-10-08 at 11 42 58 AM

Fixes https://github.com/microsoft/vscode-azurefunctions/issues/1544

### DIFF
--- a/appservice/src/createAppService/AppServicePlanSkuStep.ts
+++ b/appservice/src/createAppService/AppServicePlanSkuStep.ts
@@ -4,8 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { SkuDescription } from 'azure-arm-website/lib/models';
-import { AzureWizardPromptStep } from 'vscode-azureextensionui';
-import { IAzureQuickPickItem } from 'vscode-azureextensionui';
+import { AzureWizardPromptStep, IAzureQuickPickItem } from 'vscode-azureextensionui';
 import { ext } from '../extensionVariables';
 import { localize } from '../localize';
 import { nonNullProp } from '../utils/nonNull';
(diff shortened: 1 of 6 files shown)


appservice/src/createAppService/SiteHostingPlanStep.ts
```diff
@@ -6,17 +6,24 @@
...
 import { AppServicePlanListStep } from './AppServicePlanListStep';
 import { IAppServiceWizardContext } from './IAppServiceWizardContext';
 
 export class SiteHostingPlanStep extends AzureWizardPromptStep<IAppServiceWizardContext> {
     public async prompt(wizardContext: IAppServiceWizardContext): Promise<void> {
         const placeHolder: string = localize('selectHostingPlan', 'Select a hosting plan.');
-        const picks: IAzureQuickPickItem<boolean>[] = [
-            { label: localize('consumption', 'Consumption Plan'), data: true },
-            { label: localize('dedicated', 'App Service Plan'), data: false }
+        const picks: IAzureQuickPickItem<[boolean, RegExp | undefined]>[] = [
```
nturinski: Would it make sense to make an enum that has the different RegExps that would be used as a filter? 

(E is elastic and EP is elastic premium?)

appservice/src/createAppService/AppServicePlanSkuStep.ts
```diff
@@ -15,10 +14,14 @@ import { IAppServiceWizardContext } from './IAppServiceWizardContext';
 
 export class AppServicePlanSkuStep extends AzureWizardPromptStep<IAppServiceWizardContext> {
     public async prompt(wizardContext: IAppServiceWizardContext): Promise<void> {
-        const skus: SkuDescription[] = this.getCommonSkus();
+        let skus: SkuDescription[] = this.getCommonSkus();
         if (wizardContext.newSiteOS === WebsiteOS.windows && wizardContext.newSiteKind === AppKind.functionapp) {
             skus.push(...this.getElasticPremiumSkus());
+        }
 
+        const regExp: RegExp | undefined = wizardContext.planSkuFamilyFilter;
```
nturinski: nit: Dat variable name... 🙄

appservice/src/createAppService/SiteHostingPlanStep.ts
EricJizbaMSFT: Not sure what you mean. Are you saying an enum made up of RegExps? I don't think that's a thing

Also I don't think this is a common scenario, so no I don't think it's worth it

PR #608, opened 2019-10-09
Include subfolders of a symlinked folder

EricJizbaMSFT:
Fixes https://github.com/microsoft/vscode-azurefunctions/issues/1289

### DIFF
--- a/appservice/src/FileUtilities.ts
+++ b/appservice/src/FileUtilities.ts
@@ -66,12 +66,17 @@ export async function zipDirectoryGitignore(folderPath: string, gitignoreName: s
     });
 }
 
+const commonGlobSettings: {} = {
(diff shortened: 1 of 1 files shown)


appservice/src/FileUtilities.ts
```diff
@@ -66,12 +66,17 @@ export async function zipDirectoryGitignore(folderPath: string, gitignoreName: s
     });
 }
 
+const commonGlobSettings: {} = {
+    dot: true, // Include paths starting with '.'
+    nodir: true, // required for symlinks https://github.com/archiverjs/node-archiver/issues/311#issuecomment-445924055
```
EricJizbaMSFT: @wwlorey might be worth double checking storage also supports symlinked folders since you were just messing with the glob settings

PR #609, opened 2019-10-09
Sync triggers more often for function app deploy

EricJizbaMSFT:
More work in my endeavor to make deployments/nightly builds more reliable. Seems like linux consumption remote build in particular needs to wait 10 seconds and then sync triggers. Rather than try and figure out exactly which plans/os's don't need syncing, I just look through the deploy logs for a line like this:

> 3:04:08 PM emjjs1: Syncing 2 function triggers with payload size 158 bytes successful.

### DIFF
--- a/appservice/src/deploy/deployZip.ts
+++ b/appservice/src/deploy/deployZip.ts
@@ -14,6 +14,7 @@ import * as FileUtilities from '../FileUtilities';
 import { localize } from '../localize';
 import { SiteClient } from '../SiteClient';
 import { deployToStorageAccount } from './deployToStorageAccount';
+import { syncTriggersPostDeploy } from './syncTriggersPostDeploy';
 import { waitForDeploymentToComplete } from './waitForDeploymentToComplete';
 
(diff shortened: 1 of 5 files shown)


appservice/src/deploy/deployZip.ts
```diff
@@ -43,24 +44,34 @@ export async function deployZip(context: IActionContext, client: SiteClient, fsP
...
-        // This delay is a temporary stopgap that should be resolved with the new pipelines
-        await delayFirstWebAppDeploy(client, asp);
+        let shouldSyncTriggers: boolean;
+        if (useStorageAccountDeploy) {
+            await deployToStorageAccount(client, zipFilePath);
+            shouldSyncTriggers = true;
+        } else {
+            const kuduClient: KuduClient = await client.getKuduClient();
+            await kuduClient.pushDeployment.zipPushDeploy(fs.createReadStream(zipFilePath), { isAsync: true, author: 'VS Code' });
+            const fullLog: string = await waitForDeploymentToComplete(context, client);
```
nturinski: nit: Not sure how I feel about the function's name since it also returns the full deployment log now

EricJizbaMSFT: I don't disagree - but not gonna deal with it this close to a release

PR #610, opened 2019-10-10
Show elastic premium for linux

EricJizbaMSFT:
I originally thought elastic premium wasn't supported on linux, but turns out it just has a much smaller list of locations. Added elastic premium as an option on linux and added better logic to filter locations

Breaking change was to move setLocationsTask to it's own function

Fixes https://github.com/microsoft/vscode-azurefunctions/issues/1249

### DIFF
--- a/appservice/src/index.ts
+++ b/appservice/src/index.ts
@@ -4,29 +4,32 @@
  *--------------------------------------------------------------------------------------------*/
 
 export * from './confirmOverwriteSettings';
-export { AppKind, LinuxRuntimes, WebsiteOS } from './createAppService/AppKind';
 export * from './createAppService/AppInsightsCreateStep';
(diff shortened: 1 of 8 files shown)


appservice/src/index.ts
```diff
@@ -36,5 +39,5 @@ export * from './tree/DeploymentsTreeItem';
 export * from './tree/DeploymentTreeItem';
 export * from './tree/ISiteTreeRoot';
 export * from './TunnelProxy';
-export { registerAppServiceExtensionVariables } from './extensionVariables';
-export { IConnectToGitHubWizardContext } from './github/IConnectToGitHubWizardContext';
+
```
EricJizbaMSFT: source.organizeImports is really annoying for files like these, but I like it for other files. Anyways only actual change is adding setLocationsTask

PR #617, opened 2019-10-30
Fix casing of "Read Only" 📖

EricJizbaMSFT:
The casing here didn't match what I already had in functions:
Screen Shot 2019-10-30 at 8 59 40 AM


nturinski:
🤣 that emoji.  Don't you need to change it for Functions and Proxies as well?

EricJizbaMSFT:
Yeah but that's in the functions repo

PR #619, opened 2019-10-30
Improve retry attempt telemetry

EricJizbaMSFT:
The current telemetry only tells us the attempt number for the _last_ call in the loop. The whole point of this telemetry is to tell us how often a retry helps, so I changed it to be the max attempt of all calls

### DIFF
--- a/appservice/src/deploy/waitForDeploymentToComplete.ts
+++ b/appservice/src/deploy/waitForDeploymentToComplete.ts
@@ -45,6 +45,7 @@ export async function waitForDeploymentToComplete(context: IActionContext, clien
             logEntries = await retry(
                 async (attempt: number) => {
                     addAttemptTelemetry(context, 'getLogEntry', attempt);
+                    throw new Error('test');
                     // tslint:disable-next-line: no-non-null-assertion
                     return <LogEntry[]>await kuduClient.deployment.getLogEntry(deployment!.id!);
                 },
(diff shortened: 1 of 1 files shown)


appservice/src/deploy/waitForDeploymentToComplete.ts
```diff
@@ -44,7 +44,7 @@ export async function waitForDeploymentToComplete(context: IActionContext, clien
         try {
             logEntries = await retry(
                 async (attempt: number) => {
-                    context.telemetry.properties.getLogEntryAttempt = attempt.toString();
+                    addAttemptTelemetry(context, 'getLogEntry', attempt);
```
wwlorey: Would it be too extra to use an enum for methodName?

appservice/src/deploy/waitForDeploymentToComplete.ts
```diff
@@ -191,3 +191,11 @@ async function tryGetLatestDeployment(context: IActionContext, kuduClient: KuduC
 
     return [deployment, permanentId, initialStartTime];
 }
+
+function addAttemptTelemetry(context: IActionContext, methodName: string, attempt: number): void {
+    const key: string = methodName + 'MaxAttempt';
+    const existingValue: number | undefined = context.telemetry.measurements[key];
```
wwlorey: This is probably fine as it is but I was a bit confused reading it. Maybe this would help?

```suggestion
    const existingAttempt: number | undefined = context.telemetry.measurements[key];
```

appservice/src/deploy/waitForDeploymentToComplete.ts
EricJizbaMSFT: Yeah

PR #622, opened 2019-11-01
Remove update editor  🔼 📅📋 on updateRemote

nturinski:
Fixes https://github.com/microsoft/vscode-azureappservice/issues/1179

The editor used to update the active document with the new text once it updated the file in Azure.  I don't really understand the point of this as you have just updated the remote with your local copy, so it should be the same.

This was causing an issue that would overwrite users' documents if they happened to open a different file after starting an upload.

### DIFF
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -1,6 +1,6 @@
 {
     "name": "vscode-azureextensionui",
-    "version": "0.28.2",
+    "version": "0.28.3",
     "lockfileVersion": 1,
     "requires": true,
     "dependencies": {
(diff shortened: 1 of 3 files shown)


EricJizbaMSFT:
> The editor used to update the active document with the new text once it updated the file in Azure. I don't really understand the point of this as you have just updated the remote with your local copy, so it should be the same.

So I looked into this and I bet some of this code was copied from Cosmos. I believe in that case the cosmos documents have timestamp or etag in the document itself, so we needed to update the editor to refresh those values.

I agree this fix would help App Service - but this change is in the UI package so I'm not sure if it would break something else. Maybe take a look at what Cosmos is doing here and see if that would help: https://github.com/microsoft/vscode-cosmosdb/blob/master/src/CosmosEditorManager.ts#L99

PR #624, opened 2019-11-04
Add a suppressReportIssue option

bwateratmsft:
This seems like a nice feature, particularly for failures in `docker build` and `docker run` in debugging tasks which are 95+% user fault...

### DIFF
--- a/ui/src/callWithTelemetryAndErrorHandling.ts
+++ b/ui/src/callWithTelemetryAndErrorHandling.ts
@@ -94,8 +94,13 @@ function handleError(context: IActionContext, callbackId: string, error: unknown
             message = errorData.message;
         }
 
+        const items: MessageItem[] =
+            context.errorHandling.suppressReportIssue ?
+                [] :
+                [DialogResponses.reportAnIssue];
+
         // don't wait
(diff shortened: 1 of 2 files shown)


bwateratmsft:
@nturinski I've noticed that the build pipelines keep "losing" the macOS job, and it times out after an hour. It's happening across branches and build definitions (vscode-docker, too). Any idea what's going on?

nturinski:
@bwateratmsft Not really sure, but had noticed it has been happening across our repos too.

ui/src/callWithTelemetryAndErrorHandling.ts
```diff
@@ -94,8 +94,13 @@ function handleError(context: IActionContext, callbackId: string, error: unknown
             message = errorData.message;
         }
 
+        const items: MessageItem[] =
```
nturinski: Could you construct just an empty array and then push in the buttons as we need them?  I was thinking for the future, it'd be easier to push extra buttons based on errorHandling conditions if we do it that way.

nturinski:
Oh yeah, you'll have to bump the package.json version and then run an `npm install` to update the package-lock.json

PR #625, opened 2019-11-05
Remove SiteCreateStep 🚧

EricJizbaMSFT:
Related to https://github.com/microsoft/vscode-azurefunctions/pull/1627, moving SiteCreateStep into functions and app service repo since the amount of different code is much more than the amount of shared code at this point.

Also moved the hosting plan step since only functions uses that. I left the runtime step in here since hopefully functions will use the availableStacks api eventually.


appservice/src/createAppService/AppServicePlanListStep.ts
```diff
@@ -39,6 +39,7 @@ export class AppServicePlanListStep extends AzureWizardPromptStep<IAppServiceWiz
         const options: IAzureQuickPickOptions = { placeHolder: localize('selectPlan', 'Select a {0} App Service plan.', getWebsiteOSDisplayName(nonNullProp(wizardContext, 'newSiteOS'))), id: `AppServicePlanListStep/${wizardContext.subscriptionId}` };
         wizardContext.plan = (await ext.ui.showQuickPick(this.getQuickPicks(wizardContext), options)).data;
 
+        wizardContext.telemetry.properties.newPlan = String(!wizardContext.plan);
```
nturinski: Is there any benefit to this over `toString()`?

EricJizbaMSFT: `String(value)` works even if value is null/undefined so as a general rule I do it more often than `value.toString()`

PR #628, opened 2019-11-15
Minor fix for retrying getLogEntry

EricJizbaMSFT:
Based on telemetry, this retry loop in particular is getting to 5 (the max) most often. I want to move the catch block inside the retry logic so that we're not retrying for the situation described in the comment


EricJizbaMSFT:
Discussed offline

PR #629, opened 2019-11-19
actionContext.errorHandling.issueProperties, paste long issues to clipboard, error.category

StephenWeatherford:
1) Support actionContext.errorHandling.issueProperties
2) Paste to clipboard when issue body is long
3)  Support error.category

Comments:
a) I'm actually not 100% sure whether error.category will be helpful...  May have to experiment.  I could delay this part of the checkin.
b) I'm mixed on whether issueProperties should be on the error object or in actionContext, since in actual usage frequently you would need to place the property into the actionContext when an error occurs (or in a finally clause).  But since you'd have to do that if it were on the error object as well, going with Eric's original suggestion here.


StephenWeatherford:
I'm actually not 100% sure whether error.category will be helpful...

ui/src/callWithTelemetryAndErrorHandling.ts
```diff
@@ -34,7 +34,8 @@ function initContext(): [number, IActionContext] {
         },
```
EricJizbaMSFT: Shouldn't you be adding category to telemetry somewhere in this file?

StephenWeatherford: :-)  I think I'm going to roll it back for now anyway.  Most of the errors which I wish had categories are coming from external sources so it doesn't help in those scenarios.  Might add later.

PR #635, opened 2019-12-06
Handle deployment message as JSON in App service DeploymentTreeItem

dikhakha:
Scenario: The deployment message field is set as JSON string when logs are published from GitHub actions , VS Code, Azure Pipelines. 

PR: This PR handles parsing the JSON message to ensure that Deployment tree item have short string message and not JSON string.

Before: 
image

After: 
image


appservice/src/tree/DeploymentTreeItem.ts
```diff
@@ -156,6 +156,15 @@ export class DeploymentTreeItem extends AzureTreeItem<ISiteTreeRoot> {
         }
     }
 
+    private getDeploymentMessage(message: string): string {
```
EricJizbaMSFT: Now that you created a `getDeploymentMessage` method, I'd prefer to move all message-related logic into this method.

Something like this:
```typescript
private getDeploymentMessage(deployResult: DeployResult): string {
    let message: string = nonNullProp(deployResult, 'message');
    try {
        const messageJSON: { message?: string } = <{ message?: string }>JSON.parse(message);
        if (messageJSON.message) {
            message = messageJSON.message;
        }
    } catch {
        // ignore and assume message was not in JSON format
    }

    const firstLine: string = this.getFirstLine(message);
    // truncate long messages and add "..."
    return firstLine.length > 50 ? `${firstLine.substring(0, 50)}...` : firstLine;
}
```
(I didn't test this, but it was pretty simple copy-paste. Please double check what I did)

dikhakha: Sure, will update it.

dikhakha: Done. I have updated as suggested. Thanks for your review !

PR #637, opened 2019-12-11
Don't 🚫 mangle 🤕 function names when webpacking 📦

wwlorey:
Related to https://github.com/microsoft/vscode-azurestorage/issues/525

The above issue was caused by identifiers misaligning (due to webpack mangling function names) in [@azure/core-http](https://github.com/Azure/azure-sdk-for-js/tree/master/sdk/core/core-http).

@StephenWeatherford  I know you implemented the original webpack config logic, do you have any comments on this change?

### DIFF
--- a/dev/package-lock.json
+++ b/dev/package-lock.json
@@ -1,6 +1,6 @@
 {
     "name": "vscode-azureextensiondev",
-    "version": "0.2.3",
+    "version": "0.2.4",
     "lockfileVersion": 1,
     "requires": true,
     "dependencies": {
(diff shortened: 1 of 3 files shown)


StephenWeatherford:
Why are we just seeing this now?

StephenWeatherford:
(The reason for mangling names is mostly for size of module load, probably doesn't make much of a difference is my guess, and does make debugging easier.)

wwlorey:
This problem happened when we pulled in the new @azure/storage-blob SDK. We were thinking that the SDK team changed the way their new packages are webpacked which interfered with our webpack config.

StephenWeatherford:
Did it greatly affect the number of files that end up in the distribution from before?  Might be good to do some quick timings for loading.

wwlorey:
We only saw a module size increase from 3.3MB to 3.7MB as opposed to >9MB with no minimizing

PR #639, opened 2019-12-19
Get latest extension state during refresh

ravipal:
vs code docker extension calls refresh on the Azure Registry node which needs to reflect the latest extension state.

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -1,7 +1,7 @@
 {
     "name": "vscode-azureextensionui",
     "author": "Microsoft Corporation",
-    "version": "0.29.0",
+    "version": "0.30.0",
     "description": "Common UI tools for developing Azure extensions for VS Code",
     "tags": [
         "azure",
--- a/ui/package.json
+++ b/ui/package.json
@@ -1,7 +1,7 @@
 {
     "name": "vscode-azureextensionui",
     "author": "Microsoft Corporation",
-    "version": "0.30.0",
+    "version": "0.29.1",
     "description": "Common UI tools for developing Azure extensions for VS Code",
     "tags": [
         "azure",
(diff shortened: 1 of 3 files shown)


ui/src/treeDataProvider/AzureAccountTreeItemBase.ts
```diff
@@ -59,6 +61,8 @@ export abstract class AzureAccountTreeItemBase extends AzExtParentTreeItem imple
     }
 
     public async loadMoreChildrenImpl(_clearCache: boolean, context: types.IActionContext): Promise<AzExtTreeItem[]> {
+        // Refresh the AzureAccount, to handle the Azure account extension installed/uninstalled between refreshes
+        this._azureAccountTask = this.loadAzureAccount(this._testAccount);
```
EricJizbaMSFT: I believe this means `loadAzureAccount` will subscribe to some events multiple times. Can you double check and make sure we only subscribe to those events once

ui/package.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "vscode-azureextensionui",
     "author": "Microsoft Corporation",
-    "version": "0.29.0",
+    "version": "0.30.0",
```
EricJizbaMSFT: tl;dr please update to 0.29.1 instead. Also update the version in package-lock.json

More explanation:
Normally, you're [supposed to bump](https://semver.org/) the major version for breaking changes. Since we're not ready to commit to "1.0.0" yet, we shift the meaning of each position down. Thus we bump the minor version for breaking changes and the patch version for anything else. Since you did not introduce any breaking changes, you don't need to bump the minor version.

ui/src/treeDataProvider/AzureAccountTreeItemBase.ts
ravipal: Now I call the loadAzureAccount only when the old account is null.

ui/package.json
ravipal: Thanks! Good to know.

PR #643, opened 2020-01-13
Support parsing XML error messages

wwlorey:
This supports the XML format of @azure/storage-blob and @azure/storage-file-share error messages.

Related to https://github.com/microsoft/vscode-azurestorage/issues/538

### DIFF
--- a/ui/src/parseError.ts
+++ b/ui/src/parseError.ts
@@ -120,11 +120,7 @@ function parseIfHtml(message: string): string {
 function parseIfXml(message: string): string {
     const matches: RegExpMatchArray | null = message.match(/<\?xml.*<Message>(.*)/);
     if (matches) {
-        try {
-            return matches[1];
-        } catch (err) {
-            // ignore
-        }
+        return matches[1];
     }
 
     return message;
--- a/ui/src/parseError.ts
+++ b/ui/src/parseError.ts
@@ -118,7 +118,7 @@ function parseIfHtml(message: string): string {
 }
 
 function parseIfXml(message: string): string {
-    const matches: RegExpMatchArray | null = message.match(/<\?xml.*<Message>(.*)/);
+    const matches: RegExpMatchArray | null = message.match(/<\?xml.*<Message>(.*\nRequestId:.*\nTime:.*)<\/Message>/);
     if (matches) {
         return matches[1];
     }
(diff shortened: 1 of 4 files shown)


EricJizbaMSFT:
Add a test or two here: https://github.com/microsoft/vscode-azuretools/blob/master/ui/test/parseError.test.ts

ui/src/parseError.ts
```diff
@@ -68,6 +68,10 @@ export function parseError(error: any): IParsedError {
 
     message = parseIfHtml(message);
 
+    // Azure storage SDK errors are presented in XML
+    // https://github.com/Azure/azure-sdk-for-js/issues/6927
+    message = parseIfXml(message);
```
wwlorey: The original location for this worked but it made more sense to put it next to `parseIfHtml`

ui/test/parseError.test.ts
```diff
@@ -639,4 +639,30 @@ You do not have permission to view this directory or page using the credentials
         assert.strictEqual(pe.message, 'Incorrect authentication credentials.');
         assert.strictEqual(pe.isUserCancelledError, false);
     });
+
+    test('XML errors', () => {
+        const err1: {} = {
+            statusCode: 409,
+            message: `<?xml version="1.0" encoding="utf-8"?><Error><Code>ContainerBeingDeleted</Code><Message>The specified container is being deleted. Try operation later.
+RequestId:35eae91c-a01e-0010-1b5c-ca2979000000
```
EricJizbaMSFT: The previous release also displayed requestId and time. I think RequestId is helpful if the user wants to talk to Azure Support. Also it just makes sense to me to display the whole "Message" property

ui/src/parseError.ts
```diff
@@ -113,6 +117,15 @@ function parseIfHtml(message: string): string {
     return message;
 }
 
+function parseIfXml(message: string): string {
+    const matches: RegExpMatchArray | null = message.match(/<\?xml.*<Message>(.*\nRequestId:.*\nTime:.*)<\/Message>/);
```
EricJizbaMSFT: The RequestId stuff is a little too specific for me. Part of why I wanted the whole "Message" property is so that this code would work even if they changed the format of the message in the future. Perhaps you did this because you had problems matching a multi-line string? If so, try the "s" flag, which means `.` will also match new lines.

Here's the simplest regex that still gets the job done, IMO:
```
/<Message>(.*)<\/Message>/si
```

(I threw in the "i" flag to handle if "message" is lowercase)

wwlorey: I didn't realize that 's' flag was an option, looks very clean 🧼

PR #644, opened 2020-01-14
Don't show 'report issue' for known errors/states

ravipal:
For most of the errors, the error notification is shown to user with an option to open issue, this results into unnecessary issues being reported. This change defines a AzureToolsError with error handling properties. When this error is thrown, the appropriate behavior can be set and will be handled while displaying the notification.

### DIFF
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -583,6 +583,7 @@ export interface IParsedError {
     message: string;
     stack?: string;
     isUserCancelledError: boolean;
+    suppressReportIssue: boolean;
 }
 
 /**
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -97,6 +97,7 @@ export interface ITreeItemPickerContext extends IActionContext {
 
     /**
      * When no item is available for user to pick, this message will be displayed in the error notification.
+     * This will also suppress the report issue button.
      */
     noItemFoundErrorMessage?: string;
 }
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -94,6 +94,11 @@ export interface ITreeItemPickerContext extends IActionContext {
      * For example, this could be used when the command deletes a tree item.
      */
     suppressCreatePick?: boolean;
+
+    /**
+     * When no item is available for user to pick, this message will be displayed in the error notification.
+     */
+    noItemFoundErrorMessage?: string;
 }
 
 /**
@@ -583,7 +588,6 @@ export interface IParsedError {
     message: string;
     stack?: string;
     isUserCancelledError: boolean;
-    actionContext?: Partial<IActionContext>;
 }
 
 /**
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -583,7 +583,7 @@ export interface IParsedError {
     message: string;
     stack?: string;
     isUserCancelledError: boolean;
-    suppressReportIssue: boolean;
+    actionContext?: Partial<IActionContext>;
 }
 
 /**
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -8,7 +8,7 @@ import { Location } from 'azure-arm-resource/lib/subscription/models';
 import { StorageAccount } from 'azure-arm-storage/lib/models';
 import { ServiceClientCredentials } from 'ms-rest';
 import { AzureEnvironment, AzureServiceClientOptions } from 'ms-rest-azure';
-import { Disposable, Event, ExtensionContext, InputBoxOptions, Memento, MessageItem, MessageOptions, OpenDialogOptions, OutputChannel, Progress, QuickPickItem, QuickPickOptions, TextDocument, TreeDataProvider, TreeItem, Uri, ViewColumn } from 'vscode';
+import { Disposable, Event, ExtensionContext, InputBoxOptions, Memento, MessageItem, MessageOptions, OpenDialogOptions, OutputChannel, Progress, QuickPickItem, QuickPickOptions, TextDocument, TreeDataProvider, TreeItem, Uri } from 'vscode';
 import { AzureExtensionApi, AzureExtensionApiProvider } from './api';
 
 export type OpenInPortalOptions = {
@@ -65,11 +65,11 @@ export declare class AzExtTreeDataProvider implements TreeDataProvider<AzExtTree
     public showTreeItemPicker<T extends AzExtTreeItem>(expectedContextValues: string | RegExp | (string | RegExp)[], context: ITreeItemPickerContext, startingTreeItem?: AzExtTreeItem): Promise<T>;
 
     /**
-     * Traverses a tree to find a node matching the given fullId of a tree item. This will not "Load more..."
+     * Traverses a tree to find a node matching the given fullId of a tree item
      * @param fullId The full ID of the tree item
      * @param context The action context
      */
-    public findTreeItem<T extends AzExtTreeItem>(fullId: string, context: IActionContext): Promise<T | undefined>;
+    public findTreeItem<T extends AzExtTreeItem>(fullId: string, context: IFindTreeItemContext): Promise<T | undefined>;
 
     /**
      * Optional method to return the parent of `element`.
@@ -81,6 +81,26 @@ export declare class AzExtTreeDataProvider implements TreeDataProvider<AzExtTree
      * @return Parent of `element`.
      */
     public getParent(treeItem: AzExtTreeItem): Promise<AzExtTreeItem | undefined>;
+}
+
+export interface ILoadingTreeContext extends IActionContext {
+    /**
+     * A custom message to overwrite the default message while loading
+     */
+    loadingMessage?: string;
+
+    /**
+     * Number of seconds to delay before showing the progress message (default is 2)
+     * This is meant to avoid flashing a progress message in cases where it takes less than 2 seconds to load everything
+     */
+    loadingMessageDelay?: number;
+}
+
+export interface IFindTreeItemContext extends ILoadingTreeContext {
+    /**
+     * If true, this will load all children when searching for the tree item
+     */
+    loadAll?: boolean;
 }
 
 export interface ITreeItemPickerContext extends IActionContext {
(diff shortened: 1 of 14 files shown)


ui/src/errors.ts
```diff
@@ -6,21 +6,37 @@
...
-export class NotImplementedError extends Error {
+export class NotImplementedError extends AzureToolsError {
     constructor(methodName: string, obj: object) {
         super(localize('notImplementedError', '"{0}" is not implemented on "{1}".', methodName, obj.constructor.name));
     }
 }
+
+export class NoResouceFoundError extends AzureToolsError {
+    constructor() {
+        super(localize('noResourcesError', 'No matching resources found.'));
```
StephenWeatherford: Seems like we would want to report this error, no?

ui/src/errors.ts
```diff
@@ -6,21 +6,37 @@
 import { localize } from "./localize";
 
 // tslint:disable: max-classes-per-file
+export class AzureToolsError extends Error {
```
StephenWeatherford: This should be an interface

StephenWeatherford: Preferably with a more generic name that describes what it's adding.

ui/src/errors.ts
```diff
@@ -6,21 +6,37 @@
...
+    public suppressDisplay: boolean = false;
+    public suppressReportIssue: boolean = false;
```
EricJizbaMSFT: Instead of individual properties like `suppressReportIssue`, what if we let them specify a `Partial<IActionContext>` so that they can overwrite anything related to error handling _or_ telemetry (Stephen in particular has wanted to specify additional telemetry properties from errors for a while now haha). It could even be called `OverwriteActionContextError` or something

ui/src/errors.ts
ravipal: Based on the usage, at least in the docker extension, such error is shown when user tries to pick an item from treeview that doesn't exist, An example is user tries to stop a container, but there were no running container. This doesn't mean bug in the code that require opening the issue. We still show the error, but wouldn't offer the 'Report Issue' button.

EricJizbaMSFT: I would argue the "bug in the code" is that the error is too generic/confusing for the Docker scenario. Ideally Docker could specify a more specific error like "No running containers found."

ravipal: I agree, that the error message is generic and needs to be improved, but not part of this change.

ui/src/errors.ts
ravipal: Yes, that would be more open. Regarding name, how about ErrorWithActionContext. I don't want to use "Overwrite", because in error handling, we may not always use the value from this error, sometimes the action might define some behavior which is what we need to use.

EricJizbaMSFT: Yeah I was thinking we could do a `Partial<IActionContext>` or it could be a callback/method that takes in the current `IActionContext`. Not sure if people will want to edit the existing or just specify things to overwrite

ui/src/errors.ts
ravipal: Renamed the class.
I kept it as class as it would allow throwing just the base error and doesn't require another derivedclass.

ui/src/errors.ts
StephenWeatherford: @EricJizbaMSFT So do you agree that this should never show the "report an issue" button?

ui/src/parseError.ts
```diff
@@ -68,13 +69,19 @@ export function parseError(error: any): IParsedError {
 
     message = parseIfHtml(message);
 
+    let actionContext: Partial<IActionContext> | undefined;
+    if (error instanceof OverwriteActionContextError) {
```
StephenWeatherford: This is one reason I want this as an interface - you should simply check for the existence of the actionContext property, rather than checking for a specific derived class.  That's what duck typing is for.

ui/src/errors.ts
StephenWeatherford: What I don't like is that you're exposing this functionality via a requirement to derive from a certain class (I really don't even want users to be able to see another class that suggests this).  Instead, it should be an interface (technically not required to have an interface at all but it's usually helpful).  Users can then simply add the actionContext property to any error, even existing ones that have already been thrown.

ui/src/callWithTelemetryAndErrorHandling.ts
```diff
@@ -69,6 +69,8 @@ export async function callWithTelemetryAndErrorHandling<T>(callbackId: string, c
 
 function handleError(context: IActionContext, callbackId: string, error: unknown): void {
     const errorData: IParsedError = parseError(error);
+    context = { ...context, ...errorData.actionContext };
```
StephenWeatherford: Don't think this is what you want, it will completely replace the telemetry and errorHandling objects if they exist on the errorData, for instance completely erasing any existing telemetry properties in the context.  And I'd suggest you need a test for this.

StephenWeatherford:
*   I don't want to use "Overwrite", because in error handling, we may not always use the value from this error, sometimes the action might define some behavior which is what we need to use.
Can you expand on this with an example?

ui/src/callWithTelemetryAndErrorHandling.ts
ravipal: I see. I though it will do a deep merge. Is there any other utility/pattern exist to merge the partial type other than checking each property?

ui/src/errors.ts
EricJizbaMSFT: I think by default it should be shown and each individual extension can hide it if they want. This is a common error - we've had issues reported on Functions, too, and it's a good reminder that I need to write more specific errors for those scenarios

ui/src/callWithTelemetryAndErrorHandling.ts
EricJizbaMSFT: It might be easier if we change the type from `Partial<IActionContext>` to `(context: IActionContext) => void` so that people can edit the context themselves without us having to do a deep merge. Thoughts? Is that cumbersome?

To answer your original question - yes there are tools to do a deep merge, but I don't know offhand. Would have to look it up.

ui/src/parseError.ts
ravipal: Are you suggesting something like below? if so, how would type script ensures that the error.actionContext is of type IActionContext?
`    if (error.actionContext) {
        actionContext = error.actionContext;
    }`

StephenWeatherford: If it has the properties of a type, it *is* of that type.

I'd suggest:
const errorActionContext = (<{ actionContext?: Partial<IActionContext> }>)error.actionContext;
if (typeof errorActionContext === "object") { ... }

Now, this doesn't ensure that the properties under errorActionContext are as expected, although at that point. For instance, if error.actionContext.telemetry were a string, then referencing errorActionContext.telemetry.properties would throw an error.  That'd be pretty unlikely, you can probably just ignore it, or wrap that code in a try/catch. Or you could be explicit with type checking everything as I did above but personally I'd go with try/catch for everything under actionContext.

ui/src/callWithTelemetryAndErrorHandling.ts
StephenWeatherford: Interesting to have a function.  Maybe not a bad idea.

You can search on npmjs.com for deep merge libraries.

ui/src/errors.ts
ravipal: The extension calls ShowTreeItemPicker with certain filter. If no such items found for the given filter, then we show this error. In my opinion, this error should not show 'report issue'. Were there any scenarios in Functions that require user to open issue with this exception?

How would the individual extension overwrite the behavior?

ui/src/callWithTelemetryAndErrorHandling.ts
ravipal: @EricJizbaMSFT, I didn't get your comment on change the type. Can you please elaborate on this?
yes there are several libraries to merge and I can use one, if we are going with that route.

ui/src/errors.ts
ravipal: Not Applicable with the new design

ui/src/parseError.ts
ravipal: Not Applicable with the new design

ui/src/callWithTelemetryAndErrorHandling.ts
ravipal: Not Applicable with the new design

ui/index.d.ts
```diff
@@ -114,6 +114,11 @@ export interface ITreeItemPickerContext extends IActionContext {
      * For example, this could be used when the command deletes a tree item.
      */
     suppressCreatePick?: boolean;
+
+    /**
+     * When no item is available for user to pick, this message will be displayed in the error notification.
```
EricJizbaMSFT: I would add a line here saying it will also suppress the report issue button

ui/src/errors.ts
```diff
@@ -24,3 +25,16 @@ export class NotImplementedError extends Error {
         super(localize('notImplementedError', '"{0}" is not implemented on "{1}".', methodName, obj.constructor.name));
     }
 }
+
+export class NoResouceFoundError extends Error {
+    constructor(errorMessage?: string, context?: IActionContext) {
```
EricJizbaMSFT: Just pass in `context: ITreeItemPickerContext` as the only arg and make it required. Should simplify the code

ravipal: I thought about it. But this will support more scenarios like, a caller to throw this error without an action context. The message and context can be independent.

EricJizbaMSFT: So `index.d.ts` is essentially the public contract of this package. If a class is not listed in there, it can't be used outside of this package. Since `NoResouceFoundError` is only used _within_ the package, I prefer to go for the simpler route. If you wanted to consume this class outside of the package (which requires listing it in `index.d.ts`), then I agree it could be worth the added complexity

EricJizbaMSFT: Not a blocker, though. Approved

ravipal: It should stay within the package. Wouldn't separating the parameter be more open and allows more flexibility?

EricJizbaMSFT: Right I'm saying I vote for simplicity if it's internal and flexibility if it's used externally

PR #650, opened 2020-01-29
Better validation for app names

EricJizbaMSFT:
The default `checkNameAvailability` message sucks:
Screen Shot 2020-01-29 at 1 57 08 PM
Plus it seems innacurate since all the docs I found online (and the portal) say 60 is the max length

We really only need `checkNameAvailability` for the "globally unique" part. Followed same pattern as our [other steps](https://github.com/microsoft/vscode-azuretools/blob/master/ui/src/wizard/StorageAccountNameStep.ts#L41).

Fixes https://github.com/microsoft/vscode-azurefunctions/issues/1804

### DIFF
--- a/appservice/package-lock.json
+++ b/appservice/package-lock.json
@@ -1,6 +1,6 @@
 {
     "name": "vscode-azureappservice",
-    "version": "0.54.1",
+    "version": "0.54.2",
     "lockfileVersion": 1,
     "requires": true,
     "dependencies": {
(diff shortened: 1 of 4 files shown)


nturinski:
Do you know if this applies to App Service too?  I recall having max length issues due to the fact slots would always take the app name and postpend the slot name to that.

EricJizbaMSFT:
Yes these rules also apply to app service. I couldn't find conclusive limits on app name _plus_ slot name, but I did update the slot name validation to match [the docs](https://docs.microsoft.com/en-us/azure/architecture/best-practices/resource-naming) and the portal.

PR #658, opened 2020-02-14
Verify md5 hash during storage account deploy

EricJizbaMSFT:
And a few other naming changes to make us consistent with the func cli. See here for their implementation: https://github.com/Azure/azure-functions-core-tools/blob/master/src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs

Fixes https://github.com/microsoft/vscode-azurefunctions/issues/1877

### DIFF
--- a/appservice/src/deploy/deployToStorageAccount.ts
+++ b/appservice/src/deploy/deployToStorageAccount.ts
@@ -5,9 +5,13 @@
 
 import { StringDictionary } from 'azure-arm-website/lib/models';
 import * as azureStorage from "azure-storage";
+import * as crypto from "crypto";
+import * as fse from 'fs-extra';
+import * as moment from 'moment';
 import { ext } from '../extensionVariables';
 import { localize } from '../localize';
 import { SiteClient } from '../SiteClient';
+import { randomUtils } from '../utils/randomUtils';
 
 /**
  * Method of deployment that is only intended to be used for Linux Consumption Function apps because it doesn't support kudu pushDeployment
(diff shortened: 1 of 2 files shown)


appservice/src/deploy/deployToStorageAccount.ts
```diff
@@ -5,17 +5,23 @@
...
+import { randomUtils } from '../utils/randomUtils';
 
 /**
  * Method of deployment that is only intended to be used for Linux Consumption Function apps because it doesn't support kudu pushDeployment
  * To deploy with Run from Package on a Windows plan, create the app setting "WEBSITE_RUN_FROM_PACKAGE" and set it to "1".
  * Then deploy via "zipdeploy" as usual.
  */
 export async function deployToStorageAccount(client: SiteClient, zipFilePath: string): Promise<void> {
-    const blobName: string = azureStorage.date.secondsFromNow(0).toISOString().replace(/[.,\/#!$%\^&\*;:{}=\-_`~()]/g, '').replace(/\s/g, '');
+    const datePart: string = moment.utc(Date.now()).format('YYYYMMDDHHmmss');
```
nturinski: is it worth adding a new npm package for this?  Didn't the other solution already work?

EricJizbaMSFT: Good question, but `moment` is a very widely used package that was already in our package-lock.json from other dependencies. Plus it's a small package with 0 dependencies of it's own

PR #659, opened 2020-02-14
Refactor sync triggers logic and add ability to stop app in this package

EricJizbaMSFT:
I'm adding support for Java on linux, but it doesn't quite work as-is. Here's what happens today:
1. Stop function app
1. Deploy
1. Sync triggers
1. Start function app

I need to switch 3 and 4 because sync triggers doesn't work while it's stopped. As a part of that, it's easiest to move 1 and 4 into this repo instead of having that in the functions repo.

### DIFF
--- a/appservice/src/deploy/waitForDeploymentToComplete.ts
+++ b/appservice/src/deploy/waitForDeploymentToComplete.ts
@@ -13,12 +13,13 @@ import { localize } from '../localize';
 import { SiteClient } from '../SiteClient';
 import { delay } from '../utils/delay';
 import { nonNullProp } from '../utils/nonNull';
+import { IDeployContext } from './IDeployContext';
 
 // We get a lot of errors reported for deploy (e.g. InternalServerError), but it's possible that's just from listing the logs (which has a lot of calls to kudu), not actually a deploy failure
 // Thus we will retry a few times with exponential backoff. Each "set" of retries will take a max of about 15 seconds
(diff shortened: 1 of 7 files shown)


appservice/src/deploy/waitForDeploymentToComplete.ts
```diff
@@ -13,12 +13,13 @@ import { localize } from '../localize';
...
 import { delay } from '../utils/delay';
 import { nonNullProp } from '../utils/nonNull';
+import { IDeployContext } from './IDeployContext';
 
 // We get a lot of errors reported for deploy (e.g. InternalServerError), but it's possible that's just from listing the logs (which has a lot of calls to kudu), not actually a deploy failure
 // Thus we will retry a few times with exponential backoff. Each "set" of retries will take a max of about 15 seconds
 const retryOptions: retry.Options = { retries: 4, minTimeout: 1000 };
 
-export async function waitForDeploymentToComplete(context: IActionContext, client: SiteClient, expectedId?: string, token?: CancellationToken, pollingInterval: number = 5000): Promise<string> {
+export async function waitForDeploymentToComplete(context: IDeployContext, client: SiteClient, expectedId?: string, token?: CancellationToken, pollingInterval: number = 5000): Promise<void> {
```
nturinski: Any reason to change the return type? I assume nothing was using it?

EricJizbaMSFT: I got rid of the code using the result so there was no point to return anything anymore

PR #675, opened 2020-03-25
Update how we run tests for dev/ui/appservice packages

EricJizbaMSFT:
The VS Code team split the package `vscode` into `@types/vscode` and `vscode-test`. I based the new testing structure off of their latest hello-world extension samples. This is primarily to help reduce security vulnerabilities (still workin on that).


dev/test/extension/package.json
```diff
@@ -0,0 +1,14 @@
+{
+    "name": "azureextensiondev",
```
nturinski: What's with this file?

appservice/.vscode/launch.json
```diff
@@ -8,18 +8,16 @@
...
-                "--extensionTestsPath=${workspaceFolder}/out/test"
+                "--extensionTestsPath=${workspaceFolder}/out/test/index"
             ],
-            "stopOnEntry": false,
-            "sourceMaps": true,
             "outFiles": [
                 "${workspaceFolder}/out/test/**/*.js"
             ],
-            "preLaunchTask": "npm: compile",
+            "preLaunchTask": "${defaultBuildTask}",
```
nturinski: Is `defaultBuildTask` pretest by default?  I didn't see where this gets defined anywhere.

appservice/test/index.ts
```diff
@@ -3,64 +3,55 @@
...
+        reporterOptions: {
+            reporterEnabled: 'spec, mocha-junit-reporter',
+            mochaJunitReporterReporterOptions: {
+                mochaFile: path.join(__dirname, '..', '..', 'test-results.xml')
+            }
+        }
+    };
+
+    addEnvVarsToMochaOptions(options);
+    console.log(`Mocha options: ${JSON.stringify(options, undefined, 2)}`);
```
nturinski: Also why 2?

dev/test/extension/package.json
EricJizbaMSFT: It's a fake extension since these are npm packages, but the tests sometimes need to act as if they're running in the context of a VS Code extension. The UI and appservice packages already have one https://github.com/microsoft/vscode-azuretools/tree/master/ui/test/extension

appservice/.vscode/launch.json
EricJizbaMSFT: It's set here: https://github.com/microsoft/vscode-azureappservice/blob/master/.vscode/tasks.json#L9

This is one of the things that the new hello-world extensions from VS Code do, so I copied them. I guess it's a tiny bit better in case you rename the build task, but mostly it doesn't make a difference

appservice/test/index.ts
EricJizbaMSFT: It's a formatting thing to make the output look nicer. 2 is the number of spaces to insert as tabs

StephenWeatherford:
> This is primarily to help reduce security vulnerabilities (still workin on that).

How is this related to security vulnerabilities?

EricJizbaMSFT:
The old `vscode` package (that I got rid of in favor of `@types/vscode` and `vscode-test`) had some dependencies of dependencies that were showing up in `npm audit`

EricJizbaMSFT:
Maybe it would be more accurate to say "Security vulnerabilities is what inspired me to do this" because obviously there are other benefits in terms of staying up to date with VS Code testing packages

StephenWeatherford:
> The old `vscode` package (that I got rid of in favor of `@types/vscode` and `vscode-test`) had some dependencies of dependencies that were showing up in `npm audit`

Ah, thanks.

PR #677, opened 2020-03-30
Add support for postDeployTask

EricJizbaMSFT:
Related to https://github.com/microsoft/vscode-azurefunctions/issues/1528

### DIFF
--- a/appservice/src/index.ts
+++ b/appservice/src/index.ts
@@ -19,7 +19,7 @@ export * from './deleteSite';
 export * from './deploy/deploy';
 export * from './deploy/getDeployFsPath';
 export * from './deploy/IDeployContext';
-export * from './deploy/runPreDeployTask';
+export { handleFailedPreDeployTask, IPreDeployTaskResult, runPreDeployTask, tryRunPreDeployTask } from './deploy/runDeployTask';
 export * from './editScmType';
 export { registerAppServiceExtensionVariables } from './extensionVariables';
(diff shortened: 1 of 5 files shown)


appservice/src/index.ts
```diff
@@ -19,7 +19,7 @@ export * from './deleteSite';
 export * from './deploy/deploy';
 export * from './deploy/getDeployFsPath';
 export * from './deploy/IDeployContext';
-export * from './deploy/runPreDeployTask';
+export { handleFailedPreDeployTask, IPreDeployTaskResult, runPreDeployTask, tryRunPreDeployTask } from './deploy/runDeployTask';
```
nturinski: Why are you exporting `handleFailedPreDeployTask`?  I don't see it used in https://github.com/microsoft/vscode-azurefunctions/pull/1996/files

EricJizbaMSFT: It's used here: https://github.com/microsoft/vscode-azurefunctions/blob/master/src/commands/deploy/runPreDeployTask.ts#L35

All the functions listed here already existed and were exported previously. The only thing _not_ exported out of `runDeployTask.ts` is `startPostDeployTask` which is the function I just added

nturinski: Oh doh, you were * exporting.  Sorry 😬

EricJizbaMSFT: dontworry

PR #678, opened 2020-04-01
Provide a workaround when md5 validation isn't working

EricJizbaMSFT:
Related to https://github.com/microsoft/vscode-azurefunctions/issues/1983 - seems like this error might be happening when it shouldn't (not entirely sure). Let's provide a workaround while we investigate

### DIFF
--- a/appservice/package-lock.json
+++ b/appservice/package-lock.json
@@ -1,6 +1,6 @@
 {
     "name": "vscode-azureappservice",
-    "version": "0.57.3",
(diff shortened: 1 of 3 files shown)


EricJizbaMSFT:
> Hotfix?

I was thinking a link to a nightly build in the issue would be enough until our next planned release. It's a very small percentage of users in telemetry and the existing workaround (to use "azureFunctions.scmDoBuildDuringDeployment") is actually the recommended way of deploying

PR #679, opened 2020-04-02
Make sure we're finding those tasks correctly 🕵🏻‍♂️

EricJizbaMSFT:
This line of code was not working because sometimes the tasks were the same, but the objects were different:
> if (e.execution.task === preDeployTask) {

This isn't the first time we've had a bug when finding tasks or checking for equality, so I refactored that logic out into it's own file and spent some time adding tests.

Fixes https://github.com/microsoft/vscode-azurefunctions/issues/2001 and fixes https://github.com/microsoft/vscode-azureappservice/issues/1428

### DIFF
--- a/appservice/src/deploy/runDeployTask.ts
+++ b/appservice/src/deploy/runDeployTask.ts
@@ -27,8 +27,7 @@ export async function tryRunPreDeployTask(context: IActionContext, deployFsPath:
     let preDeployTaskResult: IPreDeployTaskResult = { taskName, exitCode: undefined, failedToFindTask: false };
 
     if (taskName && shouldExecuteTask(scmType, settingKey, taskName)) {
-        const tasks: vscode.Task[] = await vscode.tasks.fetchTasks();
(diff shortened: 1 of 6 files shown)


appservice/src/deploy/runDeployTask.ts
```diff
@@ -27,7 +27,8 @@ export async function tryRunPreDeployTask(context: IActionContext, deployFsPath:
     let preDeployTaskResult: IPreDeployTaskResult = { taskName, exitCode: undefined, failedToFindTask: false };
 
     if (taskName && shouldExecuteTask(scmType, settingKey, taskName)) {
-        const task: vscode.Task | undefined = await findTask(deployFsPath, taskName);
+        const tasks: vscode.Task[] = await vscode.tasks.fetchTasks();
+        const task: vscode.Task | undefined = taskUtils.findTask(deployFsPath, taskName, tasks);
```
nturinski, line 31: Should you just include `fetchTasks()` as part of `findTask`?  I always seem them called together anyway.

PR #681, opened 2020-04-06
Add azurewebsites placeholder to site name prompt

EricJizbaMSFT:
Screen Shot 2020-04-06 at 3 24 55 PM

Fixes https://github.com/microsoft/vscode-azurefunctions/issues/1940


nturinski:
I feel like this might be confusing.  Do you think people might think to type appname.azurewebsites.net?

EricJizbaMSFT:
If they do, it'll tell them a "." is not allowed. If not this, what do you suggest?

EricJizbaMSFT:
Gonna merge it so I can get a release candidate, but feel free to suggest something else

nturinski:
Yeah, I'll think about it 🤔 I forgot, what was the reason we're adding that to the placeholder?

EricJizbaMSFT:
https://github.com/microsoft/vscode-azurefunctions/issues/1940

PR #683, opened 2020-04-20
Accept TheamIcon as iconPath type

ravipal:
Accept TheamIcon as iconPath type

### DIFF
--- a/ui/src/treeDataProvider/AzExtTreeItem.ts
+++ b/ui/src/treeDataProvider/AzExtTreeItem.ts
@@ -3,7 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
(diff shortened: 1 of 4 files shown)


ui/src/treeDataProvider/AzExtTreeItem.ts
```diff
@@ -19,7 +19,7 @@ export abstract class AzExtTreeItem implements types.AzExtTreeItem {
     public id?: string;
     public commandId?: string;
     public commandArgs?: unknown[];
-    public iconPath?: string | Uri | { light: string | Uri; dark: string | Uri };
+    public iconPath?: string | Uri | { light: string | Uri; dark: string | Uri } | ThemeIcon;
```
EricJizbaMSFT, line 22: nit: looks like we could change these to `types.TreeItemIconPath` instead of spelling it all out

PR #684, opened 2020-04-27
Use `sendTelemetryErrorEvent` for errors

EricJizbaMSFT:
According to [this](https://github.com/microsoft/vscode-extension-telemetry#sending-errors-as-events):
> Use this method for sending error telemetry as traditional events to App Insights. This method will automatically drop events in certain environments for first party extensions.

### DIFF
--- a/ui/src/callWithTelemetryAndErrorHandling.ts
+++ b/ui/src/callWithTelemetryAndErrorHandling.ts
@@ -119,7 +119,12 @@ function handleTelemetry(context: IActionContext, callbackId: string, start: num
         const end: number = Date.now();
         context.telemetry.measurements.duration = (end - start) / 1000;
 
+        const errorPropRegExp: RegExp = /(error|stack|exception)/i;
         // Note: The id of the extension is automatically prepended to the given callbackId (e.g. "vscode-cosmosdb/")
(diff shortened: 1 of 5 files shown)


ui/src/callWithTelemetryAndErrorHandling.ts
```diff
@@ -119,7 +119,12 @@ function handleTelemetry(context: IActionContext, callbackId: string, start: num
         const end: number = Date.now();
         context.telemetry.measurements.duration = (end - start) / 1000;
 
+        const errorPropRegExp: RegExp = /(error|stack|exception)/i;
         // Note: The id of the extension is automatically prepended to the given callbackId (e.g. "vscode-cosmosdb/")
-        ext.reporter.sendTelemetryEvent(callbackId, context.telemetry.properties, context.telemetry.measurements);
+        if (Object.entries(context.telemetry.properties).some(([key, value]) => errorPropRegExp.test(key) && value)) {
+            ext.reporter.sendTelemetryErrorEvent(callbackId, context.telemetry.properties, context.telemetry.measurements);
```
StephenWeatherford, line 125: Might make sense to send two events in this case so we get *some* telemetry in VSO when errors occur, but could be changed later.

PR #686, opened 2020-04-28
Send non-error properties for error telemetry event

EricJizbaMSFT:
Stephen bothered them, so they added the ability to specify properties to drop in case of an error, instead of dropping the whole event

> The last parameter is an optional list of case-sensitive properties that should be dropped. If no array is passed, we will drop all properties but still send the event.

### DIFF
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -8,7 +8,7 @@ import { Location } from 'azure-arm-resource/lib/subscription/models';
 import { StorageAccount } from 'azure-arm-storage/lib/models';
 import { ServiceClientCredentials } from 'ms-rest';
 import { AzureEnvironment, AzureServiceClientOptions } from 'ms-rest-azure';
-import { Disposable, Event, ExtensionContext, InputBoxOptions, Memento, MessageItem, MessageOptions, OpenDialogOptions, OutputChannel, Progress, QuickPickItem, QuickPickOptions, TextDocument, TreeDataProvider, TreeItem, Uri, ThemeIcon } from 'vscode';
+import { Disposable, Event, ExtensionContext, InputBoxOptions, Memento, MessageItem, MessageOptions, OpenDialogOptions, OutputChannel, Progress, QuickPickItem, QuickPickOptions, TextDocument, ThemeIcon, TreeDataProvider, TreeItem, Uri } from 'vscode';
 import { AzureExtensionApi, AzureExtensionApiProvider } from './api';
 
 export type OpenInPortalOptions = {
@@ -582,7 +582,7 @@ export interface IErrorHandlingContext {
 export interface ITelemetryReporter {
     sendTelemetryEvent(eventName: string, properties?: { [key: string]: string | undefined }, measurements?: { [key: string]: number | undefined }): void;
 
-    sendTelemetryErrorEvent(eventName: string, properties?: { [key: string]: string | undefined }, measurements?: { [key: string]: number | undefined }): void;
+    sendTelemetryErrorEvent(eventName: string, properties?: { [key: string]: string | undefined }, measurements?: { [key: string]: number | undefined }, errorProps?: string[]): void;
 }
 
 /**
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -580,8 +580,11 @@ export interface IErrorHandlingContext {
 }
 
 export interface ITelemetryReporter {
-    sendTelemetryEvent(eventName: string, properties?: { [key: string]: string | undefined }, measurements?: { [key: string]: number | undefined }): void;
-
+    /**
+     * Use this method for sending error telemetry as traditional events to App Insights.
+     * This method will automatically drop error properties in certain environments for first party extensions.
+     * The last parameter is an optional list of case-sensitive properties that should be dropped. If no array is passed, we will drop all properties but still send the event.
+     */
     sendTelemetryErrorEvent(eventName: string, properties?: { [key: string]: string | undefined }, measurements?: { [key: string]: number | undefined }, errorProps?: string[]): void;
 }
 
--- a/ui/src/DebugReporter.ts
+++ b/ui/src/DebugReporter.ts
@@ -9,15 +9,7 @@ import { ITelemetryReporter } from '../index';
 export class DebugReporter implements ITelemetryReporter {
     constructor(private _extensionName: string, private _extensionVersion: string, private _verbose: boolean) { }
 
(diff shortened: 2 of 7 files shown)


ui/src/DebugReporter.ts
```diff
@@ -10,22 +10,22 @@ export class DebugReporter implements ITelemetryReporter {
     constructor(private _extensionName: string, private _extensionVersion: string, private _verbose: boolean) { }
 
     public sendTelemetryEvent(eventName: string, properties?: { [key: string]: string | undefined; }, measures?: { [key: string]: number | undefined; }): void {
-        this.logTelemetryEvent(eventName, properties, measures, 'TELEMETRY');
+        this.logTelemetryEvent(eventName, properties, measures);
     }
 
-    public sendTelemetryErrorEvent(eventName: string, properties?: { [key: string]: string | undefined; }, measures?: { [key: string]: number | undefined; }): void {
-        this.logTelemetryEvent(eventName, properties, measures, 'TELEMETRY-ERROR');
+    public sendTelemetryErrorEvent(eventName: string, properties?: { [key: string]: string | undefined; }, measures?: { [key: string]: number | undefined; }, _errorProps?: string[]): void {
```
StephenWeatherford, line 16: Why don't we get rid of this?  We probably want to guard against people accidentally sending error telemetry properties in the standard sendTelemetryEvent anyway, right?

StephenWeatherford:
> Stephen bothered them.

Down deep I know I've always bothered you, too.

ui/src/DebugReporter.ts
EricJizbaMSFT: You mean get rid of `sendTelemetryEvent` and force people to use `sendTelemetryErrorEvent`? I guess I went with the easiest thing that wouldn't break people, but I suppose it would make sense to remove it if we want to be the enforcers

StephenWeatherford: I meant get rid of sendTelemetryErrorEvent, so that sendTelemetryEvent always calls the telemetry's error event.  You only added sendTelemetryErrorEvent in the last PR, right?  So breakage would be minimal?

Seems better than relying on people to call the event right, or else they might violate GDPR...

EricJizbaMSFT: Hmm so you're saying in [here](https://github.com/microsoft/vscode-azuretools/blob/master/ui/src/createTelemetryReporter.ts#L25) we'd have to somehow convert the existing `DebugReporter` we get from `vscode-extension-telemetry`, to a `DebugReporter` that has a method name of `sendTelemetryErrorEvent`, but actually uses `sendTelemetryEvent` under the covers? I'm not sure how I feel about the added layer in the middle. So far, the interface `ITelemetryReporter` directly matches `TelemetryReporter` and only exists so that we can allow a `DebugReporter` with the same type.

That's not what I originally thought you meant, so it doesn't break people how I thought it would. However, it would mean any existing call to `sendTelemetryEvent` would have all properties dropped since `errorProps` is undefined by default

Tbh, I feel like `callWithTelemetryAndErrorHandling` and `registerCommand` does a lot of the "guarding" already. I can't think of a single case where I call `sendTelemetryEvent` directly

StephenWeatherford: DebugTelemetry is ours, we don't get it from `vscode-extension-telemetry`.  Instead, we'd need a middle layer like you say to implement ITelemetryReporter to call the TelemetryReporter we get from `vscode-extension-telemetry` [here](https://github.com/microsoft/vscode-azuretools/blob/3167744d999c8e2acd87bd78b8ac895257ebc924/ui/src/createTelemetryReporter.ts#L27).

> So far, the interface ITelemetryReporter directly matches TelemetryReporter and only exists so that we can allow a DebugReporter with the same type.

Well, it also exists to simplify our interface and isolate us from vscode changes.

> I can't think of a single case where I call sendTelemetryEvent directly

Exactly, so why are you *adding* a new method that isn't needed?  We should keep our interface the same (ITelemetryReporter) and have it always do the right behavior, instead of complicating our interface and having it possibly do the wrong thing?

EricJizbaMSFT: Yes I meant to say `TelemetryReporter` for the middle layer

> Exactly, so why are you adding a new method that isn't needed?

For the simplicity of matching `TelemetryReporter`. I agree it's "complicating our interface" in the sense that it has a few more characters, but it does simplify our interface concept-wise in the sense that it has the exact same behavior as the more "official" `TelemetryReporter`

If the goal is to prevent bad behaviors, I'd rather remove `sendTelemetryEvent` from the interface and keep `sendTelemetryErrorEvent`. This should avoid the middle-layer I mentioned and most of the confusion related to differences in `TelemetryReporter` and `ITelemetryReporter`. It would break a _few_ people (as mentioned above I don't think `sendTelemetryEvent` is used very often), but I'd be fine with that if we're doing it to prevent bad behavior. Would that be a satisfactory compromise?

StephenWeatherford: Seems weird but I'm okay with it.

ui/index.d.ts
```diff
@@ -580,8 +580,11 @@ export interface IErrorHandlingContext {
 }
 
 export interface ITelemetryReporter {
-    sendTelemetryEvent(eventName: string, properties?: { [key: string]: string | undefined }, measurements?: { [key: string]: number | undefined }): void;
-
+    /**
+     * Use this method for sending error telemetry as traditional events to App Insights.
+     * This method will automatically drop error properties in certain environments for first party extensions.
```
StephenWeatherford, line 585: Somewhat confusing...  Do these come straight from vscode telemetry?

> This method will automatically drop error properties in certain environments for first party extensions.

Doesn't seem automatic since last arg is required to drop error properties, otherwise all are dropped?

>   If no array is passed, we will drop all properties but still send the event.

I thought the behavior was to drop the entire event?

StephenWeatherford: BTW, thanks for handling this!

EricJizbaMSFT: Yeah I copied this straight from here: https://github.com/microsoft/vscode-extension-telemetry#sending-errors-as-events

> Doesn't seem automatic since last arg is required to drop error properties, otherwise all are dropped?

It _is_ automatic. If you forget to pass an empty array for error properties, you don't get any properties (in those certain environments). Source code [here](https://github.com/microsoft/vscode-extension-telemetry/blob/master/src/telemetryReporter.ts#L250)

> I thought the behavior was to drop the entire event?

Nope just the properties. See above link to source code

PR #688, opened 2020-04-30
Improve Wizard UI

EricJizbaMSFT:
Highlights:
1. Cache input box values so that users don't have to re-type a value if they hit back during a wizard (only cached within the single wizard)
1. Add a "back" button to warning dialogs if they're shown during a wizard
1. More reliably detect if a step actually prompted using `onDidFinishPrompt`. This fixes bugs where the step didn't prompt and then the _next_ step's back button doesn't work

### DIFF
--- a/ui/src/wizard/AzureWizard.ts
+++ b/ui/src/wizard/AzureWizard.ts
@@ -22,6 +22,9 @@ export class AzureWizard<T extends types.IActionContext> implements types.AzureW
     private _stepHideStepCount?: boolean;
     private _wizardHideStepCount?: boolean;
 
+    private _cachedInputBoxValues: { [step: string]: string | undefined } = {};
+    private _currentStepName: string | undefined;
+
     public constructor(context: T, options: types.IWizardOptions<T>) {
         // reverse steps to make it easier to use push/pop
         // tslint:disable-next-line: strict-boolean-expressions
@@ -31,6 +34,10 @@ export class AzureWizard<T extends types.IActionContext> implements types.AzureW
         this._executeSteps = options.executeSteps || [];
         this._context = context;
         this._wizardHideStepCount = options.hideStepCount;
+    }
+
+    public getCachedInputBoxValue(): string | undefined {
+        return this._currentStepName ? this._cachedInputBoxValues[this._currentStepName] : undefined;
     }
 
     public get hideStepCount(): boolean {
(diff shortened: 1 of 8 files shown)


nturinski:
Related: https://github.com/microsoft/vscode-azureappservice/issues/1226

ui/src/wizard/AzureWizardUserInput.ts
```diff
@@ -37,10 +42,10 @@ export class AzureWizardUserInput implements IWizardUserInput {
                 quickPick.step = this._wizard.currentStep;
                 quickPick.totalSteps = this._wizard.totalSteps;
             }
-            quickPick.buttons = this._wizard.currentStep > 1 ? [QuickInputButtons.Back] : [];
+            quickPick.buttons = this.showBackButton ? [QuickInputButtons.Back] : [];
 
             // Copy settings that are common between options and quickPick
-            quickPick.placeholder = options.placeHolder;
+            quickPick.placeholder = options.loadingPlaceHolder || options.placeHolder;
```
nturinski, line 48: Should we default this placeHolder to something like "Loading..." or "Retrieving..."? Or is the concern that for some quickPicks, it'd flash by so quickly that it'd be distracting?

ui/src/wizard/AzureWizard.ts
```diff
@@ -61,16 +68,27 @@ export class AzureWizard<T extends types.IActionContext> implements types.AzureW
                 if (step.shouldPrompt(this._context)) {
                     step.propertiesBeforePrompt = Object.keys(this._context).filter(k => !isNullOrUndefined(this._context[k]));
 
+                    const disposable: vscode.Disposable = ext.ui.onDidFinishPrompt((result) => {
+                        // tslint:disable-next-line: no-non-null-assertion
+                        step!.prompted = true;
+                        if (typeof result === 'string' && this._currentStepName) {
+                            this._cachedInputBoxValues[this._currentStepName] = result;
```
nturinski, line 75: Won't this break if you have duplicateSteps enabled?

ui/src/wizard/AzureWizardUserInput.ts
EricJizbaMSFT: No I didn't have that concern - it was more that the existing placeholder already makes sense in many cases. For example, if the placeholder says "Select a Web App..." and the thing is loading - I can reasonably infer that it's loading web apps. If it just says a generic "Loading..." - I almost feel like that's less information

The ideal message is probably something like "Retrieving web apps...", but that obviously can't be set in a generic way here.

ui/src/wizard/AzureWizard.ts
EricJizbaMSFT: Yeah I thought about this, but I'm not sure. The only situation I've found that requires duplicate steps is my current work to make the tree item picker a wizard, but that uses quick picks instead of input boxes so it doesn't really apply.

Hard to know until we have a situation with input boxes using this, but since you commented I'm fine to disable this behavior for duplicate steps

nturinski: I mean, worst-case scenario, it just over-writes the old cached value with a new one.  I'm fine with leaving enabled 🤷‍♂️

PR #691, opened 2020-04-30
Update reference to vscode-azureextensionui

bwateratmsft:
This is needed downstream in order to be able to consume `vscode-azureextensionui` 0.30.1.


bwateratmsft:
Dupe of #690.

PR #697, opened 2020-05-04
Fix "Error: Cannot find module 'vscode'" in dev package

EricJizbaMSFT:
I have to use `_vscode` when I call `new EventEmitter<types.PromptResult>` otherwise tests will fail

The problem was introduced in https://github.com/microsoft/vscode-azuretools/pull/687

### DIFF
--- a/dev/src/TestUserInput.ts
+++ b/dev/src/TestUserInput.ts
@@ -4,7 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import * as assert from 'assert';
-import * as vscodeTypes from 'vscode'; // `TestUserInput._vscode` should be use for anything that's not purely a type (e.g. instantiating a class)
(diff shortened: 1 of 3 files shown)


dev/src/TestUserInput.ts
```diff
@@ -4,7 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import * as assert from 'assert';
-import { Event, EventEmitter, InputBoxOptions, MessageItem, MessageOptions, OpenDialogOptions, QuickPickItem, QuickPickOptions, Uri } from 'vscode';
+import * as vscodeTypes from 'vscode'; // `TestUserInput._vscode` should be use for anything that's not purely a type (e.g. instantiating a class)
```
wwlorey, line 7: nit: just a small grammar thing:
```suggestion
import * as vscodeTypes from 'vscode'; // `TestUserInput._vscode` should be used for anything that's not purely a type (e.g. instantiating a class)
```

PR #701, opened 2020-05-06
Log web app save confirmation text

alexweininger:
Fixes [vscode-azureappservice#1509](https://github.com/microsoft/vscode-azureappservice/issues/1509)

### DIFF
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -485,6 +485,11 @@ export declare abstract class BaseEditor<ContextT> implements Disposable {
     abstract getFilename(context: ContextT): Promise<string>;
 
     /**
+     * Implement this to return the resource name from the remote
+     */
+    abstract getResourceName(context: ContextT): Promise<string>;
+
+    /**
      * Implement this to return the size in MB.
      */
     abstract getSize(context: ContextT): Promise<number>;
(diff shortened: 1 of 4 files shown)


alexweininger:
Save confirmation text is now logged as well as put in the notification.

Like this:

image

nturinski:
So for this issue (vscode-azureappservice#1509), I think we were actually more interested in updating the `Updated "file.name"...` output line.

If you look [here](https://github.com/microsoft/vscode-azuretools/blob/master/ui/src/AzExtOutputChannel.ts#L33), `appendLog` accepts a `resourceName` parameter.  We'd rather use that so that the message includes the app's name when updating the file.

For example:
`10:40:21 AM: weininger-02: Updated "hostingstart.html".`

The point of this issue is that we include the app name so that the user can see which resource was updated.  The test team just included the `Saving "hostingstart.html"` message as an example of a message that included the resource name in the notification.

alexweininger:
@nturinski Ok, I see now. I have got this working. Do we want to include the resource name in all output lines or just the "Updated {filename}" log?

nturinski:
As discussed offline, let's prefix all of the file download/uploading output with the resource name.

PR #702, opened 2020-05-07
Bump UI package version

alexweininger:


EricJizbaMSFT:
You have to update the version of the appservice package so that we can release this.

I suggest `0.58.1` -> `0.59.0`. I'm somewhat iffy on whether or not a breaking change in a dependency (ui) means a breaking change for this package (appservice), but better to be safe I guess

PR #715, opened 2020-05-18
Only show one long running notification at a time in runWithLoadingNotification

wwlorey:
Related to https://github.com/microsoft/vscode-cosmosdb/issues/1463

### DIFF
--- a/ui/src/treeDataProvider/runWithLoadingNotification.ts
+++ b/ui/src/treeDataProvider/runWithLoadingNotification.ts
@@ -3,31 +3,30 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { CancellationToken, EventEmitter, ProgressLocation, window } from 'vscode';
+import { CancellationToken, ProgressLocation, window } from 'vscode';
 import * as types from '../../index';
 import { localize } from '../localize';
 
 let notificationActive: boolean = false;
 
 export async function runWithLoadingNotification<T>(context: types.ILoadingTreeContext, callback: (cancellationToken: CancellationToken) => Promise<T>): Promise<T> {
-    if (notificationActive) {
-        return await callback({
-            isCancellationRequested: false,
-            onCancellationRequested: (new EventEmitter()).event
-        });
-    } else {
-        return await window.withProgress({ location: ProgressLocation.Notification, cancellable: true }, async (progress, cancellationToken) => {
-            notificationActive = true;
-            const message: string = context.loadingMessage || localize('loadingAll', 'Loading resources...');
-            const messageDelay: number = context.loadingMessageDelay !== undefined ? context.loadingMessageDelay : 2;
-            const timer: NodeJS.Timer = setTimeout(() => progress.report({ message }), messageDelay * 1000);
+    return await window.withProgress({ location: ProgressLocation.Notification, cancellable: true }, async (progress, cancellationToken) => {
+        const message: string = context.loadingMessage || localize('loadingAll', 'Loading resources...');
+        const messageDelay: number = context.loadingMessageDelay !== undefined ? context.loadingMessageDelay : 2;
 
-            try {
-                return await callback(cancellationToken);
-            } finally {
+        let timer: NodeJS.Timer | undefined;
+        if (!notificationActive) {
+            timer = setTimeout(() => progress.report({ message }), messageDelay * 1000);
+            notificationActive = true;
+        }
+
+        try {
+            return await callback(cancellationToken);
+        } finally {
+            if (timer) {
                 clearTimeout(timer);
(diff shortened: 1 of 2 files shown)


nturinski:
Related to https://github.com/microsoft/vscode-azurestorage/issues/671

ui/src/treeDataProvider/runWithLoadingNotification.ts
```diff
@@ -3,20 +3,31 @@
...
 
+let notificationActive: boolean = false;
+
 export async function runWithLoadingNotification<T>(context: types.ILoadingTreeContext, callback: (cancellationToken: CancellationToken) => Promise<T>): Promise<T> {
-    return await window.withProgress({ location: ProgressLocation.Notification, cancellable: true }, async (progress, cancellationToken) => {
-        const message: string = context.loadingMessage || localize('loadingAll', 'Loading resources...');
-        const messageDelay: number = context.loadingMessageDelay !== undefined ? context.loadingMessageDelay : 2;
-        const timer: NodeJS.Timer = setTimeout(() => progress.report({ message }), messageDelay * 1000);
+    if (notificationActive) {
+        return await callback({
```
EricJizbaMSFT, line 14: You can still use the below `window.withProgress` instead of creating your own cancellation token here. The only thing you need to prevent from happening with `notificationActive` is `progress.report`

ui/src/treeDataProvider/runWithLoadingNotification.ts
```diff
@@ -3,20 +3,31 @@
...
+            isCancellationRequested: false,
+            onCancellationRequested: (new EventEmitter()).event
+        });
+    } else {
+        return await window.withProgress({ location: ProgressLocation.Notification, cancellable: true }, async (progress, cancellationToken) => {
+            notificationActive = true;
```
EricJizbaMSFT, line 20: In a pattern like this, there shouldn't be any code between `notificationActive = true;` and the start of the try block so that we're 100% certain it gets reset in the `finally` block.

ui/src/treeDataProvider/runWithLoadingNotification.ts
EricJizbaMSFT: Actually scrap that. The cancellationToken you pass here should be the same cancellation token from the original callback. Since there's only one notification, I assume we want to cancel both callbacks if the user hits the "Cancel" button?

wwlorey: I may of misunderstood what you were suggesting here. Would this involve caching the cancellation token from the first `runWithLoadingNotification` call to use for subsequent calls?

EricJizbaMSFT: Yes

Looks like you changed to match my first comment in this thread, which should be completely ignored in favor of the second comment

ui/src/treeDataProvider/runWithLoadingNotification.ts
```diff
@@ -7,16 +7,24 @@ import { CancellationToken, ProgressLocation, window } from 'vscode';
...
 import { localize } from '../localize';
 
+let activeCancellationToken: CancellationToken | undefined;
+
 export async function runWithLoadingNotification<T>(context: types.ILoadingTreeContext, callback: (cancellationToken: CancellationToken) => Promise<T>): Promise<T> {
-    return await window.withProgress({ location: ProgressLocation.Notification, cancellable: true }, async (progress, cancellationToken) => {
-        const message: string = context.loadingMessage || localize('loadingAll', 'Loading resources...');
-        const messageDelay: number = context.loadingMessageDelay !== undefined ? context.loadingMessageDelay : 2;
-        const timer: NodeJS.Timer = setTimeout(() => progress.report({ message }), messageDelay * 1000);
+    if (!!activeCancellationToken) {
```
EricJizbaMSFT, line 13: I'm still a little concerned because the first callback could finish before the second callback and then the notification is gone and the user has no way to cancel. The reason we added this notification is because it's loading _all_ resources instead of doing our typical pagination with a "load more..." option. That means it could be loading a LOT of resources.

I looked into this more and I don't think we want to display the "Loading resources..." notifications _at all_ in Databases. As mentioned already, it was intended for the case when we're loading _all_ resources. Well the code [here](https://github.com/microsoft/vscode-azuretools/blob/master/ui/src/treeDataProvider/AzExtTreeDataProvider.ts#L179) is using `runWithLoadingNotification` even when `context.loadAll` is false. IMO that's what we should be fixing instead. I believe these are the two lines of code causing the notifications in Databases and I don't see either one using `loadAll`:

https://github.com/microsoft/vscode-cosmosdb/blob/master/src/postgres/commands/registerPostgresCommands.ts#L57

https://github.com/microsoft/vscode-cosmosdb/blob/master/src/mongo/registerMongoCommands.ts#L137

PR #722, opened 2020-05-22
Create base classes for AppSettingsTreeItem and AppSettingsTreeItem

alexweininger:
- Contains small breaking changes.
- Added `IAppSettingsClient` which defines what needs to be provided to the `AppSettingsTreeItemBase` in order to do all basic operations (that both trial and real sites need to do) on App Settings.
- Pretty much all code in `AppSettingsTreeItem` can be shared between sites and trial sites. So I moved the app settings logic to its base class.
- What is left in `AppSettingTreeItem` is logic surrounding slots, which trial sites will not need.
- Created base classes `AppSettingsTreeItemBase` and `AppSettingTreeItemBase`, which both contain logic that is needed for both trial sites and real sites.

Edit: after Eric's review:

- No longer splitting `AppSettingsTreeItem` and `AppSettingTreeItem` into base classes.
- Added optional properties in `IAppSettingsClient` for slots.


appservice/src/tree/AppSettingTreeItemBase.ts
```diff
@@ -0,0 +1,85 @@
...
+import { IAppSettingsClient } from '../IAppSettingsClient';
+import { validateAppSettingKey } from './AppSettingsTreeItem';
+import { AppSettingsTreeItemBase } from './AppSettingsTreeItemBase';
+import { getThemedIconPath } from './IconPath';
+
+export abstract class AppSettingTreeItemBase extends AzExtTreeItem {
+    public static contextValue: string = 'applicationSettingItem';
+    public readonly contextValue: string = AppSettingTreeItemBase.contextValue;
+
+    public readonly _client: IAppSettingsClient;
```
EricJizbaMSFT, line 18: nit: You should only prefix properties with an underscore if they're private

appservice/src/tree/AppSettingsTreeItem.ts
```diff
@@ -33,27 +33,16 @@ export function validateAppSettingKey(settings: StringDictionary, client: SiteCl
...
 
-    public hasMoreChildrenImpl(): boolean {
-        return false;
+    constructor(parent: AzureParentTreeItem<ISiteTreeRoot>) {
+        super(parent, parent.root.client);
     }
 
-    public async loadMoreChildrenImpl(_clearCache: boolean): Promise<AzureTreeItem<ISiteTreeRoot>[]> {
-        this._settings = await this.root.client.listApplicationSettings();
+    public async loadMoreChildrenImpl(_clearCache: boolean): Promise<AzExtTreeItem[]> {
```
EricJizbaMSFT, line 44: I would think you need this in the base class. Is `loadMoreChildrenImpl` different for trial apps for some reason? Same with createChildImpl

appservice/src/IAppSettingsClient.ts
```diff
@@ -0,0 +1,15 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { StringDictionary } from 'azure-arm-website/lib/models';
+
+export interface IAppSettingsClient {
```
EricJizbaMSFT, line 8: So your goal with this PR is to create a place for code that can and can't be shared. However, it feels like you ended up with _two_ places for code that can't be shared and we only need one. In other words, I don't think you need to create `IAppSettingsClient` _and_ split them out into base classes. Of the two options, I prefer `IAppSettingsClient` because I could see us using that interface for non-tree scenarios as well

alexweininger:
> Might be just my opinion, but I'm not a huge fan of "draft" PRs. What exactly is draft about this? If you can fix whatever you need to make it "real" that would be great.
> 
> Also - for refactoring PRs, I prefer to see a second PR showing how it's used at the same time. Even if you're not ready to support the trial app scenario, you could put up a PR migrating over non-trial apps to this stuff

Probably should have made it more clear in the PR comment, but I was looking for if I was heading in the right direction with these changes.

appservice/src/IAppSettingsClient.ts
```diff
@@ -0,0 +1,21 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { SlotConfigNamesResource, StringDictionary } from 'azure-arm-website/lib/models';
+
+export interface IAppSettingsClient {
```
EricJizbaMSFT, line 8: I assume you'll need to reference this interface outside the package, meaning you should export it here: https://github.com/microsoft/vscode-azuretools/blob/master/appservice/src/index.ts

appservice/src/tree/AppSettingTreeItem.ts
```diff
@@ -23,15 +23,18 @@ export class AppSettingTreeItem extends AzureTreeItem<ISiteTreeRoot> {
...
         super(parent);
+        this._client = client;
         this._key = key;
         this._value = value;
         this._hideValue = true;
     }
 
-    public static async createAppSettingTreeItem(parent: AppSettingsTreeItem, key: string, value: string): Promise<AppSettingTreeItem> {
-        const ti: AppSettingTreeItem = new AppSettingTreeItem(parent, key, value);
+    public static async createAppSettingTreeItem(parent: AppSettingsTreeItem, client: IAppSettingsClient, key: string, value: string): Promise<AppSettingTreeItem> {
```
nturinski, line 36: It's a bit late since it's merged, but why do you need to pass client as a parameter if you have it as a private property?

PR #723, opened 2020-05-27
Create IFilesClient

alexweininger:
PR much like #722, but for `FolderTreeItem` and `FileTreeItem`.

- Created `IFilesClient` which only has two properties: `fullname`, and `getKuduClient()`


appservice/src/tree/FolderTreeItem.ts
```diff
@@ -39,7 +42,7 @@ export class FolderTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
     }
 
     public async loadMoreChildrenImpl(_clearCache: boolean, _context: IActionContext): Promise<AzExtTreeItem[]> {
-        const kuduClient: KuduClient = await this.root.client.getKuduClient();
+        const kuduClient: KuduClient = await this.client.getKuduClient();
         let response: IKuduItemResponse;
         try {
             response = <IKuduItemResponse><unknown>(await kuduClient.vfs.getItemWithHttpOperationResponse(this.path)).response;
```
nturinski, line 48: Since you had the files node more or less working for Trial Apps, I'm assuming that the kuduClient API call (on line 45) worked to retrieve the files?

alexweininger: Yes, I was able to get the KuduClient, and retrieve files for the trial app.

The prototype code for that is here:

appservice/src/tree/FileTreeItem.ts
```diff
@@ -3,25 +3,28 @@
...
     public static contextValue: string = 'file';
     public readonly contextValue: string = FileTreeItem.contextValue;
     public readonly label: string;
     public readonly path: string;
     public readonly isReadOnly: boolean;
 
-    constructor(parent: AzureParentTreeItem, label: string, path: string, isReadOnly: boolean) {
+    public readonly client: IFilesClient;
+
+    constructor(parent: AzExtParentTreeItem, client: IFilesClient, label: string, path: string, isReadOnly: boolean) {
```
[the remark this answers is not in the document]
alexweininger: > I think looks good more or less! I had a couple of questions, but I don't think that they're necessarily blocking.
> 
> 1. Since we're relying on SiteClient having all of the functions/properties needed by IAppSettingsClient and IFilesClient, I think we should say that the SiteClient implements both of these clients so it's more obvious when we just pass SiteClient as the IAppSettingsClient in the actual extension repos.
> 2. As far as I know, everything that uses the Files/Log Files also uses Kudu.  But do we want to extract some of the kudu code out so we can make it more generic?  For example, I don't know if we'd be able to use this for SWA (or if we'd even want to for that matter).  But I'm pretty sure in the SWA case, we'd have to make API calls to GitHub to retrieve the vfs, etc.

For 1: I made it implement IFilesClient, will make it implement IAppSettingsClient in a new (extremely small) PR.
For 2: Is it okay if we leave this change for when we do decide if SWA is able to use this?

PR #725, opened 2020-05-29
Don't depend on AzureTreeItem

alexweininger:
Plan on hiding commands that don't work with trial apps based on the DeploymentsTreeItem's contextValue. If the context value is trial app then the commands wont show up. In the

### DIFF
--- a/appservice/src/disconnectRepo.ts
+++ b/appservice/src/disconnectRepo.ts
@@ -5,22 +5,16 @@
 
 import { SiteSourceControl } from 'azure-arm-website/lib/models';
 import { MessageItem } from 'vscode';
-import { AzureTreeItem, IActionContext } from 'vscode-azureextensionui';
-import { editScmType } from './';
+import { IActionContext, ISubscriptionContext } from 'vscode-azureextensionui';
+import { editScmType, SiteClient } from './';
 import { ext } from './extensionVariables';
 import { localize } from './localize';
 import { ScmType } from './ScmType';
-import { DeploymentsTreeItem } from './tree/DeploymentsTreeItem';
 
-export async function disconnectRepo(context: IActionContext, node: DeploymentsTreeItem): Promise<void> {
-    if (node.root && node.parent instanceof AzureTreeItem) {
-        const sourceControl: SiteSourceControl = await node.root.client.getSourceControl();
-        const disconnectButton: MessageItem = { title: localize('disconnect', 'Disconnect') };
-        const disconnect: string = localize('disconnectFromRepo', 'Disconnect from "{0}"? This will not affect your app\'s active deployment. You may reconnect a repository at any time.', sourceControl.repoUrl);
-        await ext.ui.showWarningMessage(disconnect, { modal: true }, disconnectButton);
-        await editScmType(node.root.client, node.parent, context, ScmType.None);
-        await node.refresh();
-    } else {
-        throw Error(localize('actionNotSupported', 'Action not supported.'));
-    }
+export async function disconnectRepo(context: IActionContext, client: SiteClient, subscriptionContext: ISubscriptionContext): Promise<void> {
+    const sourceControl: SiteSourceControl = await client.getSourceControl();
+    const disconnectButton: MessageItem = { title: localize('disconnect', 'Disconnect') };
+    const disconnect: string = localize('disconnectFromRepo', 'Disconnect from "{0}"? This will not affect your app\'s active deployment. You may reconnect a repository at any time.', sourceControl.repoUrl);
+    await ext.ui.showWarningMessage(disconnect, { modal: true }, disconnectButton);
+    await editScmType(client, subscriptionContext, context, ScmType.None);
 }
--- a/appservice/src/disconnectRepo.ts
+++ b/appservice/src/disconnectRepo.ts
@@ -5,7 +5,7 @@
 
 import { SiteSourceControl } from 'azure-arm-website/lib/models';
 import { MessageItem } from 'vscode';
-import { IActionContext } from 'vscode-azureextensionui';
+import { AzureTreeItem, IActionContext } from 'vscode-azureextensionui';
 import { editScmType } from './';
 import { ext } from './extensionVariables';
 import { localize } from './localize';
@@ -13,12 +13,12 @@ import { ScmType } from './ScmType';
 import { DeploymentsTreeItem } from './tree/DeploymentsTreeItem';
 
 export async function disconnectRepo(context: IActionContext, node: DeploymentsTreeItem): Promise<void> {
-    if (node.root) {
+    if (node.root && node.parent instanceof AzureTreeItem) {
         const sourceControl: SiteSourceControl = await node.root.client.getSourceControl();
         const disconnectButton: MessageItem = { title: localize('disconnect', 'Disconnect') };
         const disconnect: string = localize('disconnectFromRepo', 'Disconnect from "{0}"? This will not affect your app\'s active deployment. You may reconnect a repository at any time.', sourceControl.repoUrl);
         await ext.ui.showWarningMessage(disconnect, { modal: true }, disconnectButton);
-        await editScmType(node.root.client, node, context, ScmType.None);
+        await editScmType(node.root.client, node.parent, context, ScmType.None);
         await node.refresh();
     } else {
         throw Error(localize('actionNotSupported', 'Action not supported.'));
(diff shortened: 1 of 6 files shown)


appservice/src/tree/DeploymentsTreeItem.ts
```diff
@@ -14,22 +14,27 @@ import { ScmType } from '../ScmType';
...
 import { getThemedIconPath } from './IconPath';
+import { ISiteTreeRoot } from './ISiteTreeRoot';
 
 /**
  * NOTE: This leverages a command with id `ext.prefix + '.connectToGitHub'` that should be registered by each extension
  */
 export class DeploymentsTreeItem extends AzExtParentTreeItem {
     public static contextValueConnected: string = 'deploymentsConnected';
     public static contextValueUnconnected: string = 'deploymentsUnconnected';
+    public static contextValueTrialApp: string = 'deploymentsTrialApp';
```
EricJizbaMSFT, line 25: I would keep trial app stuff out of this repo if possible. For example, you could just create a class in the app service repo that extends DeploymentsTreeItem and overwrites the contextValue

appservice/src/disconnectRepo.ts
```diff
@@ -6,16 +6,21 @@
...
+export async function disconnectRepo(context: IActionContext, node: DeploymentsTreeItem): Promise<void> {
+    if (node.root && node.parent instanceof AzureTreeItem) {
+        const sourceControl: SiteSourceControl = await node.root.client.getSourceControl();
+        const disconnectButton: MessageItem = { title: localize('disconnect', 'Disconnect') };
+        const disconnect: string = localize('disconnectFromRepo', 'Disconnect from "{0}"? This will not affect your app\'s active deployment. You may reconnect a repository at any time.', sourceControl.repoUrl);
+        await ext.ui.showWarningMessage(disconnect, { modal: true }, disconnectButton);
+        await editScmType(node.root.client, node.parent, context, ScmType.None);
+        await node.refresh();
+    } else {
+        throw Error(localize('actionNotSupported', 'Action not supported.'));
```
EricJizbaMSFT, line 24: I feel like you should prevent this from happening in the App Service repo, not here

appservice/src/editScmType.ts
```diff
@@ -11,10 +11,11 @@ import { connectToGitHub } from './github/connectToGitHub';
 import { localize } from './localize';
 import { ScmType } from './ScmType';
 import { SiteClient } from './SiteClient';
+import { DeploymentsTreeItem } from './tree/DeploymentsTreeItem';
 import { ISiteTreeRoot } from './tree/ISiteTreeRoot';
 import { nonNullProp } from './utils/nonNull';
 
-export async function editScmType(client: SiteClient, node: AzureTreeItem<ISiteTreeRoot>, context: IActionContext, newScmType?: ScmType, showToast: boolean = true): Promise<ScmType | undefined> {
+export async function editScmType(client: SiteClient, node: AzureTreeItem<ISiteTreeRoot> | DeploymentsTreeItem, context: IActionContext, newScmType?: ScmType, showToast: boolean = true): Promise<ScmType | undefined> {
```
EricJizbaMSFT, line 18: We don't _really_ need the node. We just need `ISubscriptionContext` for `openInPortal` down the line. That would kind of separate out the tree stuff from this function

appservice/src/disconnectRepo.ts
```diff
@@ -6,16 +6,21 @@
...
+import { DeploymentsTreeItem } from './tree/DeploymentsTreeItem';
 
-export async function disconnectRepo(context: IActionContext, node: AzureTreeItem<ISiteTreeRoot>): Promise<void> {
-    const sourceControl: SiteSourceControl = await node.root.client.getSourceControl();
-    const disconnectButton: MessageItem = { title: localize('disconnect', 'Disconnect') };
-    const disconnect: string = localize('disconnectFromRepo', 'Disconnect from "{0}"? This will not affect your app\'s active deployment. You may reconnect a repository at any time.', sourceControl.repoUrl);
-    await ext.ui.showWarningMessage(disconnect, { modal: true }, disconnectButton);
-    await editScmType(node.root.client, node, context, ScmType.None);
-    await node.refresh();
+export async function disconnectRepo(context: IActionContext, node: DeploymentsTreeItem): Promise<void> {
```
EricJizbaMSFT, line 15: Yeah I don't think you actually need the node here. I would prefer if you had a call signature more similar to `editScmType`. Maybe even the exact same parameters just to be consistent. With all of these comments, I'm hoping you don't actually need to add `root` or `parent` to `DeploymentsTreeItem`

appservice/src/disconnectRepo.ts
```diff
@@ -5,17 +5,16 @@
...
-import { editScmType, ISiteTreeRoot } from './';
+import { IActionContext, ISubscriptionContext } from 'vscode-azureextensionui';
+import { editScmType, SiteClient } from './';
 import { ext } from './extensionVariables';
 import { localize } from './localize';
 import { ScmType } from './ScmType';
 
-export async function disconnectRepo(context: IActionContext, node: AzureTreeItem<ISiteTreeRoot>): Promise<void> {
-    const sourceControl: SiteSourceControl = await node.root.client.getSourceControl();
+export async function disconnectRepo(context: IActionContext, client: SiteClient, subscriptionContext: ISubscriptionContext): Promise<void> {
```
nturinski, line 14: nit: If you are also bothered by this, feel free to fix it, but if not, don't bother.

...but I hate how the order of the parameters is different between this and `editScmType` 😂

appservice/src/github/IConnectToGitHubWizardContext.ts
```diff
@@ -12,5 +12,5 @@ export interface IConnectToGitHubWizardContext extends IActionContext {
     repoData?: gitHubRepoData;
     branchData?: gitHubBranchData;
     client?: SiteClient;
-    node?: AzureTreeItem;
+    subscriptionContext?: ISubscriptionContext;
```
nturinski, line 15: Rather than having a context being a property on the `IConnectToGitHubWizardContext`, you should be able to just have it extend `ISubscriptionContext` as well.  Then whenever you you need `IConnectToGitHubWizardContext`, you can just do `Object.assign()` like this https://github.com/microsoft/vscode-azurefunctions/blob/51f4d50e978d292e7d46579c918eb229b7774178/src/tree/SubscriptionTreeItem.ts#L82-L87

nturinski: Or, just leave them as two separate entities.  For instance, you set `context.subscriptionContext = subscriptionContext` but you could also just pass in the subscriptionContext to `getGitHubJsonResponse` as well.

I wouldn't worry too much about it, I think with the new GitHub auth experience coming into VS Code stable, I'll have to come back and refactor a lot of this code so do whatever's simpler for you.

appservice/src/disconnectRepo.ts
EricJizbaMSFT: Lol Nathan I agree haha. Also we usually put `IActionContext` first - so keep that in mind assuming you make them consistent

appservice/src/github/IConnectToGitHubWizardContext.ts
alexweininger: I ended up having `IConnectToGithubWizardContext` extend `ISubscriptionContext`  👍

appservice/src/github/connectToGitHub.ts
```diff
@@ -27,10 +27,10 @@ export type gitHubBranchData = { name: string };
 export type gitHubLink = { prev?: string, next?: string, last?: string, first?: string };
 export type gitHubWebResource = requestUtils.Request & { nextLink?: string };
 
-export async function connectToGitHub(node: AzureTreeItem, client: SiteClient, context: IConnectToGitHubWizardContext): Promise<void> {
+export async function connectToGitHub(subscriptionContext: ISubscriptionContext, client: SiteClient, context: IConnectToGitHubWizardContext): Promise<void> {
     const title: string = localize('connectGitHubRepo', 'Connect GitHub repository');
     context.client = client;
-    context.node = node;
+    context = Object.assign(context, { subscriptionContext });
```
EricJizbaMSFT, line 33: If you're gonna make `IConnectToGitHubWizardContext` extend `ISubscriptionContext`, there's no point in passing in the subscriptionContext separately. Typescript will enforce that the subscription stuff is on the GithubContext before it's passed in here

I'm guessing that's why you have another Object.assign above when you call `connectToGitHub`

alexweininger: 🤦

nturinski:
🎉

PR #726, opened 2020-06-04
Add log streaming to IFilesClient

alexweininger:
Since the files and logs tree items are so closely related I added the items needed to support log streaming into `IFilesClient`.

These changes shouldn't be breaking. I tested the changes with functions, and app service.


nturinski:
Could you put up the PRs that demonstrate these changes?

appservice/src/IFilesClient.ts
```diff
@@ -4,9 +4,15 @@
 *--------------------------------------------------------------------------------------------*/
 
 import KuduClient from 'vscode-azurekudu';
+import { IHostKeys } from '.';
```
EricJizbaMSFT, line 7: nit: please watch out for this kind of thing in the tools packages and import from the actual file instead of just `'.'`. VS code does this by default because of how we export files in `index.ts`, but it looks pretty bad

appservice/package.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "vscode-azureappservice",
     "author": "Microsoft Corporation",
-    "version": "0.61.0",
+    "version": "0.61.1",
```
EricJizbaMSFT, line 4: Fyi I actually already released a version 0.61.1 today. If you rebase/merge this branch with master you'll see that this doesn't actually change the version

appservice/src/IFilesClient.ts
```diff
@@ -4,9 +4,15 @@
...
 
 export interface IFilesClient {
 
     fullName: string;
+    defaultHostUrl: string;
+    isFunctionApp: boolean;
+    id: string;
+    kuduUrl: string | undefined;
     getKuduClient(): Promise<KuduClient>;
+    listHostKeys(): Promise<IHostKeys>;
```
nturinski, line 17: This is only used by Functions, so I don't think that it should be on the interface itself.

alexweininger: Made optional

appservice/package.json
alexweininger: Bumped to 0.61.2

PR #727, opened 2020-06-08
Sort resource groups by name when listing in ResourceGroupListStep

wwlorey:
| old | new |
| - | - |
| Screen Shot 2020-06-08 at 12 58 07 PM |  Screen Shot 2020-06-08 at 1 41 41 PM |

Related to https://github.com/microsoft/vscode-cosmosdb/issues/1507

### DIFF
--- a/ui/src/wizard/ResourceGroupListStep.ts
+++ b/ui/src/wizard/ResourceGroupListStep.ts
@@ -68,7 +68,7 @@ export class ResourceGroupListStep<T extends types.IResourceGroupWizardContext> 
     }
 
     private async getQuickPicks(wizardContext: T): Promise<types.IAzureQuickPickItem<ResourceGroup | undefined>[]> {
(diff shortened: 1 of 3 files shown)


ui/src/wizard/ResourceGroupListStep.ts
```diff
@@ -79,7 +79,7 @@ export class ResourceGroupListStep<T extends types.IResourceGroupWizardContext>
         }
 
         const resourceGroups: ResourceGroup[] = await ResourceGroupListStep.getResourceGroups(wizardContext);
-        return picks.concat(resourceGroups.map((rg: ResourceGroup) => {
+        picks = picks.concat(resourceGroups.map((rg: ResourceGroup) => {
```
EricJizbaMSFT, line 82: I would sort the resource groups here before you turn them into picks, instead of sorting the picks at the end. This will ensure "Create new resource group" is always at the top

PR #733, opened 2020-06-10
Don't write zip file to disk when deploying

EricJizbaMSFT:
I was hoping the below changes would fix the [two](https://github.com/microsoft/vscode-azurefunctions/issues/2108) [bugs](https://github.com/microsoft/vscode-azureappservice/issues/1571) I've been investigating, but no dice 😭. Still think these changes are worth it, though:
1. I realized the zipper we get from `archiver` is a readable stream. I can pass that stream directly to `zipPushDeploy` instead of writing to a zip file on disk and  calling `fs.createReadStream` to get a new stream. I feel like cutting out the middle man (a zip file on disk) can only help
1. Added a line to the logs about the package size. Seemed nice to have 🤷‍♂️

    > 6:37:35 PM emjjslin: Creating zip package...
    > 6:37:35 PM emjjslin: Zip package size: 1.83 kB
    > 6:37:35 PM emjjslin: Starting deployment...

1. Removed check for md5 hash on storage account deploy. This was partially because I didn't know how to make it work with part 1 above, but more importantly I just feel like it's hurting more users than it's helping. Fixes https://github.com/microsoft/vscode-azurefunctions/issues/1983


appservice/src/deploy/getZipStream.ts
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * Adds files using glob filtering
+ */
+function addFilesGlob(zipper: archiver.Archiver, folderPath: string): void {
+    const zipDeployConfig: vscode.WorkspaceConfiguration = vscode.workspace.getConfiguration(ext.prefix, vscode.Uri.file(folderPath));
+    // tslint:disable-next-line: strict-boolean-expressions
+    const globPattern: string = zipDeployConfig.get<string>('zipGlobPattern') || '**/*';
+    const ignorePattern: string | string[] | undefined = zipDeployConfig.get<string | string[]>('zipIgnorePattern');
+    zipper.glob(globPattern, { cwd: folderPath, ignore: ignorePattern, ...commonGlobSettings });
```
nturinski, line 65: Just want to confirm, but when you add the commonGlobSettings and `dot: true` does this mean we ARE ignoring it or not?

EricJizbaMSFT: First of all - if you look at `getZipFileToDeploy.ts`, the glob behavior should be the exact same as it was before

Secondly, what is "it"?

PR #736, opened 2020-06-11
Encode publish userName and password

alexweininger:
While working on local git deploy for trial apps I discovered a pretty cool bug 😎. We weren't encoding the `publishUserName` and `publishPassword` when we create the remote Url. And the usernames for trial apps always has a `$` in it so it was throwing an invalid URL exception. I am also encoding the password because although not common I believe it might be possible for a reserved character to show up in there as well.

### DIFF
--- a/appservice/src/deploy/localGitDeploy.ts
+++ b/appservice/src/deploy/localGitDeploy.ts
@@ -18,10 +18,11 @@ import { waitForDeploymentToComplete } from './waitForDeploymentToComplete';
 export async function localGitDeploy(client: SiteClient, fsPath: string, context: IActionContext): Promise<void> {
     const publishCredentials: User = await client.getWebAppPublishCredential();
     const publishingPassword: string = nonNullProp(publishCredentials, 'publishingPassword');
+    const publishingUserName: string = nonNullProp(publishCredentials, 'publishingUserName');
 
     await callWithMaskHandling(
         async (): Promise<void> => {
-            const remote: string = `https://${nonNullProp(publishCredentials, 'publishingUserName')}:${nonNullProp(publishCredentials, 'publishingPassword')}@${client.gitUrl}`;
+            const remote: string = `https://${encodeURIComponent(publishingUserName)}:${encodeURIComponent(publishingPassword)}@${client.gitUrl}`;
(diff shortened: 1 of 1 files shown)


appservice/src/deploy/localGitDeploy.ts
```diff
@@ -18,10 +18,11 @@ import { waitForDeploymentToComplete } from './waitForDeploymentToComplete';
 export async function localGitDeploy(client: SiteClient, fsPath: string, context: IActionContext): Promise<void> {
     const publishCredentials: User = await client.getWebAppPublishCredential();
     const publishingPassword: string = nonNullProp(publishCredentials, 'publishingPassword');
+    const publishingUserName: string = nonNullProp(publishCredentials, 'publishingUserName');
 
     await callWithMaskHandling(
         async (): Promise<void> => {
-            const remote: string = `https://${nonNullProp(publishCredentials, 'publishingUserName')}:${nonNullProp(publishCredentials, 'publishingPassword')}@${client.gitUrl}`;
+            const remote: string = `https://${encodeURIComponent(publishingUserName)}:${encodeURIComponent(publishingPassword)}@${client.gitUrl}`;
```
EricJizbaMSFT, line 25: I don't think our current `callWithMaskHandling` would handle an encoded value. Can you add a test case [here](https://github.com/microsoft/vscode-azuretools/blob/master/ui/test/callWithMaskHandling.test.ts) where the error message thrown includes an encoded version of the valueToMask? And then fix the bug assuming the test fails

alexweininger: While doing this discovered that I can't seem to run our tests due to https://github.com/microsoft/vscode/issues/99492. So for now I will just fix the bug then add an issue to add the test case when I (or anyone else) is able to run the tests again.

EricJizbaMSFT: That seems to only affect running from the command line. I normally run tests inside VS Code. In fact, the default and only F5 experience for our npm packages is running tests:
Screen Shot 2020-06-11 at 12 10 19 PM

alexweininger: I see... Lol thanks

PR #737, opened 2020-06-11
Mask encoded strings

alexweininger:
* add test case for encoded string
* replace and mask encoded values

Conext: https://github.com/microsoft/vscode-azuretools/pull/736#discussion_r438984861

### DIFF
--- a/ui/src/callWithMaskHandling.ts
+++ b/ui/src/callWithMaskHandling.ts
@@ -17,9 +17,15 @@ export async function callWithMaskHandling<T>(callback: () => Promise<T>, valueT
             throw error;
         }
 
-        const maskedMessage: string = valueToMask ?
+        let maskedMessage: string = valueToMask ?
             parsedError.message.replace(new RegExp(escape(valueToMask), 'g'), '***') :
             parsedError.message;
+
+        // replace encoded value
+        maskedMessage = valueToMask ?
+            maskedMessage.replace(new RegExp(escape(encodeURIComponent(valueToMask)), 'g'), '***') :
+            maskedMessage;
+
         throw new Error(maskedMessage);
     }
 }
--- a/ui/src/callWithMaskHandling.ts
+++ b/ui/src/callWithMaskHandling.ts
@@ -17,14 +17,12 @@ export async function callWithMaskHandling<T>(callback: () => Promise<T>, valueT
             throw error;
         }
 
-        let maskedMessage: string = valueToMask ?
-            parsedError.message.replace(new RegExp(escape(valueToMask), 'g'), '***') :
-            parsedError.message;
+        const valuesToMask: string[] = [valueToMask, encodeURIComponent(valueToMask)];
 
-        // replace encoded value
-        maskedMessage = valueToMask ?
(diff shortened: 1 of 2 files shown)


ui/src/callWithMaskHandling.ts
```diff
@@ -17,9 +17,15 @@ export async function callWithMaskHandling<T>(callback: () => Promise<T>, valueT
             throw error;
         }
 
-        const maskedMessage: string = valueToMask ?
+        let maskedMessage: string = valueToMask ?
```
EricJizbaMSFT, line 20: You should be able to do this in a loop instead of copying code

Something like:
```
const valuesToMask: string[] = [originalValueToMask, encodeURIComponent(originalValueToMask)];
for (const valueToMask of valuesToMask) {
```

ui/test/callWithMaskHandling.test.ts
```diff
@@ -88,6 +88,25 @@ suite("callWithMaskHandling Tests", () => {
...
+           hint: to the same ref. You may want to first integrate the remote changes
+           hint: (e.g., 'git pull ...') before pushing again.
+           hint: See the 'Note about fast-forwards' in 'git push --help' for details.`;
+
+            await assertThrowsAsync(async (): Promise<void> => {
+                await callWithMaskHandling(async () => {
+                    throw errorMessage;
+                }, '$scHQERrAlXSmlCeN1mrhDzsHWeDz2XZt5R343HgCNmxS0xlswcaA2Cowflda');
+            }, (err: Error) => {
+                return validateError(err, credentials);
```
EricJizbaMSFT, line 107: I think the value you pass to `validateError` should be the same as the value you pass to callWithMaskHandling in order to be more robust. Right now the test kind of relies on the fact that they're the same except for one character. 

To fix that, you'll probably have to adjust `validateError` to check for the encoded values, too

ui/test/callWithMaskHandling.test.ts
```diff
@@ -8,9 +8,11 @@ import { parseError } from '../src/parseError';
 import { assertThrowsAsync } from './assertThrowsAsync';
 
 // tslint:disable: align
+// tslint:disable: max-func-body-length
 suite("callWithMaskHandling Tests", () => {
     const credentials: string = 'scHQERrAlXSmlCeN1mrhDzsHWeDz2XZt5R343HgCNmxS0xlswcaA2Cowflda';
     const credentialsSpecialChars: string = 'sc()HQ*E+RrAlXSm[CeN1$$$rhDz^^HWeDz2X[t5R343HgCN.xS0x]swc|A2CÑwf¬da';
+    const credentialsWithReservedChars: string = '$scHQERrAlXSmlCeN1mrhDzsHWeDz2XZt5R343HgCNmxS0xlswcaA2Cowflda';
```
EricJizbaMSFT, line 15: Lol `credentialsWithReservedChars` feels like a lame version of `credentialsSpecialChars`

alexweininger: I know haha. I wanted to make it clear that these weren't just your regular special chars however. Because I believe that reserved chars in URIs are a different set of characters than special characters. I'm gonna keep it in there.

PR #739, opened 2020-06-12
Local git deploy for trial apps

alexweininger:
* update simple-git dependency
* Add a few items to `IDeploymentsClient` in order to satisfy local git deploy
* Add `localGitDeploy` to the exports in index.ts so I can use it in the app service extension without having to go through `deploy.ts` which has a bunch of extra logic irrelevant to trial apps since only local git deploy is available.
* Local git deploy now has an optional parameter that can control whether or not to commit for the user (defaults to false).
* Local git deploy now also pushes the branch the user is currently on. Instead of defaulting to master. This is necessary for trial apps since the default branch is `RELEASE`.

View changes in use: https://github.com/microsoft/vscode-azureappservice/pull/1595

### DIFF
--- a/appservice/src/IDeploymentsClient.ts
+++ b/appservice/src/IDeploymentsClient.ts
@@ -3,7 +3,7 @@
 *  Licensed under the MIT License. See License.txt in the project root for license information.
 *--------------------------------------------------------------------------------------------*/
 
-import { SiteConfigResource, SiteSourceControl, StringDictionary, User } from 'azure-arm-website/lib/models';
+import { SiteConfigResource, SiteSourceControl, User } from 'azure-arm-website/lib/models';
 import { KuduClient } from 'vscode-azurekudu';
 
 export interface IDeploymentsClient {
@@ -12,8 +12,6 @@ export interface IDeploymentsClient {
     isFunctionApp: boolean;
     isLinux: boolean;
     gitUrl: string | undefined;
-    listApplicationSettings(): Promise<StringDictionary>;
-    updateApplicationSettings(appSettings: StringDictionary): Promise<StringDictionary>;
     getKuduClient(): Promise<KuduClient>;
     getSiteConfig(): Promise<SiteConfigResource>;
     getSourceControl(): Promise<SiteSourceControl>;
--- a/appservice/src/IDeploymentsClient.ts
+++ b/appservice/src/IDeploymentsClient.ts
@@ -3,14 +3,19 @@
 *  Licensed under the MIT License. See License.txt in the project root for license information.
 *--------------------------------------------------------------------------------------------*/
 
-import { SiteConfigResource, SiteSourceControl } from 'azure-arm-website/lib/models';
+import { SiteConfigResource, SiteSourceControl, StringDictionary, User } from 'azure-arm-website/lib/models';
 import { KuduClient } from 'vscode-azurekudu';
 
 export interface IDeploymentsClient {
 
     fullName: string;
     isFunctionApp: boolean;
+    isLinux: boolean;
+    gitUrl: string | undefined;
+    listApplicationSettings(): Promise<StringDictionary>;
+    updateApplicationSettings(appSettings: StringDictionary): Promise<StringDictionary>;
     getKuduClient(): Promise<KuduClient>;
     getSiteConfig(): Promise<SiteConfigResource>;
     getSourceControl(): Promise<SiteSourceControl>;
+    getWebAppPublishCredential(): Promise<User>;
 }
--- a/appservice/src/deploy/localGitDeploy.ts
+++ b/appservice/src/deploy/localGitDeploy.ts
@@ -25,9 +25,9 @@ export async function localGitDeploy(client: SiteClient, fsPath: string, context
             const remote: string = `https://${encodeURIComponent(publishingUserName)}:${encodeURIComponent(publishingPassword)}@${client.gitUrl}`;
             const localGit: git.SimpleGit = git(fsPath);
             const commitId: string = (await localGit.log()).latest.hash;
-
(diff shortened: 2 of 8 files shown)


appservice/src/IDeploymentsClient.ts
```diff
@@ -3,14 +3,19 @@
...
+import { SiteConfigResource, SiteSourceControl, StringDictionary, User } from 'azure-arm-website/lib/models';
 import { KuduClient } from 'vscode-azurekudu';
 
 export interface IDeploymentsClient {
 
     fullName: string;
     isFunctionApp: boolean;
+    isLinux: boolean;
+    gitUrl: string | undefined;
+    listApplicationSettings(): Promise<StringDictionary>;
```
nturinski, line 15: This doesn't feel right.  ApplicationSettings shouldn't be in the DeploymentsClient, I'd think.

appservice/src/deploy/localGitDeploy.ts
```diff
@@ -74,7 +74,7 @@ export async function localGitDeploy(client: SiteClient, fsPath: string, context
                         // for whatever reason, is '-f' exists, true or false, it still force pushes
                         const pushOptions: git.Options = forcePush ? { '-f': true } : {};
 
-                        localGit.push(remote, 'HEAD:master', pushOptions).catch(async (error) => {
+                        localGit.push(remote, `HEAD:${status.current}`, pushOptions).catch(async (error) => {
```
nturinski, line 77: I think that this will break the current experience.  The remotes that exist in Azure itself only have a master branch.  That is, Azure only looks at the master branch when kudu builds the website on the remote.

If I recall correctly, HEAD:master means use your current branch, and push it the remote master branch.

appservice/src/deploy/localGitDeploy.ts
```diff
@@ -74,7 +74,7 @@ export async function localGitDeploy(client: SiteClient, fsPath: string, context
                         // for whatever reason, is '-f' exists, true or false, it still force pushes
```
nturinski, line 74: Might be able to fix this now?

appservice/src/IDeploymentsClient.ts
alexweininger: Yeah it didn't feel right to me either. However `verifyNoRunFromPackageSetting` uses the client from `localGitDeploy`... I could only call verifyNoRunPackageSetting if `client instanceof IAppSettingsTreeItem` maybe.

appservice/src/deploy/localGitDeploy.ts
alexweininger: Darn I thought I had tested and it worked still for normal apps. I just tested and yeah it breaks the current experience. I think I'll add a parameter to `localGitDeploy` in which you can specify the branch.

nturinski: Is this behavior different in trial apps?  I'd figure that you would just deploy to the trial app's master branch as well.

alexweininger: Ya it has to be HEAD:RELEASE to work for trial apps.

appservice/src/deploy/localGitDeploy.ts
alexweininger: image
Since -f is just a flag and doesn't need any value. We can pass null here and it will force push. I think the behavior the original comment is explaining is actually expected based on the documentation.

appservice/src/deploy/localGitDeploy.ts
```diff
@@ -8,14 +8,15 @@ import * as git from 'simple-git/promise';
...
+import { IDeploymentsClient } from '../IDeploymentsClient';
 import { localize } from '../localize';
-import { SiteClient } from '../SiteClient';
 import { nonNullProp } from '../utils/nonNull';
 import { openUrl } from '../utils/openUrl';
 import { verifyNoRunFromPackageSetting } from '../verifyNoRunFromPackageSetting';
 import { waitForDeploymentToComplete } from './waitForDeploymentToComplete';
 
-export async function localGitDeploy(client: SiteClient, fsPath: string, context: IActionContext): Promise<void> {
+export async function localGitDeploy(client: IDeploymentsClient & IAppSettingsClient, fsPath: string, context: IActionContext, branch: string = 'master', commit: boolean = false): Promise<void> {
```
nturinski, line 19: I think it's worth passing an options parameter.  You can make it like this:
`options: localGitOptions`

```
type localGitOptions {
   fsPath: string;
   branch?: string;
   commit?: boolean;
}
```

The `localGitDeploy` is getting a little cluttered in the parameters.

appservice/src/deploy/localGitDeploy.ts
```diff
@@ -70,11 +70,15 @@ export async function localGitDeploy(client: SiteClient, fsPath: string, context
                 const tokenSource: vscode.CancellationTokenSource = new vscode.CancellationTokenSource();
                 const token: vscode.CancellationToken = tokenSource.token;
                 try {
-                    await new Promise((resolve: () => void, reject: (error: Error) => void): void => {
-                        // for whatever reason, is '-f' exists, true or false, it still force pushes
-                        const pushOptions: git.Options = forcePush ? { '-f': true } : {};
+                    await new Promise(async (resolve: () => void, reject: (error: Error) => void): Promise<void> => {
+
+                        if (commit) {
+                            await localGit.commit('Deploy using LocalGit');
```
nturinski, line 76: Since we're trying to hide that it's local git, I'd just say something like 'Deployed via Azure App Service Extension' or something like that.

appservice/src/deploy/localGitDeploy.ts
alexweininger: Good idea

PR #740, opened 2020-06-15
Fix zip stream not working on larger zips

EricJizbaMSFT:
When I did https://github.com/microsoft/vscode-azuretools/pull/733, I was testing on Node.js projects which resulted in very small zips. I did some further testing with C# projects (which have larger zips) and it wasn't working for a variety of reasons.

1. `zipStream.finalize` was never returning for larger projects. I have to set up where the stream is piped _before_ I await that. Honestly this makes sense in my head - I just didn't do it before because it was slightly more complicated and it seemed to be working anyways.
1. `zipStream.readableLength` isn't the full size of the zip. Per the docs, it "contains the number of bytes  **_in the queue_** ready to be read". This was a problem when I used `createBlockBlobFromStream` because it requires me to pass the full content length ahead of time. Instead, I found `createWriteStreamToBlockBlob`, which doesn't need the content length. Again, this makes sense in my head - there's no way to know the size of a stream ahead of time because the point of streams is you kind of queue stuff up instead of doing it all at once.

### DIFF
--- a/appservice/src/deploy/runWithZipStream.ts
+++ b/appservice/src/deploy/runWithZipStream.ts
@@ -16,11 +16,22 @@ import { localize } from '../localize';
 import { SiteClient } from '../SiteClient';
 import { getFileExtension } from '../utils/pathUtils';
 
-export async function getZipStream(context: IActionContext, fsPath: string, client: SiteClient): Promise<Readable> {
-    let zipStream: Readable;
+export async function runWithZipStream(context: IActionContext, fsPath: string, client: SiteClient, callback: (zipStream: Readable) => Promise<void>): Promise<void> {
+    function onFileSize(size: number): void {
+        context.telemetry.measurements.zipFileSize = size;
+        ext.outputChannel.appendLog(localize('zipSize', 'Zip package size: {0}', prettybytes(size)), { resourceName: client.fullName });
+    }
+
+    let zipStream: Readable & { finalize?(): Promise<void>; };
     if (getFileExtension(fsPath) === 'zip') {
         context.telemetry.properties.alreadyZipped = 'true';
         zipStream = fse.createReadStream(fsPath);
+
+        // don't wait
+        // tslint:disable-next-line: no-floating-promises
+        fse.lstat(fsPath).then(stats => {
+            onFileSize(stats.size);
+        });
     } else {
         ext.outputChannel.appendLog(localize('zipCreate', 'Creating zip package...'), { resourceName: client.fullName });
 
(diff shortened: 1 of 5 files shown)


appservice/src/deploy/runWithZipStream.ts
```diff
@@ -39,13 +50,26 @@ export async function getZipStream(context: IActionContext, fsPath: string, clie
         } else {
             zipper.file(fsPath, { name: path.basename(fsPath) });
         }
-        await zipper.finalize();
         zipStream = zipper;
+        zipper.on('end', () => {
+            onFileSize(zipper.pointer());
```
nturinski, line 55: Can this not go in the generic task where end resolves the stream (L63)?

appservice/src/deploy/runWithZipStream.ts
```diff
@@ -16,11 +16,22 @@ import { localize } from '../localize';
 import { SiteClient } from '../SiteClient';
 import { getFileExtension } from '../utils/pathUtils';
 
-export async function getZipStream(context: IActionContext, fsPath: string, client: SiteClient): Promise<Readable> {
-    let zipStream: Readable;
+export async function runWithZipStream(context: IActionContext, fsPath: string, client: SiteClient, callback: (zipStream: Readable) => Promise<void>): Promise<void> {
```
nturinski, line 19: Is there any reason you couldn't just return the zip stream at the end and pass it in as an argument for `kuduClient.pushDeployment.zipPushDeploy` and `createWriteStreamToBlockBlob`?  Just never see us use the callback pattern.

EricJizbaMSFT: The main benefit of this pattern is I can reduce copied code. If I return zipStream, I would have to duplicate some of the logic below the comment "Setup several tasks" because of the ordering of the tasks and when they're awaited

I've used the callback pattern before. `registerCommand` and `callWithTelemetryAndErrorHandling` are the most common uses, but here's a few more examples:

https://github.com/microsoft/vscode-azuretools/blob/master/ui/index.d.ts#L224

https://github.com/microsoft/vscode-azurefunctions/blob/93a8ebe455785c9264d0b235df54468eaf567252/test/runWithSetting.ts#L16

appservice/src/deploy/runWithZipStream.ts
EricJizbaMSFT: Uh I could, but `pointer` is specific to `zipper` (aka it doesn't apply in the "alreadyZipped" case). I'd have to change the type of `zipStream` to
```
Readable & { finalize?(): Promise<void>; pointer?(): number };
```
And do a check like this:
```
if (zipStream.pointer) {
    onFileSize(zipStream.pointer());
}
```

I thought it was a little simpler to handle on `zipper` itself

appservice/src/deploy/runWithZipStream.ts
nturinski: Okay fine, I will say that I feel like those use-cases are a little different, but I am okay with it

PR #746, opened 2020-06-17
Expose AzureAccount property

alexweininger:
Expose an `AzureAccount` property so that children can easily check if the user is signed in, etc.

In the Azure App Service extension we end up checking if a user is signed in to decide deploy behavior.  I didn't like how using `if (children.length > 0 && children[0] instanceof GenericTreeItem)` felt. But I see that Functions does the same thing so maybe this change is unnecessary and I can just keep checking `if (children.length > 0 && children[0] instanceof GenericTreeItem)`.

### DIFF
--- a/ui/src/treeDataProvider/AzureAccountTreeItemBase.ts
+++ b/ui/src/treeDataProvider/AzureAccountTreeItemBase.ts
@@ -37,6 +37,7 @@ export abstract class AzureAccountTreeItemBase extends AzExtParentTreeItem imple
     private _azureAccountTask: Promise<AzureAccount | undefined>;
     private _subscriptionTreeItems: SubscriptionTreeItemBase[] | undefined;
     private _testAccount: AzureAccount | undefined;
+    private _isLoggedIn: boolean = false;
 
     constructor(parent?: AzExtParentTreeItem, testAccount?: AzureAccount) {
         super(parent);
@@ -50,6 +51,10 @@ export abstract class AzureAccountTreeItemBase extends AzExtParentTreeItem imple
 
     public get iconPath(): types.TreeItemIconPath {
         return getIconPath('azure');
+    }
+
+    public get isLoggedIn(): boolean {
+        return this._isLoggedIn;
     }
 
     public dispose(): void {
@@ -79,6 +84,8 @@ export abstract class AzureAccountTreeItemBase extends AzExtParentTreeItem imple
         context.telemetry.properties.accountStatus = azureAccount.status;
         const existingSubscriptions: SubscriptionTreeItemBase[] = this._subscriptionTreeItems ? this._subscriptionTreeItems : [];
         this._subscriptionTreeItems = [];
+
+        this._isLoggedIn = azureAccount.status === 'LoggedIn';
 
         const contextValue: string = 'azureCommand';
         if (azureAccount.status === 'Initializing' || azureAccount.status === 'LoggingIn') {
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -396,6 +396,7 @@ export declare abstract class AzureAccountTreeItemBase extends AzExtParentTreeIt
     public disposables: Disposable[];
     public childTypeLabel: string;
     public autoSelectInTreeItemPicker: boolean;
+    public get isLoggedIn(): boolean;
 
     //#region Methods implemented by base class
     /**


ui/src/treeDataProvider/AzureAccountTreeItemBase.ts
```diff
@@ -32,6 +32,7 @@ export abstract class AzureAccountTreeItemBase extends AzExtParentTreeItem imple
     public readonly label: string = 'Azure';
     public childTypeLabel: string = localize('subscription', 'subscription');
     public autoSelectInTreeItemPicker: boolean = true;
+    public azureAccount: AzureAccount | undefined;
```
EricJizbaMSFT, line 35: This isn't actually exposed because you didn't update [index.d.ts](https://github.com/microsoft/vscode-azuretools/blob/master/ui/index.d.ts). That file is our public contract and affects whether or not something is "exposed".

More importantly, it's been a very conscious choice _not_ to expose this. The types for "AzureAccount" are defined [here](https://github.com/microsoft/vscode-azuretools/blob/master/ui/src/azure-account.api.d.ts#L13) and we don't want that in our public contract (index.d.ts) because we don't own those types. They are owned by the Azure Account extension [here](https://github.com/microsoft/vscode-azure-account/blob/master/src/azure-account.api.d.ts).

ui/src/treeDataProvider/AzureAccountTreeItemBase.ts
```diff
@@ -61,49 +62,49 @@ export abstract class AzureAccountTreeItemBase extends AzExtParentTreeItem imple
     }
 
     public async loadMoreChildrenImpl(_clearCache: boolean, context: types.IActionContext): Promise<AzExtTreeItem[]> {
-        let azureAccount: AzureAccount | undefined = await this._azureAccountTask;
-        if (!azureAccount) {
+        this.azureAccount = await this._azureAccountTask;
```
EricJizbaMSFT, line 65: > Expose an AzureAccount property so that children can easily check if the user is signed in, etc.

Please spend a little bit more time on your PR descriptions, especially in this repo because it affects many extensions. We can usually figure out _what_ you're doing by looking at the changes (although a short description doesn't hurt), but I want to know more about _why_ you're doing something. For example, why do children need to easily check if the user is signed in?

alexweininger: Apologies for the lack of description.

In the Azure App Service extension we end up checking if a user is signed in to decide deploy behavior.  I didn't like how using `if (children.length > 0 && children[0] instanceof GenericTreeItem)` felt. But I see that Functions does the same thing so maybe this change is unnecessary and I can just keep checking `if (children.length > 0 && children[0] instanceof GenericTreeItem)`. @nturinski and I decided that this would be a good way of checking if a user is signed in without checking the child nodes.

EricJizbaMSFT: How does it affect deploy behavior if a user is signed in?

As mentioned in my other comment, I don't want to expose the "AzureAccount" types in this package. I suggest you either check `children` like functions is doing or find another way around the problem without exposing the full `AzureAccount`. For example, maybe you just add a simple "isSignedIn" property to this class. That way, you're only exposing a boolean instead of an entire class

alexweininger: If the user has a trial app, but is not signed in, then we can automatically deploy to their trial app. If the user is signed in then we have to go through other steps to get the deploy target.

> For example, maybe you just add a simple "isSignedIn" property to this class. That way, you're only exposing a boolean instead of an entire class

Plan on doing this now.

EricJizbaMSFT: Hmm as of today, I'm guessing it would show these three options in the tree item picker when they deploy while not signed-in:
- trial app
- Sign in to azure
- Create free azure account

I agree it's a little more annoying to have to click the trial app - but isn't it important for us to highlight the other two options?

EricJizbaMSFT:
Dismissing my review since I'll be OOF the rest of this week and exposing the `AzureAccount` class was the main thing I was "requesting changes" on

ui/src/treeDataProvider/AzureAccountTreeItemBase.ts
alexweininger: I don't think this would be a the best user flow. Especially since the majority of trial app users won't be signed in (at least that is our goal). And for this user, we want the deploy button to do everything for the user. We might have to discuss this more with @fiveisprime.

PR #749, opened 2020-06-22
Only show text documents for existent tree items

wwlorey:
It's possible to call `showTextDocument` on a tree item that doesn't exist. For instance in databases, a collection that doesn't exist could be passed in [here](https://github.com/microsoft/vscode-cosmosdb/blob/140c026ea00e733aaa2d0b83b57db0e3d9855536/src/mongo/MongoScrapbook.ts#L115).

Prevents this error:
Screen Shot 2020-06-22 at 9 56 56 AM

### DIFF
--- a/ui/src/AzExtTreeFileSystem.ts
+++ b/ui/src/AzExtTreeFileSystem.ts
@@ -8,6 +8,7 @@ import { Disposable, Event, EventEmitter, FileChangeEvent, FileStat, FileSystemE
 import * as types from '../index';
 import { callWithTelemetryAndErrorHandling } from "./callWithTelemetryAndErrorHandling";
 import { localize } from "./localize";
+import { parseError } from "./parseError";
 import { AzExtTreeDataProvider } from "./treeDataProvider/AzExtTreeDataProvider";
 import { AzExtTreeItem } from "./treeDataProvider/AzExtTreeItem";
 import { nonNullProp } from "./utils/nonNull";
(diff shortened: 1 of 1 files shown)


ui/src/AzExtTreeFileSystem.ts
```diff
@@ -37,7 +38,17 @@ export abstract class AzExtTreeFileSystem<TItem extends AzExtTreeItem> implement
     public abstract getFilePath(item: TItem): string;
 
     public async showTextDocument(item: TItem): Promise<void> {
-        await window.showTextDocument(this.getUriFromItem(item));
+        await callWithTelemetryAndErrorHandling('showTextDocument', async (context) => {
+            try {
+                await window.showTextDocument(this.getUriFromItem(item));
+            } catch (err) {
+                const parsedError: types.IParsedError = parseError(err);
+                if (parsedError.message.match(/EntryNotFound \(FileSystemError\)/)) {
```
wwlorey, line 46: The `errorType` here is "Error" so I opted to check the message

EricJizbaMSFT:
So I looked into the [scenario](https://github.com/microsoft/vscode-cosmosdb/blob/140c026ea00e733aaa2d0b83b57db0e3d9855536/src/mongo/MongoScrapbook.ts#L115) you linked and I don't think there should be an error at all. Mongo lets users run commands (find, insert, etc.) on a collection even if it doesn't technically exist yet. Instead, I think we need to change `findItem` [here](https://github.com/microsoft/vscode-cosmosdb/blob/master/src/DatabasesFileSystem.ts#L86) to return a new collection tree item just like we're doing in MongoScrapbook.ts

wwlorey:
Closing in favor of https://github.com/microsoft/vscode-cosmosdb/pull/1582

PR #752, opened 2020-06-23
findSubscriptionTreeItem util

alexweininger:
We need this util to be able to access the `SubscriptionTreeItem` of any `AzExtTreeItem` (assuming it exists).

This is used in `openInPortal` so that we can pass in `AzExtTreeItem` which don't have direct access to the `root` property that `AzureTreeItem`'s do.

* Bump version
* Update the types.
* Add `findSubscriptionTreeItem` util

PR that depends on these changes: https://github.com/microsoft/vscode-azureappservice/pull/1613

### DIFF
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -459,7 +459,7 @@ export declare abstract class AzureParentTreeItem<TRoot extends ISubscriptionCon
 /**
 * Combines the root.environment.portalLink and id to open the resource in the portal.
 */
-export declare function openInPortal(root: ISubscriptionContext, id: string, options?: OpenInPortalOptions): Promise<void>;
+export declare function openInPortal(root: ISubscriptionContext | AzExtTreeItem, id: string, options?: OpenInPortalOptions): Promise<void>;
 
 export declare class UserCancelledError extends Error { }
 
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -458,6 +458,8 @@ export declare abstract class AzureParentTreeItem<TRoot extends ISubscriptionCon
 
 /**
 * Combines the root.environment.portalLink and id to open the resource in the portal.
+*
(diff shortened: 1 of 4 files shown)


ui/index.d.ts
```diff
@@ -459,7 +459,7 @@ export declare abstract class AzureParentTreeItem<TRoot extends ISubscriptionCon
 /**
 * Combines the root.environment.portalLink and id to open the resource in the portal.
```
EricJizbaMSFT, line 460: I try to keep these docs as up to date as possible. Maybe add a note like this:
> If `root` is a tree item, it will find the subscription ancestor and get environment.portalLink from there

ui/src/utils/treeUtils.ts
```diff
@@ -0,0 +1,21 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.txt in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import { SubscriptionTreeItemBase } from '..';
```
EricJizbaMSFT, line 6: The imports for this repo are weird and VS Code doesn't auto-import from the best place. I'd feel better if you fixed the `SubscriptionTreeItemBase` import to look more like the `AzExtTreeItem ` import

ui/package.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "vscode-azureextensionui",
     "author": "Microsoft Corporation",
-    "version": "0.33.6",
+    "version": "0.33.7",
```
EricJizbaMSFT, line 4: You forgot to bump package-lock.json, but I already bumped it to "0.33.7" in a PR I just merged, so it doesn't really matter. We can include everything together

PR #754, opened 2020-06-23
Deployments node contextValue based on portal support

alexweininger:
This is so that when the deploymentsNode is a child of a trial app it won't have `Open in Portal` context menu items.

View changes in use here: https://github.com/microsoft/vscode-azureappservice/pull/1615


alexweininger:
Closing since we decided to just throw an error when users click on open in portal in https://github.com/microsoft/vscode-azureappservice/pull/1615

PR #756, opened 2020-06-26
Better handle duplicate tree items

EricJizbaMSFT:
Duplicate tree items can happen for a variety of reasons and right now it completely messes up that level of the tree. Instead, I changed it to more gracefully display an error. NOTE: I faked a duplicate tree item by adding two application setting tree items to a function app.

Before:
Screen Shot 2020-06-26 at 1 53 02 PM
Screen Shot 2020-06-26 at 1 53 06 PM

After:
Screen Shot 2020-06-26 at 1 52 15 PM

Fixes https://github.com/microsoft/vscode-azurefunctions/issues/2155

### DIFF
--- a/ui/src/treeDataProvider/AzExtParentTreeItem.ts
+++ b/ui/src/treeDataProvider/AzExtParentTreeItem.ts
@@ -248,10 +248,7 @@ export abstract class AzExtParentTreeItem extends AzExtTreeItem implements types
     }
 
     private async loadMoreChildrenInternal(context: types.IActionContext): Promise<void> {
-        this._isLoadingMore = true;
         try {
-            this.treeDataProvider.refreshUIOnly(this);
-
             if (this._clearCache) {
                 // Just in case implementers of `loadMoreChildrenImpl` re-use the same child node, we want to clear those caches as well
                 for (const child of this._cachedChildren) {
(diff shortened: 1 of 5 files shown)


ui/src/treeDataProvider/AzExtParentTreeItem.ts
```diff
@@ -248,10 +248,7 @@ export abstract class AzExtParentTreeItem extends AzExtTreeItem implements types
     }
 
     private async loadMoreChildrenInternal(context: types.IActionContext): Promise<void> {
-        this._isLoadingMore = true;
         try {
-            this.treeDataProvider.refreshUIOnly(this);
```
EricJizbaMSFT, line 253: I moved this to `AzExtTreeDataProvider.loadMore` because it was causing too many refreshes to happen, making it impossible to expand an InvalidTreeItem. It has to do with the fact that VS Code saves collabsible state based on a tree item's id and we always use a [unique id](https://github.com/microsoft/vscode-azuretools/blob/master/ui/src/treeDataProvider/AzExtParentTreeItem.ts#L374) for InvalidTreeItems.

`loadMoreChildrenInternal` is called when a user clicks "Load more..." _and_ the first time a tree item is expanded, whereas `AzExtTreeDataProvider.loadMore` is _only_ called when a user clicks "Load more...".

alexweininger:
Is this related to https://github.com/microsoft/vscode-azureappservice/issues/1601?

EricJizbaMSFT:
@alexweininger yeah my changes are meant to more gracefully handle errors like that, but I don't actually fix the root cause of that specific issue

PR #757, opened 2020-06-29
Alex/select folder to deploy

alexweininger:
Changes are shown in use here: https://github.com/microsoft/vscode-azureappservice/pull/1628

### DIFF
--- a/appservice/package.json
+++ b/appservice/package.json
@@ -1,7 +1,7 @@
 {
     "name": "vscode-azureappservice",
     "author": "Microsoft Corporation",
-    "version": "0.62.3",
+    "version": "0.62.4",
     "description": "Common tools for developing Azure App Service extensions for VS Code",
     "tags": [
         "azure",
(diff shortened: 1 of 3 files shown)


alexweininger:
I'm wondering if we should just change the placeholder from "Select the folder to zip and deploy" to "Select the folder to deploy" for trial apps and real apps.

appservice/src/deploy/getDeployFsPath.ts
```diff
@@ -45,7 +45,7 @@ export async function getDeployFsPath(context: IActionContext, target: vscode.Ur
         }
 
         const selectFile: string = localize('selectDeployFile', 'Select the {0} file to deploy', fileExtensions ? fileExtensions.join('/') : '');
-        const selectFolder: string = localize('selectZipDeployFolder', 'Select the folder to zip and deploy');
+        const selectFolder: string = trialApp ? localize('selectDeployFolder', 'Select the folder to deploy') : localize('selectZipDeployFolder', 'Select the folder to zip and deploy');
```
EricJizbaMSFT, line 48: Just use "Select the folder to deploy" in all cases. Not worth the complexity to make them different

appservice/package.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "vscode-azureappservice",
     "author": "Microsoft Corporation",
-    "version": "0.62.3",
+    "version": "0.62.4",
```
EricJizbaMSFT, line 4: Normally I wouldn't release a new version for something this simple 🤷‍♂️

PR #758, opened 2020-06-29
Support appending in openReadOnlyContent

wwlorey:
Related to https://github.com/microsoft/vscode-cosmosdb/pull/1584


ui/index.d.ts
```diff
@@ -1256,8 +1256,10 @@ export function openReadOnlyJson(node: { label: string, fullId: string }, data:
  * @param node Typically (but not strictly) an `AzExtTreeItem`. `label` is used for the file name displayed in VS Code and `fullId` is used to uniquely identify this file
  * @param content The content to display
  * @param fileExtension The file extension
+ * @param appendDelimiter If provided, `content` will be appended to the document delimited by this string
+ * @param openInNewViewColumn Whether or not to open the read-only editor in a new `ViewColumn`
  */
-export function openReadOnlyContent(node: { label: string, fullId: string }, content: string, fileExtension: string): Promise<void>;
+export function openReadOnlyContent(node: { label: string, fullId: string }, content: string, fileExtension: string, appendDelimiter?: string, openInNewViewColumn?: boolean): Promise<void>;
```
EricJizbaMSFT, line 1262: When you're authoring an api (which you're basically doing), it's always good to model after existing apis when possible. I feel like the closest case is `showTextDocument` which returns an `editor`. Perhaps we should do something similar - although the editor wouldn't need to be as complex - maybe just support a few methods like `append` and `clear`

A few reasons I suggest that:
- I'm not a huge fan of `appendDelimiter`. Feels like I would just add that to `content`
- Also not a huge fan of `openInNewViewColumn`. Feels like a worse knock-off of `TextDocumentShowOptions`
- Seems like your current implementation always appends. I feel like that could get unwieldy for Databases. I would think each "round" of "Execute All" should get it's own fresh results file

ui/src/openReadOnlyContent.ts
```diff
@@ -29,13 +33,33 @@ export async function openReadOnlyJson(node: { label: string, fullId: string },
...
+        this._node = node;
+        this._fileExtension = fileExtension;
+    }
+
+    public async append(content: string, options?: TextDocumentShowOptions): Promise<void> {
+        await this._contentProvider.openReadOnlyContent(this._node, content, this._fileExtension, { append: true, ...options });
+    }
+
+    public clear(): void {
+        this._contentProvider.clearContent(this._node, this._fileExtension);
```
EricJizbaMSFT, line 61: It's a bit odd to have `ReadOnlyContentProvider` and `ReadOnlyContent` referencing each other. Instead, change `ReadOnlyContentProvider._contentMap` to `Map<string, ReadOnlyContent>` and store the content directly in the `ReadOnlyContent` class

ui/package.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "vscode-azureextensionui",
     "author": "Microsoft Corporation",
-    "version": "0.33.7",
+    "version": "0.33.8",
```
EricJizbaMSFT, line 4: I already released 0.33.8. You'll need to bump again

ui/index.d.ts
```diff
@@ -1251,13 +1251,20 @@ export function createAzExtOutputChannel(name: string, extensionPrefix: string):
  */
 export function openReadOnlyJson(node: { label: string, fullId: string }, data: {}): Promise<void>;
 
+export class ReadOnlyContent {
+    public readonly uri: Uri;
```
EricJizbaMSFT, line 1255: I would prefer not to expose `uri` outside of this package, so that we're free to change how it's structured without worrying about breaking anything

ui/src/openReadOnlyContent.ts
```diff
@@ -29,33 +29,71 @@ export async function openReadOnlyJson(node: { label: string, fullId: string },
...
+    }
+
+    public clear(): void {
+        this._content = '';
+        this._emitter.fire(this._uri);
+    }
+
+    public async isVisible(): Promise<boolean> {
+        const visibleDocuments: TextDocument[] = window.visibleTextEditors.map(editor => editor.document);
+        return visibleDocuments.includes(await workspace.openTextDocument(this._uri));
```
EricJizbaMSFT, line 68: Isn't there a uri property on `editor.document` you could check? I think that would be more reliable as you're currently relying on object equality, which isn't the best. VS Code might make different instances of the same document class under the covers

ui/index.d.ts
```diff
@@ -1251,13 +1251,21 @@ export function createAzExtOutputChannel(name: string, extensionPrefix: string):
  */
 export function openReadOnlyJson(node: { label: string, fullId: string }, data: {}): Promise<void>;
 
+export class ReadOnlyContent {
+    public append(content: string): Promise<void>;
+    public clear(): void;
+    public isVisible(): Promise<boolean>;
+    public show(options?: TextDocumentShowOptions): Promise<void>;
```
EricJizbaMSFT, line 1258: Not 100% convinced `isVisible` and `show` are needed for your current scenario, but could see them being useful in the future so 🤷‍♂️

PR #763, opened 2020-07-07
Set 'acceptLanguage' option to VS Code language when creating Azure client

JeffreyCA:
This change sets the Azure client language to the VS Code language, so that response messages are properly localized.

### DIFF
--- a/ui/src/createAzureClient.ts
+++ b/ui/src/createAzureClient.ts
@@ -20,7 +20,7 @@ export function createAzureClient<T extends IAddUserAgent>(
 export function createAzureSubscriptionClient<T extends IAddUserAgent>(
     clientInfo: { credentials: ServiceClientCredentials; environment: AzureEnvironment; },
     clientType: new (credentials: ServiceClientCredentials, baseUri?: string, options?: AzureServiceClientOptions) => T): T {
-    const client: T = new clientType(clientInfo.credentials, clientInfo.environment.resourceManagerEndpointUrl);
+    const client: T = new clientType(clientInfo.credentials, clientInfo.environment.resourceManagerEndpointUrl, { acceptLanguage: vscode.env.language });
     addExtensionUserAgent(client);
     return client;
 }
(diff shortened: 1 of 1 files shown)


ui/index.d.ts
```diff
@@ -1205,7 +1205,8 @@ export declare function addExtensionUserAgent(client: IAddUserAgent): void;
  */
 export function createAzureClient<T extends IAddUserAgent>(
```
EricJizbaMSFT, line 1206: > For my use case, I wanted to be able to create a client that used a different acceptLanguage option for localization reasons.

I assume you just pass along the VS Code language? I'd actually prefer if you just added that functionality inside `createAzureClient`. The point of the method is to ensure best practices, and localization fits that bill

JeffreyCA: Yes that's what I was going for

ui/src/createAzureClient.ts
```diff
@@ -5,13 +5,14 @@
...
 import { addExtensionUserAgent } from "./extensionUserAgent";
 
 export function createAzureClient<T extends IAddUserAgent>(
     clientInfo: { credentials: ServiceClientCredentials; subscriptionId: string; environment: AzureEnvironment; },
     clientType: new (credentials: ServiceClientCredentials, subscriptionId: string, baseUri?: string, options?: AzureServiceClientOptions) => T): T {
-    const client: T = new clientType(clientInfo.credentials, clientInfo.subscriptionId, clientInfo.environment.resourceManagerEndpointUrl);
+    const client: T = new clientType(clientInfo.credentials, clientInfo.subscriptionId, clientInfo.environment.resourceManagerEndpointUrl, { acceptLanguage: vscode.env.language });
     addExtensionUserAgent(client);
     return client;
 }
```
EricJizbaMSFT, line 18: Can you add it to `createAzureSubscriptionClient`, too? I assume it works the same way

JeffreyCA: Sure

EricJizbaMSFT:
Thanks!

PR #766, opened 2020-07-13
Add PoliCheck to build definition

EricJizbaMSFT:
Same as https://github.com/microsoft/vscode-azurefunctions/pull/2240


EricJizbaMSFT:
Not sure if we'll actually GA these per-se, but I do think these need to be GA-level quality since they're so tightly integrated with our extensions

PR #772, opened 2020-07-16
Add onAccountStatusChanged to listen for account status changes

### LINKED ISSUE #771: vscode-azureextensionui - Hooking into Azure Account events
I would like to dispose some resources that are no longer needed when the user signs out of their Azure account.

JeffreyCA:
Resolves #771

### DIFF
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -10,6 +10,7 @@ import { ServiceClientCredentials } from 'ms-rest';
 import { AzureEnvironment, AzureServiceClientOptions } from 'ms-rest-azure';
 import { Disposable, Event, ExtensionContext, FileChangeEvent, FileChangeType, FileStat, FileSystemProvider, FileType, InputBoxOptions, Memento, MessageItem, MessageOptions, OpenDialogOptions, OutputChannel, Progress, QuickPickItem, QuickPickOptions, TextDocument, TextDocumentShowOptions, ThemeIcon, TreeDataProvider, TreeItem, Uri } from 'vscode';
 import { AzureExtensionApi, AzureExtensionApiProvider } from './api';
+import { AzureLoginStatus } from './src/azure-account.api';
 
 export type OpenInPortalOptions = {
     /**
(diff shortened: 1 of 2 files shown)


EricJizbaMSFT:
Can you describe the scenario you want this for? Last time you submitted [a pr](https://github.com/microsoft/vscode-azuretools/pull/763), we wanted the functionality for ourselves, too

JeffreyCA:
I want to dispose some resources that are no longer needed when the user signs out.

JeffreyCA:
Basically I want to dispose some client connections that are no longer needed when the user signs out of their Azure account. There's more details in the referenced PR but it's pretty specific to my use case.

ui/index.d.ts
```diff
@@ -10,6 +10,7 @@ import { ServiceClientCredentials } from 'ms-rest';
 import { AzureEnvironment, AzureServiceClientOptions } from 'ms-rest-azure';
 import { Disposable, Event, ExtensionContext, FileChangeEvent, FileChangeType, FileStat, FileSystemProvider, FileType, InputBoxOptions, Memento, MessageItem, MessageOptions, OpenDialogOptions, OutputChannel, Progress, QuickPickItem, QuickPickOptions, TextDocument, TextDocumentShowOptions, ThemeIcon, TreeDataProvider, TreeItem, Uri } from 'vscode';
 import { AzureExtensionApi, AzureExtensionApiProvider } from './api';
+import { AzureLoginStatus } from './src/azure-account.api';
```
EricJizbaMSFT, line 13: I'm wondering if you should just reference the azure account directly. We try to avoid exposing anything related to "azure-account.api" in our external contract (aka this file) since those types are owned/managed by the Azure Account extension, not us.

Assuming the Azure Account extension is a direct dependency of your extension, it's actually quite simple. Our "loadAzureAccount" method is a bit more complex because we have some extensions where the Azure Account extension is an optional dependency. If it's a required dependency, you can safely assume the extension exists and is already activated when you get the exports.

JeffreyCA: Oh I see what you mean. I did what you said and it works flawlessly. I'll close this then!

PR #773, opened 2020-07-22
Allow runWithInputs to output from callback

bwateratmsft:
We have a use case for testing where it would be helpful for `runWithInputs` to be able to output something from the callback. This works and doesn't seem to cause any problems when `T` is `void`.

### DIFF
--- a/dev/src/TestUserInput.ts
+++ b/dev/src/TestUserInput.ts
@@ -32,10 +32,11 @@ export class TestUserInput implements types.TestUserInput {
         return this._onDidFinishPromptEmitter.event;
     }
 
-    public async runWithInputs(inputs: (string | RegExp | types.TestInput)[], callback: () => Promise<void>): Promise<void> {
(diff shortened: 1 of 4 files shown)


dev/src/TestUserInput.ts
```diff
@@ -32,10 +32,11 @@ export class TestUserInput implements types.TestUserInput {
         return this._onDidFinishPromptEmitter.event;
     }
 
-    public async runWithInputs(inputs: (string | RegExp | types.TestInput)[], callback: () => Promise<void>): Promise<void> {
+    public async runWithInputs<T>(inputs: (string | RegExp | types.TestInput)[], callback: () => Promise<T>): Promise<T> {
```
EricJizbaMSFT, line 35: Gotta update here, too: https://github.com/microsoft/vscode-azuretools/blob/master/dev/index.d.ts#L159

bwateratmsft: Ah! I saw src/index.ts and it doesn't need updating, thought index.d.ts was generated. I'll add it.

PR #774, opened 2020-07-23
Update some language in the app 🔑 settings prompts

nturinski:
Fixes https://github.com/microsoft/vscode-azurestaticwebapps/issues/141


appservice/src/tree/AppSettingsTreeItem.ts
```diff
@@ -12,19 +12,19 @@ import { getThemedIconPath } from './IconPath';
 
 export function validateAppSettingKey(settings: StringDictionary, client: IAppSettingsClient, newKey: string, oldKey?: string): string | undefined {
     if (client.isLinux && /[^\w\.]+/.test(newKey)) {
-        return 'App setting names can only contain letters, numbers (0-9), periods ("."), and underscores ("_")';
+        return 'App setting key names can only contain letters, numbers (0-9), periods ("."), and underscores ("_")';
```
EricJizbaMSFT, line 15: The portal doesn't use "key" anywhere that I could find. They stick with "name" and "value"

PR #775, opened 2020-07-24
Update dev package to new azure sdk for js

EricJizbaMSFT:
Related to https://github.com/microsoft/vscode-azuretools/issues/510

### DIFF
--- a/dev/src/TestAzureAccount.ts
+++ b/dev/src/TestAzureAccount.ts
@@ -3,8 +3,9 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { SubscriptionClient, SubscriptionModels } from 'azure-arm-resource';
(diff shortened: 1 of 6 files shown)


dev/src/TestAzureAccount.ts
```diff
@@ -56,7 +57,7 @@ export class TestAzureAccount implements AzureAccount, types.TestAzureAccount {
             environment: credentials.environment,
             userId: '',
             tenantId: tenantId,
-            credentials: credentials
+            credentials2: credentials
```
nturinski, line 60: They changed the property?  Is this to have backwards compatibility?  Is there any concern that they'll change this back to just credentials after support for the old SDKs are expired?

EricJizbaMSFT: Yes it is to maintain backwards compatibility, and no I'm not worried about them changing the name

https://github.com/microsoft/vscode-azure-account/issues/140#issuecomment-647969090

PR #776, opened 2020-07-24
Update ui package to new azure sdk for js

EricJizbaMSFT:
The Azure Account extension is currently restricted to VS Code Insiders, but I think there's enough change here that we can get started testing now. I used the resource groups and functions extensions as test subjects (PRs coming soon) and it all seems to be working.

For the UI package, here's the only difference noticeable to the user - if they don't have the latest account extension, we'll display a warning:
Screen Shot 2020-07-14 at 3 37 56 PM
Screen Shot 2020-07-14 at 3 38 17 PM

Fixes https://github.com/microsoft/vscode-azuretools/issues/510
Depends on https://github.com/microsoft/vscode-azuretools/pull/775

### DIFF
--- a/ui/src/createAzureClient.ts
+++ b/ui/src/createAzureClient.ts
@@ -48,13 +48,7 @@ export function createGenericClient(clientInfo?: ServiceClientCredentials | { cr
 class GenericServiceClient extends ServiceClient {
     constructor(credentials: ServiceClientCredentials | undefined, options: types.IMinimumServiceClientOptions) {
         super(credentials, options);
-
-        if (options.baseUri) {
-            if (options.baseUri.endsWith('/')) {
-                options.baseUri = options.baseUri.slice(0, -1);
-            }
-            this.baseUri = options.baseUri;
-        }
+        this.baseUri = options.baseUri?.endsWith('/') ? options.baseUri.slice(0, -1) : options.baseUri;
     }
 
     public async sendRequest(options: RequestPrepareOptions | WebResourceLike): Promise<HttpOperationResponse> {
--- a/ui/src/createAzureClient.ts
+++ b/ui/src/createAzureClient.ts
@@ -3,24 +3,73 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { ServiceClientCredentials } from "ms-rest";
-import { AzureEnvironment, AzureServiceClientOptions } from "ms-rest-azure";
+import { Environment } from '@azure/ms-rest-azure-env';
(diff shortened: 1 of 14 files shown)


ui/src/createAzureClient.ts
```diff
@@ -3,24 +3,73 @@
...
+        acceptLanguage: vscode.env.language,
+        baseUri: clientInfo.environment.resourceManagerEndpointUrl,
+        userAgent: appendExtensionUserAgent
+    });
+}
+
+export function createGenericClient(clientInfo?: ServiceClientCredentials | { credentials: ServiceClientCredentials; environment: Environment; }): ServiceClient {
+    let credentials: ServiceClientCredentials | undefined;
+    let baseUri: string | undefined;
+    if (clientInfo && 'credentials' in clientInfo) {
```
nturinski, line 35: And this has already has come in handy 😄

ui/src/createAzureClient.ts
```diff
@@ -3,24 +3,73 @@
...
-    clientType: new (credentials: ServiceClientCredentials, baseUri?: string, options?: AzureServiceClientOptions) => T): T {
-    const client: T = new clientType(clientInfo.credentials, clientInfo.environment.resourceManagerEndpointUrl, { acceptLanguage: vscode.env.language });
-    addExtensionUserAgent(client);
-    return client;
+class GenericServiceClient extends ServiceClient {
+    constructor(credentials: ServiceClientCredentials | undefined, options: types.IMinimumServiceClientOptions) {
+        super(credentials, options);
+
+        if (options.baseUri) {
+            if (options.baseUri.endsWith('/')) {
```
nturinski, line 53: nit: Ternary could reduce the lines

ui/src/createAzureClient.ts
```diff
@@ -3,24 +3,73 @@
...
+                options.baseUri = options.baseUri.slice(0, -1);
+            }
+            this.baseUri = options.baseUri;
+        }
+    }
+
+    public async sendRequest(options: RequestPrepareOptions | WebResourceLike): Promise<HttpOperationResponse> {
```
nturinski, line 60: Is this to replace the requestUtils.sendRequest?  I don't understand why it's in the `createAzureClient` file.

EricJizbaMSFT: It's a part of `GenericServiceClient`, which is returned in `createGenericClient`

EricJizbaMSFT: But anyways - yes this is related to replacing `requestUtils`. I figured `createGenericClient` was similar enough to `createAzureClient` that it could fit here. Only the function `createGenericClient` is actually exported, _not_ `GenericServiceClient`

PR #777, opened 2020-07-24
Update kudu package to new azure sdk for js

EricJizbaMSFT:
This PR is actually way fewer files than it seems because almost all of it is auto-generated by autorest (including all the ts files, tsconfig.json, rollup.config.js, and even parts of package.json). The main files we will ever edit are "autorestconfig.json" and "swagger.json"

Related to #510


kudu/rollup.config.js
```diff
@@ -0,0 +1,37 @@
+import rollup from "rollup";
```
nturinski, line 1: Does this need a license at the top too?

kudu/tsconfig.json
```diff
@@ -1,16 +1,19 @@
...
-        "node_modules",
-        "lib"
-    ]
-}
\ No newline at end of file
+  "compilerOptions": {
+    "module": "es6",
+    "moduleResolution": "node",
+    "strict": true,
+    "target": "es5",
```
nturinski, line 6: Should the target have changed especially since the module is es6?

kudu/package.json
```diff
@@ -1,43 +1,58 @@
 {
-    "name": "vscode-azurekudu",
-    "author": "Microsoft Corporation",
-    "version": "0.1.9",
-    "description": "Client used to interact with Kudu.",
-    "tags": [
```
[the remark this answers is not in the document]
EricJizbaMSFT: This change was related to the auto-generation, not intentional. In any case, the main goal of this package is for our own use, so tags don't really matter

kudu/tsconfig.json
EricJizbaMSFT: This file was generated by autorest, so I'd rather just trust them to pick this stuff

kudu/rollup.config.js
EricJizbaMSFT: I'm not worried about it. This file was generated with autorest, and the main sdk repo doesn't seem to add it to their files either - for example: https://github.com/Azure/azure-sdk-for-js/blob/master/sdk/appservice/arm-appservice/rollup.config.js

PR #778, opened 2020-07-24
Update appservice package to new azure sdk for js

EricJizbaMSFT:
Related to #510
Depends on https://github.com/microsoft/vscode-azuretools/pull/776
Depends on https://github.com/microsoft/vscode-azuretools/pull/777


appservice/src/deploy/waitForDeploymentToComplete.ts
```diff
@@ -149,7 +148,7 @@ async function tryGetLatestDeployment(context: IActionContext, kuduClient: KuduC
         // Use "latest" to get the deployment before we know the "initialReceivedTime" or "permanentId"
         try {
             deployment = await retryKuduCall(context, 'getResult', async () => {
-                return <DeployResult | undefined>await kuduClient.deployment.getResult('latest');
+                return await kuduClient.deployment.getResult('latest');
```
nturinski, line 151: Have you tried this when there's no deployment yet?  I know that use to cause an error, but the return type doesn't include "undefined" so I'm wondering what happens.

appservice/src/github/GitHubRepoListStep.ts
```diff
@@ -7,18 +7,25 @@ import { AzureWizardPromptStep, IAzureQuickPickItem } from 'vscode-azureextensio
...
+        let repoData: gitHubRepoData | string;
         const picksCache: ICachedQuickPicks<gitHubRepoData> = { picks: [] };
-        const requestOptions: gitHubWebResource = await createRequestOptions(context, nonNullProp(context, 'orgData').repos_url);
-        do {
-            repoData = (await ext.ui.showQuickPick(this.getRepositories(context, picksCache, requestOptions), { placeHolder })).data;
-        } while (!repoData);
+
+        let url: string = nonNullProp(context, 'orgData').repos_url;
+        // tslint:disable-next-line: no-constant-condition
+        while (true) {
```
nturinski, line 21: Pretty clever, I was wondering why you let repoData be a string too.

appservice/src/slotFunctionOperations.ts
```diff
@@ -3,37 +3,38 @@
...
 }
 
 /**
  * Temporary workaround because this isn't in azure sdk yet
  */
-export async function getFunctionSlot(subscription: ISubscriptionContext, id: string, functionName: string): Promise<FunctionEnvelope> {
-    const request: requestUtils.Request = await requestUtils.getDefaultAzureRequest(getUrlPath(id, functionName), subscription);
-    const response: IRawAzureResource = <IRawAzureResource>JSON.parse(await requestUtils.sendRequest(request));
-    return convertRawResource(response);
+export async function getFunctionSlot(subscription: ISubscriptionContext, id: string, functionName: string): Promise<WebSiteManagementModels.FunctionEnvelope> {
```
nturinski, line 26: This still isn't in the sdk?

appservice/src/startStreamingLogs.ts
```diff
@@ -61,7 +58,17 @@ export async function startStreamingLogs(client: ISimplifiedSiteClient, verifyLo
 
                 await new Promise((onLogStreamEnded: () => void, reject: (err: Error) => void): void => {
                     let newLogStream: ILogStream;
-                    const logsRequest: request.Request = requestApi(`${client.kuduUrl}/api/logstream/${logsPath}`);
+
+                    const logsRequest: request.Request = request(`${client.kuduUrl}/api/logstream/${logsPath}`, {
```
nturinski, line 62: This kind of seems more error prone than using the kuduClient/request API to sign and format the request.  Any reason for the change?

appservice/src/TunnelProxy.ts
```diff
@@ -187,25 +185,33 @@ export class TunnelProxy {
...
-        request.resolveWithFullResponse = true; // allows access to the status code from the response
-        const pingResponse: IncomingMessage = await requestUtils.sendRequest(request);
-        ext.outputChannel.appendLog(`[Tunnel] Ping responded with status code: ${pingResponse.statusCode}`);
+        const client: ServiceClient = createGenericClient();
+        let statusCode: number | undefined;
+        try {
+            const response: HttpOperationResponse = await client.sendRequest({ method: 'GET', url: this._client.defaultHostUrl });
+            statusCode = response.status;
+        } catch (error) {
+            if (error instanceof RestError) {
```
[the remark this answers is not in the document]
EricJizbaMSFT: Yes I was able to test this (Pretty sure I just added some bogus text to the url to make a 404 error happen). I did this to maintain consistency with the previous behavior of this line:
```
request.simple = false; // allows the call to succeed without exception, even when status code is not 2XX
```
I couldn't find a `simple` property with the new sdk, so I just catch the exceptions instead.

appservice/src/createAppService/AppInsightsCreateStep.ts
```diff
@@ -83,9 +82,12 @@ export class AppInsightsCreateStep extends AzureWizardExecuteStep<IAppServiceWiz
 
     private async getPairedRegions(locationName: string): Promise<string[]> {
         try {
-            const request: requestUtils.Request = await requestUtils.getDefaultRequest('https://appinsights.azureedge.net/portal/regionMapping.json');
-            const response: string = await requestUtils.sendRequest(request);
-            const regionMappingJson: RegionMappingJsonResponse = <RegionMappingJsonResponse>JSON.parse(response);
+            const client: ServiceClient = createGenericClient();
+            const response: HttpOperationResponse = await client.sendRequest({
+                method: 'GET',
+                url: 'https://appinsights.azureedge.net/portal/regionMapping.json'
```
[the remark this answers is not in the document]
EricJizbaMSFT: You're the one that [added this](https://github.com/microsoft/vscode-azuretools/pull/562) :P I just kept it as-is for my PR

appservice/src/deploy/waitForDeploymentToComplete.ts
EricJizbaMSFT: You just mean double check the logic for the catch block? My change here (to remove the type casting) should have no functional effect. It's possible the error thrown is different (i.e. maybe errorType ends up being "NotFound" instead of "404"), but I double checked deploying to a fresh app and it seemed to work. This scenario should be covered by nightly tests and CTI tests, so I feel like trying it once was enough for now.

appservice/src/github/GitHubRepoListStep.ts
EricJizbaMSFT: Yeah the old `createRequestOptions` didn't work with the new types for various reasons, so I cleaned it up a bit. This stuff should probably move to octokit anyways 🤷‍♂️

appservice/src/slotFunctionOperations.ts
EricJizbaMSFT: Nope 😞

appservice/src/startStreamingLogs.ts
EricJizbaMSFT: `signRequest` doesn't work like it did in the old sdk. Basically, if you want to use `signRequest`, you _have_ to send the request with a ServiceClient from "ms-rest-js". It's not compatible with the "request" module. That's a big part of why I added [createGenericClient](https://github.com/microsoft/vscode-azuretools/blob/master/ui/src/createAzureClient.ts#L32) in the UI package. I tried using that GenericClient here, but this log stream stuff is pretty unique, mostly because we keep the request open while the user is streaming logs. I plan to investigate this more in the future, because this it's pretty much the only code still using the "request" module and I'd love to remove that dependency (since it's deprecated)

appservice/src/createAppService/AppInsightsCreateStep.ts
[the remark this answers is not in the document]
nturinski: No need to assign blame!

appservice/src/github/GitHubRepoListStep.ts
nturinski: Yeah, was thinking we should move it all to appservice and just have Static Web Apps consume from there.

PR #779, opened 2020-07-29
Pick up lodash update for ui

StephenWeatherford:


ui/package.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "vscode-azureextensionui",
     "author": "Microsoft Corporation",
-    "version": "0.33.9",
+    "version": "0.33.10",
```
EricJizbaMSFT, line 4: npm packages don't ship their dependencies, so publishing a new version is unnecessary

StephenWeatherford: I figured it processed the dependent package-lock files, component governance claimed ui was one of the components bringing this in.

StephenWeatherford: I don't see it in npm ls, will look at component governance again.

PR #794, opened 2020-09-09
Fix deploy when AzureWebJobsStorage doesn't have EndpointSuffix

EricJizbaMSFT:
Fixes https://github.com/microsoft/vscode-azurefunctions/issues/2360

### DIFF
--- a/appservice/src/deploy/deployToStorageAccount.ts
+++ b/appservice/src/deploy/deployToStorageAccount.ts
@@ -4,10 +4,11 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { WebSiteManagementModels } from '@azure/arm-appservice';
+import { Environment } from '@azure/ms-rest-azure-env';
 import { BlobSASPermissions, BlobServiceClient, BlockBlobClient, ContainerClient, generateBlobSASQueryParameters, StorageSharedKeyCredential } from '@azure/storage-blob';
 import * as moment from 'moment';
 import { URL } from 'url';
-import { IActionContext } from 'vscode-azureextensionui';
+import { IActionContext, parseError } from 'vscode-azureextensionui';
 import { ext } from '../extensionVariables';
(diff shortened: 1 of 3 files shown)


appservice/src/deploy/deployToStorageAccount.ts
```diff
@@ -39,10 +40,28 @@ async function createBlobServiceClient(client: SiteClient): Promise<BlobServiceC
     // Use same storage account as AzureWebJobsStorage for deployments
     const azureWebJobsStorageKey: string = 'AzureWebJobsStorage';
     const settings: WebSiteManagementModels.StringDictionary = await client.listApplicationSettings();
-    if (settings.properties && settings.properties[azureWebJobsStorageKey]) {
-        return BlobServiceClient.fromConnectionString(settings.properties[azureWebJobsStorageKey]);
+    let connectionString: string | undefined = settings.properties && settings.properties[azureWebJobsStorageKey];
+    if (connectionString) {
+        try {
+            return BlobServiceClient.fromConnectionString(connectionString);
```
nturinski, line 46: Since it's required now, shouldn't we just always add the endpointSuffix if it doesn't have one?  If we know creating the BlobServiceClient will fail without it, seems like we're just wasting time by trying first.

nturinski (approved):

appservice/src/deploy/deployToStorageAccount.ts
EricJizbaMSFT: I consider what I'm doing to be a bit of "black box magic". When I do that, I prefer to always leave a path so that the user can avoid the magic if it doesn't work for them. In this case I can just avoid the magic by default since "fromConnectionString" does all the validation up front before trying to connect to Azure (aka it's super fast). Basically, I want to protect against the scenario where my code isn't perfect and doesn't handle all edge cases correctly haha.

PR #800, opened 2020-09-11
Refactor confirmOverwrite and downloadSettings

nturinski:
Confirm overwrite I was basically able to copy and paste in exactly from Functions.

downloadAppSettings was a little more tricky.  The refactor requires you to pass in the localFileSettings name and file path since it can vary from extension to extension (specifically App Service vs. SWA and Functions).  Also, only Functions supports encryption/decryption since it relies on the func cli core tools.  The refactor would require the Functions extension to do the decryption before the download and the encryption afterwards.

Not really sure if there's a way around that 🤔


appservice/src/appSettings/downloadAppSettings.ts
```diff
@@ -0,0 +1,49 @@
...
+import * as fse from 'fs-extra';
+import * as vscode from 'vscode';
+import { ext } from "../extensionVariables";
+import { localize } from "../localize";
+import { AppSettingsTreeItem } from "../tree/AppSettingsTreeItem";
+import { confirmOverwriteSettings } from "./confirmOverwriteSettings";
+import { IAppSettingsClient } from "./IAppSettingsClient";
+import { ILocalSettingsJson } from "./ILocalSettingsJson";
+
+export async function downloadAppSettings(node: AppSettingsTreeItem, localSettingsFileName: string, localSettingsPath: string): Promise<void> {
```
EricJizbaMSFT, line 16: > The refactor requires you to pass in the localFileSettings name and file path since it can vary from extension to extension (specifically App Service vs. SWA and Functions).

I get that app service would be different, but wouldn't SWA and Functions be the same?

> Also, only Functions supports encryption/decryption since it relies on the func cli core tools. The refactor would require the Functions extension to do the decryption before the download and the encryption afterwards.

Again, I feel like SWA should be the same as Functions in terms of encryption/decryption

appservice/src/appSettings/downloadAppSettings.ts
```diff
@@ -0,0 +1,49 @@
...
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { WebSiteManagementModels } from "@azure/arm-appservice";
+import * as fse from 'fs-extra';
+import * as vscode from 'vscode';
+import { ext } from "../extensionVariables";
+import { localize } from "../localize";
+import { AppSettingsTreeItem } from "../tree/AppSettingsTreeItem";
+import { confirmOverwriteSettings } from "./confirmOverwriteSettings";
```
EricJizbaMSFT, line 12: Why does "Download App Settings" use the shared app service package, but "Create Function" uses the functions extension API? How are you deciding which feature uses which method?

appservice/src/appSettings/downloadAppSettings.ts
nturinski: It is, but we don't have the hard dependency on the cli tools like Functions.  Technically you can still use the app settings without having the Function extension as well.  So not really sure how to handle that scenario.

appservice/src/appSettings/downloadAppSettings.ts
nturinski: Well, downloadAppSetting has a lot of shared code between App Service, Functions, and SWA.  However-- this code isn't _quite_ generic enough as it is since App Service's .env files don't rely on a JSON object.

Do you think it'd make more sense to expand the Functions API to include the download/upload app settings logic?  I really hadn't considered that.

appservice/src/appSettings/downloadAppSettings.ts
EricJizbaMSFT: Functions doesn't have a hard dependency on the cli tools, either. We only need the func cli if the settings are encrypted, and SWA would have the same problem

appservice/src/appSettings/confirmOverwriteSettings.ts
```diff
@@ -0,0 +1,81 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as vscode from 'vscode';
+import { DialogResponses } from "vscode-azureextensionui";
+import { ext } from "../extensionVariables";
+import { localize } from "../localize";
+
+export async function confirmOverwriteSettings(sourceSettings: { [key: string]: string }, destinationSettings: { [key: string]: string }, destinationName: string): Promise<void> {
```
EricJizbaMSFT, line 11: Lol turns out you already copied this code over: https://github.com/microsoft/vscode-azuretools/pull/357

nturinski: Oh right, I just moved it in this PR. lol.

appservice/src/appSettings/downloadAppSettings.ts
EricJizbaMSFT: I'm not convinced `downloadAppSettings` could be shared with App Service - would have to see how you make that generic. If it only works for Functions/SWA, that's where my question comes in. I don't want to start sticking a bunch of Functions-specific features in this package just for SWA.

nturinski: Yeah, I think it makes sense to leave OverwriteSettings in here (since it's pretty generic) but, I think I like the idea of doing all the appSetting stuff through the Functions API since it's technically all Functions under the hood.  Are you good with that?

EricJizbaMSFT: Yeah I think so

nturinski:
Closing in favor of leveraging the Functions API for this instead.

PR #801, opened 2020-09-15
Remove substeps when going back even when the prompt is skipped

nturinski:
This is to fix specifically the Function API bug where when you create a new project through the API, select C#, and then go back, the DotnetSteps were not properly getting removed.

Also, moving the code for clearing the context of going back into the loop broke like 4 tests, so I didn't want to mess with that since just moving `removeFromEnd` code solved the problem.

### DIFF
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -1,6 +1,6 @@
 {
     "name": "vscode-azureextensionui",
-    "version": "0.35.0",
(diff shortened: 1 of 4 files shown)


EricJizbaMSFT (approved):

nturinski:
😲 Is this speedy approval the power of tests?

PR #802, opened 2020-09-16
Fix App Insights location name equality

EricJizbaMSFT:
`displayName` isn't reliable because it will get translated. `name` should be more reliable - just need to do some extra checks around casing and white space

Fixes https://github.com/microsoft/vscode-azurefunctions/issues/2356

### DIFF
--- a/appservice/src/utils/azureUtils.ts
+++ b/appservice/src/utils/azureUtils.ts
@@ -1,0 +1,15 @@
+
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+// tslint:disable-next-line: export-name
+export function areLocationNamesEqual(name1: string | undefined, name2: string | undefined): boolean {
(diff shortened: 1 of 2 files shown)


appservice/src/utils/azureUtils.ts
```diff
@@ -0,0 +1,15 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+// tslint:disable-next-line: export-name
+export function areLocationNamesEqual(name1: string | undefined, name2: string | undefined): boolean {
+    return normalizeLocationName(name1) === normalizeLocationName(name2);
+}
+
+function normalizeLocationName(name: string | undefined): string {
```
nturinski, line 12: Can we somehow just use LocationListStep's `generalizeLocationName` instead of having a util file for this?

nturinski (approved):

appservice/src/utils/azureUtils.ts
EricJizbaMSFT: We could, but this code felt more like `delay` or `nonNull` - aka little utils that don't quite make sense to be exposed in the UI package, so we just copy it to every repo we need it in

PR #803, opened 2020-09-17
Add ignoreFocusOut to ITreeItemPickerContext

gnailuy:
Hi vscode-azuretools team,

I'm trying to add picker using [`AzExtTreeDataProvider.showTreeItemPicker()`](https://github.com/microsoft/vscode-azuretools/blob/57fcd807bf9140bcd433d3e9e8c11efb9ac64870/ui/src/treeDataProvider/AzExtTreeDataProvider.ts#L144). Looks like `ignoreFocusOut` is not set in [`AzExtParentTreeItem.pickChildTreeItem()`](https://github.com/microsoft/vscode-azuretools/blob/57fcd807bf9140bcd433d3e9e8c11efb9ac64870/ui/src/treeDataProvider/AzExtParentTreeItem.ts#L105), and when it is `undefined` it will be treat as `true` in [`AzureUserInput.showQuickPick()`](https://github.com/microsoft/vscode-azuretools/blob/57fcd807bf9140bcd433d3e9e8c11efb9ac64870/ui/src/AzureUserInput.ts#L36), so the picker will always ignore focus out.

I think we can add this property to the `ITreeItemPickerContext` interface, so that the caller can control this behavior. Here is my PR, please help review it. Also please inform me if there are better solutions. Thank you so much!

### DIFF
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -117,6 +117,11 @@ export interface ITreeItemPickerContext extends IActionContext {
     suppressCreatePick?: boolean;
 
     /**
+     * If set to true, the quick pick dialog will not close when focus moves out.
+     */
+    ignoreFocusOut?: boolean;
+
+    /**
      * When no item is available for user to pick, this message will be displayed in the error notification.
      * This will also suppress the report issue button.
      */
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -117,7 +117,7 @@ export interface ITreeItemPickerContext extends IActionContext {
     suppressCreatePick?: boolean;
 
     /**
-     * If set to true, the quick pick dialog will not close when focus moves out.
+     * If set to true, the quick pick dialog will not close when focus moves out. Defaults to `true`.
      */
     ignoreFocusOut?: boolean;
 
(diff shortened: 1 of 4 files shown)


ui/index.d.ts
```diff
@@ -116,6 +116,11 @@ export interface ITreeItemPickerContext extends IActionContext {
      */
     suppressCreatePick?: boolean;
 
+    /**
+     * If set to true, the quick pick dialog will not close when focus moves out.
```
EricJizbaMSFT, line 120: This should say "defaults to `true`" at the end

EricJizbaMSFT (commented):
If you want this released after you fix the minor doc comment, please bump the version in package.json and package-lock.json (i.e. 0.35.1 -> 0.35.2). Otherwise we'll just include it whenever the next release happens

ui/index.d.ts
gnailuy: Thank you, Eric! Fixed the comment.

It's best to have this released soon, so I've updated the version to `0.35.2`. Thanks a lot.

EricJizbaMSFT (approved):

EricJizbaMSFT:
@gnailuy should be released, thanks!

gnailuy:
Thanks a lot, Eric!

PR #806, opened 2020-09-30
Lazily load @azure dependencies

bwateratmsft:
I tried out this approach and with these changes (plus similar ones in `vscode-docker`) I cut the extension (initial) code loading time by about 75%, which is _huge_. Since most Docker users are not doing anything with Azure this represents a substantial perf gain for them.

I had to make this a breaking change and bump the major version number because of `createGenericClient`. These inline `import` statements need an `await`, which meant something had to become `async`. Moving `GenericServiceClient` into its own class file removes the need to do confusing (impossible?) imports to have a subclass extend imported code.

I looked at all consumers of `createGenericClient` and they all appear to be `async` already, so consuming the new version won't entail much work.

The reason for this shape of imports--where the imports are untouched at the top but changed only in one spot in the body--is that everything _except_ what I touched is just typing information--for TypeScript's benefit--but the argument given to `createAzureClient` is what actually results in an `import` in the emitted JavaScript.

### DIFF
--- a/ui/src/clients.ts
+++ b/ui/src/clients.ts
@@ -1,0 +1,22 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { ResourceManagementClient } from '@azure/arm-resources';
+import { StorageManagementClient } from '@azure/arm-storage';
+import { SubscriptionClient } from '@azure/arm-subscriptions';
+import * as types from '../index';
+import { createAzureClient, createAzureSubscriptionClient } from './createAzureClient';
+
+export async function createStorageClient<T extends types.IStorageAccountWizardContext>(wizardContext: T): Promise<StorageManagementClient> {
+    return createAzureClient(wizardContext, (await import('@azure/arm-storage')).StorageManagementClient);
+}
+
+export async function createResourcesClient<T extends types.IResourceGroupWizardContext>(wizardContext: T): Promise<ResourceManagementClient> {
+    return createAzureClient(wizardContext, (await import('@azure/arm-resources')).ResourceManagementClient);
+}
+
+export async function createSubscriptionsClient<T extends types.ISubscriptionWizardContext>(wizardContext: T): Promise<SubscriptionClient> {
+    return createAzureSubscriptionClient(wizardContext, (await import('@azure/arm-subscriptions')).SubscriptionClient);
+}
--- a/ui/src/clients.ts
+++ b/ui/src/clients.ts
@@ -9,6 +9,9 @@ import { SubscriptionClient } from '@azure/arm-subscriptions';
 import * as types from '../index';
 import { createAzureClient, createAzureSubscriptionClient } from './createAzureClient';
 
+// Lazy-load @azure packages to improve startup performance.
+// NOTE: The client is the only import that matters, the rest of the types disappear when compiled to JavaScript
+
 export async function createStorageClient<T extends types.IStorageAccountWizardContext>(wizardContext: T): Promise<StorageManagementClient> {
     return createAzureClient(wizardContext, (await import('@azure/arm-storage')).StorageManagementClient);
 }
--- a/ui/src/createAzureClient.ts
+++ b/ui/src/createAzureClient.ts
@@ -4,7 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { Environment } from '@azure/ms-rest-azure-env';
-import { HttpOperationResponse, RequestPrepareOptions, ServiceClient, ServiceClientCredentials, WebResourceLike } from '@azure/ms-rest-js';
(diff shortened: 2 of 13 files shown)


ui/src/wizard/LocationListStep.ts
```diff
@@ -50,7 +50,8 @@ export class LocationListStep<T extends ILocationWizardContextInternal> extends
 
     public static async getLocations<T extends ILocationWizardContextInternal>(wizardContext: T): Promise<SubscriptionModels.Location[]> {
         if (wizardContext._allLocationsTask === undefined) {
-            const client: SubscriptionClient = createAzureSubscriptionClient(wizardContext, SubscriptionClient);
+            const armSubscriptions: typeof import('@azure/arm-subscriptions') = await import('@azure/arm-subscriptions');
+            const client: SubscriptionClient = createAzureSubscriptionClient(wizardContext, armSubscriptions.SubscriptionClient);
```
bwateratmsft, line 54: These could all be shortened with the previous line eliminated, let me know what you think.
```suggestion
            const client: SubscriptionClient = createAzureSubscriptionClient(wizardContext, (await import('@azure/arm-subscriptions')).SubscriptionClient);
```

ui/src/createAzureClient.ts
```diff
@@ -29,7 +29,7 @@ export function createAzureSubscriptionClient<T>(
     });
 }
 
-export function createGenericClient(clientInfo?: ServiceClientCredentials | { credentials: ServiceClientCredentials; environment: Environment; }): ServiceClient {
+export async function createGenericClient(clientInfo?: ServiceClientCredentials | { credentials: ServiceClientCredentials; environment: Environment; }): Promise<ServiceClient> {
```
bwateratmsft, line 32: The other methods in this file could also be made `async` for consistency--though I haven't inspected all of their consumers--but I opted to err on the side of less breaking changes instead. I'm OK with either.

ui/src/wizard/ResourceGroupCreateStep.ts
```diff
@@ -21,7 +21,8 @@ export class ResourceGroupCreateStep<T extends types.IResourceGroupWizardContext
         const newName: string = wizardContext.newResourceGroupName!;
         // tslint:disable-next-line:no-non-null-assertion
         const newLocation: string = wizardContext.location!.name!;
-        const resourceClient: ResourceManagementClient = createAzureClient(wizardContext, ResourceManagementClient);
+        const armResources: typeof import('@azure/arm-resources') = await import('@azure/arm-resources');
```
EricJizbaMSFT, line 24: This looks pretty tedious to copy/paste if we need to create a client again. How about you add a separate "clients" file of some kind with functions like "createResourceClient" and "createStorageClient". They'll be short functions, but that way we could use autocomplete instead of having to copy/paste these lines when adding new client code

Also, I could see myself copying the "clients" file over to other repos when implementing this

ui/src/createAzureClient.ts
```diff
@@ -39,31 +39,10 @@ export function createGenericClient(clientInfo?: ServiceClientCredentials | { cr
         credentials = clientInfo;
     }
 
-    return new GenericServiceClient(credentials, <types.IMinimumServiceClientOptions>{
+    const gsc: typeof import('./GenericServiceClient') = await import('./GenericServiceClient');
```
EricJizbaMSFT, line 42: Did you test this import stuff with webpack? And no issues?

ui/src/createAzureClient.ts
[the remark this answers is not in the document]
bwateratmsft: Looks like this one is not in `index.d.ts`...

ui/src/createAzureClient.ts
bwateratmsft: Yes, and yes :smile:

ui/src/createAzureClient.ts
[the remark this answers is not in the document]
EricJizbaMSFT: https://github.com/microsoft/vscode-azuretools/blob/main/ui/index.d.ts#L1218

ui/src/clients.ts
```diff
@@ -0,0 +1,22 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { ResourceManagementClient } from '@azure/arm-resources';
+import { StorageManagementClient } from '@azure/arm-storage';
+import { SubscriptionClient } from '@azure/arm-subscriptions';
+import * as types from '../index';
+import { createAzureClient, createAzureSubscriptionClient } from './createAzureClient';
+
```
EricJizbaMSFT, line 11: How about a comment to explain what's going on - not sure this is obvious. Something like:

> // Lazy-load `@azure` packages to improve startup performance. 
> // NOTE: The client is the only import that matters, the rest of the types disappear when compiled to JavaScript

EricJizbaMSFT (approved):

ui/src/clients.ts
bwateratmsft: Good idea!

PR #807, opened 2020-10-01
Update appservice to use latest ui

bwateratmsft:


appservice/package.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "vscode-azureappservice",
     "author": "Microsoft Corporation",
-    "version": "0.66.2",
+    "version": "0.67.0",
```
bwateratmsft, line 4: I wasn't really sure what good/bad could happen if the consumer is using the older version of the `ui` package along with a newer version of this one--so I thought the safest bet is to rev the major on this one as well. Let me know what you think.

EricJizbaMSFT: Yeah this is safest

appservice/package.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "vscode-azureappservice",
     "author": "Microsoft Corporation",
```
EricJizbaMSFT, line 3: I assume you don't care about lazy-loading azure packages here because you already lazy-load "vscode-azureappservice" itself?

EricJizbaMSFT (approved):

appservice/package.json
bwateratmsft: That's correct. It certainly wouldn't hurt to apply lazy-loading in here, but I'll leave that for later.

PR #808, opened 2020-10-02
Switch vscode-azureappservice to dayjs

bwateratmsft:
`dayjs`, after webpacking, is about 7 KB compared to about 700 KB for `moment`. This will mean faster load times and smaller packed VSIXs.

### DIFF
--- a/appservice/src/deploy/deployToStorageAccount.ts
+++ b/appservice/src/deploy/deployToStorageAccount.ts
@@ -6,7 +6,9 @@
 import { WebSiteManagementModels } from '@azure/arm-appservice';
 import { Environment } from '@azure/ms-rest-azure-env';
 import { BlobSASPermissions, BlobServiceClient, BlockBlobClient, ContainerClient, generateBlobSASQueryParameters, StorageSharedKeyCredential } from '@azure/storage-blob';
-import * as moment from 'moment';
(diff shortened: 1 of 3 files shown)


appservice/src/deploy/deployToStorageAccount.ts
```diff
@@ -15,13 +17,16 @@ import { SiteClient } from '../SiteClient';
...
+dayjs.extend(utc);
+
 /**
  * Method of deployment that is only intended to be used for Linux Consumption Function apps because it doesn't support kudu pushDeployment
  * To deploy with Run from Package on a Windows plan, create the app setting "WEBSITE_RUN_FROM_PACKAGE" and set it to "1".
  * Then deploy via "zipdeploy" as usual.
  */
 export async function deployToStorageAccount(context: IActionContext, fsPath: string, client: SiteClient): Promise<void> {
-    const datePart: string = moment.utc(Date.now()).format('YYYYMMDDHHmmss');
+    const datePart: string = dayjs().utc().format('YYYYMMDDHHmmss');
```
bwateratmsft, line 29: I compared the output from the old code vs. the new code, here and both places below, to ensure the identical outcome.

EricJizbaMSFT (approved):

PR #809, opened 2020-10-07
Lazy load @azure packages

EricJizbaMSFT:
Same deal as https://github.com/microsoft/vscode-azuretools/pull/806

Had to make this a breaking version change because "setLocationsTask" changed from sync to async

### DIFF
--- a/appservice/src/utils/azureClients.ts
+++ b/appservice/src/utils/azureClients.ts
@@ -1,0 +1,24 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { ApplicationInsightsManagementClient } from '@azure/arm-appinsights';
+import { WebSiteManagementClient } from '@azure/arm-appservice';
+import { ResourceManagementClient } from '@azure/arm-resources';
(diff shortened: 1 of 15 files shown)


appservice/src/utils/azureClients.ts
```diff
@@ -0,0 +1,24 @@
...
+import { ApplicationInsightsManagementClient } from '@azure/arm-appinsights';
+import { WebSiteManagementClient } from '@azure/arm-appservice';
+import { ResourceManagementClient } from '@azure/arm-resources';
+import { createAzureClient, ISubscriptionContext } from 'vscode-azureextensionui';
+
+// Lazy-load @azure packages to improve startup performance.
+// NOTE: The client is the only import that matters, the rest of the types disappear when compiled to JavaScript
+
+export async function createWebSiteClient<T extends ISubscriptionContext>(context: T): Promise<WebSiteManagementClient> {
+    return createAzureClient(context, (await import('@azure/arm-appservice')).WebSiteManagementClient);
```
nturinski, line 15: Once it's imported, if we run createWebSiteClient again, the await should return immediately, right?

nturinski (approved):

appservice/src/utils/azureClients.ts
EricJizbaMSFT: Yes - Brandon even did some testing to verify

https://github.com/microsoft/vscode-docker/pull/2368#discussion_r498988999

PR #811, opened 2020-10-19
Implement experimentation service

bwateratmsft:
This implements the experimentation service (and does an unrelated change to expose `NoResourceFoundError` which we need for some other purposes).

Tested:
- [x] Shared telemetry properties
- [x] Experimentation service (flighting)
- [x] Experimentation service (treatment variables)

### DIFF
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -1223,7 +1223,7 @@ export interface IExperimentationServiceAdapter {
      * Gets a treatment variable from the cache (which will be ~1 session delayed)
      * @param name The variable name
      */
-    getCachedTreatmentVariable<T extends string | number | boolean>(name: string): T | undefined;
+    getCachedTreatmentVariable<T extends string | number | boolean>(name: string): Promise<T | undefined>;
 
     /**
      * Gets a treatment variable directly from the web. This is slower than cache and can result in behavior changing mid-session.
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -1201,6 +1201,8 @@ export interface UIExtensionVariables {
 export interface IExperimentationServiceAdapter {
     isCachedFlightEnabled(flight: string): Promise<boolean>;
     isLiveFlightEnabled(flight: string): Promise<boolean>;
+    getCachedTreatmentVariable<T extends string | number | boolean>(name: string): T | undefined;
+    getLiveTreatmentVariable<T extends string | number | boolean>(name: string): Promise<T | undefined>;
 }
 
 export interface IAddUserAgent {
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -10,6 +10,7 @@ import { Environment } from '@azure/ms-rest-azure-env';
 import { ServiceClient, ServiceClientCredentials } from '@azure/ms-rest-js';
 import { TokenCredentialsBase } from '@azure/ms-rest-nodeauth';
 import { Disposable, Event, ExtensionContext, FileChangeEvent, FileChangeType, FileStat, FileSystemProvider, FileType, InputBoxOptions, Memento, MessageItem, MessageOptions, OpenDialogOptions, OutputChannel, Progress, QuickPickItem, QuickPickOptions, TextDocument, TextDocumentShowOptions, ThemeIcon, TreeDataProvider, TreeItem, Uri } from 'vscode';
+import { TargetPopulation } from 'vscode-tas-client';
 import { AzureExtensionApi, AzureExtensionApiProvider } from './api';
 
 export type OpenInPortalOptions = {
@@ -1175,6 +1176,11 @@ export declare namespace DialogResponses {
 export declare function registerUIExtensionVariables(extVars: UIExtensionVariables): void;
 
 /**
+ * Call this to register the experimentation service adapter
+ */
+export declare async function registerExperimentationService(ctx: ExtensionContext, targetPopulation?: TargetPopulation): Promise<void>;
+
+/**
  * Interface for common extension variables used throughout the UI package.
  */
 export interface UIExtensionVariables {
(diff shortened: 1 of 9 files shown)


nturinski:
Closing and reopening to kick-start build checks

ui/src/createTelemetryReporter.ts
```diff
@@ -37,3 +38,31 @@ export function createTelemetryReporter(ctx: vscode.ExtensionContext): IInternal
 
     return newReporter;
 }
+
+class InternalTelemetryReporter extends TelemetryReporter implements IInternalTelemetryReporter {
```
bwateratmsft, line 42: This will not actually add the shared properties to all events (which is necessary), only to those sent via `postEvent`. Should probably internally register a `TelemetryHandler`.

ui/src/DebugReporter.ts
```diff
@@ -9,6 +9,20 @@ import { IInternalTelemetryReporter } from './createTelemetryReporter';
...
+     */
+    public postEvent(): void {
+        // Not needed
+    }
+
+    /**
+     * Implements `setSharedProperty` for `IExperimentationTelemetry`
+     */
+    public setSharedProperty(): void {
+        // Not needed
```
bwateratmsft, line 23: There's not a good way to verify that the telemetry is actually working with this in place.

ui/src/registerExperimentationService.ts
```diff
@@ -0,0 +1,95 @@
...
+            );
+        } catch {
+            // Best effort
+        }
+    }
+
+    ext.experimentationService = result;
+}
+
+class ExperimentationServiceAdapter implements IExperimentationServiceAdapter {
```
bwateratmsft, line 34: Consider moving this and `ExperimentationTelemetry` to another file and/or give them their own folder. @EricJizbaMSFT what do you think?

ui/index.d.ts
```diff
@@ -1182,12 +1190,19 @@ export interface UIExtensionVariables {
...
+    experimentationService?: IExperimentationServiceAdapter;
+
     /**
      * Set to true if not running under a webpacked 'dist' folder as defined in 'vscode-azureextensiondev'
      */
     ignoreBundle?: boolean;
 }
 
+export interface IExperimentationServiceAdapter {
+    isCachedFlightEnabled(flight: string): Promise<boolean>;
```
bwateratmsft, line 1202: Should include support for treatment variables as well.

ui/src/registerExperimentationService.ts
```diff
@@ -0,0 +1,112 @@
...
+    public setSharedProperty(name: string, value: string): void {
+        this.sharedProperties[name] = value;
+    }
+
+    /**
+     * Implements a telemetry handler that adds the shared properties to the event
+     * @param actionContext The action context
+     */
+    public handleTelemetry(actionContext: IActionContext): void {
+        actionContext.telemetry.properties = {
```
EricJizbaMSFT, line 107: I prefer `Object.assign` so that the object reference stays the same instead of creating a new object. It's unlikely, but just in case someone is holding on to a reference to the telemetry properties

ui/index.d.ts
```diff
@@ -1174,6 +1177,11 @@ export declare namespace DialogResponses {
  */
 export declare function registerUIExtensionVariables(extVars: UIExtensionVariables): void;
 
+/**
+ * Call this to register the experimentation service adapter
+ */
+export declare async function registerExperimentationService(ctx: ExtensionContext, targetPopulation?: TargetPopulation): Promise<void>;
```
EricJizbaMSFT, line 1183: Lets add some docs for `targetPopulation`. Maybe:
> Optional parameter to override default population ("Insiders" if extension version includes "alpha", otherwise "Public")

ui/index.d.ts
```diff
@@ -1182,12 +1190,21 @@ export interface UIExtensionVariables {
...
     /**
      * Set to true if not running under a webpacked 'dist' folder as defined in 'vscode-azureextensiondev'
      */
     ignoreBundle?: boolean;
 }
 
+export interface IExperimentationServiceAdapter {
+    isCachedFlightEnabled(flight: string): Promise<boolean>;
+    isLiveFlightEnabled(flight: string): Promise<boolean>;
+    getCachedTreatmentVariable<T extends string | number | boolean>(name: string): T | undefined;
```
EricJizbaMSFT, line 1204: Can you give some quick examples on how these are used, maybe an existing Docker scenario? If Docker isn't using all these methods, I would err on the side of not supporting them until someone actually wants them.

ui/src/registerExperimentationService.ts
EricJizbaMSFT: No preference

ui/index.d.ts
EricJizbaMSFT, line 1183: Idk "register" makes it sound required to me. How about we either update the docs to say "Optional function to register..." or refactor to `createExperimentationService` which just returns the `experimentationService` instead of actually registering it

bwateratmsft: That's a good idea. It also eliminates the need to deal with extension variables.

ui/index.d.ts
bwateratmsft: Cached vs live: Cached will grab the value from the cache, which means it's ~1 session delayed. TAS will "fill" its in-memory cache, then load from the web and update the disk cache, but _not_ the in-memory cache (otherwise, users could see behaviors change mid-session). So, the next session will get the new data. Live, however, goes straight to the web and forces an in-memory cache update, and also takes longer. This is useful if you need to avoid the 1 session delay.

Flights: simple yes/no of whether the flight is enabled.

Treatment variables: Any string, number, or boolean can be passed (you could also implement boolean flighting with this if you wanted to). For example, you may want to have two treatment groups and a control group, each with a different string shown somewhere in the UI, to compare the results.

I'll add the above to comments in the code. We do have plans to use the treatment variables in the near-ish future so I figured it would be best to just implement it now.

ui/src/registerExperimentationService.ts
bwateratmsft: I'll leave them here then, keeps everything together with fewer exports.

ui/index.d.ts
```diff
@@ -1188,6 +1203,35 @@ export interface UIExtensionVariables {
...
+     * Gets whether or not the flight is enabled directly from the web. This is slower than cache and can result in behavior changing mid-session.
+     * @param flight The flight variable name
+     */
+    isLiveFlightEnabled(flight: string): Promise<boolean>;
+
+    /**
+     * Gets a treatment variable from the cache (which will be ~1 session delayed)
+     * @param name The variable name
+     */
+    getCachedTreatmentVariable<T extends string | number | boolean>(name: string): T | undefined;
```
bwateratmsft, line 1226: I would like for this to be `async` like `isCachedFlightEnabled` but I can't get TypeScript to let me. It's saying it can't assign `string` to `string | number | boolean | undefined`.

bwateratmsft: Figured it out!

EricJizbaMSFT (approved):

PR #818, opened 2020-11-18
let storage account widzardContext interface accept old version storage account…

v-tsun:
… type to use for stack storage account

### DIFF
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -5,6 +5,7 @@
 
 import { ResourceManagementModels } from '@azure/arm-resources';
 import { StorageManagementModels } from '@azure/arm-storage';
+import { StorageManagementModels as StackManagementModels } from '@azure/arm-storage-profile-2019-03-01-hybrid'
 import { SubscriptionModels } from '@azure/arm-subscriptions';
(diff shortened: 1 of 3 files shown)


v-tsun:
@rakku-ms @bganapa @sijuman

ui/index.d.ts
```diff
@@ -1068,7 +1069,7 @@ export interface IStorageAccountWizardContext extends IResourceGroupWizardContex
      * If an existing storage account is picked, this value will be defined after `StorageAccountListStep.prompt` occurs
      * If a new storage account is picked, this value will be defined after the `execute` phase of the 'create' subwizard
      */
-    storageAccount?: StorageManagementModels.StorageAccount;
```
EricJizbaMSFT, line 1071: As mentioned [here](https://github.com/microsoft/vscode-azurestorage/pull/882#discussion_r527160916), I don't think we need to actually reference the types for the hybrid npm package

EricJizbaMSFT:
Per my comment, I don't think this PR will be necessary. But if it is - please re-open this or submit a new PR once you're ready to be out of "draft"

PR #819, opened 2020-11-20
Add support for "Report Issue" command

ejizba:
It'll keep track of up to 50 issues per session of VS Code and show users a quick pick like this:
Screen Shot 2020-11-20 at 12 00 40 PM

In general, this just gives users more flexibility for reporting errors. It allows them to do it if we suppressed the "Report an Issue" button or if they just dismissed the notification - all without having to reproduce it again.


ui/src/registerReportIssueCommand.ts
```diff
@@ -0,0 +1,56 @@
...
+        cachedIssues.push(issue);
+
+        // Remove any duplicate issues
+        cachedIssues = cachedIssues.filter(i => {
+            return i === issue || i.callbackId !== issue.callbackId || i.error.message !== issue.error.message;
+        });
+
+        const maxIssues: number = 50;
+        if (cachedIssues.length > maxIssues) {
+            cachedIssues.pop();
```
StephenWeatherford, line 29: Doesn't that remove the most-recently pushed item?  You want shift I think?

StephenWeatherford (approved):

PR #820, opened 2020-11-24
Add `ui` to `IActionContext`

ejizba:
and pass `IActionContext` along to a few more tree methods

My main motivation is so that I can specify one `TestUserInput` per nightly test and run them in parallel. I was able to get App Service's "Create and Deploy" tests from ~11 minutes down to ~2 minutes on my local machine. I also think this will help simplify some logic in `AzureWizard`, but haven't gone down that rabbit hole yet

### DIFF
--- a/ui/src/getPackageInfo.ts
+++ b/ui/src/getPackageInfo.ts
@@ -5,60 +5,54 @@
 
 import * as fse from 'fs-extra';
 import { ExtensionContext } from "vscode";
-import { IActionContext } from "../index";
-import { callWithTelemetryAndErrorHandling } from "./callWithTelemetryAndErrorHandling";
 import { ext } from "./extensionVariables";
-import { parseError } from "./parseError";
 
-export function getPackageInfo(ctx?: ExtensionContext): { extensionName: string, extensionVersion: string, aiKey: string, extensionId: string, bugsUrl: string | undefined } {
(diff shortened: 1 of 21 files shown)


ui/src/getPackageInfo.ts
```diff
@@ -5,60 +5,54 @@
...
+    extensionName: string;
+    extensionVersion: string;
+    aiKey: string;
+    extensionId: string;
+    bugsUrl: string | undefined;
+}
 
-    let packageJson: IPackageJson = {};
-    // tslint:disable-next-line:no-floating-promises
-    callWithTelemetryAndErrorHandling('azureTools.getPackageInfo', (context: IActionContext) => {
```
ejizba, line 20: This was causing timing problems, because of this call stack:
> registerUIExtensionVariables -> createTelemetryReporter -> getPackageInfo -> callWithTelemetryAndErrorHandling

Basically, it tries to get the extension variables, while registering/creating the extension variables. In the past we've been able to work around this, but I decided it just wasn't worth the complexity after I hit more problems. I took a look at telemetry and very few people had errors, most of which were "Too many files open". I feel like that error is much more helpful to users than the error they're seeing now, which is just "Extension's package.json is missing aiKey"

StephenWeatherford (approved):

PR #824, opened 2020-12-02
Best effort to mask user info from telemetry and "report issue"

ejizba:
Best effort to mask things like username, subscription id, and resource name from telemetry and "Report Issue". Almost all of the logic is taken care of inside this package. The only place I expect we will have to add to valuesToMask outside of this package is during wizard steps, but I feel like that's not too bad.

NOTE: I switched to hyphens instead of asterisks because it did weird things in markdown:
### Hyphens
Screen Shot 2020-12-02 at 2 30 00 PM
### Asterisks
Screen Shot 2020-12-02 at 2 29 15 PM


ui/src/masking.ts
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * Example id: /subscriptions/00000000-0000-0000-0000-00000000/resourceGroups/rg1/providers/Microsoft.Web/sites/site1
+ */
+export function addValuesToMaskFromAzureId(context: IActionContext, node: AzExtTreeItem): void {
+    const parts: string[] = node.fullId.toLowerCase().split('/');
+    if (parts[1] === 'subscriptions' && parts[3] === 'resourcegroups') { // NOTE: subscription id is already added to extValuesToMask elsewhere
+        context.valuesToMask.push(parts[4]);
+
+        if (parts[5] === 'providers' && parts[6]?.startsWith('microsoft.') && parts[8]) {
```
nturinski, line 41: Would it be possible to search the array for 'providers' index and then get the value of the index + 1?  I'm pretty sure with resource ids, that they just keep tacking on parts to the end, but not 100% confident that'll always be the case

nturinski (approved):

ui/src/masking.ts
ejizba: What problem are you trying to solve? Can you give an example id maybe?

nturinski: Never mind, I thought that it'd be possible that an id didn't follow this exact formatting.  But according to this: https://docs.microsoft.com/en-us/rest/api/resources/resources/getbyid that's always the expected structure.

PR #825, opened 2020-12-04
Mitigate "Entry not found in cache." error

ejizba:
Screen Shot 2020-12-04 at 11 29 34 AM

### DIFF
--- a/ui/src/extensionVariables.ts
+++ b/ui/src/extensionVariables.ts
@@ -4,27 +4,29 @@
  *--------------------------------------------------------------------------------------------*/
 
 import * as assert from 'assert';
-import { ExtensionContext } from "vscode";
-import { IAzExtOutputChannel, IAzureUserInput, UIExtensionVariables } from "../index";
+import { commands, ExtensionContext } from "vscode";
+import * as types from "../index";
+import { registerErrorHandler } from './callWithTelemetryAndErrorHandling';
 import { createTelemetryReporter, IInternalTelemetryReporter } from './createTelemetryReporter';
 import { localize } from "./localize";
+import { parseError } from './parseError';
 
-interface IInternalExtensionVariables extends UIExtensionVariables {
+interface IInternalExtensionVariables extends types.UIExtensionVariables {
     _internalReporter: IInternalTelemetryReporter;
 }
 
-class UninitializedExtensionVariables implements UIExtensionVariables {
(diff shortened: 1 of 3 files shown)


nturinski (approved):
This is a great idea, but I'm wondering if the message isn't kind of confusing.  I could see people wondering why sometimes this action requires a reload and other times it doesn't.  Could we say something along the lines:
> Something unexpected while performing this action.  Your VS Code window must be reloaded.

Or something to that nature?

ejizba:
I like the clean message of "You gotta reload, bro". I agree the whole situation is confusing, but I'm not sure "something unexpected" really gives any more information. In my experience, this issue only happens if you've left a VS Code window open for a long time, so I feel like that's a pretty good hint on its own. The real way to get rid of the confusion is to get the bug fixed but 🤷‍♂️

bwateratmsft (approved):
I like this implementation for custom actions, it's straightforward and very easy to consume.

ui/src/extensionVariables.ts
```diff
@@ -49,4 +51,25 @@ export function registerUIExtensionVariables(extVars: UIExtensionVariables): voi
...
+
+    registerErrorHandler(handleEntryNotFound);
+}
+
+/**
+ * Long-standing issue that is pretty common for all Azure calls, but can be fixed with a simple reload of VS Code
+ * https://github.com/microsoft/vscode-azure-account/issues/53
+ */
+async function handleEntryNotFound(context: types.IErrorHandlerContext): Promise<void> {
+    if (parseError(context.error).message === 'Entry not found in cache.') {
```
bwateratmsft, line 63: Is the message reliably exactly this, or should a regex or something a little more flexible be used?

ejizba: Unfortunately this is extremely reliable haha. The exact source is [here](https://github.com/AzureAD/azure-activedirectory-library-for-nodejs/blob/dev/lib/token-request.js#L526), there's been an open PR to fix the root cause [here](https://github.com/AzureAD/azure-activedirectory-library-for-nodejs/pull/188) for several years, and per [this comment](https://github.com/AzureAD/azure-activedirectory-library-for-nodejs/issues/199#issuecomment-566266974), I don't think they'll ever update the package again. I think the only way to fix the root cause is to move to the new auth library they mentioned, but that's up to the VS Code peeps in the Azure Account extension

PR #827, opened 2020-12-09
Minor improvements to "Report Issue" command

ejizba:
1. Skip the quick pick if there are no cached issues
1. Include tree errors in the quick pick even though they're not displayed as a notification


nturinski (approved):
So just to verify, but calling `reportAnIssue(undefined)` is the equivalent of selecting `Manually enter error`, right?

ejizba:
Yah

PR #828, opened 2020-12-10
Make duplicate IDs unique if the classes are different

wwlorey:
I chose the approach of adding a temporary ~`uniqueFullId`~ `fullIdWithContext` in the case of an ID collision as suggested in https://github.com/microsoft/vscode-azurestorage/pull/893#discussion_r533823632

Related: https://github.com/microsoft/vscode-azurestorage/issues/829

### DIFF
--- a/ui/src/tree/AzExtTreeDataProvider.ts
+++ b/ui/src/tree/AzExtTreeDataProvider.ts
@@ -40,7 +40,7 @@ export class AzExtTreeDataProvider implements IAzExtTreeDataProviderInternal, ty
         return this._onTreeItemCreateEmitter.event;
     }
 
-    public getTreeItem(treeItem: AzExtTreeItem & { uniqueFullId?: string }): TreeItem {
+    public getTreeItem(treeItem: AzExtTreeItem): TreeItem {
         return {
             label: treeItem.label,
             description: treeItem.effectiveDescription,
@@ -57,7 +57,7 @@ export class AzExtTreeDataProvider implements IAzExtTreeDataProviderInternal, ty
         };
     }
 
-    public async getChildren(arg?: AzExtParentTreeItem): Promise<(AzExtTreeItem & { uniqueFullId?: string })[]> {
+    public async getChildren(arg?: AzExtParentTreeItem): Promise<AzExtTreeItem[]> {
         try {
             return <AzExtTreeItem[]>await callWithTelemetryAndErrorHandling('AzureTreeDataProvider.getChildren', async (context: types.IActionContext) => {
                 context.errorHandling.suppressDisplay = true;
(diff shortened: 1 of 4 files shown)


ui/src/tree/AzExtTreeDataProvider.ts
```diff
@@ -40,11 +40,11 @@ export class AzExtTreeDataProvider implements IAzExtTreeDataProviderInternal, ty
         return this._onTreeItemCreateEmitter.event;
     }
 
-    public getTreeItem(treeItem: AzExtTreeItem): TreeItem {
+    public getTreeItem(treeItem: AzExtTreeItem & { uniqueFullId?: string }): TreeItem {
         return {
             label: treeItem.label,
             description: treeItem.effectiveDescription,
-            id: treeItem.fullId,
+            id: treeItem.uniqueFullId || treeItem.fullId,
```
nturinski, line 47: I'd prefer having `uniqueFullId` just being it's own property on treeItems that need it.  And then using it in the `getChildren` commmands that need it.  

I don't like replacing id with this uniqueFullId even if you have to pass it in.  I think it also makes the `getChildren` command look really awkward and clunky.

nturinski: Actually just ignore my comment.  I reread Eric's comment and this seems like what he suggested 🤔

ejizba: Well, but I do agree the `getChildren` method looks clunky now. Maybe see if you can pull the "duplicate" logic out into a separate method

ejizba: > I'd prefer having uniqueFullId just being it's own property on treeItems that need it. And then using it in the getChildren commands that need it.

Not sure if this is what Nathan is saying, but I'm okay with making `uniqueFullId` a property on `AzExtTreeItem` itself - as long as you only do it in `AzExtTreeItem.ts` and _not_ `index.d.ts` (which would expose the property outside the ui package).

nturinski: That wasn't exactly what I meant, but I saw your comment about not exposing it outside the ui package and I agree.  I think this is what I'd prefer too ^

ui/src/tree/AzExtTreeDataProvider.ts
```diff
@@ -241,6 +226,21 @@ export class AzExtTreeDataProvider implements IAzExtTreeDataProviderInternal, ty
             }
         }
     }
+
+    private isDuplicateChild(child: AzExtTreeItem, existingChildren: AzExtTreeItem[]): boolean {
+        for (const existingChild of existingChildren) {
```
nturinski, line 231: I feel like this is pretty inefficient in terms of having to iterate through an array of children for each single node.  

What if we used an object mapping?  The key value could just be the id, the value could be the node (not really sure if we even need a value if you just need to look it up once to create a unique id)

nturinski: Oh right, I forgot that the existingChildren array is just the duplicates (not the entire array of children).  So this isn't as bad as I thought-- though I think using an object can still make the logic look a little cleaner.

wwlorey: Looping through each child here should be the same time complexity as before which used `result.some` but I agree it could be improved. I implemented a map and also accounted for the case where there are multiple duplicate tree items with the same context value. For example:

```ts
treeItem1 = { fullId: "id", contextValue: "a" }
treeItem2 = { fullId: "id", contextValue: "b" }
treeItem3 = { fullId: "id", contextValue: "b" }
```

Before, two of the same uniqueFullIds would have been added to the result array - one for `treeItem2` and another for `treeItem3`. Now it marks `treeItem3` as a duplicate.

nturinski (approved):

ui/src/tree/AzExtTreeDataProvider.ts
```diff
@@ -78,12 +78,13 @@ export class AzExtTreeDataProvider implements IAzExtTreeDataProviderInternal, ty
...
 
-                const result: AzExtTreeItem[] = [];
+                const resultMap: Map<string, AzExtTreeItem> = new Map();
                 const duplicateChildren: AzExtTreeItem[] = [];
                 for (const child of children) {
-                    result.some(c => c.fullId === child.fullId) ? duplicateChildren.push(child) : result.push(child);
+                    this.isDuplicateChild(child, resultMap) ? duplicateChildren.push(child) : resultMap.set(child.uniqueFullId || child.fullId, child);
                 }
 
+                const result: AzExtTreeItem[] = Array.from(resultMap.values());
```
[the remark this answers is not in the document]
wwlorey: Yes it is - from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map#Objects_vs._Maps

> The keys in Map are ordered in a simple, straightforward way: A Map object iterates entries, keys, and values in the order of entry insertion.

ejizba (approved):

PR #830, opened 2020-12-19
Get environment for azure stack

v-tsun:


ui/src/getEnvironment.ts
```diff
@@ -0,0 +1,77 @@
...
+        const env: IAzureStackEnvironment<Environment> = filter.session.credentials2.environment;
+        env.portalUrl = metadata.portalEndpoint;
+        env.galleryEndpointUrl = metadata.galleryEndpoint;
+        env.activeDirectoryEndpointUrl = metadata.authentication.loginEndpoint.slice(0, metadata.authentication.loginEndpoint.lastIndexOf('/') + 1);
+        env.activeDirectoryResourceId = metadata.authentication.audiences[0];
+        env.activeDirectoryGraphResourceId = metadata.graphEndpoint;
+        env.storageEndpointSuffix = baseUrl.substring(baseUrl.indexOf('.'));
+        env.keyVaultDnsSuffix = '.vault'.concat(baseUrl.substring(baseUrl.indexOf('.')));
+        env.managementEndpointUrl = metadata.authentication.audiences[0];
+        (<IAzureStackSession>filter.session).environment = env;
```
ejizba, line 75: I would prefer if the environment was set in the [account extension](https://github.com/Microsoft/vscode-azure-account) since the "ppe" setting and "AzureResourceFilter" object are both owned by that extension.

v-tsun:
Move changes to azure account and get environments there

PR #835, opened 2021-01-11
Add support for Markdown tooltips

bwateratmsft:


ui/package.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "vscode-azureextensionui",
     "author": "Microsoft Corporation",
-    "version": "0.38.3",
+    "version": "0.38.4",
```
ejizba, line 4: We consider updating the VS Code types to be a "breaking change" so we'd need to bump this to "0.39.0". Also, is it worth requiring 1.52 for all our extensions just for "MarkdownString"? I lean towards no since 1.52 is so recent, but 🤷‍♂️ Even if we say "no", maybe you can work some magic to get `tooltip` to work as a MarkdownString (if the user has 1.52+) without "officially" updating the VS Code types?

bwateratmsft: That makes sense. Since all this library is doing is copying from the `AzExtTreeItem` into `TreeItem`, it's probably possible to do this without changing the VSCode dependency. We do plan to use Markdown though: https://github.com/microsoft/vscode-docker/issues/2592

I'll look into whether or not it's possible to do this with magic instead of a dependency update. :smile:

bwateratmsft: Seems like it worked fine so I rolled that stuff back out.

ui/index.d.ts
```diff
@@ -176,6 +176,7 @@ export declare abstract class AzExtTreeItem {
     public description?: string;
     public iconPath?: TreeItemIconPath;
     public commandId?: string;
+    public tooltip?: string;
```
bwateratmsft, line 179: Problematically this gets called earlier than is ideal. We could use `resolveTreeItem` to solve this issue.

ejizba (approved):

ui/src/tree/AzExtTreeDataProvider.ts
```diff
@@ -53,7 +53,16 @@ export class AzExtTreeDataProvider implements IAzExtTreeDataProviderInternal, ty
...
+            } : undefined,
+            tooltip: treeItem.resolveTooltip ? undefined : treeItem.tooltip // If `resolveTooltip` is defined, return undefined here, so that `resolveTreeItem` and `resolveTooltip` get used
+        };
+    }
+
+    public async resolveTreeItem(ti: TreeItem, treeItem: AzExtTreeItem): Promise<TreeItem> {
+        return {
+            ...ti,
+            // If `resolveTooltip` is undefined we shouldn't be here anyway, but we'll fall back to the item's tooltip if this does happen
+            tooltip: treeItem.resolveTooltip ? await treeItem.resolveTooltip() : treeItem.tooltip
```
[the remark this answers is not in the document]
bwateratmsft: Good to know. I'd be interested to hear more about those bugs as I've used it pretty heavily over in the Docker extension...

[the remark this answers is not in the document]
ejizba: Well of course I can't find any bugs now that you want examples :P

Tbf, I think my suggestion also just looks cleaner in this specific case

PR #836, opened 2021-01-13
No longer need to exclude language client from webpacking

bwateratmsft:
It looks like we no longer need to be excluding the language client from webpacking: https://github.com/microsoft/vscode-docker/issues/2601#issuecomment-759718618

Both @dbaeumer and I tried it out webpacking with this PR in place and did not encounter any obvious issues. It also unblocks us from fixing https://github.com/microsoft/vscode-docker/issues/2601.

I'd like to hear from @StephenWeatherford first on why this exclusion was present originally; I don't know the history here.

### DIFF
--- a/dev/package.json
+++ b/dev/package.json
@@ -1,7 +1,7 @@
 {
     "name": "vscode-azureextensiondev",
     "author": "Microsoft Corporation",
-    "version": "0.5.0",
+    "version": "0.5.1",
     "description": "Common dev dependency tools for developing Azure extensions for VS Code",
     "tags": [
         "azure",
--- a/dev/package.json
+++ b/dev/package.json
@@ -1,7 +1,7 @@
 {
     "name": "vscode-azureextensiondev",
     "author": "Microsoft Corporation",
-    "version": "0.5.1",
+    "version": "0.6.0",
     "description": "Common dev dependency tools for developing Azure extensions for VS Code",
     "tags": [
         "azure",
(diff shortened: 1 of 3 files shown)


dev/package.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "vscode-azureextensiondev",
     "author": "Microsoft Corporation",
-    "version": "0.5.0",
+    "version": "0.5.1",
```
bwateratmsft, line 4: Should this be a major update (i.e. `0.6.0`)?

ejizba (approved):
@neelip you might want to double check mongo support with this change (ping me if you need help), but I'm fine merging if this fixes Docker.

dev/package.json
bwateratmsft: I'll go ahead and rev it then. It costs nothing and might save somebody from headaches in the future.

StephenWeatherford:
If it works there's no reason to continue excluding.  There are various reasons for components having been excluded (usually you can tell by errors during the build, although sometimes things work anyway - frequently this is due to dynamically finding resource or source files in a way webpack can't deal with).  Given that vscode is also doing webpacking, it's quite possible and even likely that they intentionally fixed this component to be webpack-compatible.

Note that the cosmos extension was one of the most complex in that it creates two different output components from webpacking, so if you want to sanity check anything that might be the most logical (in particular mongo scrapbooks).

bwateratmsft:
Let's hope then! Nothing glaringly obvious broke; I tried out several scenarios with the webpacked build.

StephenWeatherford:
Probably fine then.

PR #837, opened 2021-01-15
Add pared down version of VS Code triage actions

ejizba:
I took VS Code's [repo](https://github.com/microsoft/vscode-github-triage-actions), deleted a bunch of code, and added the rest to the tools repo here. The code that is left should be the bare minimum needed to create my new bot (as discussed offline) and run the "locker"/"needs more info" actions. I debated including "locker"/"needs more info" here or just using the VS Code repo directly for those, but decided it'll be easier to have all the actions we use in one place. For example, I already [ran into](https://github.com/microsoft/vscode-azurefunctions/runs/1711469539) one quirk where if the "needs more info" action fails, it tries to comment the error on an issue in a "vscode-internalbacklog" repo. Since we don't have access to that repo, the only logs I see are the failure to access that repo, rather than the actual reason the action failed. When deleting code, I just deleted that functionality because we can just look at the action logs directly.

Since this is all copied code - I would prefer if the review is more about the high-level concept, rather than reviewing each line of code that I didn't author.

### DIFF
--- a/triage-actions/api/api.js
+++ b/triage-actions/api/api.js
@@ -1,0 +1,7 @@
+"use strict";
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See LICENSE in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+Object.defineProperty(exports, "__esModule", { value: true });
+//# sourceMappingURL=api.js.map
--- a/triage-actions/.editorconfig
+++ b/triage-actions/.editorconfig
@@ -1,0 +1,13 @@
+# top-most EditorConfig file
+root = true
+
+# Tab indentation
+[{*.ts}]
+indent_style = tab
+trim_trailing_whitespace = true
+
+# The indent size used in the `package.json` file cannot be changed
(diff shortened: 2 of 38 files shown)


triage-actions/api/api.js
```diff
@@ -0,0 +1,7 @@
+"use strict";
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See LICENSE in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+Object.defineProperty(exports, "__esModule", { value: true });
+//# sourceMappingURL=api.js.map
```
nturinski, line 7: This seems weird.  Did you comment this out or was it copied that way?

nturinski (approved):

triage-actions/api/api.js
ejizba: Yeah the .js files are automatically generated - any time you run a build and I think as a pre-commit hook. You would see this comment on the .js files in our extension, too - we just don't commit them to GitHub normally. In this case, we commit them because the bot runs directly from the repo, as opposed to our extension which gets packaged into a vsix

PR #838, opened 2021-01-20
Fix readonly flag for actions

ejizba:
The readonly flag wasn't working when I tried to do a dry-run. I did a [PR in the main repo](https://github.com/microsoft/vscode-github-triage-actions/pull/19) to fix that and copied the change here.

The rest of the changes aren't as important, but clean it up a little bit:
1. Fix a warning that "readonly" isn't recognized by listing it in a yml file
2. Make it so that setting readonly to "false" will work as expected
3. Fix an inaccurate log to reference "team" instead of "rando"

### DIFF
--- a/triage-actions/needs-more-info-closer/NeedsMoreInfoCloser.ts
+++ b/triage-actions/needs-more-info-closer/NeedsMoreInfoCloser.ts
@@ -45,7 +45,7 @@ export class NeedsMoreInfoCloser {
 						(await issue.hasWriteAccess(lastComment.author))
 					) {
 						if (lastComment) {
-							safeLog(`Last comment on ${hydrated.number} by rando. Closing.`)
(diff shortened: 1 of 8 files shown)


triage-actions/needs-more-info-closer/NeedsMoreInfoCloser.ts
```diff
@@ -45,7 +45,7 @@ export class NeedsMoreInfoCloser {
 						(await issue.hasWriteAccess(lastComment.author))
 					) {
 						if (lastComment) {
-							safeLog(`Last comment on ${hydrated.number} by rando. Closing.`)
+							safeLog(`Last comment on ${hydrated.number} by team. Closing.`)
```
nturinski, line 48: 😂

nturinski (approved):

PR #840, opened 2021-01-27
Bump ms-rest-js to help with ECONNRESET

ejizba:
The azure sdks have a `systemErrorRetryPolicy` that's supposed to automatically retry a request for errors like "ECONNRESET". However, it had a bug where it never retried. I found [a PR](https://github.com/Azure/azure-sdk-for-js/pull/8803) that fixed the bug in the "core-http" package a while ago, but not the "ms-rest-js" package we had been using. I filed a bug on "ms-rest-js" to get the fix ported over, but not sure when that'll happen and it sounds like "core-http" is the newer/better version of "ms-rest-js", anyways.

Only issue was some build errors around `ServiceClientCredentials`, but I mitigated that with `AzExtServiceClientCredentials`.

Related to https://github.com/microsoft/vscode-azurefunctions/issues/2390 and https://github.com/microsoft/vscode-azureappservice/issues/1932

### DIFF
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -7,8 +7,7 @@ import { ResourceManagementModels } from '@azure/arm-resources';
 import { StorageManagementModels } from '@azure/arm-storage';
 import { SubscriptionModels } from '@azure/arm-subscriptions';
 import { Environment } from '@azure/ms-rest-azure-env';
-import { RequestPolicyFactory, ServiceClient, ServiceClientCredentials } from '@azure/ms-rest-js';
-import { TokenCredentialsBase } from '@azure/ms-rest-nodeauth';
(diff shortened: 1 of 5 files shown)


nturinski (approved):
🤯

ejizba:
[ms-rest-js 2.2.1](https://github.com/Azure/ms-rest-js/blob/master/Changelog.md#221---2021-02-05) has the fix we need for ECONNRESET, so we don't need to use core-http anymore (which only helped for genericClient anyways)

I decided to keep the changes for `AzExtServiceClientCredentials` because it lets us remove the explicit dependency on "ms-rest-nodeauth" and I have seen typescript errors related to the credentials before.

PR #841, opened 2021-01-28
Replace falsy tooltip with label

bwateratmsft:
Needed to fix downstream https://github.com/microsoft/vscode-docker/issues/2679. Alternatively, if https://github.com/microsoft/vscode/issues/115337 is fixed, this will not be needed.

Because all `AzExtTreeDataProvider` now have `resolveTreeItem` defined, VSCode always attempts to resolve a tooltip. However, due to [this code](https://github.com/microsoft/vscode/blob/5b75a42575fd85c3a9f69bec76bd1afe7c2dcb33/src/vs/base/browser/ui/iconLabel/iconLabel.ts#L262-L269), if the resolved `tooltip` is falsy, then it will not replace 'Loading...' from a few lines above.

In the past the standard behavior for tooltips was that `label` would be used, if `tooltip` was falsy. So, in this code for resolving `tooltip`, we'll replace it with the `label` if it is falsy.

### DIFF
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -1,6 +1,6 @@
 {
     "name": "vscode-azureextensionui",
-    "version": "0.38.4",
+    "version": "0.38.5",
     "lockfileVersion": 1,
     "requires": true,
     "dependencies": {
--- a/ui/package.json
+++ b/ui/package.json
@@ -1,7 +1,7 @@
 {
     "name": "vscode-azureextensionui",
     "author": "Microsoft Corporation",
-    "version": "0.38.4",
+    "version": "0.38.5",
     "description": "Common UI tools for developing Azure extensions for VS Code",
     "tags": [
         "azure",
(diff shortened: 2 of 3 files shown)


nturinski (approved):

bwateratmsft:
@nturinski @ejizba @StephenWeatherford Do you know if any extensions other than Docker have started using 0.38.4+ of `vscode-azureextensionui`? Any using that version will show 'Loading...' as the tooltip forever unless `resolveTooltip` comes back with something non-falsy, or unless the VSCode team fixes https://github.com/microsoft/vscode/issues/115337. I assume 1.53.0 is due out relatively soon, so if we don't get quick traction from VSCode we should probably put this fix in place and update users from 0.38.4 to 0.38.5.

StephenWeatherford:
ARM is on an earlier version.

bwateratmsft:
If we don't hear back on that VSCode bug by say, mid next week, I'll go ahead and put this PR through. What do you think, @nturinski / @ejizba / @StephenWeatherford?

bwateratmsft:
Per @fiveisprime this will probably not be in the nearest upcoming VSCode release, so I'll go ahead and put this fix in.

When it is fixed in VSCode, I'd like to _revert_ the change here, because what we'll be doing is essentially duplicating the logic to use label when tooltip is falsy, but I'd prefer to leave the decision of what to show with VSCode in that scenario. I'll create an issue to revert it later.

PR #843, opened 2021-02-01
Revert falsy tooltip change

### LINKED ISSUE #842: Revert falsy tooltip change (#841)
Once the upstream [bug in VSCode](https://github.com/microsoft/vscode/issues/115337) is fixed, we should revert the change in #841, so we can return the decision of how to handle a falsy label to VSCode. (Currently VSCode uses the label, and we now have logic duplicating that behavior; but ideally we should remove that logic when possible)

bwateratmsft:
Reverts #841, fixes #842. The upstream VSCode [issue](https://github.com/microsoft/vscode/issues/115337) has been fixed so we should go back to our old behavior.

### DIFF
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -1,6 +1,6 @@
 {
     "name": "vscode-azureextensionui",
-    "version": "0.38.5",
+    "version": "0.38.6",
     "lockfileVersion": 1,
     "requires": true,
     "dependencies": {
--- a/ui/package.json
+++ b/ui/package.json
@@ -1,7 +1,7 @@
 {
     "name": "vscode-azureextensionui",
     "author": "Microsoft Corporation",
-    "version": "0.38.5",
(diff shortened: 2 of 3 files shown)


bwateratmsft:
Before submitting this I'd like to verify the fix in tomorrow's Insiders build.

ejizba:
Can you comment on the timing of this? When was the bug introduced in VS Code and when was it fixed? Are we at risk of people on older versions of VS Code still seeing the bug? If so, it doesn't seem that bad to leave the workaround in

bwateratmsft:
@ejizba the bug was introduced in Insiders 1.53 late last year and just fixed yesterday; it hasn't appeared in any GA builds of VSCode yet. So, removing the fix (i.e. submitting this PR) should be fine, since it would only affect a few folks on Insiders until they upgraded.

ejizba (approved):

PR #844, opened 2021-02-02
Add isCustomCloud to ISubscriptionContext

v-tsun:
Related with PR https://github.com/microsoft/vscode-azurestorage/pull/905


v-tsun:
@nturinski

ui/index.d.ts
```diff
@@ -1113,6 +1113,8 @@ export interface IStorageAccountWizardContext extends IResourceGroupWizardContex
     storageAccount?: StorageManagementModels.StorageAccount;
 
     newStorageAccountName?: string;
+
+    isStack?: boolean;
```
nturinski, line 1117: Could you add a comment describing what this will be used for?

nturinski (commented):
Looks good.  Could you update the ui package.json and bump the minor version?  And then run `npm install` to update the `package.lock.json`

ui/index.d.ts
```diff
@@ -1113,6 +1113,12 @@ export interface IStorageAccountWizardContext extends IResourceGroupWizardContex
     storageAccount?: StorageManagementModels.StorageAccount;
```
ejizba, line 1114: A bit weird to me that this is on a storage-specific context, since other resource types would use it, too.

What if we add it to `ISubscriptionContext` and set it [here](https://github.com/microsoft/vscode-azuretools/blob/main/ui/src/tree/AzureAccountTreeItemBase.ts#L124) like this:

```
isStack: filter.session.environment.name === 'AzurePPE'
```
I just feel like it's more accurate for us to let the Azure Account extension parse settings like "target_azurestack_api_profile" and we just use whatever environment they give us. With my suggest, we'd get rid of [this code](https://github.com/microsoft/vscode-azurestorage/pull/905/files#diff-f66a4412131d5a84d7f193ac464b2b219b1f63199ea192eefb1637ff9f1ea8a6R7). I'm also assuming that "AzurePPE" is a constant that's not supposed to change - regardless I feel like there should be some property on `environment` that lets us determine if it's stack.

ui/index.d.ts
v-tsun: I add some comments, could you please take a look if they are good to you

ui/index.d.ts
v-tsun: @nturinski

bganapa: @ejizba The setting that determines whether the environment is stack is  target_azurestack_api_profile.  This is equivalent to isStack.

As you know AzurePPE could point to any ARM endpoint including the Azure dogfood environments or any specific regions like https://westus.management.azure.com/metadata/endpoints?api-version=1.0

I believe what we have here is fine.. Do you think some renaming will help here?

ejizba: @bganapa I'm pretty confused by your comment. What is your specific concern with my suggestion? And no, I don't see how renaming will be helpful.

bganapa: @ejizba the environment object is from ms-rest-azure and I dont think we want to change that package. It may be misleading as well. The setting will not be used by any SDK otherthan here in the tools.  
AzurePPE is a constant and it is not changing. It is controlled by the Account Extension. 

I am saying that target_azurestack_api_profile is equivalent to IsAzureStack..

ejizba: Feels like we're on different wavelengths here. My suggested change looks like this in context: 
https://github.com/microsoft/vscode-azuretools/commit/0fafd9085d54ecf2f79bf195eb51eedb8a8fd377. It's contained completely within this repo, _not_ the "ms-rest-azure" package. Any concerns with that specific commit?

I can comment more on _why_ I prefer that change, but I would like to make sure we're on the same page about _what_ is changing first.

v-tsun: Hey @ejizba , Bala was talking if we need have property in environment to let us determine if it is stack as you said in your comment. In order to have property in environment then we need to change "ms-rest-env" since the type of environment is defined in that package.  I think for this https://github.com/microsoft/vscode-azuretools/commit/0fafd9085d54ecf2f79bf195eb51eedb8a8fd377 specific context. Bala's concern is sometimes AzurePPE might be used to set property for Azure instead of Azure Stack.  So that is why we change to get it from user setting which we think it would be a better way. Thx

ejizba: The end of my comment was not about adding a specific `isStack` property - my point was there should be an _existing_ property we can hopefully use if `environment.name === 'AzurePPE'` doesn't work (which it sounds like it won't). I don't know what that property is. You probably don't have to change "ms-rest-env", but you might need to change the Azure Account extension.

I don't think the setting is an accurate way to determine if the user is logged in to Azure Stack. For example, even if I set "target_azurestack_api_profile" to "true", I only get Azure Stack if I choose "Azure PPE" when I run "Azure: Sign in to Azure Cloud", right?:
Screen Shot 2021-02-03 at 5 11 29 PM.

My larger point is that the Azure Account extension should handle reading and parsing their own settings (including "target_azurestack_api_profile"). They should handle letting the user choose which environment to use (stack vs. not stack like in my screenshot), they should handle refreshing if the settings change, etc. and I don't want us to duplicate that work. We should be able to know if it's Azure Stack based on the information they give us on the `filter` object from the official extension API

bganapa: We are working with Azure Accoutns team to include that. Thanks eric

v-tsun:
hey @ejizba, our changes in vscode account is approved by Rachel. Could you please take a look on this PR again? Thx

ejizba (approved):

ejizba:
If you update the version in package.json & package-lock.json from "0.39.4" to "0.39.5", we can release it for you

bganapa (approved):

v-tsun:
> If you update the version in package.json & package-lock.json from "0.39.4" to "0.39.5", we can release it for you

Hey Eric, merge with main repo and update the version. Please check it out.

PR #845, opened 2021-02-05
Better handling of json in requests

ejizba:
1. Automatically remove BOM characters if they exist (related to https://github.com/microsoft/vscode-azurefunctions/issues/2490)
2. Prevent the azure sdk from assuming a response with no content-type is JSON. We will attempt to parse it and only assume it's json if the parsing succeeds (related to https://github.com/microsoft/vscode-azurefunctions/issues/2464)

### DIFF
--- a/ui/src/createAzureClient.ts
+++ b/ui/src/createAzureClient.ts
@@ -4,10 +4,11 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { Environment } from '@azure/ms-rest-azure-env';
-import { ServiceClient, ServiceClientCredentials } from '@azure/ms-rest-js';
+import { BaseRequestPolicy, HttpOperationResponse, RequestPolicy, RequestPolicyFactory, RequestPolicyOptions, ServiceClient, ServiceClientCredentials, WebResourceLike } from '@azure/ms-rest-js';
 import * as vscode from "vscode";
 import * as types from '../index';
 import { appendExtensionUserAgent } from "./extensionUserAgent";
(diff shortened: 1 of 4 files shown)


ui/src/createAzureClient.ts
```diff
@@ -43,6 +46,63 @@ export async function createGenericClient(clientInfo?: ServiceClientCredentials
...
+    }
+
+    public async sendRequest(request: WebResourceLike): Promise<HttpOperationResponse> {
+        const response: HttpOperationResponse = await this._nextPolicy.sendRequest(request);
+        if (!response.headers.get(contentTypeName) && response.bodyAsText) {
+            try {
+                parseJson(response.bodyAsText);
+                response.headers.set(contentTypeName, 'application/json');
+            } catch {
+                response.headers.set(contentTypeName, 'application/octet-stream');
```
nturinski, line 103: Is it worth checking to see if there's any binary characters before setting this as the content type rather than just using `text/plain?`  How does the sdk handle octet-stream responses?

nturinski (approved):

ui/src/createAzureClient.ts
ejizba: These are why I chose octet-stream:

https://stackoverflow.com/questions/12539058/is-there-a-default-mime-type/12560996
https://stackoverflow.com/questions/1176022/unknown-file-type-mime

As far as I can tell, the sdk handles text/plain and application/octet-stream the same.

PR #852, opened 2021-02-17
Add gulp_installResourceGroups

ejizba:
Not technically necessary, but nice to have since all our extensions will use it


wwlorey (approved):

dev/index.d.ts
```diff
@@ -88,6 +88,11 @@ export declare function getDefaultWebpackConfig(options: DefaultWebpackOptions):
  */
 export declare function gulp_installAzureAccount(): Promise<void>;
 
+/**
+ * "Installs" a fake version of the resource groups extension before running tests. The extension isn't actually used for tests, but our extension would fail to activate without this
+ */
+export declare function gulp_installResourceGroups(): Promise<void>;
```
nturinski, line 94: nit: Aren't we referring to it as Azure Resources?

ejizba: For user-visible things, yeah I think we should call it "Azure Resources". For everything else, I think it's basically a free for all. There's too many things that will stay "azureresourcegroups" forever (extension id, github repo, etc.) so it's always going to be a bit confusing and not worth making consistent

PR #856, opened 2021-02-26
Add simplified sku prompt for basic create

ejizba:
Screen Shot 2021-02-26 at 2 31 29 PM

### DIFF
--- a/appservice/src/createAppService/AppServicePlanSkuStep.ts
+++ b/appservice/src/createAppService/AppServicePlanSkuStep.ts
@@ -12,9 +12,11 @@ import { AppKind, WebsiteOS } from './AppKind';
 import { IAppServiceWizardContext } from './IAppServiceWizardContext';
 import { setLocationsTask } from './setLocationsTask';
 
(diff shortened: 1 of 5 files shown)


nturinski (approved):

appservice/src/createAppService/AppServicePlanSkuStep.ts
```diff
@@ -12,9 +12,11 @@ import { AppKind, WebsiteOS } from './AppKind';
 import { IAppServiceWizardContext } from './IAppServiceWizardContext';
 import { setLocationsTask } from './setLocationsTask';
 
+type ExtendedSkuDescription = WebSiteManagementModels.SkuDescription & { label?: string; description?: string }
```
nturinski, line 15: Why make the label/descriptions optional though?  I feel like we should enforce them if you're making an ExtendedSkuDescription.

ejizba: Well I'm using the same type for both the basic and advanced scenarios. In this case "Extended" means, "hey you can add better labels if you want to", which only applies for basic create

PR #857, opened 2021-03-03
Nat/zip

nturinski:
Potential fixes for: https://github.com/microsoft/vscode-azurefunctions/issues/2108; https://github.com/microsoft/vscode-azureappservice/issues/1571

Rather than using archiver which doesn't seem to work well on WSL/Codespaces, we switched to yazl which is the same archiver that VS Code is using.

One issue that I'm having is I can't seem to throw an error when the zip fails.  I was forcing an error by passing in an incorrect file path.  I'd see the error in the debugger if I was checking for uncaught exceptions, but I haven't been able to catch it in the code.  Any ideas?


appservice/src/utils/zip.ts
```diff
@@ -0,0 +1,34 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+import * as path from 'path';
+import { Readable } from 'stream';
+import * as yazl from 'yazl';
+
+export async function zip(deployPath: string, files: string[]): Promise<Readable> {
+    const zipFile = new yazl.ZipFile();
+    const buffer: string[] = [];
+    zipFile.outputStream.on('data', (data) => {
```
ejizba, line 13: It honestly looks like you can return `zipFile.outputStream` (If you change the return type to `NodeJS.ReadableStream`) and you have to do `.on` at all. It's already a stream.

appservice/src/deploy/runWithZipStream.ts
```diff
@@ -33,42 +35,29 @@ export async function runWithZipStream(context: IActionContext, fsPath: string,
...
-                await addFilesGitignore(zipper, fsPath, '.funcignore');
+                filesToZip = await getFilesFromGitignore(fsPath, '.funcignore');
             } else {
-                addFilesGlob(zipper, fsPath);
+                filesToZip = await getFilesFromGlob(fsPath);
             }
         } else {
-            zipper.file(fsPath, { name: path.basename(fsPath) });
+            // remove the file from the fsPath
+            fsPath = fsPath.substring(0, fsPath.lastIndexOf(path.sep));
```
ejizba, line 53: What are you trying to do here? There has to be a `path` method that will do it for you

appservice/src/deploy/runWithZipStream.ts
```diff
@@ -80,17 +69,23 @@ const commonGlobSettings: {} = {
 /**
  * Adds files using glob filtering
  */
-function addFilesGlob(zipper: archiver.Archiver, folderPath: string): void {
+export async function getFilesFromGlob(folderPath: string): Promise<string[]> {
     const zipDeployConfig: vscode.WorkspaceConfiguration = vscode.workspace.getConfiguration(ext.prefix, vscode.Uri.file(folderPath));
     const globPattern: string = zipDeployConfig.get<string>('zipGlobPattern') || '**/*';
-    const ignorePattern: string | string[] | undefined = zipDeployConfig.get<string | string[]>('zipIgnorePattern');
-    zipper.glob(globPattern, { cwd: folderPath, ignore: ignorePattern, ...commonGlobSettings });
+    const filesToInclude: string[] = await globby(globPattern, { cwd: folderPath });
```
ejizba, line 75: Did you break symlink support for App Service? I tested for Functions and it was still working, but that code is still referencing `commonGlobSettings` and this is not

appservice/src/deploy/runWithZipStream.ts
```diff
@@ -80,17 +69,23 @@ const commonGlobSettings: {} = {
...
+    return filesToInclude.filter(file => {
+        return !filesToIgnore.includes(file);
+    })
 }
 
 /**
  * Adds files using gitignore filtering
  */
-async function addFilesGitignore(zipper: archiver.Archiver, folderPath: string, gitignoreName: string): Promise<void> {
+async function getFilesFromGitignore(folderPath: string, gitignoreName: string): Promise<string[]> {
```
ejizba, line 88: Looks like `globby` has a `gitignore` method, too. Might be worth switching over to that instead of `glob-gitignore`, but I'm not sure if that should be in this PR.

appservice/src/utils/zip.ts
ejizba: Also once you get rid of the `.on` stuff, this is a pretty simple function and I guess I'd rather have the code in `runWithZipStream` instead of a utils file

nturinski: `NodeJS.ReadableStream` wasn't actually compatible with `Stream` which is the type that we are using so I didn't want to carelessly change the types.

I originally did just pass `zip.outputStream` but I don't think it was working (though that may have been related to a different issue).

appservice/src/deploy/runWithZipStream.ts
nturinski: I have to remove the file from the path name to pass into the zip function since it uses the deploy path and the file name as parameters.  I'll see if there's a path function that I can use, but this seemed like a simple enough action

appservice/src/utils/zip.ts
nturinski: It breaks this specifically https://github.com/microsoft/vscode-azuretools/blob/main/appservice/src/deploy/deployToStorageAccount.ts#L82

ejizba: Okay seems like we want this

```
return new Readable().wrap(zipFile.outputStream);
```


https://nodejs.org/api/stream.html#stream_readable_wrap_stream

appservice/src/deploy/runWithZipStream.ts
nturinski: Good catch.  By default, sym links are on, but including dots are not so I enabled it in the options.

image

appservice/src/deploy/runWithZipStream.ts
```diff
@@ -33,42 +36,33 @@ export async function runWithZipStream(context: IActionContext, fsPath: string,
...
             if (client.isFunctionApp) {
-                await addFilesGitignore(zipper, fsPath, '.funcignore');
+                filesToZip = await getFilesFromGitignore(fsPath, '.funcignore');
             } else {
-                addFilesGlob(zipper, fsPath);
+                filesToZip = await getFilesFromGlob(fsPath);
             }
         } else {
-            zipper.file(fsPath, { name: path.basename(fsPath) });
+            filesToZip.push(path.basename(fsPath));
```
ejizba, line 52: This else case still bugs me. How about something like this:
```typescript
const zipFile = new yazl.ZipFile();
if ((await fse.lstat(fsPath)).isDirectory()) {
    if (!fsPath.endsWith(path.sep)) {
        fsPath += path.sep;
    }

    if (client.isFunctionApp) {
        filesToZip = await getFilesFromGitignore(fsPath, '.funcignore');
    } else {
        filesToZip = await getFilesFromGlob(fsPath);
    }

    for (const file of filesToZip) {
        zipFile.addFile(path.join(fsPath, file), file);
    }
} else {
    zipFile.addFile(fsPath, path.basename(fsPath));
}
```

appservice/src/deploy/runWithZipStream.ts
```diff
@@ -33,42 +36,33 @@ export async function runWithZipStream(context: IActionContext, fsPath: string,
...
+            // remove the file from the fsPath
+            fsPath = path.dirname(fsPath);
+        }
+        const zipFile = new yazl.ZipFile();
+        for (const file of filesToZip) {
+            zipFile.addFile(path.join(fsPath, file), file);
         }
-        zipStream = zipper;
-        zipper.on('end', () => {
-            onFileSize(zipper.pointer());
```
ejizba, line 54: We lost this onFileSize output. If you can find a quick way to get it back that would be nice, but I would not spend much time on it

appservice/src/deploy/runWithZipStream.ts
ejizba: It looks like you added it to the `filesToIgnore` but not the `filesToInclude`

appservice/src/deploy/runWithZipStream.ts
ejizba: What were your thoughts on this? Part of me wants to just move over to be consistent while we're changing things up, but I also understand keeping changes minimal at this point has value, too

nturinski: I'd rather not mess with it right now.  Since we'd theoretically be releasing this with Functions and this purely affects Functions, I'd rather wait until after release since this is more an engineering thing (assuming functionality is the same)

appservice/package-lock.json
```diff
@@ -1,5396 +1,8 @@
 {
     "name": "vscode-azureappservice",
-    "version": "0.73.1",
-    "lockfileVersion": 2,
+    "version": "0.74.0",
+    "lockfileVersion": 1,
```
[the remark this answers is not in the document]
nturinski: Poop, I didn't realize, but my path variable was set up incorrectly.  This should be fixed now.

ejizba (approved):

PR #858, opened 2021-03-04
Use outputStreamCursor to determine size of file

nturinski:
I don't know if this is really worth it, but I think we can use the outputStreamCursor to determine the size of the zip after compression.

https://github.com/thejoshwolfe/yazl/blob/36b2a6a31a32637dca19b34e7b3295378e78a644/index.js#L142

You can see it gets incremented based on the buffer size as it is compressing.  The property isn't technically defined in the types, but I'm pretty confident that it should always exist after a successful zipping.


appservice/src/deploy/runWithZipStream.ts
```diff
@@ -55,6 +55,12 @@ export async function runWithZipStream(context: IActionContext, fsPath: string,
             zipFile.addFile(fsPath, path.basename(fsPath));
         }
 
+        zipFile.outputStream.on('finish', () => {
```
wwlorey, line 58: It looks like you could use `on.('data', ...` instead to manually add up all the chunks like shown [here](https://stackoverflow.com/a/48868017/7044639) for a similar package. I think that would be more comfy than relying on `outputStreamCursor` to be there

appservice/src/deploy/runWithZipStream.ts
```diff
@@ -34,8 +34,15 @@ export async function runWithZipStream(context: IActionContext, fsPath: string,
         });
     } else {
         ext.outputChannel.appendLog(localize('zipCreate', 'Creating zip package...'), { resourceName: client.fullName });
-        const zipFile = new yazl.ZipFile();
+        const zipFile: yazl.ZipFile & { outputStreamCursor?: number } = new yazl.ZipFile();
         let filesToZip: string[] = [];
+        let sizeOfZipFile: number = 0;
+
+        zipFile.outputStream.on('data', (chunk) => {
+            sizeOfZipFile += chunk.length
```
nturinski, line 42: With my test folder, I was seeing it zipped by 8.19 MB vs. 7.82 MB archived by winrar at normal with zip compression.

appservice/src/deploy/runWithZipStream.ts
nturinski: This ends up being slightly inaccurate due to measuring the buffer size.  While close, it's not the exact same thing as the content size.  Still investigating how measure that accurately.

nturinski: Okay, I don't think it's worth pursuing this any further.  I was able to find that when compressed, the `zipFile` object has `entries: Entry[]`.  This object type is not exposed by the public API (intentionally), but has a property called `compressedSize`.  Adding all the compressedSizes, the zipFile size difference between the two methods is like 0.02 MB.  (8.19 MB vs. 8.17 MB)

I'm not sure how Kudu is calculating their size since it reports about ~7.9 MB (it'll tell you it's deleting the old zipfile when you deploy).  But I think I've wasted enough time trying to get the numbers to match especially since they're close enough and it's probably all an estimate anyway.

appservice/src/deploy/runWithZipStream.ts
```diff
@@ -34,8 +34,17 @@ export async function runWithZipStream(context: IActionContext, fsPath: string,
         });
     } else {
         ext.outputChannel.appendLog(localize('zipCreate', 'Creating zip package...'), { resourceName: client.fullName });
-        const zipFile = new yazl.ZipFile();
+        const zipFile: yazl.ZipFile = new yazl.ZipFile();
         let filesToZip: string[] = [];
+        let sizeOfZipFile: number = 0;
+
+        zipFile.outputStream.on('data', (chunk) => {
+            if (Array.isArray(chunk)) {
```
wwlorey, line 42: It looks like it's possible for chunk to be a string or buffer as well: https://nodejs.org/api/stream.html#stream_event_data

wwlorey (approved):

appservice/src/deploy/runWithZipStream.ts
nturinski: Hmm, my understanding was Buffers are a TypedArray which are in term a type of array.  Might be safer to just do a check on length, but the issue is chunk is `any` so this how I was type-checking it.

wwlorey (approved):

appservice/src/deploy/runWithZipStream.ts
ejizba: I don't like using `unknown[]` because I don't think that's technically accurate and I do like `if (Array.isArray(chunk)) {`. What was wrong with that? Can we change back?

nturinski: I can change it back  I just did this because in the yazl types, the return type of chunk is actually `any[]` so I don't think `unknown[]` is technically inaccurate.
```
class EventEmitter {
        addListener(event: string | symbol, listener: (...args: any[]) => void): this;
        on(event: string | symbol, listener: (...args: any[]) => void): this;
```

ejizba: `(...args: any[])` is different than `(args: any[])`

The former says all the args combined are an array, the latter says just the first arg is an array

nturinski: Oh, doh you are totally right.

PR #859, opened 2021-03-17
set location if passed in default location name isn't available in location list

v-tsun:


v-tsun:
Hey @ejizba, this is the PR to fix set location issue in StorageAccount extension.

ui/src/wizard/LocationListStep.ts
```diff
@@ -40,11 +40,17 @@ export class LocationListStep<T extends ILocationWizardContextInternal> extends
     }
 
     public static async setLocation<T extends ILocationWizardContextInternal>(wizardContext: T, name: string): Promise<void> {
+        /**
+         * Get location from locations list if passed in default location name is availble
+         * Otherwise get the first location from location list.
+         */
         const locations: SubscriptionModels.Location[] = await LocationListStep.getLocations(wizardContext);
         name = generalizeLocationName(name);
```
bganapa, line 48: Do we want to do this only for CustomCloud or even for azure this approach is fine?

v-tsun: As far as I am concerned, passed in default location name would be available in Azure. Therefore, I don't think we will get into that if condition in Azure.

v-tsun:
Close this PR according to Eric's comment https://github.com/microsoft/vscode-azurestorage/pull/905#discussion_r596354356

PR #860, opened 2021-03-17
Fix problem where `parseError` fails with unexpected stack type

ejizba:
I'm seeing a decent amount of users getting this error in Functions telemetry:
```
(e.stack || "").split is not a function
```

Turns out our `parseError` method will fail if `stack` is not a string. I have no idea why it _wouldn't_ be a string, but I think it's fair to assume it's not the kind of stack we want in that case. This problem means we don't even get the root cause error tracked in telemetry which is pretty bad.

### DIFF
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "vscode-azureextensionui",
-    "version": "0.39.5",
+    "version": "0.39.6",
     "lockfileVersion": 2,
     "requires": true,
     "packages": {
         "": {
             "name": "vscode-azureextensionui",
-            "version": "0.39.4",
+            "version": "0.39.6",
             "license": "MIT",
             "dependencies": {
                 "@azure/arm-resources": "^3.0.0",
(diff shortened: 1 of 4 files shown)


nturinski (approved):
Would it be worth tracking  the type of/keys in a stack if it happens to be weird (like an object?)  Maybe we can use that to figure out how to parse the stack if it ends up returning an object.

ejizba:
If it's not a standard Node.js stack, it's not worth tracking. It's probably not a callstack at all.

bwateratmsft:
@ejizba did these errors start coming from a particular extension (i.e. was it Docker)? I've started to see them in the Docker extension due to microsoft/vscode-docker#2833, microsoft/vscode-docker#2843, microsoft/vscode-docker#2839, which appear to all actually be due to microsoft/vscode/issues/120558.

The stack in this error is an array with one value: `stack: ['Error: Decryption failed']`.

ejizba:
I saw this in telemetry for the Functions extension and I don't know what the root cause error was because I could not repro it. We do not use the Secrets API, though, so it wasn't that.

PR #862, opened 2021-03-22
Update vscode-tas-client version

bwateratmsft:
This PR updates the vscode-tas-client version. It is not a breaking change.

### DIFF
--- a/ui/src/createExperimentationService.ts
+++ b/ui/src/createExperimentationService.ts
@@ -45,20 +45,26 @@ export async function createExperimentationService(ctx: vscode.ExtensionContext,
 class ExperimentationServiceAdapter implements IExperimentationServiceAdapter {
     public wrappedExperimentationService?: tas.IExperimentationService;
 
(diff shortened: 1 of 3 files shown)


ui/src/createExperimentationService.ts
```diff
@@ -45,20 +45,26 @@ export async function createExperimentationService(ctx: vscode.ExtensionContext,
 class ExperimentationServiceAdapter implements IExperimentationServiceAdapter {
     public wrappedExperimentationService?: tas.IExperimentationService;
 
+    /**
+     * @deprecated Use `getCachedTreatmentVariable<boolean>('flight-name') instead
```
bwateratmsft, line 49: We don't necessarily have to deprecate these, but the `vscode-tas-client` library did deprecate their versions of these methods. I thought it best to match but I don't feel strongly, so please chime in if you feel differently!

ejizba (approved):

PR #866, opened 2021-03-25
Generic requests should throw an error if status code is not 2xx

ejizba:
I'm seeing a decent number of errors like "Cannot convert undefined or null to object" in telemetry. A lot of them are because the generic requests are not throwing an error when the status code says it should. Instead, it just returns the response, which doesn't have a result with the expected type and we get a bunch of weird errors like the one I mentioned.

Also added a bunch of unit tests. We've had plenty of problems with these requests haha so it's probably past-due

### DIFF
--- a/ui/src/createAzureClient.ts
+++ b/ui/src/createAzureClient.ts
@@ -4,11 +4,13 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { Environment } from '@azure/ms-rest-azure-env';
-import { BaseRequestPolicy, HttpOperationResponse, RequestPolicy, RequestPolicyFactory, RequestPolicyOptions, RequestPrepareOptions, ServiceClient, WebResource, WebResourceLike } from '@azure/ms-rest-js';
+import { BaseRequestPolicy, HttpOperationResponse, RequestPolicy, RequestPolicyFactory, RequestPolicyOptions, RequestPrepareOptions, RestError, ServiceClient, WebResource, WebResourceLike } from '@azure/ms-rest-js';
 import * as vscode from "vscode";
 import * as types from '../index';
 import { appendExtensionUserAgent } from "./extensionUserAgent";
 import { GenericServiceClient } from './GenericServiceClient';
+import { localize } from './localize';
+import { parseError } from './parseError';
 import { parseJson, removeBom } from './utils/parseJson';
 
 export function createAzureClient<T>(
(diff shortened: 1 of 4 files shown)


ui/src/createAzureClient.ts
```diff
@@ -120,3 +130,24 @@ class MissingContentTypePolicy extends BaseRequestPolicy {
...
+ * The Azure SDK will only throw errors for bad status codes if it has an "operationSpec", but none of our "generic" requests will have that
+ */
+class StatusCodePolicy extends BaseRequestPolicy {
+    constructor(nextPolicy: RequestPolicy, requestPolicyOptions: RequestPolicyOptions) {
+        super(nextPolicy, requestPolicyOptions);
+    }
+
+    public async sendRequest(request: WebResourceLike): Promise<HttpOperationResponse> {
+        const response: HttpOperationResponse = await this._nextPolicy.sendRequest(request);
+        if (response.status < 200 || response.status >= 300) {
```
nturinski, line 144: Should we be considering 300 codes as errors?  I thought most 300 codes are all still valid (just redirects) https://developer.att.com/video-optimizer/docs/best-practices/http-300-status-codes

I could see our extension not being able to handle the redirecting part of the response right now though.

ui/test/request.test.ts
```diff
@@ -0,0 +1,113 @@
...
+    test('200, json body, no content type, with bom', async () => {
+        const response = await sendTestRequest({ status: 200, body: `\ufeff{ "data": "Hello World!" }` });
+        assert.strictEqual(response.parsedBody.data, 'Hello World!');
+    });
+
+    test('200, json body, json content type, with bom', async () => {
+        const response = await sendTestRequest({ status: 200, body: `\ufeff{ "data": "Hello World!" }`, contentType: 'application/json' });
+        assert.strictEqual(response.parsedBody.data, 'Hello World!');
+    });
+
```
nturinski, line 97: No 300 code tests?

nturinski (approved):

ui/src/createAzureClient.ts
ejizba: I mainly did this because that's what ms-rest-js does:
https://github.com/Azure/ms-rest-js/blob/master/lib/policies/deserializationPolicy.ts#L140

I think we could debate what "valid" means, but 200-299 are the only ones listed as "Successful":
https://developer.mozilla.org/docs/Web/HTTP/Status

PR #867, opened 2021-03-26
Change from opn (deprecated) to open

bwateratmsft:
I've been doing some work to eliminate warnings/errors/etc. throughout our build process; one of them is that `opn` is deprecated. We can switch to `open` which is not.

~~I don't think it's needed anymore since VSCode offers this `vscode.env.openExternal` API, so this lets us eliminate opn as a dependency entirely. That API probably works better in remote cases anyway.~~

~~I tested out this change with the "Report Issue" button which seemed to be a good test sample--long URL with lots of query string--and everything worked nicely.~~

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -60,7 +60,7 @@
         "tslint": "^5.20.1",
         "tslint-microsoft-contrib": "5.0.1",
         "typescript": "^3.8.3",
-        "vscode-azureextensiondev": "^0.8.0",
+        "vscode-azureextensiondev": "^0.8.1",
         "vscode-test": "^1.3.0"
     }
 }
--- a/ui/package.json
+++ b/ui/package.json
@@ -1,7 +1,7 @@
 {
     "name": "vscode-azureextensionui",
     "author": "Microsoft Corporation",
-    "version": "0.40.0",
+    "version": "0.40.1",
     "description": "Common UI tools for developing Azure extensions for VS Code",
     "tags": [
         "azure",
--- a/ui/package.json
+++ b/ui/package.json
@@ -1,7 +1,7 @@
 {
     "name": "vscode-azureextensionui",
     "author": "Microsoft Corporation",
-    "version": "0.40.1",
+    "version": "0.41.0",
     "description": "Common UI tools for developing Azure extensions for VS Code",
     "tags": [
         "azure",
--- a/ui/package.json
+++ b/ui/package.json
@@ -40,6 +40,7 @@
         "escape-string-regexp": "^2.0.0",
         "fs-extra": "^8.0.0",
         "html-to-text": "^5.1.1",
+        "open": "^8.0.4",
         "semver": "^5.7.1",
         "vscode-extension-telemetry": "^0.1.5",
         "vscode-nls": "^4.1.1",
--- a/ui/package.json
+++ b/ui/package.json
@@ -40,7 +40,6 @@
         "escape-string-regexp": "^2.0.0",
         "fs-extra": "^8.0.0",
         "html-to-text": "^5.1.1",
-        "opn": "^6.0.0",
         "semver": "^5.7.1",
         "vscode-extension-telemetry": "^0.1.5",
         "vscode-nls": "^4.1.1",
--- a/ui/src/utils/openUrl.ts
+++ b/ui/src/utils/openUrl.ts
@@ -3,13 +3,8 @@
  *  Licensed under the MIT License. See License.md in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-// tslint:disable-next-line:no-require-imports
-import opn = require("opn");
+import * as vscode from 'vscode';
 
 export async function openUrl(url: string): Promise<void> {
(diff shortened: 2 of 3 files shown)


ui/src/utils/openUrl.ts
```diff
@@ -3,13 +3,8 @@
...
-import opn = require("opn");
+import * as vscode from 'vscode';
 
 export async function openUrl(url: string): Promise<void> {
-    // Using this functionality is blocked by https://github.com/Microsoft/vscode/issues/25852:
-    // await vscode.commands.executeCommand('vscode.open', vscode.Uri.parse(url));
-
-    // tslint:disable-next-line: no-unsafe-any
-    opn(url);
+    await vscode.env.openExternal(vscode.Uri.parse(url));
```
ejizba, line 9: "Report Issue" is a good test case, but another area I've had problems is opening the "Live Metrics stream" in the portal for Linux Function Apps. I have to do something like this:
```
`https://localhost:7071/${encodeURIComponent(JSON.stringify({ "test": "data" }))}/hi`
```

The URL looks like this:
```
https://localhost:7071/%7B%22test%22%3A%22data%22%7D/hi
```

but VS Code ends up converting it to this (which doesn't work):
```
https://localhost:7071/%7B%22test%22:%22data%22%7D/hi
```

Unfortunately I just don't think we can trust VS Code's `openExternal`. They've done a lot of work to fix edge cases, but fundamentally they're still "processing" the URL under the covers instead of just passing along the string we give them. Seems like we need to wait for https://github.com/microsoft/vscode/issues/85930 to be fixed before we can trust it

ejizba (changes requested):

ui/src/utils/openUrl.ts
```diff
@@ -3,13 +3,8 @@
...
-// tslint:disable-next-line:no-require-imports
-import opn = require("opn");
+import * as vscode from 'vscode';
 
 export async function openUrl(url: string): Promise<void> {
-    // Using this functionality is blocked by https://github.com/Microsoft/vscode/issues/25852:
-    // await vscode.commands.executeCommand('vscode.open', vscode.Uri.parse(url));
-
-    // tslint:disable-next-line: no-unsafe-any
-    opn(url);
```
ejizba, line 14: > I've been doing some work to eliminate warnings/errors/etc. throughout our build process; one of them is that opn is deprecated.

If this is your only goal, pretty sure you can just switch us to `open` to fix the warning. Should also update this, though: https://github.com/microsoft/vscode-azuretools/blob/main/dev/src/webpack/getDefaultWebpackConfig.ts#L26

ui/src/utils/openUrl.ts
```diff
@@ -4,12 +4,11 @@
...
-import opn = require("opn");
+import open = require("open");
 
 export async function openUrl(url: string): Promise<void> {
     // Using this functionality is blocked by https://github.com/Microsoft/vscode/issues/25852:
     // await vscode.commands.executeCommand('vscode.open', vscode.Uri.parse(url));
 
-    // tslint:disable-next-line: no-unsafe-any
-    opn(url);
+    await open(url);
```
bwateratmsft, line 13:
```suggestion
    void open(url);
```

Should this actually be `void` instead of `await`?

ui/src/utils/openUrl.ts
ejizba: Still need the default webpack config updated

ui/src/utils/openUrl.ts
ejizba: I don't think it really matters, but I guess I'd keep it as "await". I could imagine a scenario where we want to wait until the browser is open to continue some sort of UI flow in VS Code

ui/src/utils/openUrl.ts
bwateratmsft: Should I do vscode-azureextensiondev first then? i.e. as a separate PR?

bwateratmsft: #869

ui/package.json
```diff
@@ -40,7 +40,7 @@
         "escape-string-regexp": "^2.0.0",
         "fs-extra": "^8.0.0",
         "html-to-text": "^5.1.1",
-        "opn": "^6.0.0",
+        "open": "^8.0.4",
```
bwateratmsft, line 43: TODO: also update vscode-azureextensiondev to 0.8.1 after #869 is done

ui/src/utils/openUrl.ts
```diff
@@ -4,12 +4,11 @@
  *--------------------------------------------------------------------------------------------*/
 
 // tslint:disable-next-line:no-require-imports
-import opn = require("opn");
+import open = require("open");
 
 export async function openUrl(url: string): Promise<void> {
     // Using this functionality is blocked by https://github.com/Microsoft/vscode/issues/25852:
```
ejizba, line 10:
```suggestion
    // Using this functionality is blocked by https://github.com/Microsoft/vscode/issues/85930
```

ejizba (approved):

PR #868, opened 2021-03-26
Change from `request` (deprecated) to `genericServiceClient` / `@azure/ms-rest-js`

bwateratmsft:
Eliminates `request` (which is deprecated), along with a huge set of transitive dependencies, and makes a bunch more dev-only.

I chose `createGenericClient` / `@azure/ms-rest-js` because it's used in several other places for generic requests.


appservice/src/startStreamingLogs.ts
```diff
@@ -55,21 +56,24 @@ export async function startStreamingLogs(client: ISimplifiedSiteClient, verifyLo
...
-                            password: nonNullProp(creds, 'publishingPassword')
-                        },
-                        headers: {
-                            'User-Agent': appendExtensionUserAgent()
-                        }
-                    });
+                const buffer: Buffer = Buffer.from(`${creds.publishingUserName}${nonNullProp(creds, 'publishingPassword')}`);
+                const abortController: AbortController = new AbortController();
 
+                const logsRequest: Request = new Request(`${client.kuduUrl}/api/logstream/${logsPath}`, {
```
bwateratmsft, line 62: Maybe use @azure/core-http instead of node-fetch directly?

appservice/src/startStreamingLogs.ts
```diff
@@ -55,21 +56,28 @@ export async function startStreamingLogs(client: ISimplifiedSiteClient, verifyLo
...
+                    headers: {
+                        Authorization: `Basic ${buffer.toString('base64')}`,
+                        'User-Agent': appendExtensionUserAgent(),
+                    },
+                    signal: abortController.signal,
+                });
+
+                const logsResponse: Response = await fetch(logsRequest);
```
bwateratmsft, line 70: This isn't throwing errors for non-2xx response codes.

appservice/src/startStreamingLogs.ts
bwateratmsft: Gonna use @azure/ms-rest-js which is already used in several other places in this project.

appservice/src/startStreamingLogs.ts
```diff
@@ -55,21 +56,30 @@ export async function startStreamingLogs(client: ISimplifiedSiteClient, verifyLo
...
-                            'User-Agent': appendExtensionUserAgent()
-                        }
-                    });
+                const genericClient: ServiceClient = await createGenericClient(new BasicAuthenticationCredentials(creds.publishingUserName, nonNullProp(creds, 'publishingPassword')));
 
+                const abortController: AbortController = new AbortController();
+
+                // Due to https://github.com/Azure/ms-rest-js/issues/393 we need to use the WebResource constructor
+                // When that is fixed this can be simplified to just an options object passed to `genericClient.sendRequest`
+                const request = new WebResource(
```
bwateratmsft, line 65: I tried it out with Azure Functions, and it definitely works initially but seems to quit working after a bit. Maybe a timeout?

ejizba (approved):

PR #875, opened 2021-03-31
Fix webpack bug where it copies extra root-level dependency

ejizba:
tl;dr The databases vsix is broken when I use the latest webpack stuff. This should fix it and I think it's low risk in the sense that we'd know pretty fast if it broke anything else.

---
I ran into this error for Databases, related to which "external" dependencies it copies over to the "dist" folder

```
Activating extension 'ms-azuretools.vscode-cosmosdb' failed: Cannot find module 'lru-cache'
Require stack:
- /Users/erijiz/.vscode-insiders/extensions/ms-azuretools.vscode-cosmosdb-0.16.0/dist/node_modules/semver/classes/range.js
...
```

It's a little complicated, but I'll try to explain. Say you have "pg" and "semver" as your root dependencies in package.json and they each have their own sub dependencies like this:
- pg@1.0.0
  - semver@1.0.0
- semver@2.0.0
  - lru-cache@1.0.0

There's two versions of semver involved, and only one depends on lru-cache. You're going to get a node_modules folder like this:
Screen Shot 2021-03-31 at 11 49 30 AM

Here is what happens when you list "pg" as an "external".

### Expected
It copies this over to dist
Screen Shot 2021-03-31 at 11 49 30 AM copy 2

### Actual
It copies this over to dist
Screen Shot 2021-03-31 at 11 49 30 AM copy
The problem is it copied over the root semver@2.0.0 even though we're already copying the semver@1.0.0 under "pg". That alone might not be a problem (we just get some extra files), but it did _not_ copy lru-cache, which is where I start to see errors.

### DIFF
--- a/dev/package-lock.json
+++ b/dev/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "vscode-azureextensiondev",
-    "version": "0.9.1",
+    "version": "0.9.2",
     "lockfileVersion": 2,
     "requires": true,
     "packages": {
         "": {
             "name": "vscode-azureextensiondev",
-            "version": "0.9.1",
+            "version": "0.9.2",
             "license": "MIT",
             "dependencies": {
                 "@azure/arm-subscriptions": "^2.0.0",
--- a/dev/package.json
+++ b/dev/package.json
@@ -1,7 +1,7 @@
 {
     "name": "vscode-azureextensiondev",
     "author": "Microsoft Corporation",
-    "version": "0.9.1",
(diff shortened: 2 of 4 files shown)


nturinski (approved):
Am I crazy, or are all three of those screenshots the same?

Anyway, I'll trust you on this one.  As you said, we'll know pretty quickly if we destroyed something with this.

ejizba:
> Am I crazy, or are all three of those screenshots the same?

Yes the screenshot is the same, but the red box is different :P

PR #876, opened 2021-04-02
AppService - Added lines to Deploy output to show Zip Ignored Files

neelip:
Fixes [microsoft/vscode-azureappservice#1797](https://github.com/microsoft/vscode-azureappservice/issues/1797)

### DIFF
--- a/appservice/src/deploy/runWithZipStream.ts
+++ b/appservice/src/deploy/runWithZipStream.ts
@@ -78,13 +78,13 @@ async function getFilesFromGlob(folderPath: string, client: SiteClient): Promise
     const filesToInclude: string[] = await globby(globPattern, globOptions);
     const zipIgnorePatternStr = 'zipIgnorePattern';
 
-    const ignorePattern: string | string[] = zipDeployConfig.get<string | string[]>(zipIgnorePatternStr) || '';
-    const filesToIgnore: string[] = await globby(ignorePattern, globOptions);
+    const ignorePatternList: string | string[] = zipDeployConfig.get<string | string[]>(zipIgnorePatternStr) || '';
+    const filesToIgnore: string[] = await globby(ignorePatternList, globOptions);
 
-    ext.outputChannel.appendLog(`Ignoring files from \"${ext.prefix}.${zipIgnorePatternStr}\"`, { resourceName: client.fullName });
+    ext.outputChannel.appendLog(localize('zipIgnoreFileMsg', `Ignoring files from \"{0}.{1}\"`, ext.prefix, zipIgnorePatternStr), { resourceName: client.fullName });
 
-    for (const file of filesToIgnore) {
-        ext.outputChannel.appendLine(`\"${file}\"`);
+    for (const pattern of ignorePatternList) {
+        ext.outputChannel.appendLine(`\"${pattern}\"`);
     }
 
     return filesToInclude.filter(file => {
--- a/appservice/src/deploy/runWithZipStream.ts
+++ b/appservice/src/deploy/runWithZipStream.ts
@@ -78,13 +78,18 @@ async function getFilesFromGlob(folderPath: string, client: SiteClient): Promise
     const filesToInclude: string[] = await globby(globPattern, globOptions);
     const zipIgnorePatternStr = 'zipIgnorePattern';
 
-    const ignorePatternList: string | string[] = zipDeployConfig.get<string | string[]>(zipIgnorePatternStr) || '';
+    let ignorePatternList: string | string[] = zipDeployConfig.get<string | string[]>(zipIgnorePatternStr) || '';
     const filesToIgnore: string[] = await globby(ignorePatternList, globOptions);
 
-    ext.outputChannel.appendLog(localize('zipIgnoreFileMsg', `Ignoring files from \"{0}.{1}\"`, ext.prefix, zipIgnorePatternStr), { resourceName: client.fullName });
-
-    for (const pattern of ignorePatternList) {
-        ext.outputChannel.appendLine(`\"${pattern}\"`);
+    if (ignorePatternList) {
+        if (typeof ignorePatternList === 'string') {
+            ignorePatternList = [ignorePatternList];
+        }
+        ext.outputChannel.appendLog(localize('zipIgnoreFileMsg', `Ignoring files from \"{0}.{1}\"`, ext.prefix, zipIgnorePatternStr), { resourceName: client.fullName });
+        ext.outputChannel.appendLine(`\"${ignorePatternList}\"`);
+        for (const pattern of ignorePatternList) {
+            ext.outputChannel.appendLine(`\"${pattern}\"`);
+        }
     }
 
     return filesToInclude.filter(file => {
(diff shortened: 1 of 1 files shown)


appservice/src/deploy/runWithZipStream.ts
```diff
@@ -80,6 +80,11 @@ async function getFilesFromGlob(folderPath: string): Promise<string[]> {
     const ignorePattern: string | string[] = zipDeployConfig.get<string | string[]>('zipIgnorePattern') || '';
     const filesToIgnore: string[] = await globby(ignorePattern, globOptions);
 
+    ext.outputChannel.appendLog(localize('zipIgnoreFilesMsg', 'Ignoring files from .vscode/settings.json \"appService.zipIgnorePattern\"'), { resourceName: client.fullName });
```
ejizba, line 83: I wouldn't list the file name `.vscode/settings.json` since it could be set elsewhere, like in global user settings

ejizba, line 83: Use `ext.prefix` and make a const for `"zipIgnorePattern"`

appservice/src/deploy/runWithZipStream.ts
```diff
@@ -80,6 +80,11 @@ async function getFilesFromGlob(folderPath: string): Promise<string[]> {
...
+    for (const file of filesToIgnore) {
+        ext.outputChannel.appendLine(localize('zipIgnoreFileNames', '\"{0}\"', file));
```
ejizba, line 85: No need to use `localize` here

neelip: Do you mean for the lines or specifically the second one?
`    ext.outputChannel.appendLog(localize('zipIgnoreFilesMsg', 'Ignoring files from .vscode/settings.json \"appService.zipIgnorePattern\"'), { resourceName: client.fullName });
`

`ext.outputChannel.appendLine(localize('zipIgnoreFileNames', '\"{0}\"', file));`

ejizba: Specifically for "zipIgnoreFileNames"

appservice/src/deploy/runWithZipStream.ts
```diff
@@ -71,15 +71,22 @@ const commonGlobSettings: {} = {
...
+    const zipIgnorePatternStr = 'zipIgnorePattern';
 
-    const ignorePattern: string | string[] = zipDeployConfig.get<string | string[]>('zipIgnorePattern') || '';
+    const ignorePattern: string | string[] = zipDeployConfig.get<string | string[]>(zipIgnorePatternStr) || '';
     const filesToIgnore: string[] = await globby(ignorePattern, globOptions);
 
+    ext.outputChannel.appendLog(`Ignoring files from \"${ext.prefix}.${zipIgnorePatternStr}\"`, { resourceName: client.fullName });
+
+    for (const file of filesToIgnore) {
+        ext.outputChannel.appendLine(`\"${file}\"`);
```
ejizba, line 87: Can we get an example of what this looks like? If I ignore the "node_modules" folder - it's not gonna print out every file in "node_modules", is it?

appservice/src/deploy/runWithZipStream.ts
```diff
@@ -71,15 +71,22 @@ const commonGlobSettings: {} = {
...
     const globOptions = { cwd: folderPath, followSymbolicLinks: true, dot: true };
     const globPattern: string = zipDeployConfig.get<string>('zipGlobPattern') || '**/*';
     const filesToInclude: string[] = await globby(globPattern, globOptions);
+    const zipIgnorePatternStr = 'zipIgnorePattern';
 
-    const ignorePattern: string | string[] = zipDeployConfig.get<string | string[]>('zipIgnorePattern') || '';
+    const ignorePattern: string | string[] = zipDeployConfig.get<string | string[]>(zipIgnorePatternStr) || '';
     const filesToIgnore: string[] = await globby(ignorePattern, globOptions);
 
+    ext.outputChannel.appendLog(`Ignoring files from \"${ext.prefix}.${zipIgnorePatternStr}\"`, { resourceName: client.fullName });
```
ejizba, line 84: > Specifically for "zipIgnoreFileNames"

Meaning "zipIgnoreFiles**Msg**" should still use `localize`

appservice/src/deploy/runWithZipStream.ts
neelip: Yes, with the last commit that's what was happening, but I fixed it with the latest one. 

With the recent commit this is what it looks like...
image

appservice/src/deploy/runWithZipStream.ts
```diff
@@ -71,14 +71,21 @@ const commonGlobSettings: {} = {
...
+    const zipIgnorePatternStr = 'zipIgnorePattern';
 
-    const ignorePattern: string | string[] = zipDeployConfig.get<string | string[]>('zipIgnorePattern') || '';
-    const filesToIgnore: string[] = await globby(ignorePattern, globOptions);
+    const ignorePatternList: string | string[] = zipDeployConfig.get<string | string[]>(zipIgnorePatternStr) || '';
+    const filesToIgnore: string[] = await globby(ignorePatternList, globOptions);
+
+    ext.outputChannel.appendLog(localize('zipIgnoreFileMsg', `Ignoring files from \"{0}.{1}\"`, ext.prefix, zipIgnorePatternStr), { resourceName: client.fullName });
+
+    for (const pattern of ignorePatternList) {
```
ejizba, line 86: Can you double check what happens if `ignorePatternList` is a `string` instead of `string[]`. You might need a little extra logic to fix that case

neelip: Thank you! I missed that part

appservice/src/deploy/runWithZipStream.ts
```diff
@@ -71,14 +71,26 @@ const commonGlobSettings: {} = {
...
-    const filesToIgnore: string[] = await globby(ignorePattern, globOptions);
+    let ignorePatternList: string | string[] = zipDeployConfig.get<string | string[]>(zipIgnorePatternStr) || '';
+    const filesToIgnore: string[] = await globby(ignorePatternList, globOptions);
+
+    if (ignorePatternList) {
+        if (typeof ignorePatternList === 'string') {
+            ignorePatternList = [ignorePatternList];
+        }
+        ext.outputChannel.appendLog(localize('zipIgnoreFileMsg', `Ignoring files from \"{0}.{1}\"`, ext.prefix, zipIgnorePatternStr), { resourceName: client.fullName });
+        ext.outputChannel.appendLine(`\"${ignorePatternList}\"`);
```
ejizba, line 89: Mmm I think you just added this line for debugging purposes?

neelip: No, this part of the expected log output.

ejizba: Why? It seems duplicate with the for loop going through `ignorePatternList` again just below it

neelip: Yes, you are right. thank you!

ejizba (approved):

PR #880, opened 2021-04-05
Add "Recommend Skus" list to "Create - Advanced"

neelip:
Fixes #[1862](https://github.com/microsoft/vscode-azureappservice/issues/1862)

Here is how it looks now:
image

### DIFF
--- a/appservice/src/createAppService/AppServicePlanSkuStep.ts
+++ b/appservice/src/createAppService/AppServicePlanSkuStep.ts
@@ -42,7 +42,7 @@ export class AppServicePlanSkuStep extends AzureWizardPromptStep<IAppServiceWiza
 
         while (!wizardContext.newPlanSku) {
             const placeHolder = localize('pricingTierPlaceholder', 'Select a pricing tier');
-            wizardContext.newPlanSku = (await wizardContext.ui.showQuickPick(pricingTiers, { placeHolder })).data;
+            wizardContext.newPlanSku = (await wizardContext.ui.showQuickPick(pricingTiers, { placeHolder, suppressPersistence: true })).data;
 
             if (!wizardContext.newPlanSku) {
                 if (wizardContext.newSiteOS === WebsiteOS.linux) {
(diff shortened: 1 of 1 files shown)


ejizba:
@fiveisprime what are your thoughts? It looks a bit weird to me to have "F1 - Free" next to "Free - Try out Azure at no cost"

I'm wondering if we should either (1) suppress "recently used" for this prompt (2) leave out the "recommended" skus until VS Code support grouping and/or (3) find a better label/description/detail for each pick

fiveisprime:
Definitely not digging the two "free" entries. Can you help me understand why there are two "Free" tiers options showing up?

neelip:
**F1** `Free` is part of what already exists for advanced create. 
**Free** `Try out azure at no cost` is part of the recommended set. 

@ejizba I like option 1 and suppressing `(recently used)`, however also completely removing **F1** `Free`. 
I don't see the point of having 2 Free options. 
Also maybe appending (Recommended) to each description from the recommended set. 
What do you think @fiveisprime?

fiveisprime:
Wait so F1 is only showing up because you have picked it before, right? With a fresh install, I would only see Free. Is that right? Maybe we can go over this in our next meeting - I'm not sure exactly what this update does.

neelip:
@fiveisprime  Should we also remove `(recently used)`?

fiveisprime:
Yes and I think I like the (recommended). Can you show me how it looks with (recommended)?

neelip:
After check-in, going with no '(Recommended)' tag, and suppressing '(recently used)'.

appservice/src/createAppService/AppServicePlanSkuStep.ts
```diff
@@ -16,7 +16,11 @@ type ExtendedSkuDescription = WebSiteManagementModels.SkuDescription & { label?:
 
 export class AppServicePlanSkuStep extends AzureWizardPromptStep<IAppServiceWizardContext> {
     public async prompt(wizardContext: IAppServiceWizardContext): Promise<void> {
-        let skus: ExtendedSkuDescription[] = wizardContext.advancedCreation ? this.getCommonSkus() : this.getRecommendedSkus();
+        let skus: ExtendedSkuDescription[] = this.getRecommendedSkus()
+        if (wizardContext.advancedCreation) {
+            skus = skus.concat(this.getCommonSkus()).filter(sku => !(!sku.label && sku.name && sku.name === 'F1'));
```
ejizba, line 21: No need to do this filtering stuff when you can just remove `F1` from `getCommonSkus`

Also the name `getCommonSkus` doesn't really make sense anymore. Maybe `getAdvancedSkus`

ejizba (approved):

PR #883, opened 2021-04-07
Use VS Code ThemeIcon wherever possible

ejizba:
Benefits include:
1. VS Code owns accessibility on these instead of us (which helps for bugs like https://github.com/microsoft/vscode-azurefunctions/issues/2679)
2. Users can customize the icons by installing different themes

There were two main cases I _didn't_ use ThemeIcon:
1. Azure-specific icons - I feel like this falls into a "branding" thing and it's better to go with what we already have
2. The loading icon because of https://github.com/microsoft/vscode/issues/117859

Here's a sampling using a few different icon themes:

Screen Shot 2021-04-06 at 6 03 47 PMScreen Shot 2021-04-06 at 6 04 08 PMScreen Shot 2021-04-06 at 6 04 20 PM

Screen Shot 2021-04-06 at 6 05 47 PM
Screen Shot 2021-04-06 at 6 06 07 PM
Screen Shot 2021-04-06 at 6 05 26 PM


nturinski (approved):
😭 It's beautiful. So much red.

PR #887, opened 2021-04-16
Switch UI to to ESLint

wwlorey:


ui/src/parseError.ts
```diff
@@ -8,64 +8,76 @@ import { IParsedError } from '../index';
 import { localize } from './localize';
 import { parseJson } from './utils/parseJson';
 
-// tslint:disable:no-unsafe-any
-// tslint:disable:no-any
+// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types, @typescript-eslint/no-explicit-any
```
nturinski, line 11: I feel like you should just disable these rules for the entire file like it was originally.  While Eric previously mentioned he doesn't like turning off the rules for more than one line at a time, errors are kind of in a different ballgame because they always get caught as an "any" type and the whole point of this file was to ensure that they would have a readable message for us to use.

Also Eric wrote this file originally so if he has a problem with 🤷‍♀️

ui/src/registerEvent.ts
```diff
@@ -8,12 +8,15 @@ import * as types from '../index';
 import { callWithTelemetryAndErrorHandling } from './callWithTelemetryAndErrorHandling';
 import { ext } from './extensionVariables';
 
-// tslint:disable:no-any no-unsafe-any
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
 export function registerEvent<T>(eventId: string, event: Event<T>, callback: (context: types.IActionContext, ...args: any[]) => any): void {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
nturinski, line 13: I think you could use the `unknown` type here to get rid of the explicit-any rules

ui/src/tree/AzExtTreeDataProvider.ts
```diff
@@ -23,6 +23,7 @@ export class AzExtTreeDataProvider implements IAzExtTreeDataProviderInternal, ty
 
     private readonly _loadMoreCommandId: string;
     private readonly _rootTreeItem: AzExtParentTreeItem;
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
```
nturinski, line 26: What's causing this to be flagged??

nturinski (approved):

nturinski:
None of my comments are blockers though 👍

ui/src/parseError.ts
ejizba: I agree with what Nathan said. Please just disable for the whole file

> Eric previously mentioned he doesn't like turning off the rules for more than one line at a time

That mostly just applies to production code. Test code can be a lot more flexible when it comes to linting

ui/src/tree/AzExtTreeDataProvider.ts
wwlorey: It didn't like that `new Map()` didn't have a typecast. I added one and removed this exception

ui/src/parseError.ts
ejizba: Oh I didn't realize we _are_ talking about production code. In that case I agree less with what Nathan said, but I'm fine with it for this PR. Your main goal is to switch to eslint and keeping the status quo (disable for the whole file) is probably best in the short term. If/when we turn on lint in this file, we should fix some things rather than just disable each line. It'll take some work, but it might prevent something like this bug: https://github.com/microsoft/vscode-azuretools/pull/860

PR #888, opened 2021-04-21
Fix AuthorizationFailed Error with App Service - Basic Create

neelip:
[Update]
This PR resolves the issue of `AuthorizationFailed` with creating App Service Plans by prompting the user to select from existing. This will setup AppServicePlanCreateStep and AppServicePlanListStep to have similar behavior as ResourceGroupCreateStep and ResourceGroupListStep. 

Fixes #[2040](https://github.com/microsoft/vscode-azureappservice/issues/2040)


appservice/src/tryGetSiteResource.ts
```diff
@@ -28,11 +28,19 @@ export async function tryGetWebAppSlot(client: WebSiteManagementClient, resource
...
-    const result: T | WebSiteManagementModels.DefaultErrorResponse = await callback();
-    const regExp: RegExp = /NotFound/i;
-    if (regExp.test(parseError(result).errorType) || ('error' in result && regExp.test(parseError(result.error).errorType))) {
-        return undefined;
-    } else {
-        return <T>result;
+    try {
+        const result: T | WebSiteManagementModels.DefaultErrorResponse = await callback();
+        const regExp: RegExp = /NotFound/i;
+        if (regExp.test(parseError(result).errorType) || ('error' in result && regExp.test(parseError(result.error).errorType))) {
```
nturinski, line 34: This code was already checking for the error `NotFound` which makes me think that this API doesn't throw an error, but returns a `DefaultErrorResponse`.  Is that not the case with the `AuthorizationFailed` errors?

appservice/src/tryGetSiteResource.ts
```diff
@@ -28,11 +28,19 @@ export async function tryGetWebAppSlot(client: WebSiteManagementClient, resource
  * 3. { "Code": "NotFound", "Message": "Server farm with name appsvc_linux_centralus not found." }
  */
 async function tryGetSiteResource<T>(callback: () => Promise<T | WebSiteManagementModels.DefaultErrorResponse>): Promise<T | undefined> {
```
ejizba, line 30: I don't think `tryGetSiteResource` is the appropriate place to fix this issue. It's specifically meant to workaround an azure-sdk-for-js issue as described in the comment above and it affects a lot of code. I think your fix needs to be more targeted to specific scenarios.

What's the user experience like with your fix?  For example, if I look at [this code](https://github.com/microsoft/vscode-azuretools/blob/main/appservice/src/createAppService/AppServicePlanCreateStep.ts#L31) it appears that `tryGetAppServicePlan` will return `undefined` and we will try to create a new plan - which wouldn't work.

appservice/src/tryGetSiteResource.ts
neelip: You are right, it does return DefaultErrorResponse, but when checked for this particular scenario "The client 'xxxxxx@xxx.com' with object id 'XXXXXXXXXXXXXXXX' does not have..." is thrown rather than been returned.  I wasn't sure why that was the case for this particular error.

appservice/src/tryGetSiteResource.ts
neelip: I agree the current commit doesn't resolve the issue, when I just tested it. 
However, the modification at `tryGetSiteResource` is required as part of the fix, as `AuthorizationFailed` errors are thrown from the callback() instead of being returned as a DefaultErrorResponse as they are expected to behave.

ejizba: The only case that should return undefined for `tryGetSiteResource` is if the resource doesn't exist. Every other error (`AuthorizationFailed`, `ECONNRESET`, timeout, etc.`) _should_ be thrown as an error. Nothing should be returned as `DefaultErrorResponse`.

neelip: Ok I see what you mean, that's why only the error "NotFound" returns undefined.

ejizba:
@neelip feels like this PR has changed a lot. If/when it's ready for review again, please add repro steps and describe the new expected behavior

neelip:
This PR resolves the issue of `AuthorizationFailed` with creating App Service Plans by prompting the user to select from existing.

appservice/src/createAppService/AppInsightsCreateStep.ts
```diff
@@ -36,7 +36,6 @@ export class AppInsightsCreateStep extends AzureWizardExecuteStep<IAppServiceWiz
                 ext.outputChannel.appendLog(localize('existingNewAppInsights', 'Using existing Application Insights resource "{0}".', aiName));
             } catch (error) {
                 const pError: IParsedError = parseError(error);
-                // Only expecting a resource not found error if this is a new component
```
ejizba, line 39: revert this change

appservice/src/createAppService/AppServicePlanCreateStep.ts
```diff
@@ -4,14 +4,15 @@
...
 import { ext } from '../extensionVariables';
 import { localize } from '../localize';
 import { tryGetAppServicePlan } from '../tryGetSiteResource';
 import { createWebSiteClient } from '../utils/azureClients';
 import { nonNullProp, nonNullValueAndProp } from '../utils/nonNull';
 import { getAppServicePlanModelKind, WebsiteOS } from './AppKind';
+import { AppServicePlanListStep } from './AppServicePlanListStep';
 import { IAppServiceWizardContext } from './IAppServiceWizardContext';
 
 export class AppServicePlanCreateStep extends AzureWizardExecuteStep<IAppServiceWizardContext> {
```
ejizba, line 18: Please address the merge conflicts

appservice/src/createAppService/AppServicePlanCreateStep.ts
```diff
@@ -27,26 +28,47 @@ export class AppServicePlanCreateStep extends AzureWizardExecuteStep<IAppService
...
+                    kind: getAppServicePlanModelKind(wizardContext.newSiteKind, nonNullProp(wizardContext, 'newSiteOS')),
+                    sku: nonNullProp(wizardContext, 'newPlanSku'),
+                    location: nonNullValueAndProp(wizardContext.location, 'name'),
+                    reserved: wizardContext.newSiteOS === WebsiteOS.linux,  // The secret property - must be set to true to make it a Linux plan. Confirmed by the team who owns this API.
+                    maximumElasticWorkerCount: isElasticPremiumOrWorkflowStandard ? 20 : undefined
+                });
+                ext.outputChannel.appendLog(createdAppServicePlan);
+            }
+        } catch (e) {
+            if (wizardContext.suppress403Handling || parseError(e).errorType !== '403') {
```
ejizba, line 53: I never meant `suppress403Handling` to apply this widely - it was supposed to be unique to resource groups. The only reason I needed it was because in the rg extension [here](https://github.com/microsoft/vscode-azureresourcegroups/blob/b70b91ff4765f9ee861a4325b363b2333334296f/src/tree/SubscriptionTreeItem.ts#L39), the user is explicitly trying to create a resource group, so it would make no sense to "select an existing".

The correct fix is probably for me to move `suppress403Handling` from the `wizardContext` to a parameter in the constructor of `ResourceGroupCreateStep`. Not sure when I'll get to that, but for now I suggest you just ignore `suppress403Handling`

appservice/src/createAppService/AppServicePlanCreateStep.ts
```diff
@@ -27,26 +28,47 @@ export class AppServicePlanCreateStep extends AzureWizardExecuteStep<IAppService
...
+                throw e;
+            } else {
+                await this.selectExistingPrompt(wizardContext);
+            }
         }
+
+    }
+
+    public async selectExistingPrompt(wizardContext: IAppServiceWizardContext): Promise<void> {
+        const message: string = localize('rgForbidden', 'You do not have permission to create a app service plan in subscription "{0}".', wizardContext.subscriptionDisplayName);
```
ejizba, line 63:
```suggestion
        const message: string = localize('planForbidden', 'You do not have permission to create an app service plan in subscription "{0}".', wizardContext.subscriptionDisplayName);
```

appservice/src/createAppService/AppServicePlanCreateStep.ts
```diff
@@ -28,30 +29,52 @@ export class AppServicePlanCreateStep extends AzureWizardExecuteStep<IAppService
...
+                    location: location.name,
+                    reserved: wizardContext.newSiteOS === WebsiteOS.linux,  // The secret property - must be set to true to make it a Linux plan. Confirmed by the team who owns this API.
+                    maximumElasticWorkerCount: isElasticPremiumOrWorkflowStandard ? 20 : undefined,
+                    kubeEnvironmentProfile: getKubeProfile(wizardContext),
+                    perSiteScaling: !!wizardContext.customLocation
+                });
+                ext.outputChannel.appendLog(createdAppServicePlan);
+            }
+        } catch (e) {
+            if (parseError(e).errorType === '403' || parseError(e).errorType === 'AuthorizationFailed') {
```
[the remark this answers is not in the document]
neelip: When I tested both basic and advanced cases, only getting "AuthorizationFailed" error. 
Both [AppServicePlanCreateStep](https://github.com/microsoft/vscode-azuretools/blob/33d9ae162d4cf92f1083b3b66dedbf5454127069/appservice/src/createAppService/AppServicePlanCreateStep.ts#L32) and [ResourceGroupCreateStep](https://github.com/microsoft/vscode-azuretools/blob/33d9ae162d4cf92f1083b3b66dedbf5454127069/ui/src/wizard/ResourceGroupCreateStep.ts#L27) make calls to different resources. It makes sense for ResourceGroupCreateStep to behave differently.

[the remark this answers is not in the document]
neelip: These are the 2 possible message formats I found for error type "403".
```
Error 403 - Forbidden The web app you have attempted to reach has blocked your access.
Error 403 - This web app is stopped. The web app you have attempted to reach is currently stopped and does not accept any requests. Please try to reload the page or visit it again soon...

```
And all messages in the following format are type "AuthorizationFailed". 
image

ejizba (approved):

PR #886, opened 2021-04-13
code changes to add a new site kind

vikash-anand:
In this PR I have made changes to include a new appkind i.e workflowApp. This appkind is used to multi tenant Logic App.

PR #890, opened 2021-04-27
Adding support for WorkflowApp Sku

preetriti1:
This PR contains changes in app service plan creation and selection for creating Workflow App [Logic Apps (Preview) extension.


appservice/src/createAppService/AppServicePlanSkuStep.ts
```diff
@@ -188,4 +190,30 @@ export class AppServicePlanSkuStep extends AzureWizardPromptStep<IAppServiceWiza
             }
         ];
     }
+
+    private getWorkflowStandardSkus(): WebSiteManagementModels.SkuDescription[] {
```
tonytang-microsoft-com, line 194: Do we need to localize any display name?

tonytang-microsoft-com (approved):

appservice/src/createAppService/AppServicePlanSkuStep.ts
preetriti1: I dont see any value that is localizable here, following same pattern as Elastic Premium. I am assuming description is what we localize, which we dont have here.

appservice/src/createAppService/AppServicePlanListStep.ts
```diff
@@ -64,11 +69,13 @@ export class AppServicePlanListStep extends AzureWizardPromptStep<IAppServiceWiz
     }
 
     private async getQuickPicks(wizardContext: IAppServiceWizardContext): Promise<IAzureQuickPickItem<WebSiteManagementModels.AppServicePlan | undefined>[]> {
-        const picks: IAzureQuickPickItem<WebSiteManagementModels.AppServicePlan | undefined>[] = [{
-            label: localize('CreateNewAppServicePlan', '$(plus) Create new App Service plan'),
-            description: '',
-            data: undefined
-        }];
+        const picks: IAzureQuickPickItem<WebSiteManagementModels.AppServicePlan | undefined>[] = !wizardContext.disablePlanCreate
```
nturinski, line 72: Could you please follow this design pattern if you need to remove the `Create new App Service plan` quickpick?

https://github.com/microsoft/vscode-azuretools/blob/main/ui/src/wizard/ResourceGroupListStep.ts#L23-L29
https://github.com/microsoft/vscode-azuretools/blob/main/ui/src/wizard/ResourceGroupListStep.ts#L70-L79

appservice/src/createAppService/AppKind.ts
```diff
@@ -10,7 +10,8 @@ export enum WebsiteOS {
 
 export enum AppKind {
     app = 'app',
-    functionapp = 'functionapp'
+    functionapp = 'functionapp',
+    workflowapp = 'workflowapp'
```
nturinski, line 14: These strings don't match here: https://github.com/microsoft/vscode-azuretools/pull/886/files#diff-286a2b3db8a91df266b072104bec32dca8d8124f038dcc0df9beea633788351dR14

appservice/src/createAppService/AppServicePlanSkuStep.ts
```diff
@@ -17,7 +17,9 @@ type ExtendedSkuDescription = WebSiteManagementModels.SkuDescription & { label?:
 export class AppServicePlanSkuStep extends AzureWizardPromptStep<IAppServiceWizardContext> {
     public async prompt(wizardContext: IAppServiceWizardContext): Promise<void> {
         let skus: ExtendedSkuDescription[] = wizardContext.advancedCreation ? this.getCommonSkus() : this.getRecommendedSkus();
-        if (wizardContext.newSiteKind === AppKind.functionapp) {
+        if (wizardContext.newSiteKind.includes(AppKind.workflowapp)) {
```
nturinski, line 20: Where is `AppKind.workflowapp` set?

Have there been discussions on whether or not workflow/logic apps will be available in basic creation?  This will overwrite any skus we have from the step above.

appservice/src/createAppService/AppServicePlanSkuStep.ts
nturinski: You can optionally add a label property which should be localized, otherwise we default to the name property which doesn't need to be localized.

appservice/src/createAppService/AppServicePlanListStep.ts
```diff
@@ -80,8 +87,8 @@ export class AppServicePlanListStep extends AzureWizardPromptStep<IAppServiceWiz
             const isNewSiteLinux: boolean = wizardContext.newSiteOS === WebsiteOS.linux;
             let isPlanLinux: boolean = nonNullProp(plan, 'kind').toLowerCase().includes(WebsiteOS.linux);
 
-            if (plan.sku && plan.sku.family === 'EP') {
-                // elastic premium plans do not have the os in the kind, so we have to check the "reserved" property
+            if (plan.sku && (plan.sku.family === 'EP' || plan.sku.family === 'WS')) {
+                // elastic premium plans and workflow standard plans do not have the os in the kind, so we have to check the "reserved" property
                 // also, the "reserved" property is always "false" in the list of plans returned above. We have to perform a separate get on each plan
                 const client: WebSiteManagementClient = await createWebSiteClient(wizardContext);
                 const epPlan: WebSiteManagementModels.AppServicePlan | undefined = await tryGetAppServicePlan(client, nonNullProp(plan, 'resourceGroup'), nonNullProp(plan, 'name'));
```
nturinski, line 94: Will this work with Workflow plans still?  Are Workflow plans exclusively Linux or can they be Windows as well?

appservice/src/createAppService/AppServicePlanSkuStep.ts
preetriti1: yeah not localizing the name here.

appservice/src/createAppService/AppKind.ts
preetriti1: I have provided a comment in the other PR to just keep one value, its because kind usually is a combination of strings for workflowapp and we should not provide a strict order else string matching will not be correct.
today siteKind if it contains workflowApp it means its a logic apps in preview.

appservice/src/createAppService/AppServicePlanSkuStep.ts
preetriti1: Today workflow app (in our GA) will only support Workflow Standard Skus and nothing else.
In case of App Service Plan it will only support ASE v3 plans, for which i am providing planfilter in the context.

appservice/src/createAppService/AppServicePlanListStep.ts
preetriti1: Workflow standard sku is exactly similar to elastic plan, bit that is only specific to logic apps. Now currently these only support windows.

appservice/src/createAppService/AppServicePlanListStep.ts
preetriti1: Updating with class variable.

appservice/src/createAppService/AppServicePlanSkuStep.ts
nturinski: Could you move this to be the bottom of the if/else if conditions?  I don't think it should effect us, but I want to be safe and make sure that our check for `functionapp` happens before the `workflowapp`.

appservice/src/createAppService/AppServicePlanSkuStep.ts
```diff
@@ -88,15 +90,8 @@ export class AppServicePlanSkuStep extends AzureWizardPromptStep<IAppServiceWiza
         ];
     }
 
-    private getCommonSkus(): WebSiteManagementModels.SkuDescription[] {
+    private getAdvancedSkus(): WebSiteManagementModels.SkuDescription[] {
```
nturinski, line 93: Seems like something odd happened with the merge here and [here](https://github.com/microsoft/vscode-azuretools/pull/890/files?diff=split&w=1#diff-1487b1bd8f2d1b579960b612f89af39e94e334ac27dfca02ddfbcfcd6083e165R19)

appservice/src/createAppService/AppServicePlanListStep.ts
```diff
@@ -36,7 +43,12 @@ export class AppServicePlanListStep extends AzureWizardPromptStep<IAppServiceWiz
 
     public async prompt(wizardContext: IAppServiceWizardContext): Promise<void> {
         // Cache hosting plan separately per subscription
-        const options: IAzureQuickPickOptions = { placeHolder: localize('selectPlan', 'Select a {0} App Service plan.', getWebsiteOSDisplayName(nonNullProp(wizardContext, 'newSiteOS'))), id: `AppServicePlanListStep/${wizardContext.subscriptionId}` };
+        const options: IAzureQuickPickOptions = {
+            placeHolder: wizardContext.newSiteKind.includes(AppKind.workflowapp) && wizardContext.planSkuFamilyFilter?.test('IV2')
```
nturinski, line 47: I don't remember this filter being here before.  I hadn't seen any `IV2` plans-- just wanted to make sure that this was intentional.

appservice/src/createAppService/AppServicePlanSkuStep.ts
preetriti1: though it doesnt matter, but i will move it.

appservice/src/createAppService/AppServicePlanListStep.ts
preetriti1: Yes this was intentional, today we only support the new sku for workflow and only one specific type of sku, this is for that and only specific to workflow app. I can explain more in teams

appservice/src/createAppService/AppServicePlanSkuStep.ts
preetriti1: I copied the stuff which was different in master to my repo... didnt do a proper merge

nturinski (approved):

PR #892, opened 2021-04-28
Remove ISimplifiedSiteClient and add createKuduClient

ejizba:
Just a bit of refactor, should be no functional change.

- Remove `ISimplifiedSiteClient`, a remnant of trial apps that is no longer needed
- Move `SiteClient.getKuduClient` to it's own function `createKuduClient`, which more closely matches how we create all other clients (i.e. like these [here](https://github.com/microsoft/vscode-azurefunctions/blob/main/src/utils/azureClients.ts))


ejizba:
ping @nturinski @wwlorey @neelip

nturinski (approved):

PR #893, opened 2021-04-28
Move app setting validation on deploy out of this package

ejizba:
I want to remove the `validateLinuxFunctionAppSettings` logic from this package and put it directly in the Functions repo. Right now I have app setting validation in both repos and I want to consolidate that into one place, plus this logic is unique to Functions anyways. Should be no functional changes, just moving code around.


appservice/src/deploy/delayFirstWebAppDeploy.ts
```diff
@@ -16,6 +16,8 @@ export async function delayFirstWebAppDeploy(client: SiteClient, asp: WebSiteMan
             if (client.isFunctionApp) {
                 resolve();
             }
+
+            const asp: WebSiteManagementModels.AppServicePlan | undefined = await aspPromise;
```
nturinski, line 20: Shouldn't we still be using this logic here?

```
try {
   asp = await aspPromise;
} catch {
   asp = undefined;
}
```

appservice/src/deploy/runDeployTask.ts
```diff
@@ -70,9 +71,9 @@ export interface IPreDeployTaskResult {
     failedToFindTask: boolean;
 }
 
-function shouldExecuteTask(scmType: string | undefined, settingKey: string, taskName: string): boolean {
+function shouldExecuteTask(context: IDeployContext, scmType: string | undefined, settingKey: string, taskName: string): boolean {
     // We don't run deploy tasks for non-zipdeploy since that stuff should be handled by kudu
-    const shouldExecute: boolean = scmType !== ScmType.LocalGit && scmType !== ScmType.GitHub;
+    const shouldExecute: boolean = context.deployMethod === 'storage' || context.deployMethod === 'zip' || (scmType !== ScmType.LocalGit && scmType !== ScmType.GitHub);
```
nturinski, line 76: Did this boolean expression need to change?  Wouldn't `scmType !== ScmType.LocalGit && scmType !== ScmType.GitHub` still cover this (since if the deployMethod were `storage` or `zip` then scmType would be `ScmType.None`)?

nturinski (commented):
And don't forget to bump the version! 😁 I never get to say that.

ejizba:
> And don't forget to bump the version!

Already done: https://github.com/microsoft/vscode-azuretools/pull/892

appservice/src/deploy/runDeployTask.ts
ejizba: deployMethod is "Used to overwrite default deploy method based on scm type"

So like Linux Consumption doesn't support LocalGit/GitHub, so we always use zip regardless of scmType

appservice/src/deploy/delayFirstWebAppDeploy.ts
ejizba: It's already in a bigger try/catch

nturinski (approved):

PR #894, opened 2021-05-03
Improve location step

ejizba:
Two main changes:
1. Use a later api-version when getting locations because it has a more accurate list and has info on which locations are "recommended". I picked the same api-version that the portal uses
2. Add support for "grouping" in our quick pick. VS Code still doesn't support a separator natively (https://github.com/microsoft/vscode/issues/74967) and I got tired of waiting haha. I added our own pick with a chevron icon indicating users can expand/collapse the group and IMO it does the job just fine

Screen Shot 2021-05-03 at 11 15 59 AM
Screen Shot 2021-05-03 at 11 16 09 AM

### DIFF
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -5,7 +5,6 @@
 
 import { ResourceManagementModels } from '@azure/arm-resources';
 import { StorageManagementModels } from '@azure/arm-storage';
-import { SubscriptionModels } from '@azure/arm-subscriptions';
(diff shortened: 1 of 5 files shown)


nturinski (approved):
👏 This is pretty incredibly done.

I'm not familiar with activeItems in quickpicks.  Do they render without the quickpick prompt disappearing or will the window disappear momentarily when they expand categories?

PR #895, opened 2021-05-05
Add app service specific location step

ejizba:


appservice/src/createAppService/CustomLocationListStep.ts
```diff
@@ -0,0 +1,63 @@
...
+            wizardContext.newPlanName = await wizardContext.relatedNameTask;
+            wizardContext.newPlanSku = { name: 'U1', tier: 'ANY', size: 'U1', family: 'U', capacity: 1 };
+        } else {
+            wizardContext.location = result;
+        }
+    }
+
+    protected async getCustomQuickPicks(wizardContext: T): Promise<IAzureQuickPickItem<AzExtLocation | CustomLocation>[]> {
+        const picks: IAzureQuickPickItem<AzExtLocation | CustomLocation>[] = await super.getQuickPicks(wizardContext);
+        if (wizardContext.newSiteOS === 'linux') {
```
preetriti1, line 29: Is this necessary today? I would assume custom locations to show up all the time unless the OS is picked by user.
In case of logic apps, we dont have a OS picker and if the user selects custom location we will set the os of that type.
Can we have a way to show custom location in all cases and skip this check, either by having a class variable set in constructor of this step?

preetriti1 (changes requested):

appservice/src/createAppService/AppServicePlanCreateStep.ts
```diff
@@ -84,3 +85,13 @@ interface ExtendedAppServicePlan extends WebSiteManagementModels.AppServicePlan
...
+
+export function getPlanKind(wizardContext: IAppServiceWizardContext): string {
+    if (wizardContext.customLocation) {
+        return 'linux,kubernetes';
+    } else if (wizardContext.newSiteOS === WebsiteOS.linux) {
+        return WebsiteOS.linux;
+    } else {
+        return AppKind.app;
+    }
+}
```
neelip, line 97: Can it be moved to AppKind.ts file? Is there a specific reason for this to be here?

appservice/src/createAppService/CustomLocationListStep.ts
```diff
@@ -0,0 +1,63 @@
...
+import { CustomLocation, IAppServiceWizardContext } from './IAppServiceWizardContext';
+
+export class CustomLocationListStep<T extends IAppServiceWizardContext> extends LocationListStep<T> {
+    public async prompt(wizardContext: T): Promise<void> {
+        const options: IAzureQuickPickOptions = { placeHolder: localize('selectLocation', 'Select a location for new resources.'), enableGrouping: true };
+        const result: AzExtLocation | CustomLocation = (await wizardContext.ui.showQuickPick(this.getCustomQuickPicks(wizardContext), options)).data;
+        if ('kubeEnvironment' in result) {
+            wizardContext.customLocation = result;
+            // For any resources other than the app, we still need a non-custom location, so we'll use the kubeEnvironment's location
+            await LocationListStep.setLocation(wizardContext, result.kubeEnvironment.location);
```
nturinski, line 18: Is there any negative impact of setting the location if this is app? Though the comment mentions it, there's no specific check to verify that this isn't an app.

appservice/src/createAppService/CustomLocationListStep.ts
```diff
@@ -0,0 +1,63 @@
...
+            // Plan has very little effect when a custom location is used, so default this info instead of prompting
+            wizardContext.newPlanName = await wizardContext.relatedNameTask;
+            wizardContext.newPlanSku = { name: 'K1', tier: 'Kubernetes', size: 'K1', family: 'K', capacity: 1 };
+        } else {
+            wizardContext.location = result;
+        }
+    }
+
+    protected async getCustomQuickPicks(wizardContext: T): Promise<IAzureQuickPickItem<AzExtLocation | CustomLocation>[]> {
```
nturinski, line 27: Why call this `getCustomQuickPicks` when this is already the `CustomLocationListStep`?

appservice/src/createAppService/IAppServiceWizardContext.ts
```diff
@@ -93,4 +93,16 @@ export interface IAppServiceWizardContext extends IResourceGroupWizardContext, I
      * Indicates advanced creation should be used
      */
     advancedCreation?: boolean;
+
+    customLocation?: CustomLocation;
+}
+
+export type CustomLocation = {
```
nturinski, line 100: Feels like this should extend `Location` and just add the `kubeEnvironment` property.

```suggestion
export type CustomLocation = Location & {
    kubeEnvironment: {
        name: string;
        id: string;
        location: string;
    }
}
```

nturinski (approved):

appservice/src/createAppService/AppServicePlanCreateStep.ts
ejizba: It's only used in this file, so I thought it made the most sense to be in this file (I can remove the "export"). IMO, it should only go in AppKind.ts if it's used by other files

appservice/src/createAppService/CustomLocationListStep.ts
ejizba: To differentiate between `super.getQuickPicks` which is getting the other ones

PR #896, opened 2021-05-05
Fix generalizeLocationName

ejizba:
I'm testing some stuff in the "North Central US (Stage)" location and our current logic doesn't quite work because the location sometimes shows up with or without "(Stage)" and we need to know they're the same.

### DIFF
--- a/ui/src/wizard/LocationListStep.ts
+++ b/ui/src/wizard/LocationListStep.ts
@@ -11,7 +11,10 @@ import { nonNullProp } from '../utils/nonNull';
 import { AzureWizardPromptStep } from './AzureWizardPromptStep';
 
 function generalizeLocationName(name: string | undefined): string {
-    return (name || '').toLowerCase().replace(/\s/g, '');
+    return (name || '')
+        .toLowerCase()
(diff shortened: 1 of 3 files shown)


ui/src/wizard/LocationListStep.ts
```diff
@@ -11,7 +11,10 @@ import { nonNullProp } from '../utils/nonNull';
 import { AzureWizardPromptStep } from './AzureWizardPromptStep';
 
 function generalizeLocationName(name: string | undefined): string {
-    return (name || '').toLowerCase().replace(/\s/g, '');
+    return (name || '')
+        .toLowerCase()
+        .replace(/\([^\)]*\)/g, '') // remove parentheticals, like in "North Central US (Stage)"
+        .replace(/[^a-z0-9]/g, ''); // remove anything else other than letters/numbers
```
ravipal, line 17: `/gi` to handle Upper case?

ravipal (approved):

ejizba:
This problem is more complicated than I thought. Will file a new PR separately

PR #897, opened 2021-05-06
Adding string changes during workflow app create [from Vikash's PR]

preetriti1:
This also includes null check for siteKind property in other areas


appservice/src/createAppService/AppServicePlanListStep.ts
```diff
@@ -43,8 +43,10 @@ export class AppServicePlanListStep extends AzureWizardPromptStep<IAppServiceWiz
 
     public async prompt(wizardContext: IAppServiceWizardContext): Promise<void> {
         // Cache hosting plan separately per subscription
+        // NOTE(psamband): Logic Apps only supports Workflow Standard sku and for App Service Plan it only supports one Isolated sku.
```
nturinski, line 46:
```suggestion
        // Logic Apps only supports Workflow Standard sku and for App Service Plan it only supports one Isolated sku.
```

appservice/src/createAppService/AppServicePlanListStep.ts
```diff
@@ -43,8 +43,10 @@ export class AppServicePlanListStep extends AzureWizardPromptStep<IAppServiceWiz
...
+        // Since create is not enabled in isolated skus so we explicitly call out the type of plan picked in the picker text .
```
nturinski, line 47:
```suggestion
        // Since create is not enabled for isolated skus, we explicitly reference the type of plan picked in the placeHolder.
```

nturinski (approved):

PR #899, opened 2021-05-07
Allow location filtering on a per-provider basis

ejizba:
Basically, there's a decent number of weird scenarios where different providers handle locations differently. For now I just focused on "Microsoft.Web" and "Microsoft.Resources/resourceGroups", but in the future it would be a good idea to expand this to include storage accounts and app insights.

For example:
1. Say web apps support locations [a,b,c] and resource groups support locations [b,c,d]. We need to show [b,c], but the existing logic only allowed one `locationsTask` (meaning [a,b,c] was usually shown because web apps set the task last). Now we can properly filter to [b,c]
2. There's another edge case where some providers support "stage" locations and some don't. Even if a "stage" location is set, though, I can find a non-stage "related" location when possible

### DIFF
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -1038,7 +1038,7 @@ export declare class LocationListStep<T extends ILocationWizardContext> extends 
      * @param wizardContext The context of the wizard
      * @param provider If specified, this will check against that provider's supported locations and attempt to find a "related" location if the selected location is not supported.
      */
-    public static async getLocation<T extends ILocationWizardContext>(wizardContext: T, provider?: string): Promise<AzExtLocation>;
+    public static getLocation<T extends ILocationWizardContext>(wizardContext: T, provider?: string): Promise<AzExtLocation>;
 
     /**
      * Returns true if a location has been set on the context
@@ -1054,7 +1054,7 @@ export declare class LocationListStep<T extends ILocationWizardContext> extends 
     public prompt(wizardContext: T): Promise<void>;
     public shouldPrompt(wizardContext: T): boolean;
 
-    protected async getQuickPicks(wizardContext: T): Promise<IAzureQuickPickItem<AzExtLocation>[]>;
+    protected getQuickPicks(wizardContext: T): Promise<IAzureQuickPickItem<AzExtLocation>[]>;
 }
 
 export interface IAzureNamingRules {
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -999,18 +999,11 @@ export type AzExtLocation = {
     }
 }
 
+/**
+ * Currently no location-specific properties on the wizard context, but keeping this interface for backwards compatability and ease of use
+ * Instead, use static methods on `LocationListStep` like `getLocation` and `setLocationSubset`
+ */
 export interface ILocationWizardContext extends ISubscriptionWizardContext {
-    /**
-     * The location to use for new resources
-     * This value will be defined after `LocationListStep.prompt` occurs or after you call `LocationListStep.setLocation`
-     */
-    location?: AzExtLocation;
-
-    /**
-     * Optional task to describe the subset of locations that should be displayed.
-     * If not specified, all locations supported by the user's subscription will be displayed.
-     */
-    locationsTask?: Promise<{ name?: string }[]>;
 }
 
 export declare class LocationListStep<T extends ILocationWizardContext> extends AzureWizardPromptStep<T> {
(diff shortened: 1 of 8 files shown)


ui/index.d.ts
```diff
@@ -1021,16 +1014,37 @@ export declare class LocationListStep<T extends ILocationWizardContext> extends
...
+     * @param provider The relevant provider (i.e. 'Microsoft.Web')
+     */
+    public static setLocationSubset<T extends ILocationWizardContext>(wizardContext: T, task: Promise<string[]>, provider: string): void;
+
+    /**
+     * Gets the selected location for this wizard.
+     * @param wizardContext The context of the wizard
+     * @param provider If specified, this will check against that provider's supported locations and attempt to find a "related" location if the selected location is not supported.
+     */
+    public static async getLocation<T extends ILocationWizardContext>(wizardContext: T, provider?: string): Promise<AzExtLocation>;
```
preetriti1, line 1041: async is not supported in d.ts 
Similarly in line: 1057, today the latest azureextensionui (0.42.0) package has build error in extension due to this.
Can you please not include async?

preetriti1 (changes requested):

wwlorey (approved):

ui/index.d.ts
nturinski: According to this, just having a Promise<> return type is enough: https://stackoverflow.com/questions/48448151/how-can-i-add-an-async-function-to-a-type-def-file

nturinski (approved):

ui/src/wizard/LocationListStep.ts
```diff
@@ -22,50 +19,108 @@ interface ILocationWizardContextInternal extends types.ILocationWizardContext {
...
+                    // Some providers prefer their version of the name over the standard one, so we'll create a shallow clone using theirs
+                    return { ...loc, name: nonNullValue(providerLocations?.find(name => matchesLocation(loc, name), 'providerName')) };
+                }
+
+                if (isSupportedByProvider(location)) {
+                    return useProviderName(location);
+                }
+
+                if (location.name.toLowerCase().endsWith('stage')) {
+                    const nonStageName = location.name.replace(/stage/i, '');
```
ravipal, line 93: I don't think we have any actual location name with stage in it. But this could potentially remove the 'stage' which is not at the end.

ui/src/wizard/LocationListStep.ts
```diff
@@ -82,24 +137,31 @@ export class LocationListStep<T extends ILocationWizardContextInternal> extends
     }
 }
 
-async function getAllLocations(wizardContext: ILocationWizardContextInternal): Promise<types.AzExtLocation[]> {
+function generalizeLocationName(name: string | undefined): string {
+    return (name || '').toLowerCase().replace(/[^a-z0-9]/gi, '');
```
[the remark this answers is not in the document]
ravipal: with toLowerCase, the 'i' flag is unnecessary.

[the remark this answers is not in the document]
ejizba: @ravipal I didn't have the `i` here and you commented about it! 😝  https://github.com/microsoft/vscode-azuretools/pull/896#discussion_r626975720. I figured it made it more readable and doesn't hurt so I included it

@wwlorey `generalizeLocationName` is used to test equality of locations because different APIs use slightly different naming. I want "North Central US (Stage)" to be equal to "northcentralusstage", so my main reason was to remove those parentheses. I've never seen any other special characters in the location names

ui/src/wizard/StorageAccountCreateStep.ts
```diff
@@ -22,8 +23,7 @@ export class StorageAccountCreateStep<T extends types.IStorageAccountWizardConte
     }
 
     public async execute(wizardContext: T, progress: Progress<{ message?: string; increment?: number }>): Promise<void> {
-        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-        const newLocation: string = wizardContext.location!.name!;
+        const newLocation: string = (await LocationListStep.getLocation(wizardContext)).name;
```
preetriti1, line 26: Can this be null or undefined?

ui/src/wizard/LocationListStep.ts
```diff
@@ -22,50 +19,108 @@ interface ILocationWizardContextInternal extends types.ILocationWizardContext {
...
     protected constructor() {
         super();
     }
 
-    public static addStep<T extends ILocationWizardContextInternal>(wizardContext: types.IActionContext & Partial<ILocationWizardContextInternal>, promptSteps: AzureWizardPromptStep<T>[]): void {
+    public static addStep<T extends ILocationWizardContextInternal>(wizardContext: types.ISubscriptionContext & Partial<ILocationWizardContextInternal>, promptSteps: AzureWizardPromptStep<T>[]): void {
         if (!wizardContext._alreadyHasLocationStep) {
             promptSteps.push(new this());
             wizardContext._alreadyHasLocationStep = true;
+            this.getInternalVariables(wizardContext); // initialize
```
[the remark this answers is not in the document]
ejizba: The problem is there are a number of static methods on here that people can use without actually adding the step. `getInternalVariables` just ensures the variables are initialized by whichever one gets called first.

The easiest example is that in App Service, we use `LocationListStep.setLocation` to set the location to "centralus" during basic create without ever adding the step

ui/src/wizard/LocationListStep.ts
```diff
@@ -22,50 +19,108 @@ interface ILocationWizardContextInternal extends types.ILocationWizardContext {
...
+            if (providerLocations) {
+                function isSupportedByProvider(loc: types.AzExtLocation): boolean {
+                    return !!providerLocations?.find(name => matchesLocation(loc, name));
+                }
+                function useProviderName(loc: types.AzExtLocation): types.AzExtLocation {
+                    // Some providers prefer their version of the name over the standard one, so we'll create a shallow clone using theirs
+                    return { ...loc, name: nonNullValue(providerLocations?.find(name => matchesLocation(loc, name), 'providerName')) };
+                }
+
+                if (isSupportedByProvider(location)) {
```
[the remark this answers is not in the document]
ejizba: Yes and it depends on the user's subscription. For me, web resources are supported in the staging locations, but resource groups are not.

ui/src/wizard/LocationListStep.ts
ejizba: I think I'm fine with that. For example, may there's "North Central US (Stage) 2" or "(Stage) North Central US"

In any case, the important thing to note is that I still check `isSupportedByProvider` before actually using this `nonStageName`. I don't think there's much possible harm since I'm doing that. Although I suppose I could add a warning in the output channel

ui/src/wizard/StorageAccountCreateStep.ts
ejizba: no

PR #908, opened 2021-05-11
Filter storage accounts by selected location

ejizba:
Users should always create an app in the same location as their plan/storage account. Since the location prompt is now _before_ these prompts, the portal filters the existing resources based on location. I don't really like the flow of this, but I think we should do the same at least for now.
Screen Shot 2021-05-11 at 11 10 50 AM

The storage account step is even more annoying because we _already_ have a filter by sku. Thankfully this only happens in advanced create and I hope both filters happening at once isn't that common.

Screen Shot 2021-05-11 at 11 17 09 AM

### DIFF
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -838,6 +838,7 @@ export interface IAzureQuickPickItem<T = undefined> extends QuickPickItem {
     /**
      * Callback to use when this item is picked, instead of returning the pick
      * Only applies when used as part of an `AzureWizard`
+     * This is not compatible with `canPickMany`
      */
     onPicked?: () => void | Promise<void>;
 
@@ -876,6 +877,7 @@ export interface IAzureQuickPickOptions extends QuickPickOptions {
     /**
      * If true, you must specify a `group` property on each `IAzureQuickPickItem` and the picks will be grouped into collapsible sections
      * Only applies when used as part of an `AzureWizard`
+     * This is not compatible with `canPickMany`
      */
     enableGrouping?: boolean;
 
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -836,7 +836,14 @@ export interface IAzureQuickPickItem<T = undefined> extends QuickPickItem {
     data: T;
 
     /**
+     * Callback to use when this item is picked, instead of returning the pick
+     * Only applies when used as part of an `AzureWizard`
+     */
+    onPicked?: () => void | Promise<void>;
+
+    /**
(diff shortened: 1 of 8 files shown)


ui/src/wizard/StorageAccountListStep.ts
```diff
@@ -139,12 +141,22 @@ export class StorageAccountListStep<T extends types.IStorageAccountWizardContext
...
+        if (hasFilteredAccountsBySku && this._filters.learnMoreLink) {
             picks.push({
-                label: localize('filtered', '$(info) Some storage accounts were filtered. Learn more...'),
-                description: '',
-                suppressPersistence: true,
-                data: this._filters.learnMoreLink
+                label: localize('hasFilteredAccountsBySku', '$(info) Some storage accounts were filtered because of their sku. Learn more...'),
+                onPicked: async () => {
+                    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+                    await openUrl(this._filters.learnMoreLink!);
```
nturinski, line 149: Why not just use our `nonNullProp`?

ui/index.d.ts
```diff
@@ -835,8 +835,15 @@ export interface IAzureQuickPickItem<T = undefined> extends QuickPickItem {
 
     data: T;
 
+    /**
+     * Callback to use when this item is picked, instead of returning the pick
+     * Only applies when used as part of an `AzureWizard`
+     */
+    onPicked?: () => void | Promise<void>;
```
nturinski, line 842: Maybe add a little note about it not being compatible with `canPickMany`?

ui/src/wizard/StorageAccountListStep.ts
```diff
@@ -139,12 +141,22 @@ export class StorageAccountListStep<T extends types.IStorageAccountWizardContext
...
+                },
+                data: undefined
+            });
+        }
+
+        if (hasFilteredAccountsByLocation && location) {
+            picks.push({
+                label: localize('hasFilteredAccountsByLocation', '$(warning) Only storage accounts in the selected region "{0}" are shown.', location.displayName),
```
wwlorey, line 157: No strong preference but I think the "info" and "warning" icons look a little strange right next to each other. Maybe the info icon could suffice for both?

wwlorey (approved):

ui/src/wizard/StorageAccountListStep.ts
nturinski: I had thought that too, but I think it'd be hard for a user to parse that they're different messages if the icons were the same.  Especially since the content of the messages are similar.

PR #912, opened 2021-05-12
Rename artifact to app.jar for java se deployment

Flanker32:
Rename artifact to app.jar for java se runtime deployment


appservice/src/deploy/deploy.ts
```diff
@@ -80,9 +81,17 @@ export async function deploy(client: SiteClient, fsPath: string, context: IDeplo
                 if (!(await fse.pathExists(fsPath))) {
                     throw new Error(localize('pathNotExist', 'Failed to deploy path that does not exist: {0}', fsPath));
                 }
-
-                if (config.linuxFxVersion && /^(tomcat|wildfly)/i.test(config.linuxFxVersion)) {
+                const javaRutime = client.isLinux ? config.linuxFxVersion : config.javaContainer;
+                if (javaRutime && /^(tomcat|wildfly|jobss)/i.test(javaRutime)) {
```
nturinski, line 85: Isn't it supposed to be `jboss`? https://www.redhat.com/en/technologies/jboss-middleware/application-platform

Flanker32: Yes, its a typo, thanks a lot!

appservice/src/deploy/deploy.ts
```diff
@@ -80,9 +81,17 @@ export async function deploy(client: SiteClient, fsPath: string, context: IDeplo
...
                 }
-
-                if (config.linuxFxVersion && /^(tomcat|wildfly)/i.test(config.linuxFxVersion)) {
+                const javaRutime = client.isLinux ? config.linuxFxVersion : config.javaContainer;
+                if (javaRutime && /^(tomcat|wildfly|jboss)/i.test(javaRutime)) {
                     await deployWar(context, client, fsPath);
+                } else if (javaRutime && /^java/i.test(javaRutime)) {
+                    // For Java SE runtime, need rename the artifact to app.jar
+                    let javaArtifact: string = fsPath;
+                    if (path.basename(javaArtifact) !== "app.jar") {
```
nturinski, line 90: `.jar` files can be quite large.  This copy could potentially take a long time and we would also be writing it in a temporary folder where it would be taking up space.  Unless explicitly deleted, I don't believe temp files are deleted automatically either. (source https://www.makeuseof.com/tag/why-temp-files-not-deleted-automatically/)

I think that we could just change the name of the metadata of the zip file as it's being zipped [here](https://github.com/microsoft/vscode-azuretools/blob/main/appservice/src/deploy/runWithZipStream.ts#L55)

nturinski (changes requested):

appservice/src/deploy/deploy.ts
```diff
@@ -80,9 +81,17 @@ export async function deploy(client: SiteClient, fsPath: string, context: IDeplo
                 if (!(await fse.pathExists(fsPath))) {
                     throw new Error(localize('pathNotExist', 'Failed to deploy path that does not exist: {0}', fsPath));
                 }
-
-                if (config.linuxFxVersion && /^(tomcat|wildfly)/i.test(config.linuxFxVersion)) {
+                const javaRutime = client.isLinux ? config.linuxFxVersion : config.javaContainer;
```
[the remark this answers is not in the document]
Flanker32: Nice catch.

appservice/src/deploy/deploy.ts
Flanker32: @nturinski Thanks for your advices, I'll refactor the logic to do the rename in package.

nturinski: I think I have a potential solution for this refactor.

We could let `runWithZipStream` have an optional parameter that is of type `{ [key: string]: string }`.  The key would be the real file name and the value would be the desired file name.

So for example, if the object is `{ main.jar: app.jar }`, this [line](https://github.com/microsoft/vscode-azuretools/blob/main/appservice/src/deploy/runWithZipStream.ts#L55) could look like this:
```
const zipFileMetadata= { main.jar : "app.jar" } // this would be passed in as a parameter
zipFile.addFile(path.join(fsPath, file), zipFileMetadata[file] || file);
```

appservice/src/deploy/deployJar.ts
```diff
@@ -0,0 +1,31 @@
...
+import { WebSiteManagementModels } from '@azure/arm-appservice';
+import { createKuduClient } from '../createKuduClient';
+import { SiteClient } from '../SiteClient';
+import { delayFirstWebAppDeploy } from './delayFirstWebAppDeploy';
+import { IDeployContext } from './IDeployContext';
+import { Readable } from 'stream';
+import * as yazl from 'yazl';
+import { waitForDeploymentToComplete } from './waitForDeploymentToComplete';
+
+export async function deployJar(context: IDeployContext, client: SiteClient, fsPath: string, aspPromise: Promise<WebSiteManagementModels.AppServicePlan | undefined>): Promise<void> {
```
nturinski, line 15: I would prefer if you used the solution I suggested in this comment https://github.com/microsoft/vscode-azuretools/pull/912#discussion_r635337289.

This solution results in a lot of duplicated code.

appservice/src/deploy/deploy.ts
Flanker32: @nturinski Thanks a lot for your advices, I create a new `deployJar` method because renaming the artifact to app.jar is only the requirement for java se deployment, and there seems no similar requirement for other languages cause only java requires an artifact file while others require to deploy the project, so I'm not sure whether I should add extra parameters to the shared `runWithZipStream`.

appservice/src/deploy/deployJar.ts
Flanker32: Thanks for your advices, pr updated

nturinski (approved):
Thanks for incorporating my feedback! Looks good

appservice/src/deploy/deploy.ts
```diff
@@ -80,9 +81,14 @@ export async function deploy(client: SiteClient, fsPath: string, context: IDeplo
                 if (!(await fse.pathExists(fsPath))) {
                     throw new Error(localize('pathNotExist', 'Failed to deploy path that does not exist: {0}', fsPath));
                 }
-
-                if (config.linuxFxVersion && /^(tomcat|wildfly)/i.test(config.linuxFxVersion)) {
+                const javaRuntime = client.isLinux ? config.linuxFxVersion : config.javaContainer;
```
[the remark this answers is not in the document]
Flanker32: @nturinski Thanks for your feedback, this is to get the runtime info for java app service; for linux app service, I need to get the runtime data from linuxFxVersion, for windows, java runtime stored in javaContainer.

PR #919, opened 2021-05-14
Fix "AuthorizationFailed" errors with AppInsightsCreateStep

neelip:
Fixes #[2050](https://github.com/microsoft/vscode-azureappservice/issues/2050)

### DIFF
--- a/appservice/src/createAppService/AppInsightsCreateStep.ts
+++ b/appservice/src/createAppService/AppInsightsCreateStep.ts
@@ -47,7 +47,7 @@ export class AppInsightsCreateStep extends AzureWizardExecuteStep<IAppServiceWiz
                     ext.outputChannel.appendLog(createdNewAppInsights);
                 } else if (pError.errorType === 'AuthorizationFailed') {
                     if (!wizardContext.advancedCreation) {
-                        const createAppInsightsNotAvailable: string = localize('appInsightsNotAvailable', 'Skipping creating an Application Insights because the user doesn\'t have the right permissions.');
+                        const createAppInsightsNotAvailable: string = localize('appInsightsNotAvailable', 'Skipping Application Insights resource because you do not have permission to create one in this subscription.');
                         ext.outputChannel.appendLog(createAppInsightsNotAvailable);
                     } else {
                         await this.selectExistingPrompt(wizardContext);
@@ -57,19 +57,19 @@ export class AppInsightsCreateStep extends AzureWizardExecuteStep<IAppServiceWiz
                 }
             }
         } else {
-            const appInsightsNotAvailable: string = localize('appInsightsNotAvailable', 'Skipping creating an Application Insights resource because it isn\'t compatible with this location.');
+            const appInsightsNotAvailable: string = localize('appInsightsNotAvailable', 'Skipping Application Insights resource because it isn\'t compatible with this location.');
             ext.outputChannel.appendLog(appInsightsNotAvailable);
         }
     }
 
     public async selectExistingPrompt(wizardContext: IAppServiceWizardContext): Promise<void> {
-        const message: string = localize('planForbidden', 'You do not have permission to create an app insights resource in subscription "{0}".', wizardContext.subscriptionDisplayName);
+        const message: string = localize('aiForbidden', 'You do not have permission to create an app insights resource in subscription "{0}".', wizardContext.subscriptionDisplayName);
         const selectExisting: MessageItem = { title: localize('selectExisting', 'Select Existing') };
         wizardContext.telemetry.properties.cancelStep = 'AppInsightsNoPermissions';
         await wizardContext.ui.showWarningMessage(message, { modal: true }, selectExisting);
 
         wizardContext.telemetry.properties.cancelStep = undefined;
-        wizardContext.telemetry.properties.forbiddenResponse = 'SelectExistingAsp';
+        wizardContext.telemetry.properties.forbiddenResponse = 'SelectExistingAppInsights';
         const step: AppInsightsListStep = new AppInsightsListStep(true /* suppressCreate */);
         await step.prompt(wizardContext);
     }
--- a/appservice/src/createAppService/AppInsightsCreateStep.ts
+++ b/appservice/src/createAppService/AppInsightsCreateStep.ts
@@ -65,14 +65,21 @@ export class AppInsightsCreateStep extends AzureWizardExecuteStep<IAppServiceWiz
     public async selectExistingPrompt(wizardContext: IAppServiceWizardContext): Promise<void> {
         const message: string = localize('aiForbidden', 'You do not have permission to create an app insights resource in subscription "{0}".', wizardContext.subscriptionDisplayName);
         const selectExisting: MessageItem = { title: localize('selectExisting', 'Select Existing') };
+        const skipForNow: MessageItem = { title: localize('skipForNow', 'Skip for Now') };
         wizardContext.telemetry.properties.cancelStep = 'AppInsightsNoPermissions';
-        await wizardContext.ui.showWarningMessage(message, { modal: true }, selectExisting);
+        const result = await wizardContext.ui.showWarningMessage(message, { modal: true }, selectExisting, skipForNow);
 
-        wizardContext.telemetry.properties.cancelStep = undefined;
-        wizardContext.telemetry.properties.forbiddenResponse = 'SelectExistingAppInsights';
-        const step: AppInsightsListStep = new AppInsightsListStep(true /* suppressCreate */);
-        await step.prompt(wizardContext);
+        if (result === skipForNow) {
+            wizardContext.telemetry.properties.aiSkipForNow = 'true';
+            wizardContext.appInsightsSkip = true;
+        } else {
+            wizardContext.telemetry.properties.cancelStep = undefined;
+            wizardContext.telemetry.properties.forbiddenResponse = 'SelectExistingAppInsights';
+            const step: AppInsightsListStep = new AppInsightsListStep(true /* suppressCreate */);
+            await step.prompt(wizardContext);
+        }
     }
+
 
     public shouldExecute(wizardContext: IAppServiceWizardContext): boolean {
         return !wizardContext.appInsightsComponent && !!wizardContext.newAppInsightsName;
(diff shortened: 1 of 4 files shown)


appservice/src/createAppService/AppInsightsCreateStep.ts
```diff
@@ -44,6 +44,9 @@ export class AppInsightsCreateStep extends AzureWizardExecuteStep<IAppServiceWiz
                     wizardContext.appInsightsComponent = await client.components.createOrUpdate(rgName, aiName, { kind: 'web', applicationType: 'web', location: appInsightsLocation });
                     const createdNewAppInsights: string = localize('createdNewAppInsights', 'Successfully created Application Insights resource "{0}".', aiName);
                     ext.outputChannel.appendLog(createdNewAppInsights);
+                } else if (pError.errorType === 'AuthorizationFailed') {
+                    const createAppInsightsNotAvailable: string = localize('appInsightsNotAvailable', 'Skipping creating an Application Insights because the user doesn\'t have the right permissions.');
```
ejizba, line 48: For resource groups and plans we let them select an existing one - how come we're not doing that here?

appservice/src/createAppService/AppInsightsCreateStep.ts
```diff
@@ -44,6 +44,9 @@ export class AppInsightsCreateStep extends AzureWizardExecuteStep<IAppServiceWiz
                     wizardContext.appInsightsComponent = await client.components.createOrUpdate(rgName, aiName, { kind: 'web', applicationType: 'web', location: appInsightsLocation });
```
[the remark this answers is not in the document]
neelip: Sorry about that. I accidentally pasted the wrong image, the repro steps should be correct.

appservice/src/createAppService/AppInsightsCreateStep.ts
neelip: We can do that here as well. I think we were going with this as it might cause too many prompts and we do have an option of creating without the AppInsights. 
Do you think it is better to prompt to select the existing?

ejizba: How many prompts is it? I could maybe see skipping it for basic create, but for advanced create it seems weird to silently skip it when they explicitly tried to create one

neelip: I guess with Web Plan Contributor, there is only one prompt and that is with selecting existing resource groups. 
There shouldn't be an error with creating an App Service Plan. 

I agree I think we should skip it for basic create and prompt for advanced create.

appservice/src/createAppService/AppInsightsCreateStep.ts
```diff
@@ -44,6 +45,13 @@ export class AppInsightsCreateStep extends AzureWizardExecuteStep<IAppServiceWiz
...
                     const createdNewAppInsights: string = localize('createdNewAppInsights', 'Successfully created Application Insights resource "{0}".', aiName);
                     ext.outputChannel.appendLog(createdNewAppInsights);
+                } else if (pError.errorType === 'AuthorizationFailed') {
+                    if (!wizardContext.advancedCreation) {
+                        const createAppInsightsNotAvailable: string = localize('appInsightsNotAvailable', 'Skipping creating an Application Insights because the user doesn\'t have the right permissions.');
```
ejizba, line 50: Trying to match the other prompts/logs better:
```suggestion
                        const createAppInsightsNotAvailable: string = localize('appInsightsNotAvailable', 'Skipping Application Insights resource because you do not have permission to create one in this subscription.');
```

appservice/src/createAppService/AppInsightsCreateStep.ts
```diff
@@ -54,6 +62,18 @@ export class AppInsightsCreateStep extends AzureWizardExecuteStep<IAppServiceWiz
         }
     }
 
+    public async selectExistingPrompt(wizardContext: IAppServiceWizardContext): Promise<void> {
+        const message: string = localize('planForbidden', 'You do not have permission to create an app insights resource in subscription "{0}".', wizardContext.subscriptionDisplayName);
```
ejizba, line 66:
```suggestion
        const message: string = localize('aiForbidden', 'You do not have permission to create an app insights resource in subscription "{0}".', wizardContext.subscriptionDisplayName);
```

appservice/src/createAppService/AppInsightsCreateStep.ts
```diff
@@ -54,6 +62,18 @@ export class AppInsightsCreateStep extends AzureWizardExecuteStep<IAppServiceWiz
...
+        const selectExisting: MessageItem = { title: localize('selectExisting', 'Select Existing') };
+        wizardContext.telemetry.properties.cancelStep = 'AppInsightsNoPermissions';
+        await wizardContext.ui.showWarningMessage(message, { modal: true }, selectExisting);
+
+        wizardContext.telemetry.properties.cancelStep = undefined;
+        wizardContext.telemetry.properties.forbiddenResponse = 'SelectExistingAsp';
```
ejizba, line 72:
```suggestion
        wizardContext.telemetry.properties.forbiddenResponse = 'SelectExistingAppInsights';
```

appservice/src/createAppService/AppInsightsCreateStep.ts
```diff
@@ -54,6 +62,18 @@ export class AppInsightsCreateStep extends AzureWizardExecuteStep<IAppServiceWiz
         }
     }
 
+    public async selectExistingPrompt(wizardContext: IAppServiceWizardContext): Promise<void> {
+        const message: string = localize('planForbidden', 'You do not have permission to create an app insights resource in subscription "{0}".', wizardContext.subscriptionDisplayName);
+        const selectExisting: MessageItem = { title: localize('selectExisting', 'Select Existing') };
+        wizardContext.telemetry.properties.cancelStep = 'AppInsightsNoPermissions';
+        await wizardContext.ui.showWarningMessage(message, { modal: true }, selectExisting);
```
ejizba, line 69: It feels like this prompt should have the "Skip for Now" option because it's not obvious that "Select Existing" will eventually let me skip

appservice/src/createAppService/AppInsightsCreateStep.ts
neelip: I was trying to match this line (60), 
            `const appInsightsNotAvailable: string = localize('appInsightsNotAvailable', 'Skipping creating an Application Insights resource because it isn\'t compatible with this location.');<br class="Apple-interchange-newline">`. 
Should this be modified as well?

ejizba: Oh didn't see that one. Yeah how about we change that to:
> Skipping Application Insights resource because it isn\'t compatible with this location.


I just don't like "skipping" and "creating" right next to each other. Too many "ing"s

appservice/src/createAppService/AppInsightsCreateStep.ts
neelip: Do you mean have "Skip for Now" directly in the warning along with "Select Existing"?

neelip: Something like this...? 
image

ejizba: Yes exactly

appservice/src/createAppService/AppInsightsCreateStep.ts
```diff
@@ -44,16 +45,42 @@ export class AppInsightsCreateStep extends AzureWizardExecuteStep<IAppServiceWiz
                     wizardContext.appInsightsComponent = await client.components.createOrUpdate(rgName, aiName, { kind: 'web', applicationType: 'web', location: appInsightsLocation });
                     const createdNewAppInsights: string = localize('createdNewAppInsights', 'Successfully created Application Insights resource "{0}".', aiName);
                     ext.outputChannel.appendLog(createdNewAppInsights);
+                } else if (pError.errorType === 'AuthorizationFailed') {
+                    if (!wizardContext.advancedCreation) {
+                        const createAppInsightsNotAvailable: string = localize('appInsightsNotAvailable', 'Skipping Application Insights resource because you do not have permission to create one in this subscription.');
+                        ext.outputChannel.appendLog(createAppInsightsNotAvailable);
```
ejizba, line 51:
```suggestion
                        const appInsightsNotAuthorized: string = localize('appInsightsNotAuthorized', 'Skipping Application Insights resource because you do not have permission to create one in this subscription.');
                        ext.outputChannel.appendLog(appInsightsNotAuthorized);
```

appservice/src/createAppService/AppInsightsCreateStep.ts
```diff
@@ -44,16 +45,42 @@ export class AppInsightsCreateStep extends AzureWizardExecuteStep<IAppServiceWiz
...
+        const selectExisting: MessageItem = { title: localize('selectExisting', 'Select Existing') };
+        const skipForNow: MessageItem = { title: localize('skipForNow', 'Skip for Now') };
+        wizardContext.telemetry.properties.cancelStep = 'AppInsightsNoPermissions';
+        const result = await wizardContext.ui.showWarningMessage(message, { modal: true }, selectExisting, skipForNow);
+
+        if (result === skipForNow) {
+            wizardContext.telemetry.properties.aiSkipForNow = 'true';
+            wizardContext.appInsightsSkip = true;
+        } else {
+            wizardContext.telemetry.properties.cancelStep = undefined;
```
ejizba, line 76: Move this cancelStep line just before `if (result === skipForNow) {` so that we also clear it for the skipForNow case

appservice/src/createAppService/AppInsightsCreateStep.ts
```diff
@@ -44,16 +45,42 @@ export class AppInsightsCreateStep extends AzureWizardExecuteStep<IAppServiceWiz
...
 
+    public async selectExistingPrompt(wizardContext: IAppServiceWizardContext): Promise<void> {
+        const message: string = localize('aiForbidden', 'You do not have permission to create an app insights resource in subscription "{0}".', wizardContext.subscriptionDisplayName);
+        const selectExisting: MessageItem = { title: localize('selectExisting', 'Select Existing') };
+        const skipForNow: MessageItem = { title: localize('skipForNow', 'Skip for Now') };
+        wizardContext.telemetry.properties.cancelStep = 'AppInsightsNoPermissions';
+        const result = await wizardContext.ui.showWarningMessage(message, { modal: true }, selectExisting, skipForNow);
+
+        if (result === skipForNow) {
+            wizardContext.telemetry.properties.aiSkipForNow = 'true';
```
ejizba, line 73: Add
```
wizardContext.telemetry.properties.forbiddenResponse = 'SkipAppInsights';
```

ejizba (approved):

PR #920, opened 2021-05-15
Add grouping to App Service Skus list 

neelip:
Fixes #[2012](https://github.com/microsoft/vscode-azureappservice/issues/2012)

Advanced Create: 
image
Basic Create: 
image

Couldn't find a way to display "Show Pricing Information..." without the group. 
Any suggestions how grouping can be disabled for just that one pick? 

@fiveisprime What do you think about this layout?

### DIFF
--- a/appservice/src/createAppService/AppServicePlanSkuStep.ts
+++ b/appservice/src/createAppService/AppServicePlanSkuStep.ts
@@ -88,7 +88,7 @@ export class AppServicePlanSkuStep extends AzureWizardPromptStep<IAppServiceWiza
                 size: 'P1v2',
                 family: 'Pv2',
                 capacity: 1,
-                label: localize('premiumLabel', 'Premium (P1V2)'),
+                label: localize('premiumLabel', 'Premium (P1v2)'),
                 description: localize('premiumDescription', 'Use in production'),
                 group: recommendedGroup
             }
--- a/appservice/src/createAppService/AppServicePlanSkuStep.ts
+++ b/appservice/src/createAppService/AppServicePlanSkuStep.ts
@@ -16,47 +16,32 @@ type ExtendedSkuDescription = WebSiteManagementModels.SkuDescription & { label?:
 
 export class AppServicePlanSkuStep extends AzureWizardPromptStep<IAppServiceWizardContext> {
     public async prompt(wizardContext: IAppServiceWizardContext): Promise<void> {
-        let recommendedSkus = this.getRecommendedSkus();
+        let skus: ExtendedSkuDescription[] = wizardContext.advancedCreation ? this.getRecommendedSkus().concat(this.getAdvancedSkus()) : this.getRecommendedSkus();
+        if (wizardContext.newSiteKind === AppKind.functionapp) {
+            skus.push(...this.getElasticPremiumSkus());
+        } else if (wizardContext.newSiteKind?.includes(AppKind.workflowapp)) {
+            skus = this.getWorkflowStandardSkus();
+        }
+
         const regExp: RegExp | undefined = wizardContext.planSkuFamilyFilter;
         if (regExp) {
-            recommendedSkus = recommendedSkus.filter(s => !s.family || regExp.test(s.family));
+            skus = skus.filter(s => !s.family || regExp.test(s.family));
         }
-        const pricingTiers: IAzureQuickPickItem<WebSiteManagementModels.SkuDescription | undefined>[] = recommendedSkus.map(s => {
+
+        const pricingTiers: IAzureQuickPickItem<WebSiteManagementModels.SkuDescription | undefined>[] = skus.map(s => {
             return {
                 label: s.label || nonNullProp(s, 'name'),
                 description: s.description || s.tier,
                 data: s,
-                group: localize('recommendedSkusLabel', 'Recommended')
+                group: s.label && s.description ? localize('recommendedLabel', 'Recommended') : localize('additionalOptionsLabel', 'Additional Options')
             };
         });
-
-        let advancedSkus: ExtendedSkuDescription[] = [];
-        if (wizardContext.advancedCreation) {
-            advancedSkus.push(...this.getAdvancedSkus());
-            if (regExp) {
-                advancedSkus = advancedSkus.filter(s => !s.family || regExp.test(s.family));
-            }
-            if (wizardContext.newSiteKind === AppKind.functionapp) {
-                advancedSkus.push(...this.getElasticPremiumSkus());
-            } else if (wizardContext.newSiteKind?.includes(AppKind.workflowapp)) {
-                advancedSkus = this.getWorkflowStandardSkus();
-            }
-
-            pricingTiers.push(...advancedSkus.map(s => {
-                return {
-                    label: s.label || nonNullProp(s, 'name'),
-                    description: s.description || s.tier,
-                    data: s,
-                    group: localize('additionalOptionsLabel', 'Additional Options'),
-                };
-            }));
-        }
 
         pricingTiers.push({ label: localize('ShowPricingCalculator', '$(link-external) Show pricing information...'), data: undefined, suppressPersistence: true });
 
         while (!wizardContext.newPlanSku) {
             const placeHolder = localize('pricingTierPlaceholder', 'Select a pricing tier');
-            wizardContext.newPlanSku = <WebSiteManagementModels.SkuDescription | undefined>(await wizardContext.ui.showQuickPick(pricingTiers, { placeHolder, suppressPersistence: true, enableGrouping: wizardContext.advancedCreation })).data;
+            wizardContext.newPlanSku = (await wizardContext.ui.showQuickPick(pricingTiers, { placeHolder, suppressPersistence: true, enableGrouping: wizardContext.advancedCreation })).data;
 
             if (!wizardContext.newPlanSku) {
                 if (wizardContext.newSiteOS === WebsiteOS.linux) {
--- a/appservice/src/createAppService/AppServicePlanSkuStep.ts
+++ b/appservice/src/createAppService/AppServicePlanSkuStep.ts
@@ -21,7 +21,7 @@ export class AppServicePlanSkuStep extends AzureWizardPromptStep<IAppServiceWiza
                 label: s.label || nonNullProp(s, 'name'),
                 description: s.description || s.tier,
                 data: s,
-                group: 'Recommended'
+                group: wizardContext.advancedCreation ? 'Recommended' : ''
             };
         });
 
@@ -45,7 +45,7 @@ export class AppServicePlanSkuStep extends AzureWizardPromptStep<IAppServiceWiza
                     label: s.label || nonNullProp(s, 'name'),
                     description: s.description || s.tier,
                     data: s,
-                    group: s.tier,
+                    group: 'Additional Options',
                 };
             }));
         }
--- a/appservice/src/createAppService/AppServicePlanSkuStep.ts
+++ b/appservice/src/createAppService/AppServicePlanSkuStep.ts
@@ -16,28 +16,30 @@ type ExtendedSkuDescription = WebSiteManagementModels.SkuDescription & { label?:
 
 export class AppServicePlanSkuStep extends AzureWizardPromptStep<IAppServiceWizardContext> {
     public async prompt(wizardContext: IAppServiceWizardContext): Promise<void> {
-        const pricingTiers: IAzureQuickPickItem<WebSiteManagementModels.SkuDescription | undefined>[] = this.getRecommendedSkus().map(s => {
+        let recommendedSkus = this.getRecommendedSkus();
+        const regExp: RegExp | undefined = wizardContext.planSkuFamilyFilter;
+        if (regExp) {
+            recommendedSkus = recommendedSkus.filter(s => !s.family || regExp.test(s.family));
+        }
+        const pricingTiers: IAzureQuickPickItem<WebSiteManagementModels.SkuDescription | undefined>[] = recommendedSkus.map(s => {
             return {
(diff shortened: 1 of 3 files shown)


fiveisprime:
I like the groups, but I really dislike the empty group for the pricing info option. I'm very interested if there's a way to avoid that.

Are the screenshots in the right order? Looks like advanced is on top, no?

ejizba:
> I like the groups, but I really dislike the empty group for the pricing info option. I'm very interested if there's a way to avoid that.

Yes, @neelip should be able to fix that.

Another option is to do "Recommended" and "Additional Options" kind of like the portal, rather than group them by "Basic", "Standard", "Premium"

Screen Shot 2021-05-17 at 10 52 13 AM

appservice/src/createAppService/AppServicePlanSkuStep.ts
```diff
@@ -16,31 +16,46 @@ type ExtendedSkuDescription = WebSiteManagementModels.SkuDescription & { label?:
...
-        }
-
-        const pricingTiers: IAzureQuickPickItem<WebSiteManagementModels.SkuDescription | undefined>[] = skus.map(s => {
+        const pricingTiers: IAzureQuickPickItem<WebSiteManagementModels.SkuDescription | undefined>[] = this.getRecommendedSkus().map(s => {
             return {
                 label: s.label || nonNullProp(s, 'name'),
                 description: s.description || s.tier,
-                data: s
+                data: s,
+                group: wizardContext.advancedCreation ? 'Recommended' : ''
```
ejizba, line 24: localize

appservice/src/createAppService/AppServicePlanSkuStep.ts
```diff
@@ -16,31 +16,46 @@ type ExtendedSkuDescription = WebSiteManagementModels.SkuDescription & { label?:
...
+            if (regExp) {
+                advancedSkus = advancedSkus.filter(s => !s.family || regExp.test(s.family));
+            }
+
+            pricingTiers.push(...advancedSkus.map(s => {
+                return {
+                    label: s.label || nonNullProp(s, 'name'),
+                    description: s.description || s.tier,
+                    data: s,
+                    group: 'Additional Options',
```
ejizba, line 48: localize

appservice/src/createAppService/AppServicePlanSkuStep.ts
```diff
@@ -16,31 +16,46 @@ type ExtendedSkuDescription = WebSiteManagementModels.SkuDescription & { label?:
...
+
+            if (wizardContext.newSiteKind === AppKind.functionapp) {
+                advancedSkus.push(...this.getElasticPremiumSkus());
+            } else if (wizardContext.newSiteKind?.includes(AppKind.workflowapp)) {
+                advancedSkus = this.getWorkflowStandardSkus();
+            }
+
+            const regExp: RegExp | undefined = wizardContext.planSkuFamilyFilter;
+            if (regExp) {
+                advancedSkus = advancedSkus.filter(s => !s.family || regExp.test(s.family));
```
ejizba, line 40: This needs to be outside `if (wizardContext.advancedCreation)` so that it also filters the recommended skus

appservice/src/createAppService/AppServicePlanSkuStep.ts
```diff
@@ -16,31 +16,46 @@ type ExtendedSkuDescription = WebSiteManagementModels.SkuDescription & { label?:
...
+                };
+            }));
+        }
+
         pricingTiers.push({ label: localize('ShowPricingCalculator', '$(link-external) Show pricing information...'), data: undefined, suppressPersistence: true });
 
         while (!wizardContext.newPlanSku) {
             const placeHolder = localize('pricingTierPlaceholder', 'Select a pricing tier');
-            wizardContext.newPlanSku = (await wizardContext.ui.showQuickPick(pricingTiers, { placeHolder, suppressPersistence: true })).data;
+            // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
```
ejizba, line 57: How come you have to disable this lint rule now? You added `enableGrouping`, but that doesn't seem related

appservice/src/createAppService/AppServicePlanSkuStep.ts
ejizba, line 24: Rather than check "advancedCreation" here, how about we do `enableGrouping = wizardContext.advancedCreation` below

appservice/src/createAppService/AppServicePlanSkuStep.ts
```diff
@@ -16,31 +16,46 @@ type ExtendedSkuDescription = WebSiteManagementModels.SkuDescription & { label?:
...
+                return {
+                    label: s.label || nonNullProp(s, 'name'),
+                    description: s.description || s.tier,
+                    data: s,
+                    group: 'Additional Options',
+                };
+            }));
+        }
+
         pricingTiers.push({ label: localize('ShowPricingCalculator', '$(link-external) Show pricing information...'), data: undefined, suppressPersistence: true });
```
ejizba, line 53: You should update the dependency of the ui package in package.json because we don't want anyone to use this unless they have your corresponding fix there.

ejizba (commented):
Please post updated screenshots

appservice/src/createAppService/AppServicePlanSkuStep.ts
neelip: I am getting a "Unsafe assignment of an any value." lint error.

appservice/src/createAppService/AppServicePlanSkuStep.ts
neelip: That's a good idea!

neelip:
**Update**
Basic:
image
Advanced: 
image

nturinski:
In the screenshot, it looks like there's still a blank group.  Weren't we trying to get rid of all of those?

appservice/src/createAppService/AppServicePlanSkuStep.ts
```diff
@@ -16,31 +16,47 @@ type ExtendedSkuDescription = WebSiteManagementModels.SkuDescription & { label?:
...
-            skus.push(...this.getElasticPremiumSkus());
-        } else if (wizardContext.newSiteKind?.includes(AppKind.workflowapp)) {
-            skus = this.getWorkflowStandardSkus();
-        }
-
+        let recommendedSkus = this.getRecommendedSkus();
         const regExp: RegExp | undefined = wizardContext.planSkuFamilyFilter;
         if (regExp) {
-            skus = skus.filter(s => !s.family || regExp.test(s.family));
+            recommendedSkus = recommendedSkus.filter(s => !s.family || regExp.test(s.family));
```
ejizba, line 22: You've duplicated this family filter logic here and several lines down - just do it once for all skus

appservice/src/createAppService/AppServicePlanSkuStep.ts
```diff
@@ -16,31 +16,47 @@ type ExtendedSkuDescription = WebSiteManagementModels.SkuDescription & { label?:
...
+                };
+            }));
+        }
+
         pricingTiers.push({ label: localize('ShowPricingCalculator', '$(link-external) Show pricing information...'), data: undefined, suppressPersistence: true });
 
         while (!wizardContext.newPlanSku) {
             const placeHolder = localize('pricingTierPlaceholder', 'Select a pricing tier');
-            wizardContext.newPlanSku = (await wizardContext.ui.showQuickPick(pricingTiers, { placeHolder, suppressPersistence: true })).data;
+            wizardContext.newPlanSku = <WebSiteManagementModels.SkuDescription | undefined>(await wizardContext.ui.showQuickPick(pricingTiers, { placeHolder, suppressPersistence: true, enableGrouping: wizardContext.advancedCreation })).data;
```
ejizba, line 59: This cast should not be necessary. The type of `data` should already be `WebSiteManagementModels.SkuDescription | undefined` because the type of `pricingTiers` is `IAzureQuickPickItem<WebSiteManagementModels.SkuDescription | undefined>[]`

appservice/src/createAppService/AppServicePlanSkuStep.ts
```diff
@@ -16,31 +16,47 @@ type ExtendedSkuDescription = WebSiteManagementModels.SkuDescription & { label?:
...
             };
         });
 
+        let advancedSkus: ExtendedSkuDescription[] = [];
+        if (wizardContext.advancedCreation) {
+            advancedSkus.push(...this.getAdvancedSkus());
+            if (regExp) {
+                advancedSkus = advancedSkus.filter(s => !s.family || regExp.test(s.family));
+            }
+            if (wizardContext.newSiteKind === AppKind.functionapp) {
```
ejizba, line 39: Previously the elastic premium and workflow skus were being added regardless of `advancedCreation`, but I don't remember if that matters or not.

As a general note, you should make sure to test both App Service and Functions when modifying code like this that affects both. In particular, I'm curious if the sku prompt still looks right if I chose "Premium" for the hosting plan

appservice/src/createAppService/AppServicePlanSkuStep.ts
neelip: This is to fix the lint error that was occurring before. Either this or getting a "Unsafe assignment of an any value" lint error.

appservice/src/createAppService/AppServicePlanSkuStep.ts
neelip: In order to set different group names for recommended and advanced sku list, I need to add both sku lists to `pricingTiers` separately. Combining the skus early on will not allow me to do this.

appservice/src/createAppService/AppServicePlanSkuStep.ts
ejizba: If I take your branch and remove the cast:
https://github.com/microsoft/vscode-azuretools/commit/5e5ca3559bdb39caecf8537902e992dfb4038f0b

The build passes with no lint error:
https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=34421&view=logs&j=0da5d1d9-276d-5173-c4c4-9d4d4ed14fdb&t=4c091adb-ab1c-50df-3a5f-280100199621

appservice/src/createAppService/AppServicePlanSkuStep.ts
```diff
@@ -12,7 +12,7 @@ import { AppKind, WebsiteOS } from './AppKind';
 import { IAppServiceWizardContext } from './IAppServiceWizardContext';
 import { setLocationsTask } from './setLocationsTask';
 
-type ExtendedSkuDescription = WebSiteManagementModels.SkuDescription & { label?: string; description?: string }
+type ExtendedSkuDescription = WebSiteManagementModels.SkuDescription & { label?: string; description?: string; group?: string }
```
ejizba, line 15: Much better after the refactor, thanks!

appservice/src/createAppService/AppServicePlanSkuStep.ts
```diff
@@ -32,15 +32,16 @@ export class AppServicePlanSkuStep extends AzureWizardPromptStep<IAppServiceWiza
             return {
                 label: s.label || nonNullProp(s, 'name'),
                 description: s.description || s.tier,
-                data: s
+                data: s,
+                group: s.group || wizardContext.advancedCreation ? localize('additionalOptionsLabel', 'Additional Options') : ''
```
ejizba, line 36: I don't think you need to check `advancedCreation` here since enableGrouping will be set to false below anyways

appservice/src/createAppService/AppServicePlanSkuStep.ts
```diff
@@ -67,7 +68,8 @@ export class AppServicePlanSkuStep extends AzureWizardPromptStep<IAppServiceWiza
                 family: 'F',
                 capacity: 1,
                 label: localize('freeLabel', 'Free'),
-                description: localize('freeDescription', 'Try out Azure at no cost')
+                description: localize('freeDescription', 'Try out Azure at no cost'),
+                group: 'Recommended'
```
ejizba, line 72: Make this a `const` and localize

appservice/src/createAppService/AppServicePlanSkuStep.ts
```diff
@@ -67,7 +68,8 @@ export class AppServicePlanSkuStep extends AzureWizardPromptStep<IAppServiceWiza
                 family: 'F',
                 capacity: 1,
                 label: localize('freeLabel', 'Free'),
```
ejizba, line 70: Now that we have "Additional Options" as the label, it feels like we shouldn't include B1 and P1v2 in "Additional Options" since they're not really...additional... (they're already in recommended). However, people won't know the "Recommended" "Basic" is "B1" necessarily, so here is my suggestion:

1. Change the Recommended labels to "Free (F1)", "Basic (B1)" and "Premium (P1v2)"
2. Remove B1 and P1v2 from advanced skus since they're already in recommended

And posting a screenshot would be nice if you do that :)

neelip:
Update:
image

appservice/src/createAppService/AppServicePlanSkuStep.ts
```diff
@@ -59,46 +60,43 @@ export class AppServicePlanSkuStep extends AzureWizardPromptStep<IAppServiceWiza
...
             },
             {
                 name: 'P1v2',
                 tier: 'Premium V2',
                 size: 'P1v2',
                 family: 'Pv2',
                 capacity: 1,
-                label: localize('premiumLabel', 'Premium'),
-                description: localize('premiumDescription', 'Use in production')
+                label: localize('premiumLabel', 'Premium (P1V2)'),
```
ejizba, line 91:
```suggestion
                label: localize('premiumLabel', 'Premium (P1v2)'),
```

So that we match the other picks below like P2v2

ejizba (approved):

PR #922, opened 2021-05-17
Update behavior showQuickPick() - groups without groupName

neelip:
Related #920


ui/src/wizard/AzureWizardUserInput.ts
```diff
@@ -165,12 +165,16 @@ export class AzureWizardUserInput implements IWizardUserInput {
...
-                picks.push(<types.IAzureQuickPickItem<QuickPickGroup>>{
-                    label: `$(chevron-${group.isCollapsed ? 'right' : 'down'}) ${group.name || ''}`,
-                    data: group
-                });
-                if (!group.isCollapsed) {
+                if (!group.name) {
                     picks.push(...group.picks);
+                } else {
+                    picks.push(<types.IAzureQuickPickItem<QuickPickGroup>>{
+                        label: `$(chevron-${group.isCollapsed ? 'right' : 'down'}) ${group.name || ''}`,
```
ejizba, line 172: We can simplify now that you check if it's defined above:

```suggestion
                        label: `$(chevron-${group.isCollapsed ? 'right' : 'down'}) ${group.name}`,
```

ejizba (approved):

PR #925, opened 2021-05-19
allow customize recommendation picks in artifacts/folder combobox when deploying

wangmingliang-ms:
* allow customize recommendations in deployment artifacts/folder combobox
* Rename artifact to app.jar for java se deployment
image

this PR is related to https://github.com/microsoft/vscode-azureappservice/pull/2061

### DIFF
--- a/appservice/src/deploy/getDeployFsPath.ts
+++ b/appservice/src/deploy/getDeployFsPath.ts
@@ -19,7 +19,7 @@ const deploySubpathSetting: string = 'deploySubpath';
  * In App Service, users can deploy specific artifact files (such as .jar) which is handled by selectWorkspaceFile
  * This enforces that the `effectiveDeployFsPath` is currently open in a workspace
  */
-export async function getDeployFsPath(context: IActionContext, target: vscode.Uri | string | AzExtTreeItem | undefined, fileExtensions?: string | string[], recommendations?: IAzureQuickPickItem<string>[]): Promise<IDeployPaths> {
+export async function getDeployFsPath(context: IActionContext, target: vscode.Uri | string | AzExtTreeItem | undefined, fileExtensions?: string | string[], suggestions?: IAzureQuickPickItem<string>[]): Promise<IDeployPaths> {
     let originalDeployFsPath: string | undefined;
     let effectiveDeployFsPath: string | undefined;
     if (target instanceof vscode.Uri) {
@@ -48,8 +48,8 @@ export async function getDeployFsPath(context: IActionContext, target: vscode.Ur
         const selectFolder: string = localize('selectDeployFolder', 'Select the folder to deploy');
 
         originalDeployFsPath = fileExtensions ?
-            await workspaceUtil.selectWorkspaceFile(context, selectFile, fileExtensions, recommendations) :
-            await workspaceUtil.selectWorkspaceFolder(context, selectFolder, recommendations);
+            await workspaceUtil.selectWorkspaceFile(context, selectFile, fileExtensions, suggestions) :
+            await workspaceUtil.selectWorkspaceFolder(context, selectFolder, suggestions);
         effectiveDeployFsPath = await appendDeploySubpathSetting(context, originalDeployFsPath);
     }
 
--- a/appservice/src/deploy/getDeployFsPath.ts
+++ b/appservice/src/deploy/getDeployFsPath.ts
@@ -5,7 +5,7 @@
 
 import * as path from 'path';
 import * as vscode from 'vscode';
-import { AzExtTreeItem, callWithTelemetryAndErrorHandling, DialogResponses, IActionContext, UserCancelledError } from 'vscode-azureextensionui';
+import { AzExtTreeItem, callWithTelemetryAndErrorHandling, DialogResponses, IActionContext, IAzureQuickPickItem, UserCancelledError } from 'vscode-azureextensionui';
 import { ext } from '../extensionVariables';
 import { localize } from '../localize';
 import { isPathEqual, isSubpath } from '../utils/pathUtils';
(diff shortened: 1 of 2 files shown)


appservice/src/deploy/deployJar.ts
```diff
@@ -0,0 +1,31 @@
+/*---------------------------------------------------------------------------------------------
```
nturinski, line 1: Please remove these files/changes related to this PR: https://github.com/microsoft/vscode-azuretools/pull/912.  I don't believe that this PR is dependent on these changes so I'd rather keep them separate for ease of reviewing.

appservice/src/deploy/getDeployFsPath.ts
```diff
@@ -19,7 +19,7 @@ const deploySubpathSetting: string = 'deploySubpath';
  * In App Service, users can deploy specific artifact files (such as .jar) which is handled by selectWorkspaceFile
  * This enforces that the `effectiveDeployFsPath` is currently open in a workspace
  */
-export async function getDeployFsPath(context: IActionContext, target: vscode.Uri | string | AzExtTreeItem | undefined, fileExtensions?: string | string[]): Promise<IDeployPaths> {
+export async function getDeployFsPath(context: IActionContext, target: vscode.Uri | string | AzExtTreeItem | undefined, fileExtensions?: string | string[], recommendations?: IAzureQuickPickItem<string>[]): Promise<IDeployPaths> {
```
nturinski, line 22: It's strange to me that `recommendations` is a `IAzureQuickPickItem[]`. In general, I'd expect that this would accept a `string[]` with the fspath of the recommendations and that it should be converted into quickPickItems by selectWorkspaceFile.

appservice/src/utils/workspace.ts
```diff
@@ -35,18 +35,20 @@ export async function selectWorkspaceFile(context: IActionContext, placeHolder:
...
     if (vscode.workspace.workspaceFolders) {
-        const folderPicks: IAzureQuickPickItem<string | undefined>[] = await Promise.all(vscode.workspace.workspaceFolders.map((f: vscode.WorkspaceFolder) => {
-            return { label: path.basename(f.uri.fsPath), description: f.uri.fsPath, data: f.uri.fsPath };
+        const defaultRecommendations: IAzureQuickPickItem<string | undefined>[] = await Promise.all(vscode.workspace.workspaceFolders.map((f: vscode.WorkspaceFolder) => {
+            return {label: path.basename(f.uri.fsPath), description: f.uri.fsPath, data: f.uri.fsPath};
         }));
-
-        folderPicks.push({ label: localize('azFunc.browse', '$(file-directory) Browse...'), description: '', data: undefined });
-        folder = await context.ui.showQuickPick(folderPicks, { placeHolder });
+        const providedItems: IAzureQuickPickItem<string | undefined>[] = recommendations?.length ? recommendations : defaultRecommendations;
```
nturinski, line 49: This might be up for debate, but I don't like the idea of totally replacing the folderPicks with the recommendations.  They feel less like recommendations if I have to use browse to deploy my current workspace.  I would rather that they get unshifted into the current folderPicks array so that they appear at the top, but the other quickPicks are still available.

Also, since you are using this to recommend maven artifacts, have you considered using the `deploySubpath` setting available in App Service to deploy the `.jar` file rather than implementing this recommendation feature?  Would that accomplish what you are trying to do?

appservice/src/deploy/deployJar.ts
wangmingliang-ms: the corresponding commit has been dropped from this branch.

appservice/src/deploy/getDeployFsPath.ts
wangmingliang-ms: @nturinski  

**I updated to use a "better" word `suggestion` instead of `recommendation`.**

first, `string` is not enough to meet my requirements, because the `label` & `data` & `description` of the final `IAzureQuickPickItem` are different in my case.
e.g. 
image

the view model for the first suggested item is actually:
```
{
  label: 'Maven: com.example:spring-boot:0.0.1-SNAPSHOT:jar'
  description: 'initial\complete\spring-boot-0.0.1-SNAPSHOT:jar'
  data: `$workspace\gs-spring-boot\complete`
}
```

appservice/src/utils/workspace.ts
wangmingliang-ms: **I updated to use a "better" word suggestion instead of recommendation.**

I updated part of the code in the latest commit.
```js
const defaultSuggestions: IAzureQuickPickItem<string | undefined>[] = await Promise.all(vscode.workspace.workspaceFolders.map((f: vscode.WorkspaceFolder) => {
    return { label: path.basename(f.uri.fsPath), description: f.uri.fsPath, data: f.uri.fsPath };
}));
suggestions = suggestions?.length ? suggestions : defaultSuggestions;
const fileChooser = { label: localize('azFunc.browse', '$(file-directory) Browse...'), description: '', data: undefined };
folder = await context.ui.showQuickPick([...suggestions, fileChooser], { placeHolder });
```

let's call the action item with label `Browse...`  -> **fileChooser**.
and I would like to call the other value items(except *fileChooser*) shown directly in dropdown box `suggested items`/`suggestions`. Because **they are there only because we think user is likely to select them when deploying**. Which is quite similar to the suggestions when you google.

**I think it's a bug to always show the path of workspace on suggestion box** even if we have implicitly indicated only file can be selected(by specifying the `fileExtensions` of `getDeployFsPath`).
as you know, both `selectWorkspaceFile` and `selectWorkspaceFolder` finally delegates to `selectWorkspaceItem`. but `selectWorkspaceItem` always add the *path of current workspace* into suggestion list.

**`deploySubpath` is used, but not as you suggested.** A maven project may have multiple modules. user can choose any module to deploy. when user deploy from a maven project, they are likely to deploy any maven module in the project, so the suggestion feature will detect all the deployable maven modules and suggest them to user. 
`deploySubpath` is used to remember the default module to deployed. and a `preDeployTask: package` is added to build artifact(realy jar/war file) from the module.

PR #927, opened 2021-05-20
Only show groups if there's more than one group with a name

ejizba:


ui/src/wizard/AzureWizardUserInput.ts
```diff
@@ -177,8 +174,14 @@ export class AzureWizardUserInput implements IWizardUserInput {
                     }
                 }
             }
-            return <TPick[]>picks;
+        } else {
+            picks = picks.concat(...groups.map(g => g.picks));
```
neelip, line 178: Can't we directly return here without `else`?  Also I think we can do a picks.push() and return that.

neelip (approved):

PR #928, opened 2021-05-20
Fix zip file size showing as 0

ejizba:
Seems like Buffer is not an array so I'm seeing "0 B" every time
Screen Shot 2021-05-20 at 11 02 37 AM

The docs say it should be a string or Buffer in all cases except objectMode, which I don't think is ever really used. Honestly in other repos we just state the type as `string | Buffer` and don't even worry about type checking, but it can't hurt to do it 🤷‍♂️

### DIFF
--- a/appservice/package-lock.json
+++ b/appservice/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "vscode-azureappservice",
-    "version": "0.80.4",
+    "version": "0.80.5",
     "lockfileVersion": 2,
     "requires": true,
     "packages": {
         "": {
             "name": "vscode-azureappservice",
-            "version": "0.80.4",
+            "version": "0.80.5",
             "license": "MIT",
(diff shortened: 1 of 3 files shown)


neelip (changes requested):
could you please post a reference to the doc?

ejizba:
@neelip this is related to Nathan's recent PR https://github.com/microsoft/vscode-azuretools/pull/858. Nathan and Will already have context here, but you're welcome to take a look at that discussion if you want more info.

neelip (approved):

nturinski:
This is bizarre-- I'm almost 100% certain that it was working on my machine and that while maybe Buffer wasn't an array, I was getting an array of Buffers then.  I don't disagree with the change, I think that this is better, but I'm just shocked since I spent a painstaking amount of time testing this for size comparison reasons 👀

PR #933, opened 2021-06-07
Convert one fs-extra to workspace.fs

nturinski:
There were only 3 instances in the UI package where fs-extra is used.  However, the workspace.fs API is severely lacking in features and I was unable to remove two of the calls.

Here is a list of the API calls:
image

- https://github.com/microsoft/vscode-azuretools/blob/main/ui/src/getPackageInfo.ts#L26
For this call, I can get the Uri and open it as a file, but it doesn't ensure it's a JSON and it returns a `Uint8Array` so I'd have to treat it as a Buffer.  Kind of annoying and not nearly as convenient.
- https://github.com/microsoft/vscode-azuretools/blob/main/ui/src/utils/createTemporaryFile.ts#L16
I could definitely fandangle some logic to have similar behavior to ensure, but I feel like that defeats the point of using `ensureFile` in the first place.

Futhermore, the call I did replace ends up looking quite a bit messier that I'm not sure that it's worth converting over just yet.  I'll look into other extensions though, where `readFile` is more commonly used, but thought I'd put this up

### DIFF
--- a/ui/src/BaseEditor.ts
+++ b/ui/src/BaseEditor.ts
@@ -3,7 +3,6 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import * as fse from 'fs-extra';
 import * as path from 'path';
 import * as vscode from 'vscode';
 import { IActionContext } from '..';
@@ -30,7 +29,7 @@ export abstract class BaseEditor<ContextT> implements vscode.Disposable {
     public async showEditor(context: ContextT, sizeLimit?: number /* in Megabytes */): Promise<void> {
         const fileName: string = await this.getFilename(context);
         const resourceName: string = await this.getResourceName(context);
-        this.appendLineToOutput(localize('opening', 'Opening "{0}"...', fileName), {resourceName: resourceName});
+        this.appendLineToOutput(localize('opening', 'Opening "{0}"...', fileName), { resourceName: resourceName });
         if (sizeLimit !== undefined) {
             const size: number = await this.getSize(context);
             if (size > sizeLimit) {
@@ -57,7 +56,7 @@ export abstract class BaseEditor<ContextT> implements vscode.Disposable {
 
     public async dispose(): Promise<void> {
         // eslint-disable-next-line @typescript-eslint/no-misused-promises
-        Object.keys(this.fileMap).forEach(async (key: string) => await fse.remove(path.dirname(key)));
+        Object.keys(this.fileMap).forEach(async key => await vscode.workspace.fs.delete(this.fileMap[key][0].uri));
     }
 
     public async onDidSaveTextDocument(actionContext: IActionContext, globalState: vscode.Memento, doc: vscode.TextDocument): Promise<void> {
@@ -90,9 +89,9 @@ export abstract class BaseEditor<ContextT> implements vscode.Disposable {
     private async updateRemote(context: ContextT, doc: vscode.TextDocument): Promise<void> {
         const filename: string = await this.getFilename(context);
         const resourceName: string = await this.getResourceName(context);
-        this.appendLineToOutput(localize('updating', 'Updating "{0}" ...', filename), {resourceName: resourceName});
+        this.appendLineToOutput(localize('updating', 'Updating "{0}" ...', filename), { resourceName: resourceName });
         const updatedData: string = await this.updateData(context, doc.getText());
-        this.appendLineToOutput(localize('done', 'Updated "{0}".', filename), {resourceName: resourceName});
+        this.appendLineToOutput(localize('done', 'Updated "{0}".', filename), { resourceName: resourceName });
         if (doc.isClosed !== true) {
             const visibleDocument: vscode.TextEditor | undefined = vscode.window.visibleTextEditors.find((ed) => ed.document === doc);
             if (visibleDocument) {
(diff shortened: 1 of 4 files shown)


ui/src/BaseEditor.ts
```diff
@@ -30,7 +29,7 @@ export abstract class BaseEditor<ContextT> implements vscode.Disposable {
     public async showEditor(context: ContextT, sizeLimit?: number /* in Megabytes */): Promise<void> {
         const fileName: string = await this.getFilename(context);
         const resourceName: string = await this.getResourceName(context);
-        this.appendLineToOutput(localize('opening', 'Opening "{0}"...', fileName), {resourceName: resourceName});
+        this.appendLineToOutput(localize('opening', 'Opening "{0}"...', fileName), { resourceName: resourceName });
         if (sizeLimit !== undefined) {
             const size: number = await this.getSize(context);
```
ejizba, line 34: The whole point of using the VS Code fs API is [because](https://github.com/microsoft/vscode/blob/f207b3830717f2c354ea12ead37baff21f33a680/src/vs/vscode.d.ts#L7197) 
> It allows extensions to work with files from the local disk as well as files from remote places, like the remote extension host or ftp-servers.

 which you just can't get from "fs-extra" AFAIK. There are other benefits like "useTrash" during the "delete" method, which is pretty cool IMO. That being said, yes it won't be a simple drop-in replacement for "fs-extra" which has a lot of _extra_ stuff. I think we need more of a "big picture" strategy first rather than trying to replace individual fs-extra calls like this.  Something that lets us benefit from the VS Code API without making all of our code ugly in the process. For example, one potential strategy is to follow the pattern of `AzureUserInput`/`IAzureUserInput`/`TestUserInput` which wraps VS Code APIs and adds some of our own stuff on top

ui/src/utils/workspace.fs.ts
```diff
@@ -0,0 +1,54 @@
...
+            throw err
+        }
+    }
+
+    await workspace.fs.writeFile(file, new Uint8Array());
+}
+
+export async function readJson(path: string, options?: { throws?: boolean }): Promise<unknown> {
+    const jsonUri = Uri.file(path);
+    const dataBuffer = await workspace.fs.readFile(jsonUri);
```
nturinski, line 42: Throws this error if it doesn't exist:

image

ui/src/utils/workspace.fs.ts
```diff
@@ -0,0 +1,54 @@
...
+            await workspace.fs.createDirectory(folder);
+        } else {
+            throw err
+        }
+    }
+
+    await workspace.fs.writeFile(file, new Uint8Array());
+}
+
+export async function readJson(path: string, options?: { throws?: boolean }): Promise<unknown> {
```
nturinski, line 40: `throws` based off of this behavior: https://github.com/jprichardson/node-fs-extra/blob/master/docs/readJson.md

ui/src/utils/workspace.fs.ts
```diff
@@ -0,0 +1,54 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { dirname } from 'path';
+import { FileStat, FileType, Uri, workspace } from 'vscode';
+import { parseError } from '../parseError';
+
+export async function ensureFile(path: string): Promise<void> {
```
nturinski, line 10: Based off of https://github.com/jprichardson/node-fs-extra/blob/master/lib/ensure/file.js

ui/src/utils/workspace.fs.ts
```diff
@@ -0,0 +1,54 @@
...
+    let stats: FileStat | undefined;
+    const file = Uri.file(path);
+    try {
```
ejizba, line 13: So what's your big picture plan? You kinda pushed a bunch of changes without any description/summary. Are we going to copy/paste this code to every repo or is it worth exporting some sort of `AzExtWorkspaceFs` or `AzExtFsExtra` from the UI package?

Also before we go hog wild and do this everywhere, can we do two things:
1. Actually confirm this work is worth it - I quoted the docs for reasons why I think it's better, but I haven't actually tested those features myself
2. Write some unit tests. When we're writing code that will affect every single extension, I just feel like it's a lot easier to write tests and find bugs _before_ we convert every extension. Even if we're just copying/pasting this code, I still like to have one "source" repo with tests

ui/src/utils/workspace.fs.ts
ejizba: I don't think we've ever used the "throws" option, so I don't think it's worth implementing. I feel like it would be more readable for the caller of `readJson` to just wrap their code in a try/catch

ui/src/BaseEditor.ts
```diff
@@ -3,7 +3,6 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import * as fse from 'fs-extra';
 import * as path from 'path';
```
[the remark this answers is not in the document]
ejizba: I went ahead and put up a PR to remove this code: https://github.com/microsoft/vscode-azuretools/pull/934

nturinski:
Closing in favor of https://github.com/microsoft/vscode-azurestorage/pull/986/files/8f9cba8a79eb067e700315f50303f0aedff0b5ef..35eeffafacc583681721fa57cd7dd101ad0fe205

PR #934, opened 2021-06-11
Remove BaseEditor

ejizba:
Only reason I left it in after https://github.com/microsoft/vscode-azuretools/pull/734 was to give a little bit of transition time to `AzExtTreeFileSystem`, but we moved all our repos over a while ago.


nturinski (approved):
😭😂

PR #937, opened 2021-06-19
Add optional message to quick pick when no picks are found

ejizba:


nturinski (approved):
Should we just have a generic message even if it's not provided?  I feel like in general, when there's no quickpicks, it'd be better to have a generic message indicating so rather than just showing an empty quickpick list.

ejizba:
The value in this PR is so that I can tell users _why_ there are no picks (because it's functions v4). The value of a generic message is a lot less clear to me. I can already see there are no picks, I'm not sure I need a message saying "There are no picks".

Also minor nitpick, but if you're going to comment about messaging, please give a suggestion for the text you want

nturinski:
Well, there have been cases specifically in Static Web Apps where no branches show up because it's a new repo.  But from some user's perspective, it seemed like the process had errored or frozen up because it's just blank.  

I can of course now add my own custom message to alert users so it doesn't really matter, but I was thinking maybe something like:

`No quick pick items could be generated.  Please go back a step.`

ejizba:
Yeah I think it's just hard to come up with a generic message that works in all cases. "go back a step" could be interpreted as "go back and retry with the same value", which won't help. In my case, Functions v4 is a setting applied before they even start the wizard so going back a step would never have helped there either.

I think a custom message is the only way to truly reduce confusion

PR #938, opened 2021-06-21
Mask client name in deployment error

wwlorey:


appservice/src/deploy/deploy.ts
```diff
@@ -75,6 +75,7 @@ export async function deploy(client: SiteClient, fsPath: string, context: IDeplo
         ext.outputChannel.appendLog(localize('deployStart', 'Starting deployment...'), { resourceName: client.fullName });
         try {
             if (!context.deployMethod && config.scmType === ScmType.GitHub) {
+                context.valuesToMask.push(client.fullName);
```
ejizba, line 78: This should be tackled on a more generic basis - we want to mask `client.fullName` in as many cases as possible. Here are a few examples of where I've added "more generic" fixes:
https://github.com/microsoft/vscode-azuretools/blob/main/ui/src/registerCommand.ts#L34
https://github.com/microsoft/vscode-azuretools/blob/main/ui/src/tree/AzExtTreeDataProvider.ts#L177

ejizba:
Closing in favor of https://github.com/microsoft/vscode-azuretools/pull/941 and https://github.com/microsoft/vscode-azureappservice/pull/2087

PR #939, opened 2021-06-21
A few improvements to masking

ejizba:
Building off of https://github.com/microsoft/vscode-azuretools/pull/824, here are a few improvements

- Mask basically anything I can see on the `filter` object we get from the account extension, instead of just the subscription id
- Automatically mask emails (we might catch a few non-emails, but I think that's fine)
- Fix a bug if one `valueToMask` is a substring of another. For example, if `valuesToMask` contains both `eric` and `ericjizba`, the old behavior might have resulted in `---jizba` instead of masking the whole thing
- Make sure the vscode telemetry package is on the latest
- Use `addValuesToMaskFromAzureId` during `findTreeItem`, another main entry point for any tree-related commands of ours

### DIFF
--- a/ui/test/.eslintrc.js
+++ b/ui/test/.eslintrc.js
@@ -1,0 +1,8 @@
+module.exports = {
+    "extends": [
+        "../.eslintrc.js"
+    ],
+    "rules": {
+        "@typescript-eslint/no-explicit-any": "off"
+    }
+};
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "vscode-azureextensionui",
-    "version": "0.44.1",
+    "version": "0.44.2",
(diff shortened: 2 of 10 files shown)


ui/test/.eslintrc.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
+    "extends": [
+        "../.eslintrc.js"
+    ],
+    "rules": {
+        "@typescript-eslint/no-explicit-any": "off"
```
neelip, line 6: Do we want turn this off for all files?

neelip (approved):

ui/test/.eslintrc.js
ejizba: I only turned it off for files in the "test" folder. I prefer to have it on for other files

PR #940, opened 2021-06-22
Proactively mask more potential user info

ejizba:
Similar to how I just added masking for all emails, this should mask all guids, ip addresses, and urls.

Open to ideas for more things to mask and/or your thoughts on making this more or less aggressive


nturinski (approved):
I agree with this level of aggression.  This masking only applies to the error messages, right?  So we could still use machine IDs for retention, etc.?

ejizba:
Right, just error messages

PR #941, opened 2021-06-22
Add tree item valuesToMask

ejizba:
Since tree items are core to basically anything we do, adding valuesToMask to the tree item itself should allow us to mask its values for pretty much any action. This means rather than @wwlorey using extension values in https://github.com/microsoft/vscode-azuretools/pull/938, he could mask in the SiteTreeItem instead. Also since the tree item label is the resource name in most cases, we will get a lot of masking for free.

The only case it doesn't help is when we pass along a tree item to a new telemetry event, like [postCreateWebApp](https://github.com/microsoft/vscode-azureappservice/blob/main/src/commands/createWebApp/showCreatedWebAppMessage.ts#L20). I'm still brainstorming how we can address that, but it's not _that_ many cases.

### DIFF
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -208,6 +208,17 @@ export declare abstract class AzExtTreeItem {
     public readonly treeDataProvider: AzExtTreeDataProvider;
 
     /**
+     * Values to mask in error messages whenever an action uses this tree item
+     * NOTE: Some values are automatically masked without the need to add anything here, like the label and parts of the id if it's an Azure id
+     */
+    public readonly valuesToMask: string[];
+
+    /**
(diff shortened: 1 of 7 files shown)


nturinski (approved):

nturinski:
I guess we could keep a global list of all the resource names/ids that we should mask and push them in load/create them?

ejizba:
> I guess we could keep a global list of all the resource names/ids that we should mask and push them in load/create them?

We already have the concept of extension-level things to mask (see [here](https://github.com/microsoft/vscode-azuretools/blob/main/ui/src/masking.ts#L23)), but I worry that could get out of hand quickly. We run the mask logic for every single telemetry event and people may have a ton of resources

PR #946, opened 2021-06-23
Mask all telemetry properties, not just error ones

ejizba:
The error props are definitely the most likely to have potential user information, but there's a small risk other properties might, too. I think there's value in masking all properties, but I definitely want to be less aggressive for non-error props. In particular, I don't want to do the "id" check because we leverage guids in a few places. I'm also skipping the "email" check, but I don't have a particular example for that - it just feels aggressive how we're really only checking for one character ("@").

Finally, I beefed up the url check to cover a few more cases.

### DIFF
--- a/ui/src/masking.ts
+++ b/ui/src/masking.ts
@@ -64,9 +64,7 @@ export async function callWithMaskHandling<T>(callback: () => Promise<T>, valueT
  */
 export function maskUserInfo(data: string, actionValuesToMask: string[], lessAggressive: boolean = false): string {
     // Mask longest values first just in case one is a substring of another
-    let valuesToMask = actionValuesToMask.concat(getExtValuesToMask()).sort((a, b) => b.length - a.length);
-    // de-dupe
-    valuesToMask = valuesToMask.filter((v, index) => valuesToMask.indexOf(v) === index);
+    const valuesToMask = actionValuesToMask.concat(getExtValuesToMask()).sort((a, b) => b.length - a.length);
     for (const value of valuesToMask) {
         data = maskValue(data, value);
     }
--- a/ui/src/masking.ts
+++ b/ui/src/masking.ts
@@ -60,8 +60,9 @@ export async function callWithMaskHandling<T>(callback: () => Promise<T>, valueT
 
 /**
  * Best effort to mask all data that could potentially identify a user
+ * @param lessAggressive If set to true, the most aggressive masking will be skipped
  */
-export function maskUserInfo(data: string, actionValuesToMask: string[]): string {
+export function maskUserInfo(data: string, actionValuesToMask: string[], lessAggressive: boolean = false): string {
     // Mask longest values first just in case one is a substring of another
     let valuesToMask = actionValuesToMask.concat(getExtValuesToMask()).sort((a, b) => b.length - a.length);
     // de-dupe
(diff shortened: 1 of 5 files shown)


ui/src/masking.ts
```diff
@@ -60,19 +60,24 @@ export async function callWithMaskHandling<T>(callback: () => Promise<T>, valueT
...
-    data = data.replace(/\S+@\S+/gi, getRedactedLabel('email'));
-    data = data.replace(/\b[0-9a-f\-\:\.]{4,}\b/gi, getRedactedLabel('id')); // should cover guids, ip addresses, etc.
-    data = data.replace(/http(s|):\/\/\S*/gi, getRedactedLabel('url'));
+    if (!lessAggressive) {
+        data = data.replace(/\S+@\S+/gi, getRedactedLabel('email'));
+        data = data.replace(/\b[0-9a-f\-\:\.]{4,}\b/gi, getRedactedLabel('id')); // should cover guids, ip addresses, etc.
+    }
+
+    data = data.replace(/[a-z]+:\/\/\S*/gi, getRedactedLabel('url'));
+    data = data.replace(/\S*\.(com|org|net)\S*/gi, getRedactedLabel('url'));
```
neelip, line 78: There might be other possible cases in this list, such as 'in/eu/au/ca...'. I think it would be better to make it more generic.

ui/src/masking.ts
```diff
@@ -60,19 +60,24 @@ export async function callWithMaskHandling<T>(callback: () => Promise<T>, valueT
 
 /**
  * Best effort to mask all data that could potentially identify a user
+ * @param lessAggressive If set to true, the most aggressive masking will be skipped
```
neelip, line 63: nit: maybe include a little bit more detail here, i.e. "skipping email and id if true" or something in those lines. lessAggressive and most aggressive read a little vague.

ejizba: Well, I want to keep it a big vague so we don't have to update the comment every time the logic changes

ui/src/masking.ts
ejizba: Do you have a suggestion for a regex? I realize there are other domains, but it's a balance of how aggressive we want to be. A lot of urls will be caught by the line above (assuming they have a scheme) and things like [this](https://github.com/microsoft/vscode-azurefunctions/pull/2866/files#diff-09dcf08abf7ed01ce8e63ec67340384757790b6881bfa36b5f8eb5a818dcba55R51) will also help.

neelip: Also, I think urls in this format are usually public, in case they do appear in a error message or other wise, and I am guessing should be ok if not redacted.

neelip: I found a regex that covers a good bit of cases but I don't think it's necessary in this case. 
If interested, https://www.regextester.com/96504

neelip (approved):

ui/src/masking.ts
ejizba: > Also, I think urls in this format are usually public, in case they do appear in a error message or other wise, and I am guessing should be ok if not redacted.

Yeah the main case I'm worried about are Azure-specific urls for a user's resource, like this:

storage: `.blob.core.windows.net`
app service: .`azurewebsites.net`
cosmos: `.documents.azure.com`

The full url can often be used to identify a user (which we want to avoid), whether or not it's technically "public".

> I found a regex that covers a good bit of cases but I don't think it's necessary in this case.

Yeah I found a lot of regex's online, but they all seemed to be focused on _validating_ a url. We don't really need to validate it, we just need a quick check to see if it's _probably_ a url, hence we I like the simpler regex's

PR #950, opened 2021-07-03
Use action-specific user input in appservice package

ejizba:
Similar to https://github.com/microsoft/vscode-azurestorage/pull/987

### DIFF
--- a/appservice/src/startStreamingLogs.ts
+++ b/appservice/src/startStreamingLogs.ts
@@ -45,8 +45,8 @@ export async function startStreamingLogs(context: IActionContext, client: SiteCl
 
         return await new Promise((onLogStreamCreated: (ls: ILogStream) => void): void => {
             // Intentionally setting up a separate telemetry event and not awaiting the result here since log stream is a long-running action
-            void callWithTelemetryAndErrorHandling('appService.streamingLogs', async (context: IActionContext) => {
-                context.errorHandling.suppressDisplay = true;
+            void callWithTelemetryAndErrorHandling('appService.streamingLogs', async (streamContext: IActionContext) => {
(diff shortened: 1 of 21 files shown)


nturinski (approved):

appservice/src/startStreamingLogs.ts
```diff
@@ -26,12 +26,12 @@ function getLogStreamId(client: SiteClient, logsPath: string): string {
...
+export async function startStreamingLogs(context: IActionContext, client: SiteClient, verifyLoggingEnabled: () => Promise<void>, logStreamLabel: string, logsPath: string = ''): Promise<ILogStream> {
     const logStreamId: string = getLogStreamId(client, logsPath);
     const logStream: ILogStream | undefined = logStreams.get(logStreamId);
     if (logStream && logStream.isConnected) {
         logStream.outputChannel.show();
-        void ext.ui.showWarningMessage(localize('logStreamAlreadyActive', 'The log-streaming service for "{0}" is already active.', logStreamLabel));
+        void context.ui.showWarningMessage(localize('logStreamAlreadyActive', 'The log-streaming service for "{0}" is already active.', logStreamLabel));
         return logStream;
     } else {
         await verifyLoggingEnabled();
```
nturinski, line 37: I can't add a comment on line 48, but do you not need to change the `context` name here as well?  Sort of like how you did here: https://github.com/microsoft/vscode-azuretools/pull/950/files#diff-e9047f508bb43125f06a6038716be014c3a833be7736051e70366cddf865e89fR36

ejizba: We don't strictly _need_ to, but yes it's a good idea to do that

PR #952, opened 2021-07-07
Leverage onActionStart handlers for test runWithInputs

ejizba:
Related to https://github.com/microsoft/vscode-azuretools/pull/949, this is how I will leverage the onActionStart handler to set test user inputs

The functions [test case](https://github.com/microsoft/vscode-azurefunctions/blob/main/test/api.test.ts#L26) will end up looking like this, instead of using the extension-wide `testUserInput` which has been removed:
```typescript
await runWithInputs('api.createFunction', [language], registerOnActionStartHandler, async () => {
    await api.createFunction({
        folderPath,
        functionName,
        templateId: 'HttpTrigger',
        languageFilter: /^(Java|Type)Script$/i,
        functionSettings: { authLevel: 'anonymous' },
        suppressCreateProjectPrompt: true,
        suppressOpenFolder: true
    });
});
```

NOTE: I have to pass `registerOnActionStartHandler` in because the ui packages relies on the dev package, so I can't have the dev package rely on the ui package or I get circular dependencies.

### DIFF
--- a/dev/src/TestUserInput.ts
+++ b/dev/src/TestUserInput.ts
@@ -28,15 +28,27 @@ export class TestUserInput implements types.TestUserInput {
         this._onDidFinishPromptEmitter = new this._vscode.EventEmitter<types.PromptResult>();
     }
 
+    public static async create(): Promise<TestUserInput> {
+        return new TestUserInput(await import('vscode'));
+    }
+
     public get onDidFinishPrompt(): vscodeTypes.Event<types.PromptResult> {
         return this._onDidFinishPromptEmitter.event;
     }
 
     public async runWithInputs<T>(inputs: (string | RegExp | types.TestInput)[], callback: () => Promise<T>): Promise<T> {
-        this._inputs = <(string | RegExp | TestInput)[]>inputs;
+        this.setInputs(inputs);
         const result: T = await callback();
-        assert.equal(this._inputs.length, 0, `Not all inputs were used: ${this._inputs.toString()}`);
+        this.validateAllInputsUsed();
         return result;
+    }
+
+    public setInputs(inputs: (string | RegExp | types.TestInput)[]): void {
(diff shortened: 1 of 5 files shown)


wwlorey (approved):

dev/src/TestUserInput.ts
```diff
@@ -147,3 +159,25 @@ export class TestUserInput implements types.TestUserInput {
         return result;
     }
 }
+
+
+export async function runWithInputs<T>(callbackId: string, inputs: (string | RegExp | types.TestInput)[], registerOnActionStartHandler: types.registerOnActionStartHandlerType, callback: () => Promise<T>): Promise<T> {
+    const testUserInput = await TestUserInput.create();
+    testUserInput.setInputs(inputs);
+    const disposable = registerOnActionStartHandler((context) => {
+        if (context.callbackId === callbackId) {
```
nturinski, line 168: Just want to make sure I understand what's going on here.

So because `registerOnActionStartHandler` will let the handlers run in an arbitrary order, we're checking `callbackId` to ensure that this is the correct context to set the inputs for?

What will happen if it's not the same `callbackId`?  Will it just drop through and try to run the test anyway?

ejizba: No this doesn't have to do with multiple handlers, it has to do with multiple callbacks happening around the same time. Until this handler is disposed, every call to `callWithTelemetryAndErrorHandling` will also call this handler. If it's the expected callback id, we'll set the test user input and dispose the handler because we know we're done. If it's not, it'll just use whatever ui it already has and this does nothing.

For example, a single action can have multiple callbacks. The deploy action has at least 'azureFunctions.deploy' and 'postDeploy'. Plus there might be other random callbacks that happen like 'azureFunctions.validateFuncCoreToolsInstalled'

PR #957, opened 2021-07-12
Use `lastStep` in app service package

ejizba:
Depends on https://github.com/microsoft/vscode-azuretools/pull/956

### DIFF
--- a/appservice/src/deploy/getDeployFsPath.ts
+++ b/appservice/src/deploy/getDeployFsPath.ts
@@ -127,8 +127,7 @@ function getContainingWorkspace(context: IActionContext, fsPath: string): vscode
             });
         });
 
-        context.telemetry.properties.cancelStep = 'openInNewWindow';
-        throw new UserCancelledError();
+        throw new UserCancelledError('openInNewWindow');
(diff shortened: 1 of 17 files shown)


appservice/src/deploy/getDeployFsPath.ts
```diff
@@ -127,8 +127,7 @@ function getContainingWorkspace(context: IActionContext, fsPath: string): vscode
             });
         });
 
-        context.telemetry.properties.cancelStep = 'openInNewWindow';
```
neelip, line 130: 'cancelStep' is reset to undefined in some places but just removed in others?

neelip (approved):

appservice/src/deploy/getDeployFsPath.ts
ejizba: Well, you're describing the old behavior and yes it was a bit weird/inconsistent. The new behavior does not use `cancelStep` at all

neelip: oh right, didn't notice.

Issue #805, opened 2020-09-24
TypeScript 4 error: accessors overriding properties in AzExt*TreeItem


bwateratmsft:
Thanks to a [new error](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-0.html#properties-overriding-accessors-and-vice-versa-is-an-error) in TypeScript 4, many errors get thrown when trying to build `vscode-azureextensionui` (or, a consumer of its `TreeView` framework) against TS 4.0.

![image](https://user-images.githubusercontent.com/36966225/94154384-cfc4b980-fe4b-11ea-939f-e72d2dded25b.png)

I think fixing this would require defining these "properties" as abstract, which means consumers would have to implement them (which would make this a breaking change). If they are abstract, the implementer can implement them either as properties or as accessors.

bwateratmsft:
@EricJizbaMSFT I'm happy to help with the fix here. Do you have any thoughts on how to go about it? I think that changing the class declarations to interfaces would work, but my understanding of libraries / `declare` / etc. is not very strong.

EricJizbaMSFT:
Yeah we don't have any immediate plans to upgrade to TypeScript 4, so I probably wouldn't get to this any time soon. You're welcome to take a shot at it and feel free to make breaking changes if necessary

bwateratmsft:
I got started and immediately realized this is going to be a nightmare. There are probably 50+ files in the Docker extension alone that would need changes...

bwateratmsft:
I found a somewhat less-intrusive solution that unblocked Docker without needing changes here: https://github.com/microsoft/vscode-docker/pull/2348

In short, I have intermediate abstract classes that redefine the parent (`AzExtTreeItem`, `AzExtParentTreeItem`) properties as abstracts, allowing inheriting classes to implement the abstract as either a property or an accessor. I don't think that solution is applicable to this repo though.

PR #958, opened 2021-07-13
Upgrade ui to Typescript 4

nturinski:
Fixes https://github.com/microsoft/vscode-azuretools/issues/805

See the changes from TypeScript 3 to 4: https://devblogs.microsoft.com/typescript/announcing-typescript-4-0/

TS 4 enforces that a property can't overwrite an accessor.  Considering a lot of treeItems use `get __()` to retrieve a dynamic icon path, id, and description, I figured it made more sense to have the base classes use accessors than trying to rewrite a lot of our code to have a way to retrieve a property dynamically.

~~I tested my changes on Functions, however I can't verify because the new appservice package relies on the UI package, so I think I'm getting a circular dependency.~~

Also just fixed some spelling errors I saw along the way.

Edit: Also, is it worth specifying all errors in catch blocks as unknown?  Apparently that's an option now, but I'm not sure if we had any type-safety issues.

Edit edit: Never mind, I see why it wasn't working on Functions.  Hooray tests! 🥳

### DIFF
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -188,10 +188,15 @@ export declare abstract class AzExtTreeItem {
     /**
      * Additional information about a tree item that is appended to the label with the format `label (description)`
      */
-    public description?: string;
+    public set description(description: string | undefined);
+    public get description(): string | undefined;
+
     public set iconPath(iconPath: TreeItemIconPath | undefined);
     public get iconPath(): TreeItemIconPath | undefined;
-    public commandId?: string;
+
+    public set commandId(id: string | undefined);
+    public get commandId(): string | undefined;
+
     public tooltip?: string;
 
     /**
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -181,14 +181,16 @@ export declare abstract class AzExtTreeItem {
      *
      * If not provided, an id is generated using the treeItem's label. **Note** that when labels change, ids will change and that selection and expansion state cannot be kept stable anymore.
      */
-    public id?: string;
+    public set id(id: string | undefined)
+    public get id(): string;
     public abstract label: string;
 
     /**
      * Additional information about a tree item that is appended to the label with the format `label (description)`
      */
     public description?: string;
-    public iconPath?: TreeItemIconPath;
+    public set iconPath(iconPath: TreeItemIconPath | undefined);
+    public get iconPath(): TreeItemIconPath | undefined;
     public commandId?: string;
     public tooltip?: string;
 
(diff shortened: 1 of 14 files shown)


ui/index.d.ts
```diff
@@ -181,15 +181,20 @@ export declare abstract class AzExtTreeItem {
      *
      * If not provided, an id is generated using the treeItem's label. **Note** that when labels change, ids will change and that selection and expansion state cannot be kept stable anymore.
      */
-    public id?: string;
+    public set id(id: string | undefined)
```
ejizba, line 184: Seems like Brandon got it to work by making them abstract instead of forcing people to use accessors

https://github.com/microsoft/vscode-docker/blob/main/src/tree/AzExtTreeItemIntermediate.ts#L10

ui/index.d.ts
```diff
@@ -471,7 +479,7 @@ export declare abstract class AzureTreeItem<TRoot extends ISubscriptionContext =
     /**
      * Contains subscription information specific to the root of this branch of the tree.
      */
-    public readonly root: TRoot;
```
nturinski, line 474: Having only a get method should behave as readonly.  Also, the implemtation was already using a get accessor: https://github.com/microsoft/vscode-azuretools/blob/main/ui/src/tree/AzureTreeItem.ts#L15

ui/src/tree/AzExtTreeItem.ts
```diff
@@ -44,6 +46,14 @@ export abstract class AzExtTreeItem implements types.AzExtTreeItem {
         return this._temporaryDescription || this.description;
     }
 
+    public get id(): string {
+        return this._id || this.label;
```
ejizba, line 50: Let's not do `|| this.label` here since it's handled elsewhere and since this accessor may get overridden by child classes.

The return type should be `string | undefined` like all the other optional properties

ejizba (approved):

ui/index.d.ts
```diff
@@ -178,19 +178,25 @@ export declare abstract class AzExtTreeItem {
     /**
      * This is is used for the AzureTreeItem.openInPortal action. It is also used per the following documentation copied from VS Code:
      * Optional id for the treeItem that has to be unique across tree. The id is used to preserve the selection and expansion state of the treeItem.
-     *
-     * If not provided, an id is generated using the treeItem's label. **Note** that when labels change, ids will change and that selection and expansion state cannot be kept stable anymore.
```
nturinski, line 182: It makes sense to get rid of this comment then, right?  Since the label is being added only in `fullId`

ejizba: No we should keep this line. First of all it's copied straight from VS Code docs (as line 179 states): https://github.com/microsoft/vscode/blob/main/src/vs/vscode.d.ts#L9307

Secondly, they never say this specific property will return the label. They say "_an_ id is generated", meaning some under-the-covers id will use the label. Also, they say "if not provided", meaning this property can return undefined.

PR #962, opened 2021-07-21
Expose getScheme to use prior to opening read-only files

nturinski:
I be scheming.

I need the scheme to provide to `registerFoldingRangerProvider` before I start opening files.  Since it's random, I needed a way to generate the scheme before I start opening files so that I can provide the folding ranges for the GitHub log files.

### DIFF
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -1616,3 +1616,9 @@ export declare abstract class AzExtTreeFileSystem<TItem extends AzExtTreeItem> i
  * Registers a command that will prompt users with a list of issues they can report from that session of VS Code
  */
 export function registerReportIssueCommand(commandId: string): void;
+
+/**
+ * Gets or generates a unique scheme so that multiple extensions using this same code don't conflict with each other
+ * Use to provide scheme for various TextDocument providers
+ */
+export function getOrGenerateContentScheme(): string;
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "vscode-azureextensionui",
-    "version": "0.46.0",
+    "version": "0.46.1",
     "lockfileVersion": 2,
     "requires": true,
     "packages": {
         "": {
             "name": "vscode-azureextensionui",
-            "version": "0.46.0",
+            "version": "0.46.1",
             "license": "MIT",
             "dependencies": {
                 "@azure/arm-resources": "^3.0.0",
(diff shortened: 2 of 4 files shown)


ui/index.d.ts
```diff
@@ -1616,3 +1616,9 @@ export declare abstract class AzExtTreeFileSystem<TItem extends AzExtTreeItem> i
  * Registers a command that will prompt users with a list of issues they can report from that session of VS Code
  */
 export function registerReportIssueCommand(commandId: string): void;
+
+/**
+ * Gets or generates a unique scheme so that multiple extensions using this same code don't conflict with each other
+ * Use to provide scheme for various TextDocument providers
+ */
+export function getOrGenerateContentScheme(): string;
```
ejizba, line 1624: Can I see how you're using this? This isn't something I ever thought would be exported from this package. It's very much an internal implementation detail

nturinski: Yeah, so I'm registering it here (on activate): https://github.com/microsoft/vscode-azurestaticwebapps/blob/nat/actionLog/src/extension.ts#L36

For this FoldingProvider https://github.com/microsoft/vscode-azurestaticwebapps/blob/nat/actionLog/src/commands/github/GitHubLogFoldingProvider.ts

For any file opened here: https://github.com/microsoft/vscode-azurestaticwebapps/blob/nat/actionLog/src/commands/github/openActionLog.ts#L26

nturinski: I think the alternative would be to implement my own version of `openReadOnlyContent` within SWA

ejizba: I don't like re-implementing `openReadOnlyContent`

The `DocumentSelector` you pass to `registerFoldingRangeProvider` doesn't have to filter by scheme. What if it filters by file path and you add some unique string to the file path for github logs? That's theoretically better in the sense that the `FoldingProvider` shouldn't actually apply to _all_ readonly content, just the github logs. In practice I admit it probably doesn't matter since other readonly content (like "View Properties") probably won't have the regions anyways.

Let me know if that doesn't work for some reason. If we end up going with your current PR, I would like to "hide" some of the internal details (like the fact that we generate the scheme) and do something a bit more generic like this:
```
/**
 * Gets the scheme used for read only content
 */
export function getReadonlyContentScheme(): string;
```
Also, it should be put right next to `openReadOnlyContent` in `index.d.ts`

nturinski:
Okay, I was able to make it work with the glob pattern stuff.  Will go ahead and close this then.

PR #963, opened 2021-07-23
Add correlation id to all requests

ejizba:
For every action, we will generate a unique guid, add that guid to our telemetry, and send it in all requests for that action as the `x-ms-correlation-request-id` header. This was a request from the app service team, but in general it will help us coordinate telemetry/issues across the tooling and platform teams.

In order to make sure a unique id is used per action, I had to do some refactoring:
1. Always take in `IActionContext` when creating a client.
2. Make sure clients aren't cached and used across multiple actions. This happens the most often in relation to `AzureTreeItem`/`AzureParentTreeItem` with their `root` property (especially in app service where we cache `SiteClient`) and I've been wanting to get rid of those classes for a while now so I went ahead and removed them.


ui/src/wizard/LocationListStep.ts
```diff
@@ -167,9 +167,9 @@ function generalizeLocationName(name: string | undefined): string {
     return (name || '').toLowerCase().replace(/[^a-z0-9]/gi, '');
 }
 
-async function getAllLocations(wizardContext: types.ISubscriptionContext): Promise<types.AzExtLocation[]> {
+async function getAllLocations(wizardContext: types.ISubscriptionActionContext): Promise<types.AzExtLocation[]> {
     // NOTE: Using a generic client because the subscriptions sdk is pretty far behind on api-version
-    const client = await createGenericClient(wizardContext);
+    const client = await createGenericClient(wizardContext, wizardContext);
```
neelip, line 172: calling createGenericClient() with wizardContext for both parameters?

neelip (approved):

ui/src/createAzureClient.ts
```diff
@@ -3,52 +3,62 @@
...
 import { maskValue } from './masking';
 import { parseError } from './parseError';
+import { AzExtTreeItem } from './tree/AzExtTreeItem';
 import { parseJson, removeBom } from './utils/parseJson';
 
-export function createAzureClient<T>(
-    clientInfo: { credentials: types.AzExtServiceClientCredentials; subscriptionId: string; environment: Environment; },
-    clientType: new (credentials: types.AzExtServiceClientCredentials, subscriptionId: string, options?: types.IMinimumServiceClientOptions) => T): T {
-    return new clientType(clientInfo.credentials, clientInfo.subscriptionId, {
+export type InternalAzExtClientContext = types.ISubscriptionActionContext | [types.IActionContext, types.ISubscriptionContext | AzExtTreeItem];
```
nturinski, line 18: I don't totally understand why we export the `InternalAzExtClientContext`, but not `type AzExtClientContext = ISubscriptionActionContext | [IActionContext, ISubscriptionContext | AzExtTreeItem];`  Why do we even need both since it doesn't seem like `AzExtClientContext` is exposed?

ui/src/tree/AzExtTreeItem.ts
```diff
@@ -116,6 +117,15 @@ export abstract class AzExtTreeItem implements types.AzExtTreeItem {
         this._commandId = id;
     }
 
+    public get subscription(): types.ISubscriptionContext {
+        const result = this._subscription || this.parent?.subscription;
+        if (!result) {
+            throw Error(localize('noSubscriptionFound', 'No subscription found for this tree item.'));
```
nturinski, line 123: I think it wouldn't hurt to specify Azure subscription for clarity since Static Web Apps also leverage GitHub.

nturinski (approved):

ui/src/createAzureClient.ts
ejizba: The difference between `InternalAzExtClientContext` and `AzExtClientContext` is that the former references the actual `AzExtTreeItem` class in the `src/tree` folder and the latter references the class declaration from index.d.ts. I need to do this otherwise I get build errors related to `instanceof AzExTreeItem`

> it doesn't seem like AzExtClientContext is exposed?

That's...interesting, I assume you mean because I don't use `export` in index.d.ts? That was not intentional, and it actually still gets exported and can be referenced outside of this package. Maybe that's a bug in TypeScript? Lol anyways, I can add the `export` to make it explicit - there's a few others missing it as well.

PR #964, opened 2021-07-23
Support correlation id in app service package

ejizba:
Depends on https://github.com/microsoft/vscode-azuretools/pull/963

As mentioned in that PR, we can't cache a `client` anymore because then the correlation id isn't unique per action. My solution was to break off a big chunk of `SiteClient` into a new class `ParsedSite` which has all the static information plus a new `createClient` method used to create the client once per action.


appservice/package-lock.json
```diff
@@ -26,7 +26,7 @@
                 "portfinder": "^1.0.25",
                 "pretty-bytes": "^5.3.0",
                 "simple-git": "1.132.0",
-                "vscode-azureextensionui": "^0.46.0",
+                "vscode-azureextensionui": "file:../ui/vscode-azureextensionui-0.47.0.tgz",
```
neelip, line 29: Is this supposed to be set to local?

ejizba: I will update it as soon as 0.47.0 is released. That is why I said "Depends on #963" in the PR description. Also the PR build will fail until I fix it, so there's no risk of merging it on accident.

neelip (approved):

PR #968, opened 2021-07-27
Fix suppressPersistence bug

ejizba:
We take `suppressPersistence` into account when setting the cache for 'recently used', but not the next time around when reading the cache

See https://github.com/microsoft/vscode-azureappservice/pull/2100 for the use-case that inspired this

### DIFF
--- a/ui/src/userInput/AzExtUserInput.ts
+++ b/ui/src/userInput/AzExtUserInput.ts
@@ -3,7 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { Event, EventEmitter, MessageItem, QuickPickItem, Uri } from 'vscode';
+import { Event, EventEmitter, MessageItem, Uri } from 'vscode';
 import * as types from '../../index';
 import { IInternalActionContext, IInternalAzureWizard } from './IInternalActionContext';
 import { showInputBox } from './showInputBox';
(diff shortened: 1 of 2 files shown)


ui/src/userInput/AzExtUserInput.ts
```diff
@@ -24,7 +24,7 @@ export class AzExtUserInput implements types.IAzureUserInput {
         return this._onDidFinishPromptEmitter.event;
     }
 
-    public async showQuickPick<TPick extends QuickPickItem>(picks: TPick[] | Promise<TPick[]>, options: types.IAzureQuickPickOptions): Promise<TPick | TPick[]> {
+    public async showQuickPick<TPick extends types.IAzureQuickPickItem<unknown>>(picks: TPick[] | Promise<TPick[]>, options: types.IAzureQuickPickOptions): Promise<TPick | TPick[]> {
```
wwlorey, line 27: Could you explain the `QuickPickItem` ->  `types.IAzureQuickPickItem<unknown>` changes?

ejizba: Inside `bumpRecentlyUsedPick` I needed to access `suppressPersistence`, but it only exists on our interface `IAzureQuickPickItem`, not VS Code's `QuickPickItem`. I could've used a cast, but I'm not a fan of casting. So I changed the type in that method and kept changing it up the stack until all the build errors went away

Also I just realized I could get rid of some old casts due to my changes so I just cleaned that up.

nturinski (approved):

PR #974, opened 2021-08-03
Fix wrong deployment method for linux java function app

Flanker32:
We used to check java se runtime by check whether `linuxFxVersion` or `javaContainer` start with `java`, and do correspond deployment in this case. However, Linux Java Function App also fit the condition, so filter out function app cases in this PR.

fixes https://github.com/microsoft/vscode-azuretools/issues/975

### DIFF
--- a/appservice/src/deploy/deploy.ts
+++ b/appservice/src/deploy/deploy.ts
@@ -87,7 +87,7 @@ export async function deploy(site: ParsedSite, fsPath: string, context: IDeployC
                 const javaRuntime = site.isLinux ? config.linuxFxVersion : config.javaContainer;
                 if (javaRuntime && /^(tomcat|wildfly|jboss)/i.test(javaRuntime)) {
                     await deployWar(context, site, fsPath);
-                } else if (javaRuntime && /^java/i.test(javaRuntime)) {
+                } else if (javaRuntime && /^java/i.test(javaRuntime) && !site.isFunctionApp) {
                     const pathFileMap = new Map<string, string>([
                         [path.basename(fsPath), 'app.jar']
                     ]);


appservice/src/deploy/deploy.ts
```diff
@@ -87,7 +87,7 @@ export async function deploy(site: ParsedSite, fsPath: string, context: IDeployC
                 const javaRuntime = site.isLinux ? config.linuxFxVersion : config.javaContainer;
                 if (javaRuntime && /^(tomcat|wildfly|jboss)/i.test(javaRuntime)) {
                     await deployWar(context, site, fsPath);
-                } else if (javaRuntime && /^java/i.test(javaRuntime)) {
+                } else if (javaRuntime && /^java/i.test(javaRuntime) && !site.isFunctionApp) {
```
nturinski, line 90: Would we want this behavior if the function app was on a premium plan?  If so, we need to check the service plan

Flanker32: For premium plan(`EP`), `deployToStorageAccount` is the recommended deployment method by service team, we may add the validation in function extension deployment [link](https://github.com/microsoft/vscode-azurefunctions/blob/main/src/commands/deploy/deploy.ts#L68)

nturinski: Based off of this, would we not want to utilize the logic for Fuctions on premium plans where the basename of the deployment path is changed to app.jar in the zipfile?

Flanker32: We should not apply the deployment of java se web app to function. As rename to app.jar is only the requirement for web app, and rename the artifact will break the configuration in `function.json`. Besides premium plans should leverage the storage deployment.

nturinski (approved):

PR #977, opened 2021-08-06
Updates for the newer python debugging

tonybaloney:
The Python debugger uses a new configuration structure for mapping ports and paths.

Also, the path mapping `/home/wwwroot/` is invalid, because the runtime path is `/tmp/<hash>`, so using `.` instead should provide automated mapping from the cwd.

I've split the function out so that the configuration for nodeJS is the same

### DIFF
--- a/appservice/package.json
+++ b/appservice/package.json
@@ -60,7 +60,7 @@
         "@types/mocha": "^7.0.2",
         "@types/node": "^12.0.0",
         "@types/p-retry": "^2.0.0",
-        "@types/vscode": "1.59.0",
+        "@types/vscode": "1.53.0",
         "@types/websocket": "^1.0.0",
         "@types/yazl": "^2.4.2",
         "@typescript-eslint/eslint-plugin": "^4.28.3",
--- a/appservice/package.json
+++ b/appservice/package.json
@@ -60,7 +60,7 @@
         "@types/mocha": "^7.0.2",
         "@types/node": "^12.0.0",
         "@types/p-retry": "^2.0.0",
-        "@types/vscode": "1.53.0",
+        "@types/vscode": "1.59.0",
         "@types/websocket": "^1.0.0",
         "@types/yazl": "^2.4.2",
         "@typescript-eslint/eslint-plugin": "^4.28.3",
--- a/appservice/src/remoteDebug/startRemoteDebug.ts
+++ b/appservice/src/remoteDebug/startRemoteDebug.ts
@@ -6,7 +6,6 @@
 import { WebSiteManagementModels } from '@azure/arm-appservice';
 import * as portfinder from 'portfinder';
 import * as vscode from 'vscode';
-import * as path from 'path';
 import { callWithTelemetryAndErrorHandling, IActionContext } from 'vscode-azureextensionui';
 import { localize } from '../localize';
 import { ParsedSite } from '../SiteClient';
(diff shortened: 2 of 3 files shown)


karthiknadig (approved):

appservice/package.json
```diff
@@ -60,7 +60,7 @@
         "@types/mocha": "^7.0.2",
         "@types/node": "^12.0.0",
         "@types/p-retry": "^2.0.0",
-        "@types/vscode": "1.53.0",
+        "@types/vscode": "1.59.0",
```
ejizba, line 63: Why has this changed?

appservice/src/remoteDebug/startRemoteDebug.ts
```diff
@@ -124,3 +110,38 @@ async function getDebugConfiguration(context: IActionContext, language: RemoteDe
...
+        name: sessionId,
+        type: 'python',
+        request: 'attach',
+        connect: {
+            host: host,
+            port: portNumber,
+        },
+        pathMappings: [
+            {
+                localRoot: '${workspaceFolder}',
```
ejizba, line 125: This doesn't work for multi-root workspaces. The logic around `workspaceFolders` which you put into `getNodeDebugConfiguration` should still be used for both node _and_ python
Screen Shot 2021-08-10 at 11 02 08 AM

appservice/src/remoteDebug/startRemoteDebug.ts
```diff
@@ -124,3 +110,38 @@ async function getDebugConfiguration(context: IActionContext, language: RemoteDe
 
     return config;
 }
+
+function getPythonDebugConfiguration(sessionId: string, portNumber: number, host: string): vscode.DebugConfiguration {
```
ejizba, line 114: Node.js remote debugging worked for me when I tested your changes, but I get this for Python:
Screen Shot 2021-08-10 at 11 03 09 AM

Any ideas?

tonybaloney: It needs to be an ANT-95 image, I've had success with West Central US. Not all regions have been updated yet

appservice/src/remoteDebug/startRemoteDebug.ts
tonybaloney: I've refactored this to allow multi root workspaces and show a pick list when there is more than one instead of just throwing an error

appservice/src/remoteDebug/startRemoteDebug.ts
```diff
@@ -83,44 +84,92 @@ async function startRemoteDebugInternal(context: IActionContext, site: ParsedSit
...
-    },
-    [RemoteDebugLanguage.Python]: {
-        type: 'python'
-    }
-};
-
-async function getDebugConfiguration(context: IActionContext, language: RemoteDebugLanguage): Promise<vscode.DebugConfiguration> {
-    if (!(language in baseConfigs)) {
-        throw new Error(localize('remoteDebugLanguageNotSupported', 'The language "{0}" is not supported for remote debugging.', language));
+async function selectMultirootWorkspaceTarget(): Promise<vscode.Uri | undefined> {
```
ejizba, line 87: You don't have to implement this yourself, VS Code has an existing method `vscode.window.showWorkspaceFolderPick`

appservice/src/remoteDebug/startRemoteDebug.ts
```diff
@@ -83,44 +84,92 @@ async function startRemoteDebugInternal(context: IActionContext, site: ParsedSit
...
+                localRoot: localRoot,
+                remoteRoot: '.',
+            },
+        ],
+    };
+
+    return config;
+}
+
+async function getDebugConfiguration(language: RemoteDebugLanguage): Promise<vscode.DebugConfiguration> {
```
ejizba, line 162: Please move this back to the top and try to minimize refactoring stuff which is making the diff more difficult to read than it needs to be

tonybaloney: I forget that declaration order doesn't matter. Fixed

appservice/src/remoteDebug/startRemoteDebug.ts
tonybaloney: Ah, fixed.

appservice/package.json
tonybaloney: rolled back

tonybaloney:
@ejizba please can you review updates.

ejizba (approved):
Thanks! I still couldn't get it to work in West Central US, but I'll trust you for now. We will just need to wait for ANT-95 to roll out more broadly before we can remove the feature flag (related to https://github.com/microsoft/vscode-azureappservice/issues/2074)

PR #981, opened 2021-08-11
Add Learn More ( ? ) button to quickpick/input window if url is provided

nturinski:
Sometimes it doesn't feel like we have enough space to give users the context to understand what the prompt is asking for.  Specifically, in Static Web Apps, there are a lot of terms such as output location, etc that are foreign but essential to understand.  Hopefully with documentation readily available, we can help users understand what the prompt is actually asking for.


ui/src/constants.ts
```diff
@@ -3,5 +3,12 @@
...
  *--------------------------------------------------------------------------------------------*/
 
+import { QuickInputButton, ThemeIcon } from "vscode";
+import { localize } from "./localize";
+
 export const resourcesProvider: string = 'Microsoft.Resources';
 export const storageProvider: string = 'Microsoft.Storage';
+
+export namespace AzExtQuickInputButtons {
+    export const LearnMore: QuickInputButton = { iconPath: new ThemeIcon('question'), tooltip: localize('clickHere', 'Click here to learn more about this step.') }
```
ejizba, line 13: I don't like "Click here" because I feel like people know how buttons work. I would suggest we just go with `Learn more` which matches [this](https://github.com/microsoft/vscode-azuretools/blob/main/ui/src/DialogResponses.ts#L14)

Also, I anticipate people will want to customize this text anyways, so we might want `learnMoreTooltip` added as an option next to `learnMoreLink`

ui/src/userInput/showInputBox.ts
```diff
@@ -40,9 +44,14 @@ export async function showInputBox(context: IInternalActionContext, options: Inp
...
-                inputBox.onDidTriggerButton(_btn => {
-                    // Only back button is supported for now
-                    reject(new GoBackError());
+                inputBox.onDidTriggerButton(async btn => {
+                    if (btn === QuickInputButtons.Back) {
+                        reject(new GoBackError());
+                    } else if (btn === AzExtQuickInputButtons.LearnMore) {
+                        await openUrl(nonNullProp(options, 'learnMoreLink'));
+                        context.telemetry.properties.learnMoreStep = context.telemetry.properties.lastStep;
+                        context.telemetry.properties.clickedLearnMore = 'true';
```
ejizba, line 53: We don't need `clickedLearnMore` because we know it was clicked if `learnMoreStep` is defined

ui/src/userInput/showInputBox.ts
```diff
@@ -40,9 +44,14 @@ export async function showInputBox(context: IInternalActionContext, options: Inp
...
                 }),
-                inputBox.onDidTriggerButton(_btn => {
-                    // Only back button is supported for now
-                    reject(new GoBackError());
+                inputBox.onDidTriggerButton(async btn => {
+                    if (btn === QuickInputButtons.Back) {
+                        reject(new GoBackError());
+                    } else if (btn === AzExtQuickInputButtons.LearnMore) {
+                        await openUrl(nonNullProp(options, 'learnMoreLink'));
+                        context.telemetry.properties.learnMoreStep = context.telemetry.properties.lastStep;
```
ejizba, line 52: Maybe add `learnMoreStep` here as well just to be consistent:
https://github.com/microsoft/vscode-azuretools/blob/df7af8b62562e7c25ba132d9aa3fc848cb6f2971/ui/src/userInput/showWarningMessage.ts#L34

ui/src/constants.ts
nturinski: > Also, I anticipate people will want to customize this text anyways, so we might want learnMoreTooltip added as an option next to learnMoreLink

So I originally had an optional tooltip property to customize it, but it makes it impossible to have a constant LearnMore button.  That's not really a problem per se, but does make the button handling check a little less reliable/clean.  

I also thought about it, and I'm not really sure how much value there is in having the tooltip be customizable.  At most, I feel like you could just say "Learn more about _____" which isn't that useful to me.

ejizba (approved):

PR #982, opened 2021-08-11
Add '@microsoft' scope to eslint-config-azuretools package

wwlorey:
According to the [npm package publishing docs](https://docs.opensource.microsoft.com/releasing/publish/npm/):

> Do NOT publish packages from your own account.

But later in the document they add:

> The Microsoft NPM account has been converted to an organization so accounts that wish to publish @microsoft scoped packages must be added to the Microsoft NPM organization.

My interpretation is that the preferred way to publish MS packages is to do so under the "@microsoft" scope. And that publishing can be done from a personal account that belongs to the MS org. So it makes sense to add that scope to this new package

### DIFF
--- a/eslint-config-azuretools/README.md
+++ b/eslint-config-azuretools/README.md
@@ -2,25 +2,28 @@
 
 This package provides extensible ESLint configs used by the Azure Tools for VS Code Team.
 
-* `eslint-config-azuretools` (for production code)
-* `eslint-config-azuretools/test` (for test code)
+* `@microsoft/eslint-config-azuretools` (for production code)
+* `@microsoft/eslint-config-azuretools/test` (for test code)
 
 ## Usage
 
 1. Install this package and its peer dependencies. Peer dependencies can be listed with the following command:
 
     ```bash
-    npm info eslint-config-azuretools peerDependencies
+    npm info @microsoft/eslint-config-azuretools peerDependencies
     ```
 
 2. Depending on which config you want to use, add the following to your `.eslintrc`:
 
     ```json
-    "extends": "azuretools"
+    "extends": "@microsoft/eslint-config-azuretools"
     ```
     or
     ```json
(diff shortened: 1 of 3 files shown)


eslint-config-azuretools/README.md
```diff
@@ -2,25 +2,28 @@
...
     ```bash
-    npm info eslint-config-azuretools peerDependencies
+    npm info @microsoft/eslint-config-azuretools peerDependencies
     ```
 
 2. Depending on which config you want to use, add the following to your `.eslintrc`:
 
     ```json
-    "extends": "azuretools"
+    "extends": "@microsoft/eslint-config-azuretools"
```
wwlorey, line 19: Sadly adding a scope to the package name prevents the cool shortcut 😢

neelip, line 19: Why the change in the name here from azuretools to "...eslint-config-azuretools"?

eslint-config-azuretools/README.md
```diff
@@ -2,25 +2,28 @@
...
     ```
     or
     ```json
-    "extends": [ "azuretools", "azuretools/test" ]
+    "extends": [
+        "@microsoft/eslint-config-azuretools",
```
neelip, line 24: Same as above, what is the reason for name change?

eslint-config-azuretools/README.md
ejizba: Per these docs: https://eslint.org/docs/developer-guide/shareable-configs#npm-scoped-modules

Without a scope in the name:

> You can omit the `eslint-config-` and it will be automatically assumed by ESLint

With a scope in the name:

> The module name can also be customized, just note that when using scoped modules it is not possible to omit the `eslint-config-` prefix. Doing so would result in package naming conflicts, and thus in resolution errors in most of cases

ejizba (approved):

PR #984, opened 2021-08-13
Bump minimum VS Code version to 1.57 in ui package

ejizba:
The primary reason I want to do this is because VS Code changed to Node 14 in v1.56 (see [here](https://code.visualstudio.com/updates/v1_56#_electron-12-update)) and I prefer to keep our Node version in line with them.

However, if I bump just one more version to 1.57 (released on June 10), then I can use VS Code's built-in loading icon since https://github.com/microsoft/vscode/issues/117859 was fixed.


ui/src/utils/randomUtils.ts
```diff
@@ -6,7 +6,7 @@
 import * as crypto from "crypto";
 
 export namespace randomUtils {
-    export function getPseudononymousStringHash(s: string, encoding: crypto.HexBase64Latin1Encoding = 'base64'): string {
+    export function getPseudononymousStringHash(s: string, encoding: crypto.BinaryToTextEncoding = 'base64'): string {
```
nturinski, line 9: Why was this changed?

nturinski (approved):

ui/src/utils/randomUtils.ts
ejizba: I think they just changed the name of the type between Node 12 and Node 14

PR #989, opened 2021-08-16
Change masking to not filter "ASP.NET" or ".NET"

bwateratmsft:
This slight change to the URL masking makes it so "ASP.NET" or ".NET" are not masked. However, something like "fooASP.net" or "foo-ASP.net" will still be masked.

Playground on regexr:
https://regexr.com/63s3k

### DIFF
--- a/ui/src/masking.ts
+++ b/ui/src/masking.ts
@@ -77,7 +77,7 @@ export function maskUserInfo(unkonwnArg: unknown, actionValuesToMask: string[], 
     }
 
     data = data.replace(/[a-z]+:\/\/\S*/gi, getRedactedLabel('url'));
-    data = data.replace(/\S+(?<!\basp)\.(com|org|net)\S*/gi, getRedactedLabel('url'));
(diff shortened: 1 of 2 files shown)


ui/src/masking.ts
```diff
@@ -77,7 +77,7 @@ export function maskUserInfo(unkonwnArg: unknown, actionValuesToMask: string[],
     }
 
     data = data.replace(/[a-z]+:\/\/\S*/gi, getRedactedLabel('url'));
-    data = data.replace(/\S*\.(com|org|net)\S*/gi, getRedactedLabel('url'));
+    data = data.replace(/\S+(?<!(?<!\-)\basp)\.(com|org|net)\S*/gi, getRedactedLabel('url'));
```
ejizba, line 80: Please add ".NET" and "ASP.NET" to the unit tests - https://github.com/microsoft/vscode-azuretools/blob/main/ui/test/masking.test.ts

ejizba (approved):

ui/src/masking.ts
bwateratmsft: I added tests for ".NET", "ASP.NET", "microsoft-asp.net", "microsoftasp.net", and a few flavors of those.

PR #990, opened 2021-08-16
Fix invalid port assignment for remote debugging

tonybaloney:
This fixes the ECONNREFUSED 127.0.0.1:8000 bug.

`debugConfig.port` is no longer the valid attribute for Python but it is for Node.

Instead of hardcoding the field property and muting the linter, this assigns the port in the host function and then sets it in the debug configuration model.


tonybaloney:
FYI @ejizba this explains the issue you were seeing.

ejizba (approved):
Yep worked for me this time - cool!

PR #991, opened 2021-08-17
Show loading prompts in-between wizard steps

alexweininger:
There can be long delays between wizard steps when doing async work. Ex: getting the default value of an input. In less than ideal network conditions the delays can become a lot longer and are awkward since we don't have any UI showing.

If option is set as true, show loading prompt during long delays between wizard steps.
Below are videos showing before & after. I've added delays to the prompt of the steps to exaggerate for display. I've tested on wizards in both SWA and Functions and haven't seen any visual glitches or times when it appears and disappears quickly.

Before:

https://user-images.githubusercontent.com/12476526/129630850-71612c70-3018-4fad-bace-890782bcb481.mov

After:

https://user-images.githubusercontent.com/12476526/129630846-b859d010-2ef0-41b5-924f-1498a411fc6e.mov

Play both simultaneously for best experience 😄

### DIFF
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -718,6 +718,11 @@ export interface IAzureUserInput {
     readonly onDidFinishPrompt: Event<PromptResult>;
 
     /**
+    * If true, a prompt is currently being shown. Excludes loading prompts.
+    */
+    readonly isPrompting: boolean;
+
+    /**
     * Shows a multi-selection list.
     *
     * @param items An array of items, or a promise that resolves to an array of items.
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -905,11 +905,6 @@ export interface IWizardOptions<T extends IActionContext> {
      * If true, step count will not be displayed for the entire wizard. Defaults to false.
      */
     hideStepCount?: boolean;
-
-    /**
-    * If true, a loading prompt will be displayed if there are long delays between wizard steps.
-    */
-    showLoadingPrompt?: boolean;
 }
 
 /**
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -718,11 +718,6 @@ export interface IAzureUserInput {
     readonly onDidFinishPrompt: Event<PromptResult>;
 
     /**
-    * If true, a prompt is currently being shown. Excludes loading prompts.
-    */
-    readonly isPrompting: boolean;
-
-    /**
     * Shows a multi-selection list.
     *
     * @param items An array of items, or a promise that resolves to an array of items.
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -905,6 +905,11 @@ export interface IWizardOptions<T extends IActionContext> {
      * If true, step count will not be displayed for the entire wizard. Defaults to false.
      */
     hideStepCount?: boolean;
+
+    /**
+    * If true, a loading prompt will be displayed if there are long delays between wizard steps.
+    */
+    showLoadingPrompt?: boolean;
 }
 
 /**
--- a/ui/src/userInput/AzExtUserInput.ts
+++ b/ui/src/userInput/AzExtUserInput.ts
@@ -16,6 +16,7 @@ export class AzExtUserInput implements types.IAzureUserInput {
     public wizard?: IInternalAzureWizard;
     private _onDidFinishPromptEmitter: EventEmitter<types.PromptResult> = new EventEmitter<types.PromptResult>();
     private _context: IInternalActionContext;
+    private _isPrompting: boolean = false;
 
     public constructor(context: IInternalActionContext) {
         this._context = context;
@@ -25,14 +26,23 @@ export class AzExtUserInput implements types.IAzureUserInput {
         return this._onDidFinishPromptEmitter.event;
     }
 
+    public get isPrompting(): boolean {
+        return this._isPrompting;
+    }
+
     public async showQuickPick<TPick extends types.IAzureQuickPickItem<unknown>>(picks: TPick[] | Promise<TPick[]>, options: types.IAzureQuickPickOptions): Promise<TPick | TPick[]> {
         addStepTelemetry(this._context, options.stepName, 'quickPick', options.placeHolder);
         if (this._context.ui.wizard?.cancellationToken.isCancellationRequested) {
             throw new UserCancelledError();
         }
-        const result = await showQuickPick(this._context, picks, options);
-        this._onDidFinishPromptEmitter.fire({ value: result });
-        return result;
+        try {
+            this._isPrompting = true;
+            const result = await showQuickPick(this._context, picks, options);
+            this._onDidFinishPromptEmitter.fire({ value: result });
+            return result;
+        } finally {
+            this._isPrompting = false;
+        }
     }
 
     public async showInputBox(options: types.AzExtInputBoxOptions): Promise<string> {
@@ -40,19 +50,30 @@ export class AzExtUserInput implements types.IAzureUserInput {
         if (this._context.ui.wizard?.cancellationToken.isCancellationRequested) {
             throw new UserCancelledError();
         }
-        const result = await showInputBox(this._context, options);
-        this._onDidFinishPromptEmitter.fire({
-            value: result,
-            matchesDefault: result === options.value
-        });
-        return result;
+        try {
+            this._isPrompting = true;
+            const result = await showInputBox(this._context, options);
+            this._onDidFinishPromptEmitter.fire({
+                value: result,
+                matchesDefault: result === options.value
+            });
+            return result;
+        } finally {
+            this._isPrompting = false;
+        }
     }
 
     public async showOpenDialog(options: types.AzExtOpenDialogOptions): Promise<Uri[]> {
         addStepTelemetry(this._context, options.stepName, 'openDialog', options.title);
-        const result = await showOpenDialog(options);
-        this._onDidFinishPromptEmitter.fire({ value: result });
-        return result;
+
+        try {
+            this._isPrompting = true;
+            const result = await showOpenDialog(options);
+            this._onDidFinishPromptEmitter.fire({ value: result });
+            return result;
+        } finally {
+            this._isPrompting = false;
+        }
     }
 
     public async showWarningMessage<T extends MessageItem>(message: string, ...items: T[]): Promise<T>;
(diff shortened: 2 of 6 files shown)


alexweininger:
Since only a few wizards have async work between steps, should we disable this behavior by default and add an option to the IWizardOptions to enable it?

nturinski:
Discussed offline, but I think it makes sense to include an option parameter to enable the async loading step.  I didn't see any weird glitches/flashes with it, but on slower performance machines, it could possibly have negative impact, so unless we anticipate a long delay between input steps, we shouldn't enable it for every wizard.

That will also lower the risk of releasing this into the wild.

ui/src/wizard/AzureWizard.ts
```diff
@@ -80,10 +83,16 @@ export class AzureWizard<T extends IInternalActionContext> implements types.Azur
                         }
                     });
 
+                    const loadingQuickPick = this._showLoadingPrompt ? createQuickPick(this._context, {
+                        loadingPlaceHolder: 'Loading...'
+                    }) : undefined;
                     try {
                         this.currentStepId = getEffectiveStepId(step);
+                        loadingQuickPick?.show();
                         await step.prompt(this._context);
```
ejizba, line 92: I would suggest you create a test wizard (like for manual testing, not unit testing) with fake delays in basically any place you can possibly think of. I suspect your current implementation does not show the Loading quick pick in as many places as it should. For example, if I have prompt code like this:
```
public async prompt(context: IActionContext): Promise<void> {
    await delay(5000);
    await context.ui.showQuickPick(...);
    await delay(5000);
}
```
I don't think the loading quick pick will be shown during the second delay. Also `getSubWizard` can return a promise, so there's potential for delay there as well

ui/index.d.ts
```diff
@@ -905,6 +905,11 @@ export interface IWizardOptions<T extends IActionContext> {
      * If true, step count will not be displayed for the entire wizard. Defaults to false.
      */
     hideStepCount?: boolean;
+
+    /**
+    * If true, a loading prompt will be displayed if there are long delays between wizard steps.
+    */
+    showLoadingPrompt?: boolean;
```
ejizba, line 912: One alternative is to only show the loading prompt if the delay is more than a certain interval (500ms? 1s?). It would prevent a flashing "Loading" in the vast majority of cases but still show it for most cases that need it

That being said, I'm fine with a boolean like this while we do more testing and see how it works in our various wizards

ui/src/wizard/AzureWizard.ts
alexweininger: Unless we add a way to detect exactly when the UI inside of `step.prompt` is shown/hidden then I don't think handling these cases is possible only from the AzureWizard (and without changing the wizard step implementations).

An idea I had was to pass a function `runWithLoadingPrompt` to the step. And then the step can use this however it needs to. Instead of trying to know when the UI is shown/hidden.

ejizba: Can you use `ui.onDidFinishPrompt` somehow? We're already using it a few lines up for other reasons.

alexweininger: Looks like that will work perfectly. Thanks!

alexweininger: Ok, here is what it looks like with a `delay(3000)` after the prompt. I'm not sure how I feel about the step # in the title not incrementing right away. Thoughts?

https://user-images.githubusercontent.com/12476526/129802447-4687b8c6-0ad5-40e2-9fc9-46269fd8bec4.mov

ejizba: Meh probably not worth the effort to mess with the step number

ui/src/wizard/AzureWizard.ts
```diff
@@ -72,18 +75,26 @@ export class AzureWizard<T extends IInternalActionContext> implements types.Azur
...
                         if (typeof result.value === 'string' && !result.matchesDefault && this.currentStepId && !step?.supportsDuplicateSteps) {
                             this._cachedInputBoxValues[this.currentStepId] = result.value;
                         }
                     });
 
                     try {
                         this.currentStepId = getEffectiveStepId(step);
+                        loadingQuickPick?.show();
                         await step.prompt(this._context);
+                        loadingQuickPick?.hide();
```
ejizba, line 95: Just do this once in in the `finally` block instead of duplicated here and in the `catch` block. Also, should probably be after `disposable.dispose()` just to be safe on timing issues

ui/src/wizard/AzureWizard.ts
```diff
@@ -72,18 +75,26 @@ export class AzureWizard<T extends IInternalActionContext> implements types.Azur
                 if (step.shouldPrompt(this._context)) {
                     step.propertiesBeforePrompt = Object.keys(this._context).filter(k => !isNullOrUndefined(this._context[k]));
 
+                    const loadingQuickPick = this._showLoadingPrompt ? createQuickPick(this._context, {
+                        loadingPlaceHolder: 'Loading...'
```
ejizba, line 79: Use `localize`

ejizba (dismissed):

alexweininger:
I just found a bug, and I think I'm going to need some help solving it. If you press escape during a loading prompt, then the loading prompt hides but the next step still appears.

Fixing it by throwing a `UserCancelledError` when in the `onDidHide` handler of the loading prompt won't work since when it I throw in the handler, catch/finally is not hit.

I'm thinking I'll have to use a Promise and reject it somewhere.

ui/index.d.ts
alexweininger: I think I'll stay with the boolean for two reasons:

1. Flashing loading here won't 

> I just found a bug, and I think I'm going to need some help solving it. If you press escape during a loading prompt, then the loading prompt hides but the next step still appears.
> 
> Fixing it by throwing a `UserCancelledError` when in the `onDidHide` handler of the loading prompt won't work since when it I throw in the handler, catch/finally is not hit.
> 
> I'm thinking I'll have to use a Promise and reject it somewhere.

Talked with Eric offline, disabling cancelling during loading sounds like a good solution 😄

alexweininger:
> I just found a bug, and I think I'm going to need some help solving it. If you press escape during a loading prompt, then the loading prompt hides but the next step still appears.
> 
> Fixing it by throwing a `UserCancelledError` when in the `onDidHide` handler of the loading prompt won't work since when it I throw in the handler, catch/finally is not hit.
> 
> I'm thinking I'll have to use a Promise and reject it somewhere.

Talked with Eric offline, disabling cancelling during loading sounds like a good solution 😄

ui/src/wizard/AzureWizard.ts
```diff
@@ -72,16 +82,26 @@ export class AzureWizard<T extends IInternalActionContext> implements types.Azur
                 if (step.shouldPrompt(this._context)) {
                     step.propertiesBeforePrompt = Object.keys(this._context).filter(k => !isNullOrUndefined(this._context[k]));
 
+                    const loadingQuickPick = this._showLoadingPrompt ? createQuickPick(this._context, {
```
ejizba, line 85: Let's check `cancellationToken.isCancellationRequested` in a few more spots, at least once per `await` call. I would suggest
1. At the beginning of the loop right after `while (step) `
1. Right before `await step.getSubWizard`

ui/src/wizard/AzureWizard.ts
```diff
@@ -72,16 +82,26 @@ export class AzureWizard<T extends IInternalActionContext> implements types.Azur
...
+                        loadingPlaceHolder: localize('loading', 'Loading...')
+                    }) : undefined;
+
+                    loadingQuickPick?.onDidHide(() => {
```
ejizba, line 89: This returns a disposable that should be handled somehow. Also I can't remember - is `onDidHide` only called when the user escapes the pick or is it also called when a new pick is displayed and causes the old one to hide?

alexweininger: > is onDidHide only called when the user escapes the pick or is it also called when a new pick is displayed and causes the old one to hide?

It's called on both, user escape and when a new pick is displayed.

ejizba: So isn't that a problem? I would think a new pick being displayed would cause the wizard to eventually get cancelled because of the token.

alexweininger: Yeah it's a problem. It gets cancelled when it gets to the next step. Thinking of solutions 🤔

ui/src/userInput/AzExtUserInput.ts
```diff
@@ -26,13 +27,19 @@ export class AzExtUserInput implements types.IAzureUserInput {
...
+            throw new UserCancelledError();
+        }
         const result = await showQuickPick(this._context, picks, options);
         this._onDidFinishPromptEmitter.fire({ value: result });
         return result;
     }
 
     public async showInputBox(options: types.AzExtInputBoxOptions): Promise<string> {
         addStepTelemetry(this._context, options.stepName, 'inputBox', options.prompt);
+        if (this._context.ui.wizard?.cancellationToken.isCancellationRequested) {
```
ejizba, line 40: We need this logic for `showOpenDialog` and `showWarningMessage` as well - they can be used in a wizard step, too.

ui/src/wizard/AzureWizard.ts
```diff
@@ -72,16 +86,31 @@ export class AzureWizard<T extends IInternalActionContext> implements types.Azur
...
+                        disposables.push(loadingQuickPick?.onDidHide(() => {
+                            this._cancellationTokenSource.cancel();
+                        }));
+                    }
+
+                    disposables.push(this._context.ui.onDidFinishPrompt((result) => {
                         // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
                         step!.prompted = true;
+                        loadingQuickPick?.show();
+                        this._cancellationTokenSource = new vscode.CancellationTokenSource();
```
ejizba, line 105: First of all, each `CancellationTokenSource` should be disposed of properly. Secondly, I think there's a small chance of timing issues

I would actually approach this in a different way, though - rather than create a new token source each time, I would want to prevent the existing token source from cancelling itself when we don't want it to. In other words, `loadingQuickPick?.onDidHide` above would somehow need to know when `AzExtUserInput` is about to prompt so that they can ignore that specific event. Something along the lines of `AzExtUserInput.onDidStartPrompt: Event<whatever>` or `AzExtUserInput.isPrompting: boolean` would probably help accomplish that.

ejizba (approved):

PR #993, opened 2021-08-20
Add the AzExtFsExtra namespace to the ui package

nturinski:
Copied straight from https://github.com/microsoft/vscode-azurestorage/blob/main/src/utils/AzExtFsExtra.ts

I'll get rid of the fse call in another PR since I'll need to add a readFileSync to the namespace.

### DIFF
--- a/ui/src/utils/AzExtFsExtra.ts
+++ b/ui/src/utils/AzExtFsExtra.ts
@@ -1,0 +1,82 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { dirname } from 'path';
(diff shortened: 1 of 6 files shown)


ui/src/utils/AzExtFsExtra.ts
```diff
@@ -0,0 +1,82 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { dirname } from 'path';
+import { FileStat, FileType, Uri, workspace } from 'vscode';
+import { parseError } from '../parseError';
+
+export namespace AzExtFsExtra {
```
ejizba, line 10: You should move the unit tests over as well

nturinski: I forgot I even wrote those... 😅

ejizba (approved):

PR #994, opened 2021-08-24
Remove fs-extra from dependencies

nturinski:
I tried to make a readFileSync for AzExtFsExtra, but it was being a lot more trouble than I think it's worth (since we almost use it for nothing).


alexweininger (approved):
Nice

PR #995, opened 2021-08-24
Support `rejectUnauthorized` during `sendRequestWithTimeout`

ejizba:


ui/src/createAzureClient.ts
```diff
@@ -56,12 +58,20 @@ export function createAzureSubscriptionClient<T>(clientContext: InternalAzExtCli
...
 
-export async function sendRequestWithTimeout(context: types.IActionContext, options: RequestPrepareOptions, timeout: number, clientInfo: types.AzExtGenericClientInfo): Promise<HttpOperationResponse> {
+export async function sendRequestWithTimeout(context: types.IActionContext, options: types.AzExtRequestPrepareOptions, timeout: number, clientInfo: types.AzExtGenericClientInfo): Promise<HttpOperationResponse> {
     let request: WebResource = new WebResource();
     request = request.prepare(options);
     request.timeout = timeout;
+
+    if (options.rejectUnauthorized !== undefined) {
+        request.agentSettings = {
+            http: new Agent(),
```
nturinski, line 68: What agent does the request default to if you don't set it?  Do we not need to set it even if rejectUnauthorized is undefined?

nturinski (approved):
Any context/issue that this fixes?

nturinski:
Never mind, I see: https://github.com/microsoft/vscode-azurefunctions/issues/2887, https://github.com/microsoft/vscode-azurefunctions/pull/2930

ui/src/createAzureClient.ts
ejizba: So `agentSettings` is an optional property. However, if you _do_ set it, you have to set both `http` and `https` (I'm not entirely sure why). I assume it just defaults to the standard Agent that I'm using, which is a core part of Node.js

PR #996, opened 2021-08-24
Warn if default app is not found

ejizba:
Instead of just silently removing the setting

Fixes https://github.com/microsoft/vscode-azurefunctions/issues/2789

### DIFF
--- a/appservice/package-lock.json
+++ b/appservice/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "vscode-azureappservice",
-    "version": "0.86.2",
+    "version": "0.86.3",
     "lockfileVersion": 2,
     "requires": true,
     "packages": {
         "": {
             "name": "vscode-azureappservice",
-            "version": "0.86.2",
+            "version": "0.86.3",
             "license": "MIT",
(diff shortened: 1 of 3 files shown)


nturinski (approved):
Do we have any idea how often this was happening?  I kind of think it'd be nice if the error was a little more obvious/visible since there are a lot of things being printed to output during a deployment.  

Also, you don't think we should clear the setting still?  I feel like besides that weird casing issue, if the app is gone, then we shouldn't keep that stale id around.

ejizba:
> Do we have any idea how often this was happening? I kind of think it'd be nice if the error was a little more obvious/visible since there are a lot of things being printed to output during a deployment.

Doesn't seem like a big enough deal to me. The only downside is they have one or two more clicks to pick their app

> Also, you don't think we should clear the setting still? I feel like besides that weird casing issue, if the app is gone, then we shouldn't keep that stale id around.

I don't think we should modify a user's settings without letting them know first, and I don't think this scenario is worth a prompt. I also think there could be a number of reasons we don't find the app, so I don't agree that clearing the setting should be our default behavior even if we did prompt

PR #1002, opened 2021-09-17
Add script to sort package.json and package.nls.json

alexweininger:
Sorts:
* activationEvents
* contributes.commands
* contributes.menus (sorts items in each menu by command)

And also sorts the package.nls.json file.

We can then invoke this in a pre-commit hook in the extensions. [Example of adding this in functions](https://github.com/microsoft/vscode-azurefunctions/pull/2969) 

`npx vscode-sort-package-json` to invoke the script.


dev/scripts/sortPackageJson.js
```diff
@@ -0,0 +1,73 @@
...
+ * package.json
+ *  - activationEvents
+ *  - contributes.commands
+ *  - contributes.menus
+ * 
+ * package.nls.json
+ */
+
+const path = require('path');
+const existsSync = require('fs').existsSync;
```
nturinski, line 20: Should be able to replace these with:

```
import * as path from 'path';
import * as fs from 'fs/promises';
```

In general, we usually use imports the main reason being
> 3. Import statements are asynchronous
Being synchronous or asynchronous might not play a major role in small-scale applications. But, if we think of large-scale applications, there are hundreds of modules are used. So,
if you use require, modules will be loaded and processed one by one.
On the other hand, import statements solve this issue by working asynchronously, and they are known to perform well compared to require functions in large-scale applications.

Also just stylistically, I prefer it.

dev/scripts/sortPackageJson.js
```diff
@@ -0,0 +1,73 @@
...
+const fs = require('fs/promises');
+
+async function sortPackageJson(path) {
+    if (!existsSync(path)) {
```
nturinski, line 24: If you change the import statement:

```suggestion
    if (!fs.existsSync(path)) {
```

dev/scripts/sortPackageJson.js
```diff
@@ -0,0 +1,73 @@
+#! /usr/bin/env node
+
+/*---------------------------------------------------------------------------------------------
```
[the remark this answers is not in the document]
alexweininger: I'm not super familiar with shebangs but according to [this](https://alexewerlof.medium.com/node-shebang-e1d4b02f731d) it seems like shebangs must be on the first line. Since the kernel checks if the first two characters are `#!`.

> Node.js will happily ignore this as a comment only if it is the very first line of the file (it won’t work even if there’s an empty line or //comment line before it). Browsers ignore it too (Chrome 74+, FF 67+).

dev/package.json
```diff
@@ -62,4 +65,4 @@
         "ts-loader": "^5.3.3",
         "webpack": "5.28.0"
     }
-}
+}
```
nturinski, line 68: nit: No newline here

nturinski (approved):

PR #1007, opened 2021-10-01
Bump dev package

alexweininger:
Going to publish a release of the dev package so we can use the new sort script in the extensions.

### DIFF
--- a/dev/package.json
+++ b/dev/package.json
@@ -1,7 +1,7 @@
 {
     "name": "vscode-azureextensiondev",
     "author": "Microsoft Corporation",
-    "version": "0.10.1",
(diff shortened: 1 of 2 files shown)


dev/package.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "vscode-azureextensiondev",
     "author": "Microsoft Corporation",
-    "version": "0.10.1",
+    "version": "0.10.2",
```
bwateratmsft, line 4: Though it isn't strictly necessary, maybe get package-lock.json too?

bwateratmsft (approved):

PR #1015, opened 2021-10-20
Setup for adding databases to CI/CD

neelip:


alexweininger:
Database package doesn't currently have a test script, see https://github.com/microsoft/vscode-azuretools/blob/main/databases/package.json#L28-L35. The `package-steps.yml` runs `npm test` https://github.com/microsoft/vscode-azuretools/blob/main/.azure-pipelines/package-steps.yml#L59-L70. I think this will make it fail, but I may be wrong.

Ways around it:

- Add a temporary `test` script to `databases/package.json` for now
- Alter the `package-steps.yml` steps in some way to not run `npm test` on this package.

wwlorey (approved):
Discussed offline with @neelip: I think it's fine to wait to implement tests but to @alexweininger's point we should add a placeholder test script. The build is erroring out without it

bwateratmsft (dismissed):

PR #1025, opened 2021-11-17
Reduce aggressiveness of masking logic

### LINKED ISSUE #1024: `maskValue` is corrupting telemetry data seemingly at random
https://github.com/microsoft/vscode-azuretools/blob/main/ui/src/masking.ts#L89 is causing apparently-random data corruption:

image

bwateratmsft:
Fixes #1024. Usernames shorter than 5 characters (i.e., 4 or less) will no longer be masked out.

Longer term, I'd like to see zero masking logic here, all of it really belongs in `vscode-extension-telemetry`. I opened https://github.com/microsoft/vscode-extension-telemetry/issues/80 to get the custom values-to-mask feature added there.


ui/src/masking.ts
```diff
@@ -12,7 +12,14 @@ let _extValuesToMask: string[] | undefined;
 function getExtValuesToMask(): string[] {
     if (!_extValuesToMask) {
         try {
-            _extValuesToMask = [os.userInfo().username];
+            const username = os.userInfo().username;
+
+            if (username?.length > 4) {
```
alexweininger, line 17: I read #1024 and understand why we should not mask usernames that are too short. However, would it be sufficient and maybe a bit better in terms of PII to mask usernames that are longer than 3 characters instead of longer than 4?

The odds that a 4 letter username gets masked in relevant data is much lower than 1, 2, and 3 letter usernames. I'm just worried that a lot of usernames will be 4 letters long, and we might get flagged for PII.

At what length of username is it considered PII?

alexweininger (approved):

ui/src/masking.ts
alexweininger: Could we just prevent values like "ImageLinkProvider" from being masked at all?

bwateratmsft: We can though that would be tough. It's one of those problems where we usually don't know what we don't know, until it's too late.

wwlorey: > Maybe we should be masking for /\bMyUsername\b/i instead? Then it would not grab things like the string above no matter how long the name is.

I like this approach - that way we cover our bases by continuing to mask all usernames

bwateratmsft: I guess what I'd really prefer is for https://github.com/microsoft/vscode-extension-telemetry/issues/80 and https://github.com/microsoft/vscode-extension-telemetry/issues/81 to be fixed, then we can just remove this altogether.

bwateratmsft:
I'm going to switch this to a draft, let's see if we can get https://github.com/microsoft/vscode-extension-telemetry/issues/80 and https://github.com/microsoft/vscode-extension-telemetry/issues/81 instead of needing to do this ourselves...

ui/src/masking.ts
```diff
@@ -8,11 +8,20 @@ import * as os from 'os';
...
-let _extValuesToMask: string[] | undefined;
-function getExtValuesToMask(): string[] {
+type MaskMatcher = string | RegExp;
+
+let _extValuesToMask: MaskMatcher[] | undefined;
+function getExtValuesToMask(): MaskMatcher[] {
     if (!_extValuesToMask) {
         try {
-            _extValuesToMask = [os.userInfo().username];
+            const username = os.userInfo().username;
```
bwateratmsft, line 17: Is there any good way to mock this so we can write tests?

PR #1026, opened 2021-11-30
Run "npm audit fix --production" in dev package

alexweininger:
This should fix some of the CG alerts

Packages upgraded:
@xmldom/xmldom
axios
tar
adal-node


bwateratmsft (approved):
I think we should also consider an `npm update` and `npm audit fix` (without `--production`) as well. That can be a separate PR.

PR #1027, opened 2021-11-30
Upgrade axios in ui package

alexweininger:


bwateratmsft (approved):
Copypasta of my other comment :smile:

I think we should also consider an `npm update` and `npm audit fix` (without `--production`) as well. That can be a separate PR.

PR #1031, opened 2021-12-01
Mask username with one-time-computed regexp

### LINKED ISSUE #1024: `maskValue` is corrupting telemetry data seemingly at random

bwateratmsft:
An alternative to #1025, this fixes #1024. Personally I think I like this PR better. This will mask usernames 4 or more characters long but only with word boundaries on both sides. Case insensitive. Note, word boundaries includes the beginning or end of a string.

So, for example, if the username was "Dave", the following would be masked:
`I'm sorry Dave, I'm afraid can't do that`
`Dave, this conversation can serve no purpose`
`Access http://discovery.one/podbaydoors?user=dave: HTTP 403: Forbidden`

The following would _not_ be masked:
`Davedave`
`Daisy davesy`
`/home/davepod/baydoors`

I spent a completely unreasonable amount of time crafting this PR description.

### DIFF
--- a/ui/src/masking.ts
+++ b/ui/src/masking.ts
@@ -20,12 +20,12 @@ function getExtValuesToMask(): string[] {
 }
 
 let _usernameMask: RegExp | undefined | null = undefined;
-function getUsernameMask(): RegExp | undefined | null {
+function getUsernameMask(getUsername: () => string): RegExp | undefined | null {
     // _usernameMask starts out as undefined, and is set to null if building it fails or it is too short
     // Specifically checking for undefined here ensures we will only run the code once
     if (_usernameMask === undefined) {
         try {
-            const username = os.userInfo().username;
+            const username = getUsername();
 
             if (username.length <= UnmaskedUsernameMaxLength) {
                 // Too short to mask
@@ -39,6 +39,13 @@ function getUsernameMask(): RegExp | undefined | null {
     }
 
     return _usernameMask;
+}
+
+/**
+ * To be used ONLY by test code.
+ */
+export function resetUsernameMask(): void {
+    _usernameMask = undefined;
 }
 
 export function addExtensionValueToMask(...values: (string | undefined)[]): void {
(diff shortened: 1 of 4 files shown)


ui/src/masking.ts
```diff
@@ -8,16 +8,37 @@ import * as os from 'os';
...
+}
+
+let _usernameMask: RegExp | undefined | null = undefined;
+function getUsernameMask(): RegExp | undefined | null {
+    // _usernameMask starts out as undefined, and is set to null if building it fails or it is too short
+    // Specifically checking for undefined here ensures we will only run the code once
+    if (_usernameMask === undefined) {
         try {
-            _extValuesToMask = [os.userInfo().username];
+            const username = os.userInfo().username;
```
bwateratmsft, line 28: Is there any good way of mocking this so we can write tests? I ran tests locally (using my own username, and also by overwriting at runtime with a shorter one) and everything worked but obviously that's only useful once.

alexweininger: We could refactor these functions so that we pass in username as an argument. That way we can pass it in for testing. And add a wrapper function that we use at runtime that passes in `os.userInfo().username` for us.

bwateratmsft: In order to preserve the compute-one-time-ness of it, we'd also need to expose `_usernameMask` in some way or allow it to be rest.

alexweininger: Yeah, I think that sounds good. A function to set the value of `_usernameMask` to `undefined`. We can be clear about stating the function should be only used in tests.

bwateratmsft: Added some tests with that approach.

alexweininger (approved):

ui/src/masking.ts
```diff
@@ -8,16 +8,44 @@ import * as os from 'os';
 import { IActionContext, IParsedError } from "../index";
 import { parseError } from "./parseError";
 
+// No attempt will be made to mask usernames that are this length or less
+const UnmaskedUsernameMaxLength: number = 3;
```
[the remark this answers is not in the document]
bwateratmsft: Yeah, I figured since it was a constant (like, not just a `const`, but truly a constant).

PR #1036, opened 2021-12-09
Add optional telemetryId to registerCommand

StephenWeatherford:
This is to customize the telemetry IDs that are created for commands.

### DIFF
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -498,9 +498,12 @@ export type CommandCallback = (context: IActionContext, ...args: any[]) => any;
 /**
  * Used to register VSCode commands. It wraps your callback with consistent error and telemetry handling
  * Use debounce property if you need a delay between clicks for this particular command
+ * A telemetry event is automatically sent whenever a command is executed. The telemetry event ID will default to the same as the
+ *   commandId passed in, but can be overridden per command with telemetryId
+ * The telemetry event for this command will be named telemetryId if specified, otherwise it defaults to the commandId
  * NOTE: If the environment variable `DEBUGTELEMETRY` is set to a non-empty, non-zero value, then telemetry will not be sent. If the value is 'verbose' or 'v', telemetry will be displayed in the console window.
  */
-export declare function registerCommand(commandId: string, callback: CommandCallback, debounce?: number): void;
+export declare function registerCommand(commandId: string, callback: CommandCallback, debounce?: number, telemetryId?: string): void;
 
 /**
  * Used to register VSCode events. It wraps your callback with consistent error and telemetry handling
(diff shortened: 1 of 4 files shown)


ui/src/registerCommand.ts
```diff
@@ -23,7 +21,7 @@ export function registerCommand(commandId: string, callback: (context: types.IAc
         }
         // eslint-disable-next-line @typescript-eslint/no-unsafe-return
         return await callWithTelemetryAndErrorHandling(
-            commandId,
+            telemetryId ?? commandId,
```
bwateratmsft, line 24: I'd do `||` here, so if an empty string is inadvertently passed, it will not be used--instead, the `commandId` will be used.
```suggestion
            telemetryId || commandId,
```

ui/index.d.ts
```diff
@@ -498,9 +498,10 @@ export type CommandCallback = (context: IActionContext, ...args: any[]) => any;
 /**
  * Used to register VSCode commands. It wraps your callback with consistent error and telemetry handling
  * Use debounce property if you need a delay between clicks for this particular command
+ * The telemetry event for this command will be named telemetryId if specified, otherwise it defaults to the commandId
  * NOTE: If the environment variable `DEBUGTELEMETRY` is set to a non-empty, non-zero value, then telemetry will not be sent. If the value is 'verbose' or 'v', telemetry will be displayed in the console window.
  */
-export declare function registerCommand(commandId: string, callback: CommandCallback, debounce?: number): void;
+export declare function registerCommand(commandId: string, callback: CommandCallback, debounce?: number, telemetryId?: string): void;
```
bwateratmsft, line 504: For my own education, can you share more about the use case for this?

StephenWeatherford: We don't want "bicep.showVisualizer"/etc to be the name of the telemetry - our PM wants it something like "command.showVisualizer" so we can easily sort/query on just commands.

ui/package.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "vscode-azureextensionui",
     "author": "Microsoft Corporation",
-    "version": "0.49.0",
+    "version": "0.50.0",
```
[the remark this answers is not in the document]
StephenWeatherford: There have been some cases in the past where this type of thing was a breaking change because the tests were implementing an interface or faking a method, It's unlikely for this scenario, though, so I don't really care.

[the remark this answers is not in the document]
StephenWeatherford: done

ui/src/registerCommand.ts
StephenWeatherford: done

PR #1050, opened 2022-01-27
Upgrade Resource & Storage SDK to T2

nturinski:
A couple of changes to implement this:
- `list()` now uses an `PagedAsyncIterableIterator` instead of `Promise<IPartialList<T>`.  This means I can't just do an `await list()` anymore, which is why I have the helper function to help load all of the resources.  Might be worth to do a `loadByPage(numOfResources)` helper function later too.
- All T2 packages use a new type of credential called `coreAuth.TokenCredential`.  This uses `getToken` rather than `signRequest` so I made a type compatible with both so `createAzureClient` could be used with both T1 and T2 `ManagementClients`.
- They changed how to import the `Models` now to import individual models rather than the entire Model namespace.

Tested this with `Functions Create (Advanced)` and Container Apps.  It still lists and creates resource groups and storage accounts, so I think everything is working?  Container Apps uses the new T2 appservice package and those seem to be working as well, meaning the credentials are compatible with T1 and T2 azure sdk packages.


ui/src/wizard/StorageAccountNameStep.ts
```diff
@@ -38,13 +38,14 @@ export class StorageAccountNameStep<T extends types.IStorageAccountWizardContext
 
     private async validateStorageAccountName(client: StorageManagementClient, name: string): Promise<string | undefined> {
         name = name.trim();
-
+
```
bwateratmsft, line 41:
```suggestion

```
😸

ui/src/wizard/ResourceGroupCreateStep.ts
```diff
@@ -42,7 +42,11 @@ export class ResourceGroupCreateStep<T extends types.IResourceGroupWizardContext
             } else {
                 // if we suspect that this is a Concierge account, only pick the rg if it begins with "learn" and there is only 1
                 if (/concierge/i.test(wizardContext.subscriptionDisplayName)) {
-                    const rgs = await resourceClient.resourceGroups.list();
+                    const rgs: ResourceGroup[] = [];
+                    for await (const rg of resourceClient.resourceGroups.list()) {
```
bwateratmsft, line 46: Is `listAllIterator` an option here? Its implementation looks more or less like this.

ui/index.d.ts
```diff
@@ -154,6 +153,29 @@ export interface AzExtServiceClientCredentials {
...
+     * @param options - The options used to configure any requests this
+     *                TokenCredential implementation might make.
+     */
+     getToken(scopes: string | string[], options?: any): Promise<any | null>;
+}
+
+/**
+ * Loose type to allow for the use of T1 and T2 Azure ARM SDKs
+ */
+export type AzExtServiceClientCredentials = AzExtServiceClientCredentialsT1 | AzExtServiceClientCredentialsT2
```
bwateratmsft, line 177: For some requests in the Docker extension we call into `signRequest` (e.g. https://github.com/microsoft/vscode-docker/blob/main/src/utils/azureUtils.ts#L63). It would not be hard to change to `getToken`, but one would only do that if they knew they had to. Should we make a class that still implements `signRequest`?

Alternatively, we could mark `AzExtServiceClientCredentialsT1` deprecated or something like that, or remove it entirely, so that we get a build error, forcing the update to `getToken`.

ui/src/wizard/StorageAccountNameStep.ts
```diff
@@ -38,13 +38,14 @@ export class StorageAccountNameStep<T extends types.IStorageAccountWizardContext
...
         name = name.trim();
-
+        
         if (!name || name.length < storageAccountNamingRules.minLength || name.length > storageAccountNamingRules.maxLength) {
             return localize('invalidLength', 'The name must be between {0} and {1} characters.', storageAccountNamingRules.minLength, storageAccountNamingRules.maxLength);
         } else if (name.match(storageAccountNamingRules.invalidCharsRegExp) !== null) {
             return localize('invalidChars', "The name can only contain lowercase letters and numbers.");
         } else {
-            const nameAvailabilityResult: StorageManagementModels.CheckNameAvailabilityResult = await client.storageAccounts.checkNameAvailability(name);
+            const type = "Microsoft.Storage/storageAccounts";
```
alexweininger, line 47: We can extract this string out to a constant since we use it here and in StorageAccountListStep.ts

ui/index.d.ts
nturinski: As far as I know, the new credentials from the Azure Account are backwards compatible.  @wwlorey ?

The type `AzExtServiceClientCredentials ` could be either T1 or T2, but that's mostly so `createAzureClient` doesn't scream at us when we use it to lazy load clients.  We could also just have credentials and credentials2 on the `SubscriptionContext` so we can use whichever one typing dictates at the time (though my understanding is that they are actually the same credential? Again @wwlorey)

ui/src/wizard/ResourceGroupCreateStep.ts
nturinski: Oh yes, it totally is.  I made that helper function after I started doing it in multiple places and forgot to convert this back 👌

ui/index.d.ts
nturinski: https://github.com/microsoft/vscode-azuretools/blob/main/ui/src/tree/AzureAccountTreeItemBase.ts#L135 Yeah, if you follow credentials down to its base classes, it's => `TokenCredentialsBase` which implements `TokenClientCredentials` that extends `ServiceClientCredentials`.  `ServiceClientCredentials` still has `signRequest` and `TokenClientCredentials` has `getToken`

bwateratmsft: Just to make sure my understanding is correct, it's one object implementing two interfaces, assigned to both `ISubscriptionContext.credentials` and `ISubscriptionContext.credentials2` and the caller can choose which to use based on SDK requirements?

alexweininger (approved):

ui/index.d.ts
nturinski: Yeah, though I need to do a little work in the account extension to make the ADAL token fully forwards compatible.  As of now, if you use the `createAzureClient` helper, it doesn't seem to always work if you are logged into Azure via ADAL tokens. 

However, I believe that this update should be a non-breaking change and only effect our repos where we upgrade the Azure SDK packages.

bwateratmsft (approved):

ui/index.d.ts
wwlorey: We discussed some of this in triage yesterday but I wanted to respond here for posterity:

> It would not be hard to change to getToken, but one would only do that if they knew they had to. Should we make a class that still implements signRequest?

We can definitely still support `signRequest` and in fact that work has already been done in the account extension for MSAL (T2 SDK): https://github.com/microsoft/vscode-azure-account/pull/388. When we fix https://github.com/microsoft/vscode-azure-account/issues/415 we'll have to keep support for `signRequest` in mind. 

> As far as I know, the new credentials from the Azure Account are backwards compatible.

`credentials2` has types for both T1 and T2.

> We could also just have credentials and credentials2 on the SubscriptionContext so we can use whichever one typing dictates at the time (though my understanding is that they are actually the same credential?)

[`credentials`](https://github.com/microsoft/vscode-azure-account/blob/30056757606d4fa73f38b9b521f14c34d7455b11/src/azure-account.legacy.api.d.ts#L42) and [`credentials2`](https://github.com/microsoft/vscode-azure-account/blob/30056757606d4fa73f38b9b521f14c34d7455b11/src/azure-account.api.d.ts#L39) are actually different types. And `credentials` will be deprecated when we release 0.10.0 of the extension. It'll still be included in the legacy API though.

nturinski:
Changed quite a bit so I won't honor these previous approvals 😅

Relies on this PR https://github.com/microsoft/vscode-azure-account/pull/418/files

bwateratmsft (approved):

ui/src/utils/uiUtils.ts
```diff
@@ -20,4 +23,13 @@ export namespace uiUtils {
 
         return all;
     }
+
+    export async function listAllIterator<T>(list: (options?: coreClient.OperationOptions) => PagedAsyncIterableIterator<T>, options?: coreClient.OperationOptions): Promise<T[]> {
```
bwateratmsft, line 27: Can this be something we export for reuse by other packages/extensions?

ui/src/azure-account.api.d.ts
```diff
@@ -4,40 +4,61 @@
...
+}
+
+export interface CloudShell {
+	readonly status: CloudShellStatus;
+	readonly onStatusChanged: Event<CloudShellStatus>;
+	readonly waitForConnection: () => Promise<boolean>;
+	readonly terminal: Promise<Terminal>;
+	readonly session: Promise<AzureSession>;
+	readonly uploadFile: (filename: string, stream: ReadStream, options?: UploadOptions) => Promise<void>;
+}
```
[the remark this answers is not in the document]
nturinski: Ah, sure.  I just copied the new api from the Azure Account extension because I didn't really want to mess with it 😅

ui/index.d.ts
```diff
@@ -152,6 +151,18 @@ export interface AzExtServiceClientCredentials {
...
+     * Gets the token provided by this credential.
+     *
+     * This method is called automatically by Azure SDK client libraries. You may call this method
+     * directly, but you must also handle token caching and token refreshing.
+     *
+     * @param scopes - The list of scopes for which the token will have access.
+     * @param options - The options used to configure any requests this
+     *                TokenCredential implementation might make.
+     */
+     getToken(scopes?: string | string[], options?: any): Promise<any | null>;
```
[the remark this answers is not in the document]
nturinski: It's to keep it more loose so that other versions of credentials could be accepted.  For example, `BasicAuthenticationCredentials` returns  `WebResourceLike` whereas the `ServiceClientCredentials` returns `WebResource`.   We originally did that for `signRequest` because different versions of "@azure/ms-rest-js" would change the credential interface.  I figured the same could happen with `getToken`.

Here's an example with the typing screwing up here: https://github.com/Azure/azure-sdk-for-js/issues/10045

ui/index.d.ts
```diff
@@ -1409,7 +1420,8 @@ export interface IMinimumServiceClientOptions {
     requestPolicyFactories?: any[] | ((defaultRequestPolicyFactories: any[]) => (void | any[]));
 }
 
-export type AzExtGenericClientInfo = AzExtServiceClientCredentials | { credentials: AzExtServiceClientCredentials; environment: Environment; } | undefined;
+export type AzExtGenericCredentials = BasicAuthenticationCredentials | TokenCredentials | AzExtServiceClientCredentials;
```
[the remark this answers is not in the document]
nturinski: Because we have two different createClient helper functions, `createGenericClient` and `createAzureClient`.

Originally the azureCredentials were similar enough to basic credential types, but since we're introducing `getToken` as a requirement to create azure clients to ensure T1&T2 compatability, it would force those requirements on any generic client we create 

For example, we used to make a GitHub client with `createGenericClient` but it wouldn't have required `getToken`.

I'm not sure if that 100% made sense so feel free to ping me and I can explain it better 😓

ui/src/azure-account.api.d.ts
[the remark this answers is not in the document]
bwateratmsft: I don't feel strongly, but unless these are reexported from `vscode-azureextensionui`, I'd just leave them...

ui/index.d.ts
```diff
@@ -5,12 +5,14 @@
 
 import type { SubscriptionModels } from '@azure/arm-resources-subscriptions';
 import type { ExtendedLocation, ResourceGroup } from '@azure/arm-resources';
+import * as coreClient from '@azure/core-client';
```
bwateratmsft, line 8: I'd make these two `import type`

ui/index.d.ts
```diff
@@ -5,12 +5,14 @@
...
+import { PagedAsyncIterableIterator } from "@azure/core-paging";
 import type { Environment } from '@azure/ms-rest-azure-env';
 import type { BasicAuthenticationCredentials, HttpOperationResponse, RequestPrepareOptions, ServiceClient, TokenCredentials } from '@azure/ms-rest-js';
+import { StorageAccount } from '@azure/arm-storage';
```
bwateratmsft, line 12: And this one if it can

ui/index.d.ts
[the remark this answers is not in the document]
wwlorey: Is it possible that version of `signRequest` that returns `WebResource` is older? It looks like the current version returns `WebResourceLike`: https://github.com/Azure/ms-rest-js/blob/f6f0d92d79a1dfa8d92ff0891b88bc6b7a349e69/lib/credentials/serviceClientCredentials.ts

ui/index.d.ts
```diff
@@ -140,20 +140,34 @@ export abstract class SubscriptionTreeItemBase extends AzExtParentTreeItem {
     constructor(parent: AzExtParentTreeItem, subscription: ISubscriptionContext);
 }
 
+/**
+ * Loose type to use for T1 and T2 versions of "@azure/ms-rest-js".  The Azure Account extension returns
+ * credentials that will satisfy both T1 and T2 requirements
+ */
+export type AzExtServiceClientCredentials = AzExtServiceClientCredentialsT1 | AzExtServiceClientCredentialsT2;
```
bwateratmsft, line 147: If it satisfies both should it be this?
```suggestion
export type AzExtServiceClientCredentials = AzExtServiceClientCredentialsT1 & AzExtServiceClientCredentialsT2;
```

(Or maybe it's `AzExtGenericCredentials` that should be the union)

nturinski: Absolutely right.  Thanks, this is why you don't code during stand up lmao

wwlorey (approved):
giphy

PR #1054, opened 2022-02-01
Split ui package into utils and azureutils packages

alexweininger:
Working towards #1051 

### Work timeline

* Create @microsoft/vscdoe-azext-azureutils package **done in this PR**
* Rename @vscode-azureextensionui to @microsoft/vscode-azext-utils **done in this PR**
* Publish both packages to npm, and begin migrating extensions
* Rename utils (ui package) exports  to more generic names. ex: `AzureNameStep` → `AzExtNameStep`
* Further refactoring if needed...

### Changes

`ui/src/index.ts` shows a nice summary of what's been relocated (see `ui/index.d.ts` for more detailed summary).

The following has moved from ui to new azure utils package:
* createAzureClient
* openInPortal
* AzureAccountTreeItemBase
* SubscriptionTreeItemBase
* LocationListStep
* ResourceGroupCreateStep
* ResourceGroupListStep
* ResourceGroupNameStep
* StorageAccountCreateStep
* StorageAccountListStep
* StorageAccountNameStep
* VerifyProvidersStep
* uiUtils (listAllIterator)

 The following have been exported from ui and imported by azure utils package:
* openUrl
* addExtensionValueToMask
* callWithMaskHandling
* maskValue
* nonNull utils

And I added `registerAzureUtilsExtensionVariables` which is a carbon copy of `registerAppServiceExtensionVariables`.

### Migrating

I've migrated and tested locally Docker and VMs to use both of the new packages. See what those changes look like:
* https://github.com/microsoft/vscode-docker/compare/alex/azureUtils?expand=1
* https://github.com/microsoft/vscode-azurevirtualmachines/compare/main...alex/t2-split?expand=1 **updated**

### Notes

I plan for the new `@microsoft/vscode-azext-utils` package to live in `utils`. However, I've left the changes in the `ui` folder so you can see diff on GitHub. Once approved I'll revert the changes made in the `ui` folder. That way we're left with an unchanged version of the `vscode-azureextensionui` package code in case we need to make fixes/changes in there later.

PR #1057, opened 2022-02-07
uiUtils.listAllIterator accepts PagedIterator as param rather than function

nturinski:
Realized that this is the way more straightforward approach to a helper function.  Basically, just pass the iterator rather than trying to call it in the helper-- it just complicates things and this probably how it was intended to be used.

The iterator can also be used to load small batches using `.next()`.  I'm sure there will be cases where we don't want to load everything all at once for perf reasons.

### DIFF
--- a/ui/index.d.ts
+++ b/ui/index.d.ts
@@ -1683,5 +1683,5 @@ export declare namespace AzExtFsExtra {
 }
 
 export declare namespace uiUtils {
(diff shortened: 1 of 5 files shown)


alexweininger:
Changes look good, but after #1054 is merged you'll have to make the changes in both the ui (old package) and utils/azure (new) packages. I hope we don't need to maintain both packages for very long.

nturinski:
I feel like this change is small enough to slip in really quickly, but I'm at the mercy of the approval system 🥺

alexweininger (approved):

PR #1062, opened 2022-02-10
Add direct dependency on `@azure/ms-rest-nodeauth`

bwateratmsft:
While working on the integration for `vscode-docker`, I hit a webpacking error, because `@azure/ms-rest-nodeauth` was not making it through. Evidently, it _was_ a dev dependency for `@microsoft/vscode-azext-azureutils`, which is why build etc. worked, but it _should_ be a regular dependency, because it is directly referenced [here](https://github.com/microsoft/vscode-azuretools/blob/main/azure/src/tree/AzureAccountTreeItemBase.ts#L14).

Additionally, while looking into it I saw that an `import` in the Account API could be changed to `import type`.

### DIFF
--- a/azure/package-lock.json
+++ b/azure/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "@microsoft/vscode-azext-azureutils",
-    "version": "0.0.2",
+    "version": "0.0.3",
     "lockfileVersion": 2,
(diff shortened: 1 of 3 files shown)


alexweininger (approved):
Oops!

PR #1067, opened 2022-02-14
Set target population for insiders

bwateratmsft:
And get rid of this detestable git hook in triage-actions


wwlorey (approved):

bwateratmsft:
I'm going to do one more rather important change, so I'm going to reset @wwlorey's feedback. I bumped the version of the utils package to `0.1.0`. The caret dependency syntax considers the leftmost nonzero number to be like a major version. For changes like this that aren't breaking, I'd rather do a version number like `0.1.1`, so that `npm update` would pick it up automatically, instead of `0.0.3` where `npm update` will _not_ pick it up.

wwlorey (approved):

Issue #1073, opened 2022-02-14
Don't double export from the split UI packages


wwlorey:
I noticed this when working on migrating the storage extension to use the split packages. We can easily work around it for now but long term it's something we should fix.

<img width="834" alt="Screen Shot 2022-02-14 at 2 23 36 PM" src="https://user-images.githubusercontent.com/22795803/153958725-9ecded8e-8a0b-4bc5-92fd-7ee0a4eb4249.png">

bwateratmsft:
Sounds good. I don't think any of those things have actual Azure dependencies so we should be able to include them only in `@microsoft/vscode-azext-utils`.

PR #1074, opened 2022-02-15
Change to make `@azure/ms-rest-nodeauth` a dev dependency again

bwateratmsft:
This change allows us to make `@azure/ms-rest-nodeauth` a dev dependency once again, which I like because it depends on `adal-node` (which is deprecated although it doesn't say so on NPM), and `uuid@3` which is also deprecated. Changing this to a dev dependency means significantly less packages transit as dependencies into users of this package.

Since I was in the neighborhood I added a fix for #1073 as well.


azure/src/tree/AzureAccountTreeItemBase.ts
```diff
@@ -120,19 +119,12 @@ export abstract class AzureAccountTreeItemBase extends AzExtParentTreeItem imple
...
-
-                    // these properties don't exist on TokenCredentials
-                    if (filter.session.credentials2 instanceof DeviceTokenCredentials) {
-                        addExtensionValueToMask(
-                            filter.session.credentials2.clientId,
-                            filter.session.credentials2.domain);
-                    }
+                    // these properties don't exist on TokenCredentials, but do exist on DeviceTokenCredentials
+                    // addExtensionValueToMask gracefully handles `undefined` input by ignoring it
+                    addExtensionValueToMask(
```
bwateratmsft, line 124: Turns out, we didn't need `instanceof DeviceTokenCredentials` since, if the properties aren't there they will simply be `undefined`, which `addExtensionValueToMask` gracefully handles.

Aside from that, the type was already `TokenCredentialsBase & TokenCredential`, emphasis **&**, which suggests to me it should always be defined anyway. Though if it isn't, that's OK, it will still work (because `addExtensionValueToMask` handles falsy values gracefully).

Issue #1052, opened 2022-01-28
Publish packages under the @microsoft scope


alexweininger:
Hoping to make this change together with #1051 

- [x] utils → [@microsoft/vscode-azext-utils](https://www.npmjs.com/package/@microsoft/vscode-azext-utils)
- [x] azure → [@microsoft/vscode-azext-azureutils](https://www.npmjs.com/package/@microsoft/vscode-azext-azureutils)
- [x] appservice → [@microsoft/vscode-azext-azureappservice](https://www.npmjs.com/package/@microsoft/vscode-azext-azureappservice)
- [x] dev
- [x] ui (going to deprecate it so not sure if we need to move)
- [ ] kudu
- [ ] database (also unsure of where we are with this package)

bwateratmsft:
Can we mark this done or do we want to get a few more of them under @microsoft? e.g. dev, appservice, etc.

alexweininger:
> Can we mark this done or do we want to get a few more of them under @microsoft? e.g. dev, appservice, etc.

Sounds good, I added a task list.

bwateratmsft:
I think the `ui` folder should probably just get deleted. If we need to create a servicing update for some reason we can go back to a Git tag. If we keep the folder it will just continually generate CG alerts.

nturinski:
> I think the `ui` folder should probably just get deleted. If we need to create a servicing update for some reason we can go back to a Git tag. If we keep the folder it will just continually generate CG alerts.

Yeah, I think we may as well.  Since the package exist in npm, it's not like deleting the folder is going to break users who happen to be consuming.

alexweininger:
I'm always down for some deleting! I'll get on it 🏃‍♂️

alexweininger:
Closing as we've done the main ones

PR #1079, opened 2022-02-16
Remove ui directory

alexweininger:
As discussed in #1052, we decided to delete the ui directory.

I *think* I removed everything that referenced the ui folder besides the root readme. It's pretty outdated since we split the packages and I'll update that later.


.azure-pipelines/compliance/CredScanSuppressions.json
```diff
@@ -30,16 +34,20 @@
             "_justification": "No need to scan VS Code test folder."
         },
         {
-            "folder": "ui\\.vscode-test\\",
+            "folder": "utils\\.vscode-test\\",
```
bwateratmsft, line 37: Out of curiosity, did CredScan actually find something there?

bwateratmsft (approved):

.azure-pipelines/compliance/CredScanSuppressions.json
alexweininger: Good point, it hasn't. Should I remove that until it does?

bwateratmsft: I vote yes. Can't hurt.

bwateratmsft (approved):

PR #1087, opened 2022-03-02
Fix getting locations by using version 2020-09-01

alexweininger:
Fixes https://github.com/microsoft/vscode-azurefunctions/issues/3054 and probably a yet-to-be-discovered bug in vscode-azureappservice.

I think this is a bug with the newer versions of the Azure API used by default in the new SDKs. I filed an issue: https://github.com/Azure/azure-rest-api-specs/issues/18071

As a temporary fix, I'm making a request using a generic service client and setting the `api-version` to `2020-09-01`.

### DIFF
--- a/appservice/src/createAppService/setLocationsTask.ts
+++ b/appservice/src/createAppService/setLocationsTask.ts
@@ -36,6 +36,7 @@ export async function getWebLocations(context: IAppServiceWizardContext): Promis
         options.sku = <SkuName>context.newPlanSku.tier.replace(/\s/g, '');
     }
 
+    // Temporary fix for https://github.com/Azure/azure-rest-api-specs/issues/18071
     const genericClient = await createGenericClient(context, context);
     const result: HttpOperationResponse = await genericClient.sendRequest({
         method: 'GET',
--- a/appservice/src/createAppService/setLocationsTask.ts
+++ b/appservice/src/createAppService/setLocationsTask.ts
@@ -47,5 +47,5 @@ export async function getWebLocations(context: IAppServiceWizardContext): Promis
         }
     });
 
-    return (<GeoRegionJsonResponse>result.parsedBody).value.map((l: GeoRegion) => nonNullProp(l, 'name')) as string[];
+    return (<GeoRegionJsonResponse>result.parsedBody).value.map((l: GeoRegion) => nonNullProp(l, 'name'));
 }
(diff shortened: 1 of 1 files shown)


appservice/src/createAppService/setLocationsTask.ts
```diff
@@ -32,7 +32,15 @@ export async function getWebLocations(context: IAppServiceWizardContext): Promis
         options.sku = <SkuName>context.newPlanSku.tier.replace(/\s/g, '');
     }
 
-    const client: WebSiteManagementClient = await createWebSiteClient(context);
-    const locations = await uiUtils.listAllIterator(client.listGeoRegions(options));
-    return locations.map(l => nonNullProp(l, 'name'));
+    const genericClient = await createGenericClient(context, context);
```
nturinski, line 35: Add the github issue as a comment so we know to revisit this when it's resolved.

appservice/src/createAppService/setLocationsTask.ts
```diff
@@ -32,7 +36,15 @@ export async function getWebLocations(context: IAppServiceWizardContext): Promis
...
+    const result: HttpOperationResponse = await genericClient.sendRequest({
+        method: 'GET',
+        pathTemplate: `/providers/Microsoft.Web/geoRegions`,
+        queryParameters: {
+            'api-version': '2020-09-01',
+            ...options,
+        }
+    });
+
+    return (<GeoRegionJsonResponse>result.parsedBody).value.map((l: GeoRegion) => nonNullProp(l, 'name')) as string[];
```
wwlorey, line 50: This line scares me a bit 😱 I think this accomplishes the same thing with less typecasting:
```ts
    return result.parsedBody?.value.map((l: any) => nonNullProp(l, 'name')) || [];
```

Or maybe we throw an error if `parsedBody` doesn't have a `value` property so we can track it

alexweininger: I removed the `as string[]`, it wasn't actually needed. We cast `parsedBody` quite a bit:
https://github.com/microsoft/vscode-azuretools/blob/83d87d1a9a51849cbdf93a7837c02045853c47a7/appservice/src/SiteClient.ts#L316-L323

nturinski (approved):

nturinski:
Don't forget to bump the version though

alexweininger:
> Don't forget to bump the version though

Oh, I did it a separate PR just to give you more work 🤪
https://github.com/microsoft/vscode-azuretools/pull/1088

PR #1092, opened 2022-03-07
Update dev to 1.0.0 to avoid GitHub tag conflicts

nturinski:
Because the dev package is still technically "dev" in our pipelines, it causes issues with the GitHub releases since the dev-0.1.x tag is already in use.  We could either a.) just keep incrementing it from where we were before (0.10.4) or start over at 1.0.0.  @alexweininger and I thought it'd be less confusing to just start at 1.0.0

### DIFF
--- a/dev/package-lock.json
+++ b/dev/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "@microsoft/vscode-azext-dev",
-    "version": "0.1.2",
+    "version": "1.0.0",
     "lockfileVersion": 2,
     "requires": true,
     "packages": {
         "": {
             "name": "@microsoft/vscode-azext-dev",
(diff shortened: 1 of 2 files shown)


nturinski:
Should we just rename the folders to something else? I think that's how the tag is being generated

bwateratmsft:
I'd rather alter the tag generation logic. We could add something at the beginning of all future tags.

nturinski:
> Same issue with the appservice package. I've been manually altering the tags after the release with the intention to fix it someday.

Also not a huge deal, but does that mean the old releases (the real appservice v0.1.0) doesn't have a tag anymore?

nturinski:
Closing due to other proposed solution

PR #1095, opened 2022-03-23
Nat/resolved children for picker

nturinski:
Allows parent trees to provide their own array of `AzExtTreeItem` for when `showTreeItemPicker` is invoked.  This is mostly useful for the `SubscriptionTreeItem` because its children are technically the `orderedBy` groups.  It's more useful to just have the children be the actual resources so that we can more easily filter for all function apps or static web apps.

### DIFF
--- a/appservice/package-lock.json
+++ b/appservice/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "@microsoft/vscode-azext-azureappservice",
-    "version": "0.1.5",
+    "version": "0.1.7",
     "lockfileVersion": 2,
     "requires": true,
     "packages": {
         "": {
             "name": "@microsoft/vscode-azext-azureappservice",
-            "version": "0.1.5",
+            "version": "0.1.7",
             "license": "MIT",
             "dependencies": {
                 "@azure/abort-controller": "^1.0.4",
@@ -27,7 +27,7 @@
                 "p-retry": "^3.0.1",
                 "portfinder": "^1.0.25",
                 "pretty-bytes": "^5.3.0",
-                "simple-git": "^1.132.0",
+                "simple-git": "^3.3.0",
                 "vscode-azurekudu": "^0.2.0",
                 "vscode-nls": "^5.0.0",
                 "websocket": "^1.0.31",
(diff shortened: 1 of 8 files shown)


alexweininger:
Do you have a branch showing an example using the new API?

nturinski:
> Do you have a branch showing an example using the new API?

https://github.com/microsoft/vscode-azureresourcegroups/tree/nat/treeItemPicker

bwateratmsft (changes requested):
I guess I'd like to see the experiences side-by-side; in the Docker extension it still requires the user to go through the groups.

In any case, can't we implement this using the existing `pickTreeItemImpl` method? It should be simple enough to implement that on the `SubscriptionTreeItem` within the RG extension.

bwateratmsft:
Can't `pickTreeItemImpl` look at anything it wants to? `SubscriptionTreeItem` is in possession of all of the resource nodes (`this.rgsItem`), so we can make any quick pick we want out of that.

> To me, this seemed like an implementation that should live in the RG extension. Its `resolveChildrenForTreePicker` would make sure that all resources for the provider being searched would be resolved before returning its array of children.

Agreed, it should live in the RG extension, but I think it might be one _more_ reason to use `pickTreeItemImpl`--since we control the implementation, we can resolve that node after they pick it, but before returning the result from `pickTreeItemImpl`.

nturinski:
> Can't `pickTreeItemImpl` look at anything it wants to? `SubscriptionTreeItem` is in possession of all of the resource nodes (`this.rgsItem`), so we can make any quick pick we want out of that.

Technically, yes but when `showTreeItemPicker` is invoked, it calls `pickChildTreeItem` which calls this block
```
if (this.pickTreeItemImpl) {
            const children: AzExtTreeItem[] = await this.getCachedChildren(context);
            const pickedItem: AzExtTreeItem | undefined = await this.pickTreeItemImpl(expectedContextValues);
            if (pickedItem) {
                const child: AzExtTreeItem | undefined = children.find((ti: AzExtTreeItem) => ti.fullId === pickedItem.fullId);
                if (child) {
                    return child;
                }
            }
        }
```

If it can't find the node we're looking for in its immediate children which is generated by `loadMoreChildren` or `_cachedChildren` , it just drops out of this conditional and goes through the normal `treeItemPicker` logic of listing any children that are `AzExtParentTreeItems` of the treeItem.  I can't safely mess with these children arrays because they're being used by the `treeProvider` to create the `treeView`.

`pickTreeItemImpl` is expecting only one tree item to be picked so that it can drill into it so it wouldn't resolve the issue where if the `orderBy` property is anything other than `resourceType`, the user would have to drill into a resource group or a location node before seeing any of their resource apps.

> As it is, if we resolve all of the nodes, that means we're resolving everything in the subscription...that's quite a bit.

I was thinking that we would only resolve the nodes of the extension that called `pickTreeItem`.  So, if you use the command palette to call up `Azure Functions: Start Function App`, the Subscription node would only resolve nodes from the registered function provider.

bwateratmsft:
> If it can't find the node we're looking for in its immediate children which is generated by `loadMoreChildren` or `_cachedChildren` , it just drops out of this conditional and goes through the normal `treeItemPicker` logic

Interesting; I'll take a look into the code some. It seems strange to me that `pickChildTreeItem` is allowing the tree item to supply any arbitrary node, but then just using its ID to find a matching child node and discarding it. What would happen if we just accepted the return result?

> I was thinking that we would only resolve the nodes of the extension that called `pickTreeItem`

Well, unless we add an argument to `showTreeItemPicker` we won't know which extension is calling...just the context value(s) it's looking for, which aren't a strong indicator. I don't think this will be too hard of an issue either way; at some point we'll end up with either a list of `AppResourceTreeItem` or a single one; they know their own resolver so we can just call `resolve` on them.

bwateratmsft:
Looking back the code was basically always like this--initially introduced in [this PR](https://github.com/microsoft/vscode-azuretools/pull/18/files#diff-4b3b2e2ba4e3548c2007f779a2aead6defa7f04e54e23ffbe15a95c4a76fb2d1). But, back then, the method was called `pickChildNode` and had to return an `AzureNode<T>`, which was _not_ a `TreeItem`. Now, it's `pickChildTreeItem` and it _is_ returning a `TreeItem`--I think we can alter the body of `pickChildTreeItem` to unconditionally accept the return result of `pickChildTreeItemImpl` (if that method is defined of course).

nturinski:
Discussed offline: closing this because we decided to not by-pass group nodes when not grouped by resourceType

PR #1106, opened 2022-04-05
Remove `websocket` in favor of `ws` (non-stream version)

bwateratmsft:
`websocket` has a malicious dependency which we are required to remove. No already-shipped code uses the malicious version so there is no need to service any existing releases.

One option would be to pin to the non-malicious version of the dependent package, but I feel like that's not a viable long-term solution; it pollutes our `package.json` and doesn't solve the general problem of poorly-maintained dependencies. It doesn't look like `websocket` is super well-maintained--been more than a year since it's had any meaningful changes, and a year since a release--so I'm more inclined to switch to `ws`, which is more freshly-maintained and has zero dependencies (yay!).

~~Also bumps `simple-git` to `^3.5.0`.~~ Done in #1108 

This PR corresponds to #1109, and is an alternate form of it that uses the older non-stream version.

### DIFF
--- a/appservice/src/TunnelProxy.ts
+++ b/appservice/src/TunnelProxy.ts
@@ -10,7 +10,7 @@ import { IActionContext, IParsedError, nonNullProp, parseError, UserCancelledErr
 import * as EventEmitter from 'events';
 import { createServer, Server, Socket } from 'net';
 import { CancellationToken, Disposable } from 'vscode';
-import * as websocket from 'websocket';
+import * as ws from 'ws';
 import { ext } from './extensionVariables';
 import { localize } from './localize';
 import { ParsedSite } from './SiteClient';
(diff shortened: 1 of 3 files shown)


appservice/src/TunnelProxy.ts
```diff
@@ -59,64 +57,59 @@ class TunnelSocket extends EventEmitter {
             this.emit('error', err);
         });
 
-        this._wsClient.on('connect', (connection: websocket.connection) => {
+        this._websocket = new ws.WebSocket(
```
bwateratmsft, line 60: @nturinski I don't have a clue how to test this properly; do you mind doing so?

bwateratmsft, line 60: Rather than all this, would it be possible to use `ws.createWebSocketStream` and tie up the streams to the input `Socket` object?

nturinski (approved):
Based on the testing, I think it worked:

image

Just the SSH command itself isn't working.  This is probably more related to the proposed api not letting us use `onDidWriteTerminalData`

nturinski:
Confirmed remote debugging doesn't work on this.  SSH probably doesn't either since it never actually connected end-to-end, but I think we're proceeding in favor of #1109 anyway

appservice/src/TunnelProxy.ts
bwateratmsft: Yes: #1109

PR #1109, opened 2022-04-06
Remove `websocket` in favor of `ws` (stream version)

bwateratmsft:
`websocket` has a malicious dependency which we are required to remove. No already-shipped code uses the malicious version so there is no need to service any existing releases.

One option would be to pin to the non-malicious version of the dependent package, but I feel like that's not a viable long-term solution; it pollutes our `package.json` and doesn't solve the general problem of poorly-maintained dependencies. It doesn't look like `websocket` is super well-maintained--been more than a year since it's had any meaningful changes, and a year since a release--so I'm more inclined to switch to `ws`, which is more freshly-maintained and has zero dependencies (yay!).

This PR corresponds to #1106, and is an alternate form of it that uses streams.

### DIFF
--- a/appservice/package-lock.json
+++ b/appservice/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "@microsoft/vscode-azext-azureappservice",
-    "version": "0.1.7",
+    "version": "0.1.8",
     "lockfileVersion": 2,
     "requires": true,
     "packages": {
         "": {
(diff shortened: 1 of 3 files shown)


bwateratmsft:
Verified it is working, including getting a working SSH window.

nturinski:
Also verified that remote debugging is working with a Node v16 application.

nturinski (approved):
:shipit:

PR #1110, opened 2022-04-11
Replace `portfinder` with `get-port`

bwateratmsft:
Fixes CVE-2021-43138 by making `async` a dev dependency only. `get-port` has no dependencies (YAY) and appears better maintained. `portfinder` has had no meaningful updates in two years so I'm pessimistic for a fix there.

### DIFF
--- a/appservice/package-lock.json
+++ b/appservice/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "@microsoft/vscode-azext-azureappservice",
-    "version": "0.1.8",
+    "version": "0.1.9",
     "lockfileVersion": 2,
     "requires": true,
     "packages": {
         "": {
             "name": "@microsoft/vscode-azext-azureappservice",
-            "version": "0.1.8",
+            "version": "0.1.9",
             "license": "MIT",
             "dependencies": {
                 "@azure/abort-controller": "^1.0.4",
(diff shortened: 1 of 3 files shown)


frank-dspeed:
Sorry to say that here but i saw this linked in a other issue where i am stakeholder and your conclusion that you should not use and can not use ESM is wrong. ESM Works as long as you follow the ESM Rules 

dynamic import inside CJS Context works without problems as long as you do not use NATIV Code bindings in the ESM Code

when you plan to work with NATIV Code + ESM Code Bootstrap in CJS Like this

my.cjs or my.cts
```
const electronNativ = require('electron')
// use electron some how assign const

const aElectronMethodForESM 
globalThis.aElectronMethodForESM = aElectronMethodForESM
/* if this would be in a CJS file for electron bootstrap this already would be a global var without use strict and could already get accessed by esm*/

const myESM = import('./my-esm.js') // or .mjs depends on your package.json defaults (type) fild
myESM.then( moduleNamespace => moduleNamespace.init(aElectronMethodForESM))
```

my-esm.js
```
const init = (aElectronMethodForESM) => // using electron in msn
export { init}
```

for the context and browser problems always use messaging and as we know from the example above esm and CJS do share the same globalThis inside the instance hope that helps and makes some sense to you

PR #1111, opened 2022-04-11
Add utility for finding free port

bwateratmsft:
Needed to remove `portfinder` dependency in `appservice` package.

### DIFF
--- a/utils/src/utils/randomUtils.ts
+++ b/utils/src/utils/randomUtils.ts
@@ -14,4 +14,12 @@ export namespace randomUtils {
         const buffer: Buffer = crypto.randomBytes(Math.ceil(length / 2));
         return buffer.toString('hex').slice(0, length);
     }
+
+    export function getRandomInteger(minimumInclusive: number, maximumExclusive: number): number {
+        if (maximumExclusive <= minimumInclusive) {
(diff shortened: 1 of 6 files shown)


utils/src/utils/randomUtils.ts
```diff
@@ -14,4 +14,12 @@ export namespace randomUtils {
         const buffer: Buffer = crypto.randomBytes(Math.ceil(length / 2));
         return buffer.toString('hex').slice(0, length);
     }
+
+    export function getRandomInteger(minimumInclusive: number, maximumExclusive: number): number {
+        if (maximumExclusive <= minimumInclusive) {
+            throw new Error(`Maximum must be strictly greater than minimum`);
```
wwlorey, line 20: Would it be better to fall back to a default range in this case instead of throwing an error?

wwlorey (approved):

utils/src/utils/randomUtils.ts
bwateratmsft: IMO no. If the goal is to get some random number and they provide invalid inputs, I'd rather it fail entirely (and thus, be more likely to get caught during development) than pick a range. I also don't think there's enough information to arbitrarily pick a range, since there could be valid reasons for picking 1-100 just as much as 1-65535.

PR #1113, opened 2022-04-11
Add ActivityBase class

alexweininger:
https://github.com/microsoft/vscode-azureresourcegroups/pull/160 depends on these changes.


utils/index.d.ts
```diff
@@ -880,6 +880,55 @@ export interface IWizardOptions<T extends IActionContext> {
     showLoadingPrompt?: boolean;
 }
 
+export interface ActivityTreeItemOptions {
+    label: string;
+    contextValuePostfix?: string;
```
bwateratmsft, line 885: Still prefer the `contextValuesToAdd` shape...

utils/index.d.ts
```diff
@@ -880,6 +880,55 @@ export interface IWizardOptions<T extends IActionContext> {
...
+    collapsibleState?: TreeItemCollapsibleState;
+    children?: (parent: AzExtParentTreeItem) => AzExtTreeItem[];
```
bwateratmsft, line 887: Maybe `getChildren` since it's a method?
```suggestion
    getChildren?: (parent: AzExtParentTreeItem) => AzExtTreeItem[];
```

utils/index.d.ts
```diff
@@ -880,6 +880,55 @@ export interface IWizardOptions<T extends IActionContext> {
...
+export type ActivityTask = (progress: Progress<{ message?: string, increment?: number }>) => Promise<void>;
+
+export abstract class ActivityBase implements Activity, ActivityType {
+    public readonly onStart: Event<OnStartActivityData>;
+    public readonly onProgress: Event<OnProgressActivityData>;
+    public readonly onSuccess: Event<OnSuccessActivityData>;
+    public readonly onError: Event<OnErrorActivityData>;
+
+    public readonly task: ActivityTask;
+    public startedAtMs: number;
```
bwateratmsft, line 920: No longer need to track `startedAtMs` here.

utils/index.d.ts
```diff
@@ -891,7 +940,14 @@ export declare class AzureWizard<T extends IActionContext> {
     public constructor(wizardContext: T, options: IWizardOptions<T>);
 
     public prompt(): Promise<void>;
-    public execute(): Promise<void>;
+    public execute(options?: AzureWizardExecuteOptions): Promise<void>;
```
bwateratmsft, line 943: Can we do the changes to `AzureWizard` as a separate PR?

utils/src/activityLog/Activity.ts
```diff
@@ -0,0 +1,87 @@
...
+    public readonly onProgress: typeof this._onProgressEmitter.event;
+    public readonly onSuccess: typeof this._onSuccessEmitter.event;
+    public readonly onError: typeof this._onErrorEmitter.event;
+
+    private readonly _onStartEmitter: EventEmitter<types.OnStartActivityData>;
+    private readonly _onProgressEmitter: EventEmitter<types.OnProgressActivityData>;
+    private readonly _onSuccessEmitter: EventEmitter<types.OnSuccessActivityData>;
+    private readonly _onErrorEmitter: EventEmitter<types.OnErrorActivityData>;
+
+    public running: boolean;
```
bwateratmsft, line 31: To force data consistency, the state should be tracked as one enum, (e.g. `'NotStarted' | 'Running' | 'Success' | 'Failed' | 'Cancelled'`), rather than two booleans.

utils/src/activityLog/Activity.ts
```diff
@@ -0,0 +1,87 @@
...
+    public done: boolean;
+    public error?: types.IParsedError;
+    public readonly task: types.ActivityTask;
+    public startedAtMs: number;
```
bwateratmsft, line 35: Don't need to track `startedAtMs` anymore

utils/src/activityLog/Activity.ts
```diff
@@ -0,0 +1,87 @@
...
+export interface ActivityTreeItemOptions {
+    label: string;
+    contextValuePostfix?: string;
+    collapsibleState?: TreeItemCollapsibleState;
+    children?: (parent: types.AzExtParentTreeItem) => AzExtTreeItem[];
+}
+
+export abstract class ActivityBase implements types.Activity {
+
+    public readonly onStart: typeof this._onStartEmitter.event;
```
bwateratmsft, line 21: This is valid, but another way that saves a few lines here and in the constructor is to init like this:
```typescript
private readonly _onStartEmitter = new EventEmitter<types.OnStartActivityData>();
// ...
private readonly onStart = this._onStartEmitter.event;
```
I don't feel strongly one way or the other so do whichever you prefer :smile:

utils/index.d.ts
bwateratmsft, line 887: I wonder if it should be `AzExtTreeItem[] | Promise<AzExtTreeItem[]>`. There might be a case where they need to do something async.

utils/index.d.ts
nturinski: This just occurred to me, but the `contextValuesToAdd` approach could be difficult in scenarios where we want to change a specific state.  For example, if I `restart` an app, we add the context value `restarting`.  But when it resolves, it should have the context value of `running`.  In that case, how do we easily figure out that we need to remove the `restarting` context value as well?

I thought about this because I can imagine Alex was using this approach for similar reasons.

utils/src/activityLog/Activity.ts
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    public report(progress: { message?: string; increment?: number }): void {
+        this.progress.unshift(progress);
+        this._onProgressEmitter.fire({ ...this.getState(), message: progress.message });
+    }
+
+    public async run(): Promise<void> {
+        try {
+            this._onStartEmitter.fire(this.getState());
```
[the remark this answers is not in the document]
nturinski: One thing about using `CancellationToken`-- a lot of ARM calls don't support being canceled as far as I know. So I think a cancel in most cases would be just canceling it in the UI.

utils/index.d.ts
```diff
@@ -880,6 +880,55 @@ export interface IWizardOptions<T extends IActionContext> {
...
+export interface ActivityTreeItemOptions {
+    label: string;
+    contextValuePostfix?: string;
+    collapsibleState?: TreeItemCollapsibleState;
+    children?: (parent: AzExtParentTreeItem) => AzExtTreeItem[];
+}
+
+type ActivityEventData<T> = ActivityTreeItemOptions & T;
+
+export type OnStartActivityData = ActivityEventData<{}>;
```
nturinski, line 892: Do we need to have a type for each activity event?  VS Code seems to just have all the `onEvents` have the typing of an `Event<any>` or `Event<void>`

alexweininger: The specific types are for the event datas, I wrote it like this with the intention on making it easy to add things to the data of each event later if needed.

utils/src/activityLog/Activity.ts
nturinski: Sorry, I'm not really following why we wouldn't need to track this anymore. We could use this for timeouts of activities.

utils/src/activityLog/Activity.ts
```diff
@@ -0,0 +1,87 @@
...
+            await this.task({ report: (progress) => this.report(progress) });
+            this.done = true;
+            this._onSuccessEmitter.fire(this.getState());
+        } catch (e) {
+            this.error = parseError(e);
+            this.done = true;
```
nturinski, line 75: Small thing, but you can put the `this.done` in a finally clause.

utils/src/activityLog/activities/ExecuteActivity.ts
```diff
@@ -0,0 +1,80 @@
...
+import { TreeItemCollapsibleState } from "vscode";
+import * as types from '../../../index';
+import { AppResource } from "../../../unified";
+import { AzExtParentTreeItem } from "../../tree/AzExtParentTreeItem";
+import { GenericTreeItem } from "../../tree/GenericTreeItem";
+import { nonNullProp } from "../../utils/nonNull";
+import { ActivityBase, ActivityTreeItemOptions } from "../Activity";
+
+
+interface ExecuteActivityData<C extends types.IActionContext> {
```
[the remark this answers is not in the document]
nturinski: That makes sense though we generally end up doing that in the `context` as a property that the user can define.  Most of our wizards follow this pattern and I wonder if that was what Alex had in mind.

utils/src/activityLog/activities/ExecuteActivity.ts
```diff
@@ -0,0 +1,80 @@
...
+                    const result = this.data.context['activityResult'];
+                    const appResource: AppResource = {
+                        id: nonNullProp(result, 'id') as string,
+                        name: nonNullProp(result, 'name') as string,
+                        type: nonNullProp(result, 'type') as string,
+                    }
+
+                    const ti = new GenericTreeItem(parent, {
+                        contextValue: 'executeResult',
+                        label: "Click to view resource",
```
nturinski, line 49: I think this should be a flag that the user passes in.  It wouldn't always make sense to have this, like for a delete as an example.

Don't forget to localize!

utils/src/activityLog/activities/ExecuteActivity.ts
```diff
@@ -0,0 +1,80 @@
...
+    public initialState(): ActivityTreeItemOptions {
+        return {
+            label: this.data.title,
+            collapsibleState: TreeItemCollapsibleState.None,
+        }
+    }
+
+    public successState(): ActivityTreeItemOptions {
+        return {
+            label: this.labelOnDone,
```
nturinski, line 35: I'm not sure I totally understand what the difference between `this.data.title` and `this.labelOnDone `is.  Is it in case `data.title` changes at some point?

utils/src/activityLog/activities/ExecuteActivity.ts
```diff
@@ -0,0 +1,80 @@
...
+import { nonNullProp } from "../../utils/nonNull";
+import { ActivityBase, ActivityTreeItemOptions } from "../Activity";
+
+
+interface ExecuteActivityData<C extends types.IActionContext> {
+    title: string;
+    context: C;
+}
+
+export class ExecuteActivity<C extends types.IActionContext> extends ActivityBase {
```
nturinski, line 20: Would it make sense to make an `ActivityContext` that has `activityResult` as a defined property?

It might easier to define a lot of the state and stuff as well on the context so it's easier to pass around instead of having it all live on the Activity itself.

utils/src/activityLog/Activity.ts
alexweininger: The finally clause isn't guaranteed to be executed since we re-throw the error in the catch block. To guarantee the finally block is always executed regardless of if an error is thrown in the catch block, you must return inside it. I just learned this, and I plan on us forgetting this behavior in the future, so I'm gonna leave it as-is.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch#returning_from_a_finally-block

Finally block will not execute:
```ts
try {
  	throw new Error('hi');
} catch (error) {
  	console.error(error);
  	throw error;
} finally {
  console.log('Finally block executed');
}
```

If we add a `return` to the finally block, it will always execute:
```ts
try {
  	throw new Error('hi');
} catch (error) {
  	console.error(error);
  	throw error;
} finally {
  	console.log('Finally block executed');
	return;
}
```

utils/src/activityLog/Activity.ts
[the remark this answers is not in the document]
bwateratmsft: Yeah. I think we have to leave it up to the client extension though, to decide if/how to deal with cancellation.

utils/src/activityLog/activities/ExecuteActivity.ts
alexweininger: I'm splitting the AzureWizard changes into a new PR, I'll respond to this feedback when I do that.

utils/src/activityLog/Activity.ts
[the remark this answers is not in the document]
alexweininger: - [x] 1. Relating to cancellation, add a `this.cancellationTokenSource` field, which would provide the implementation for the CTS that got defined on the `Activity` interface
- [x] 2. We can remove `this.progress`
- [x] 3. `this.report` can be private
- [x] 4. Pass `this.report` as the first argument to the task, and `this.cancellationTokenSource.token` as the second

alexweininger:
@bwateratmsft I could see us wanting to have completed activities persist across VS Code sessions. Do you think worth rewriting parts of the API to accommodate for that?

We'd have to be able to hydrate completed activities from a saved state in local storage. We could come up with a serializable state that a completed activity can save and be rehydrated from. I can see this getting complicated if we want to persist the child nodes between sessions (a big reason why we might want to have activities persist in the first place).

utils/index.d.ts
bwateratmsft: That's a good point Nathan. We may need to expose the underlying `Set<string>`, with instruction to not remove things you didn't add.

utils/src/activityLog/Activity.ts
bwateratmsft: The purpose of it here was for sorting the tree; which is now handled by `ActivityTreeItem` (in the RG extension). Consumers could still implement their own `Activity` child class that tracks start time and handles timeout.

utils/src/activityLog/Activity.ts
bwateratmsft: Maybe a comment to reduce the chance of us forgetting this? :smile:

bwateratmsft:
@alexweininger that's a good idea.

I think it would be prohibitively difficult to handle _incomplete_ items; I don't think we could come up with any reasonable way to serialize how to update the status after restarting VSCode.

For _complete_ items, as you mention it would not be too difficult to keep just the activity node, but when you start introducing children it gets more complicated quickly.

I'll try to think about it more today. I don't want to bite off more than we can chew right now but I also don't want to paint us into a corner where we can't do that in the future.

utils/index.d.ts
```diff
@@ -880,6 +880,56 @@ export interface IWizardOptions<T extends IActionContext> {
     showLoadingPrompt?: boolean;
 }
 
+export interface ActivityTreeItemOptions {
+    label: string;
+    contextValuesToAdd?: string;
```
bwateratmsft, line 885:
```suggestion
    contextValuesToAdd?: string[];
```

bwateratmsft: (Unless we do the open `Set<string>` thing, in which case this does not matter)

bwateratmsft (approved):

utils/src/activityLog/activities/ExecuteActivity.ts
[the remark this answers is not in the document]
bwateratmsft: My concern is that without looking into the code here, a user of `ExecuteActivity` would not know that filling in that specific property would be the thing that results in a "Click to view resource" child node/command/clickable/doodad.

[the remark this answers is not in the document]
bwateratmsft: (Copied this conversation over to #1117

utils/src/activityLog/activities/ExecuteActivity.ts
bwateratmsft: (Copied to #1117)

utils/index.d.ts
alexweininger: > This just occurred to me, but the contextValuesToAdd approach could be difficult in scenarios where we want to change a specific state. For example, if I restart an app, we add the context value restarting. But when it resolves, it should have the context value of running. In that case, how do we easily figure out that we need to remove the restarting context value as well?

I'm not sure I understand the problem, if the context value string is constructed in the context value getter, we'll always just add `contextValuesToAdd`. I'm going to leave it for now and we can discuss further.

utils/index.d.ts
```diff
@@ -880,6 +880,56 @@ export interface IWizardOptions<T extends IActionContext> {
...
+export declare interface Activity {
+    id: string;
+    cancellationTokenSource?: CancellationTokenSource;
+    onStart: Event<OnStartActivityData>;
+    onProgress: Event<OnProgressActivityData>;
+    onSuccess: Event<OnSuccessActivityData>;
+    onError: Event<OnErrorActivityData>;
+}
+
+interface ActivityType {
```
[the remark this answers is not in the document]
alexweininger: Good point, I removed it

utils/index.d.ts
alexweininger: @nturinski I'm gonna leave the event data names as they are... I like them

nturinski: Do you as you will.... kids these days just don't understand grammar. 🧓

alexweininger: Well to me, it's the data for the "OnStart" event. I guess I think of the event being "OnStart" rather than just "Start"

bwateratmsft: > Do you as you will.... kids these days just don't understand grammar. 🧓

Ok boomer :wink:

utils/src/activityLog/Activity.ts
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    public getState(): types.ActivityTreeItemOptions {
+        switch (this.status) {
+            case ActivityStatus.Failed:
+                return this.errorState(this.error);
+            case ActivityStatus.Succeeded:
+                return this.successState();
+            default:
+                return this.initialState();
```
nturinski, line 77: Wouldn't we also want a "running" or "progressing" state?

nturinski (approved):

PR #1119, opened 2022-04-14
Bump async dependency in appservice package

wwlorey:
Mitigates CVE-2021-43138

### DIFF
--- a/appservice/package.json
+++ b/appservice/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@microsoft/vscode-azext-azureappservice",
     "author": "Microsoft Corporation",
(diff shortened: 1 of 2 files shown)


appservice/package.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "@microsoft/vscode-azext-azureappservice",
     "author": "Microsoft Corporation",
-    "version": "0.1.10",
+    "version": "0.1.11",
```
bwateratmsft, line 4: Since it's a dev dependency, we don't need to release a new version to clear out the CG alert.

bwateratmsft (approved):

PR #1122, opened 2022-04-18
Update collapsibleState value on expand/collapse

alexweininger:
For the unified design in resource groups we check the value of collapsibleState to determine if tree items are visible. This value was never actually being updated, and was only representing the initial collapsibleState of the tree item.

I set it up to listen to the onDidCollapseElement and onDidExpandElement events provided by [TreeView](https://code.visualstudio.com/api/references/vscode-api#TreeView%3CT%3E).

### DIFF
--- a/utils/index.d.ts
+++ b/utils/index.d.ts
@@ -6,7 +6,7 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 
 import type { Environment } from '@azure/ms-rest-azure-env';
-import { CancellationToken, CancellationTokenSource, Disposable, Event, ExtensionContext, FileChangeEvent, FileChangeType, FileStat, FileSystemProvider, FileType, InputBoxOptions, MarkdownString, MessageItem, MessageOptions, OpenDialogOptions, OutputChannel, Progress, QuickPickItem, QuickPickOptions, TextDocumentShowOptions, ThemeIcon, TreeDataProvider, TreeItem, TreeItemCollapsibleState, Uri } from 'vscode';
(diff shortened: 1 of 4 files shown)


utils/index.d.ts
```diff
@@ -25,6 +25,14 @@ export declare class AzExtTreeDataProvider implements TreeDataProvider<AzExtTree
      */
     public constructor(rootTreeItem: AzExtParentTreeItem, loadMoreCommandId: string);
 
+    /**
+     * Allows the data provider to listen to the {@link TreeView.onDidExpandElement} and {@link TreeView.onDidCollapseElement} events.
+     *
+     * This will ensure that the {@link AzExtParentTreeItem.collapsibleState} property is updated for the element.
+     * @param treeView The {@link TreeView} associated with this data provider
+     */
+    public setupCollapsibleStateListeners(treeView: TreeView<AzExtTreeItem>): Disposable;
```
bwateratmsft, line 34: So is the issue that VSCode does not update the `collapsibleState` property on the `TreeItem`, when the tree item gets expanded by the user?

alexweininger: Correct

bwateratmsft, line 34: Seems like this should be implemented in an internal fashion; i.e. creating the tree data provider would also create these listeners and it would automatically keep the state of the object up-to-date, without consumers needing to call this. That leaves dealing with the disposables which I'm not immediately sure how to do.

alexweininger:
Closing in favor of #1150

PR #1124, opened 2022-04-20
Split out deleteSite into wizard steps

alexweininger:
Splitting this into steps makes it a lot simpler to integrate with the activity panel. We'll use these steps along with the new ConfirmationStep in azureutils to delete Web Apps and Function apps.

* Bumped minor version, since this is a breaking change
* Split out into two steps:
	* DeleteLastServicePlanStep
	* DeleteSiteStep


appservice/src/deleteSite/IDeleteSiteWizardContext.ts
```diff
@@ -0,0 +1,12 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.txt in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import { ExecuteActivityContext, IActionContext } from "@microsoft/vscode-azext-utils";
+import { ParsedSite } from "../SiteClient";
+
+export interface IDeleteSiteWizardContext extends IActionContext, ExecuteActivityContext {
```
bwateratmsft, line 9: In [Nathan's PR](https://github.com/microsoft/vscode-azurefunctions/pull/3118/files#diff-ebdac05616fdc2bd7e3e037cbf5b4e6899f00a1a0742acd7045f066825b4612e) there's a similar interface that also includes `resourceGroupToDelete`, it may be helpful to combine these two in this PR where they can be shared.

appservice/src/deleteSite/DeleteSiteStep.ts
```diff
@@ -0,0 +1,41 @@
...
+import { IDeleteSiteWizardContext } from "./IDeleteSiteWizardContext";
+
+export class DeleteSiteStep extends AzureWizardExecuteStep<IDeleteSiteWizardContext> {
+    public priority: number = 100;
+
+    public async execute(context: IDeleteSiteWizardContext, progress: Progress<{ message?: string | undefined; increment?: number | undefined; }>): Promise<void> {
+
+        const site = nonNullProp(context, "site");
+
+        const deleting: string = site.isSlot ?
```
bwateratmsft, line 19: This is a little tough to follow, might be better to do something like this:

```typescript
let deleting: string;
let deleteSucceeded: string;

if (site.isSlot) {
    deleting = ...
    deleteSucceeded = ...
} else if (site.isFunctionApp) {
    ...
} else {
    ...
}
```

appservice/src/deleteSite.ts
```diff
@@ -1,56 +0,0 @@
...
-
-import type { AppServicePlan } from '@azure/arm-appservice';
-import { DialogResponses, IActionContext } from '@microsoft/vscode-azext-utils';
-import { isNullOrUndefined } from 'util';
-import * as vscode from 'vscode';
-import { ext } from './extensionVariables';
-import { localize } from './localize';
-import { ParsedSite } from './SiteClient';
-
-export async function deleteSite(context: IActionContext, site: ParsedSite): Promise<void> {
```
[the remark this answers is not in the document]
alexweininger: I'll sync with Nathan offline, I think we'll be able to easily use these changes in https://github.com/microsoft/vscode-azurefunctions/pull/3118

appservice/src/deleteSite/DeleteSiteStep.ts
alexweininger: Agreed, and fixed 🤠

bwateratmsft (approved):
LGTM, just check with Nathan on that one PR.

PR #1125, opened 2022-04-20
Add ConfirmationStep

alexweininger:


azure/index.d.ts
```diff
@@ -418,3 +418,14 @@ export function createAzureSubscriptionClient<T>(context: AzExtClientContext, cl
 export declare namespace uiUtils {
     export function listAllIterator<T>(iterator: PagedAsyncIterableIterator<T>): Promise<T[]>
 }
+
+export declare class ConfirmationStep<T extends IActionContext> extends AzureWizardPromptStep<T> {
```
bwateratmsft, line 422: IMO this should go into the `utils` package. This could be useful to any extension using the wizard framework, irrespective of Azure.

bwateratmsft: We could do a generic `ConfirmationStep` that has "Yes" and "No" options, along with a `DeleteConfirmationStep` with a "Delete" and "Cancel" or whatever the preferred two buttons are.

alexweininger: 👍 moved them into utils

utils/index.d.ts
```diff
@@ -1374,3 +1374,23 @@ export function nonNullOrEmptyValue(value: string | undefined, propertyNameOrMes
...
+    public constructor(message: string);
+    public prompt(wizardContext: IActionContext): Promise<void>;
+    public shouldPrompt(wizardContext: IActionContext): boolean;
+}
+
+/**
+ * @param message Message to display in the confirmation modal
+ * ex: `Are you sure you want to delete function app "{0}"?`
+ */
+export declare class DeleteConfirmationStep extends AzureWizardPromptStep<IActionContext> {
```
bwateratmsft, line 1392: It doesn't really make much of a difference but I guess I was envisioning that `DeleteConfirmationStep` would inherit from `ConfirmationStep`, with it just overriding `prompt()` to show different buttons.

utils/src/wizard/ConfirmationStep.ts
```diff
@@ -0,0 +1,21 @@
...
+import * as types from "../../index";
+import { DialogResponses } from "../DialogResponses";
+import { AzureWizardPromptStep } from "./AzureWizardPromptStep";
+
+export class ConfirmationStep extends AzureWizardPromptStep<types.IActionContext> {
+
+    public constructor(private readonly message: string) { super(); }
+
+    public async prompt(context: types.IActionContext): Promise<void> {
+        await context.ui.showWarningMessage(this.message, { modal: true, stepName: 'confirmation' }, DialogResponses.yes, DialogResponses.no);
```
bwateratmsft, line 15: Now that I'm looking at it, I'm not sure how this in particular would work; the `yes` / `no` response isn't carried anywhere. It's different from `DeleteConfirmationStep` which would simply pass if yes, and otherwise throw `UserCancelledError`.

In retrospect maybe we should just do the `DeleteConfirmationStep`.

bwateratmsft (approved):

utils/src/wizard/ConfirmationStep.ts
alexweininger: Good catch. It's too early in the morning 🤣, I mindlessly added this step after reading your feedback without realizing that it would never work.

bwateratmsft (approved):

utils/src/wizard/ConfirmationStep.ts
bwateratmsft: I can't even use the "too early in the morning" excuse, it's after noon!

PR #1126, opened 2022-04-20
Add the RG API declarations to the shared utils package

bwateratmsft:
This PR does a couple of things:

- Adds the RG API declarations to the shared utils package, so we can stop copying the API file everywhere
    - I added it as a separate file but I'm not married to that idea. `index.d.ts` was getting really long (not that it's that weird for declaration files to be long).
    - Also added a ton of code comments
- Adds some `declares` where they were missing
- Adds the `SealedAzExtTreeItem` and `AbstractAzExtTreeItem` interfaces and ties `AzExtTreeItem` and `AzExtParentTreeItem` to them
- Moves activity stuff alongside the RG API file
    - NOTE: I intentionally did not move `ActivityBase`, `ExecuteActivity` etc. as these are helpful implementations but not part of the interface per se


utils/rgapi.d.ts
```diff
@@ -0,0 +1,266 @@
...
+    /**
+     * The `AzExtTreeDataProvider` for the shared app resource view
+     * @deprecated Use `appResourceTree` instead
+     */
+    readonly tree: AzExtTreeDataProvider;
+
+    /**
+     * The `AzExtTreeDataProvider` for the shared app resource view
+     */
+    readonly appResourceTree: AzExtTreeDataProvider;
```
bwateratmsft, line 22: This will still need to be implemented in the RG extension, but since there's two trees I thought it best we specify here _which_ tree.

utils/rgapi.d.ts
```diff
@@ -0,0 +1,266 @@
...
+    readonly localResourceTreeView: vscode.TreeView<AzExtTreeItem>;
+
+    /**
+     * Version of the API
+     */
+    readonly apiVersion: string;
+
+    /**
+     * Reveals an item in the shared app resource tree
+     * @param resourceId The ARM resource ID to reveal
```
bwateratmsft, line 52: It's the ARM resource ID right?

utils/rgapi.d.ts
```diff
@@ -0,0 +1,266 @@
...
+    label: string;
+
+    /**
+     * Optionally, context values to add to the tree item. The overall context value string is a deduped, alphabetized, semicolon-separated list of individual context values.
+     */
+    contextValuesToAdd?: string[];
+
+    /**
+     * The collapsibleState of the tree item
+     * TODO: is this needed?
```
bwateratmsft, line 236: TODO

utils/index.d.ts
```diff
@@ -176,12 +177,157 @@ export interface ISubscriptionContext {
...
+     */
+    isAncestorOfImpl?(contextValue: string | RegExp): boolean;
+
+    /**
+     * If implemented, resolves the tooltip at the time of hovering, and the value of the `tooltip` property is ignored. Otherwise, the `tooltip` property is used.
+     */
+    resolveTooltip?(): Promise<string | MarkdownString>;
+}
+
+export type IAzExtTreeItem = AbstractAzExtTreeItem & SealedAzExtTreeItem;
```
bwateratmsft, line 308: I know we generally avoid the `IFoo` names but `AzExtTreeItem` is taken

utils/rgapi.d.ts
alexweininger: Correct

utils/rgapi.d.ts
alexweininger: Just to make sure I understand, the other tree would be the workspace tree?

utils/rgapi.d.ts
alexweininger: This is not needed anymore, there's no reason we should have to access it, and since the tree item is already created when resolving, there's no point in changing this value on resolve.

utils/index.d.ts
alexweininger: ISoundsGood

alexweininger (dismissed):
I vote we keep the rgapi.d.ts file separate from the index.d.ts at least for now. 
* The definitions in index.d.ts are implemented in the utils, while rgapi.d.ts implementations live in resource groups, so it makes sense to have them separate
* Easier to make changes if we have to
* File loads faster 😄

utils/rgapi.d.ts
bwateratmsft: Yeah. Should I rename it "workspaceTree" instead of "localResourceTree"?

alexweininger: Yeah go with "workspaceTree" for now, we may end up changing the wording later since workspace is already a separate thing in VS Code-land. Could be confusing for users.

bwateratmsft: I did that; `local` is replaced with `workspace`. I also sectioned out the deprecated parts so it's trivial to remove them later.

alexweininger (dismissed):

alexweininger (dismissed):

alexweininger (approved):

PR #1130, opened 2022-04-20
Consume new utils package

bwateratmsft:
No actual code changes so this is mostly to prevent consumers from accidentally getting both 0.2.0 utils (via direct dependency) and 0.1.0 utils (via dependency of this package).

### DIFF
--- a/azure/package-lock.json
+++ b/azure/package-lock.json
@@ -535,9 +535,8 @@
         },
         "node_modules/@microsoft/vscode-azext-utils": {
             "version": "0.2.0",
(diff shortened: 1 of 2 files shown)


alexweininger (approved):

alexweininger (approved):
How did a bad sha get into the lockfile?

bwateratmsft:
> How did a bad sha get into the lockfile?

I had done `npm install` of a local v0.2.0 file. When I changed the entry in `package.json` back to `^0.2.0`, and did `npm install`, it updated part of the `package-lock.json`...just not the SHA. (Thanks NPM)

PR #1133, opened 2022-04-21
Use contextValuesToAdd shape for AppSettings tree items

alexweininger:
Added option to pass in `contextValuesToAdd: string[]` when creating an `AppSettingsTreeItem` so that we can make them have unique context values. The children also use this array to create their context value. 

Needed to fix https://github.com/microsoft/vscode-azurestaticwebapps/issues/672

Bumped minor version since this is a breaking change.

### DIFF
--- a/appservice/src/tree/AppSettingsTreeItem.ts
+++ b/appservice/src/tree/AppSettingsTreeItem.ts
@@ -31,6 +31,12 @@ export function validateAppSettingKey(settings: StringDictionary, client: IAppSe
     return undefined;
 }
 
+interface AppSettingsTreeItemOptions {
+    supportsSlots?: boolean;
+    settingsToHide?: string[];
+    contextValuesToAdd?: string[];
+}
+
 export class AppSettingsTreeItem extends AzExtParentTreeItem {
     public static contextValue: string = 'applicationSettings';
     public readonly label: string = 'Application Settings';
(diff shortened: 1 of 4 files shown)


appservice/src/tree/AppSettingsTreeItem.ts
```diff
@@ -41,12 +47,15 @@ export class AppSettingsTreeItem extends AzExtParentTreeItem {
...
-    constructor(parent: AzExtParentTreeItem, clientProvider: AppSettingsClientProvider, supportsSlots: boolean = true, settingsToHide?: string[]) {
+    constructor(parent: AzExtParentTreeItem, clientProvider: AppSettingsClientProvider, options?: AppSettingsTreeItemOptions) {
         super(parent);
         this.clientProvider = clientProvider;
-        this.supportsSlots = supportsSlots;
-        this._settingsToHide = settingsToHide;
+        this.supportsSlots = options?.supportsSlots ?? true;
+        this._settingsToHide = options?.settingsToHide;
+        this.contextValuesToAdd = options?.contextValuesToAdd ?? [];
+        this.contextValue = Array.from(new Set([AppSettingsTreeItem.contextValue, ...(options?.contextValuesToAdd ?? [])])).sort().join(';');
```
bwateratmsft, line 58: Can this be a `get` with the `Set<string>` being constructed dynamically, like some of the other places?

bwateratmsft (dismissed):

appservice/src/tree/AppSettingsTreeItem.ts
alexweininger: I think it's worth making this a util function, we're starting to use it a lot.

bwateratmsft: Agreed, good idea.

bwateratmsft (approved):

PR #1134, opened 2022-04-25
Add contextValuesToAdd option to tree items

alexweininger:
Changes needed to fix: 
https://github.com/microsoft/vscode-azurefunctions/issues/3139
https://github.com/microsoft/vscode-azurefunctions/issues/3140
https://github.com/microsoft/vscode-azureappservice/issues/2221

### DIFF
--- a/appservice/src/tree/DeploymentTreeItem.ts
+++ b/appservice/src/tree/DeploymentTreeItem.ts
@@ -30,19 +30,23 @@ enum DeployStatus {
  */
 export class DeploymentTreeItem extends AzExtTreeItem {
     public static contextValue: RegExp = new RegExp('deployment\/.*');
-    public readonly contextValue: string;
     public label: string;
     public receivedTime: Date;
     public parent: DeploymentsTreeItem;
     private _deployResult: KuduModels.DeployResult;
+    private _scmType?: string;
 
     constructor(parent: DeploymentsTreeItem, deployResult: KuduModels.DeployResult, scmType: string | undefined) {
         super(parent);
-        this.contextValue = `deployment/${scmType}`.toLocaleLowerCase();
+        this._scmType = scmType;
         this._deployResult = deployResult;
         this.receivedTime = nonNullProp(deployResult, 'receivedTime');
         const message: string = this.getDeploymentMessage(deployResult);
         this.label = `${this.id.substring(0, 7)} - ${message}`;
+    }
+
+    public get contextValue(): string {
+        return Array.from(new Set([`deployment/${this._scmType}`.toLocaleLowerCase(), ...(this.parent.contextValuesToAdd ?? [])])).sort().join(';');
     }
 
     public get iconPath(): TreeItemIconPath {
--- a/appservice/src/tree/DeploymentTreeItem.ts
+++ b/appservice/src/tree/DeploymentTreeItem.ts
@@ -4,7 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import type { SiteSourceControl } from '@azure/arm-appservice';
-import { AzExtTreeItem, IActionContext, nonNullProp, openReadOnlyContent, TreeItemIconPath } from '@microsoft/vscode-azext-utils';
+import { AzExtTreeItem, createContextValue, IActionContext, nonNullProp, openReadOnlyContent, TreeItemIconPath } from '@microsoft/vscode-azext-utils';
 import * as os from 'os';
 import { ProgressLocation, ThemeIcon, window } from 'vscode';
 import { KuduModels } from 'vscode-azurekudu';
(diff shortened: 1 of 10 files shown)


appservice/src/tree/DeploymentsTreeItem.ts
```diff
@@ -25,15 +32,17 @@ export class DeploymentsTreeItem extends AzExtParentTreeItem {
...
-    public constructor(parent: AzExtParentTreeItem, site: ParsedSite, siteConfig: SiteConfig, sourceControl: SiteSourceControl) {
+    public constructor(parent: AzExtParentTreeItem, options: DeploymentsTreeItemOptions) {
         super(parent);
-        this.site = site;
-        this._scmType = siteConfig.scmType;
-        this._repoUrl = sourceControl.repoUrl;
+        this.site = options.site;
+        this._scmType = options.siteConfig.scmType;
+        this._repoUrl = options.sourceControl.repoUrl;
+        this.contextValuesToAdd = options?.contextValuesToAdd ?? [];
```
wwlorey, line 45: I think this is a misuse of the nullish coalescing operator. We don't want any falsy values other than `[]` making their way into `contextValuesToAdd`

appservice/src/tree/DeploymentTreeItem.ts
```diff
@@ -30,21 +30,25 @@ enum DeployStatus {
...
-        this.contextValue = `deployment/${scmType}`.toLocaleLowerCase();
+        this._scmType = scmType;
         this._deployResult = deployResult;
         this.receivedTime = nonNullProp(deployResult, 'receivedTime');
         const message: string = this.getDeploymentMessage(deployResult);
         this.label = `${this.id.substring(0, 7)} - ${message}`;
     }
 
+    public get contextValue(): string {
+        return Array.from(new Set([`deployment/${this._scmType}`.toLocaleLowerCase(), ...(this.parent.contextValuesToAdd ?? [])])).sort().join(';');
```
wwlorey, line 49: Same comment about `??` here and elsewhere regarding `contextValuesToAdd`

appservice/src/tree/FileTreeItem.ts
```diff
@@ -3,33 +3,37 @@
...
+    constructor(parent: FolderTreeItem, site: ParsedSite, label: string, path: string, isReadOnly: boolean) {
         super(parent);
         this.site = site;
         this.label = label;
         this.path = path;
         this.isReadOnly = isReadOnly;
     }
 
+    public get contextValue(): string {
+        return Array.from(new Set([FileTreeItem.contextValue, ...(this.parent.contextValuesToAdd ?? [])])).sort().join(';');
```
wwlorey, line 34: I recommend making a utility function for this "create a set of strings then sort them and join them on `;`" code since it's used in several places

appservice/src/tree/FolderTreeItem.ts
```diff
@@ -10,23 +10,37 @@ import { ParsedSite } from '../SiteClient';
 import { ISiteFileMetadata, listFiles } from '../siteFiles';
 import { FileTreeItem } from './FileTreeItem';
 
+export interface FolderTreeItemOptions {
+    site: ParsedSite;
+    label: string;
+    path: string;
+    isReadOnly: boolean;
+    contextValuesToAdd?: string[];
+}
```
wwlorey, line 19: Nice touch adding these options interfaces. Much safer than passing a bunch of parameters directly into the constructors 👍👍

appservice/src/tree/DeploymentsTreeItem.ts
alexweininger: If someone passed in a falsy value here and ignored the typings, I think I'd want it to throw an error (which it would when we do `...this.contextValuesToAdd`.

If they obey the typings, the left side of that expression can only be `string[] | undefined` so there will never be a falsy value.

appservice/src/tree/FileTreeItem.ts
alexweininger: https://github.com/microsoft/vscode-azuretools/pull/1135

appservice/src/tree/DeploymentsTreeItem.ts
wwlorey: > If someone passed in a falsy value here and ignored the typings, I think I'd want it to throw an error (which it would when we do ...this.contextValuesToAdd.

I don't think throwing an error in this case would be the best course of action.

If a falsy value (other than `[]`) was somehow passed in, it was definitely a mistake and not worth erroring. Especially because it would likely be an ugly error like "found non-callable iterator". The `??` operator is great for some situations but `||` is a better fit here.

I bring this up not only for this situation but as something to keep in mind for future code too. In my estimation the number of times we'll need to use `??` is relatively small compared to the need for `||`.

wwlorey (approved):

PR #1141, opened 2022-04-29
Add pickAppResource to hostapi

alexweininger:


utils/hostapi.d.ts
```diff
@@ -81,6 +86,18 @@ export interface AzureHostExtensionApi {
     //#endregion
 }
 
+export interface PickAppResourceOptions extends IAzureQuickPickOptions {
+    /**
+     * App resource type to filter the app resource picks
+     */
+    type?: string;
```
nturinski, line 93: I think we should go ahead and add "kind" to the options as well.

nturinski (approved):

utils/hostapi.d.ts
alexweininger: Updated it to have a filter property with both type and kind, with kind optional.

nturinski (approved):

PR #1142, opened 2022-05-03
Expose a command to be able to set `this._cachedChildren`

nturinski:
I think that this isn't ideal, but because of the way we're doing the mapping of the subscription tree item, we can't actually rely on `cachedChildren` to be a source of truth for that tree's children since not only do we have the tree mapping, but also the group tree nodes.

The way that the new `pickAppResource` works now too skips the group tree nodes since the sub tree has direct access to all App Resources, we don't actually use the cacheChildren for anything but to display the tree items in the UI.

I suspect that we'll probably be re-designing the way the tree data provider interacts with `getChildren` anyway, so I'm kind of okay with this solution.


bwateratmsft:
I'd almost prefer to just have `SubscriptionTreeItem` do a cast-to-`unknown` and alter `_cachedChildren`, rather than add this to the API. Caching is kind of a thing that doesn't belong on the API to begin with; leaning the API into it further would make it harder to step back from...

PR #1146, opened 2022-05-05
Add tags to pickAppResource filter

alexweininger:
Needed to support some database scenarios where the resources can only be differentiated based on tag values.

### DIFF
--- a/utils/hostapi.d.ts
+++ b/utils/hostapi.d.ts
@@ -86,20 +86,17 @@ export interface AzureHostExtensionApi {
     //#endregion
 }
 
+export interface AppResourceFilter {
+    type: string;
+    kind?: string;
+    tags?: Record<string, string>;
+}
+
 export interface PickAppResourceOptions extends IAzureQuickPickOptions {
     /**
-     * Options to filter the picks
+     * Options to filter the picks to resources that match any of the provided filters
      */
-    filter?: {
-        /**
-         * App resource type to filter the app resource picks
-         */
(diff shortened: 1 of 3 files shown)


utils/hostapi.d.ts
```diff
@@ -86,20 +86,17 @@ export interface AzureHostExtensionApi {
     //#endregion
 }
 
+export interface AppResourceFilter {
+    type: string;
+    kind?: string;
+    tags?: Record<string, string>;
```
bwateratmsft, line 92: ARM tags I assume? Does a resource need to match one or all of them? (Should probably add a comment to clarify that too)

bwateratmsft (approved):

utils/hostapi.d.ts
alexweininger: ARM tags. It needs to match all of them, will add comments.

utils/hostapi.d.ts
```diff
@@ -86,20 +86,20 @@ export interface AzureHostExtensionApi {
     //#endregion
 }
 
+export interface AppResourceFilter {
+    type: string;
+    kind?: string;
+    /**
+     * ARM tags that must be present on the resource. Resouce must match all provided tags.
```
bwateratmsft, line 93: Nit
```suggestion
     * ARM tags that must be present on the resource. Resource must match all provided tags.

```

bwateratmsft (approved):

PR #1148, opened 2022-05-06
Resolve tree item in `AzExtTreeDataProvider.resolveTreeItem`

nturinski:
We were experiencing a bug where a tree child would be registered twice when we resolved an `AppResourceTreeItem`.  It seems like the reason that this was happening was because tree items load their children regardless of their collapsed state.

So when the ResolvedAppResource was created, its `loadMoreChildrenImpl` would be called automatically by the TreeDataProvider, but simultaneously the `AppResourceTreeItem` would also call `loadMoreChildrenImpl`.  They would both register the same tree id and cause an error like so
image

This wasn't occurring before because of these [lines](https://github.com/microsoft/vscode-azureresourcegroups/blob/unified/src/tree/GroupTreeItemBase.ts#L65-L69).  Since everything was being resolved as soon as the `GroupTreeItems` were created, we never encountered this issue before.

We still will want to resolve all the `AppResourceTreeItems` because if we don't, everything is considered to be a parent tree node until it is resolved.


utils/src/tree/AzExtTreeDataProvider.ts
```diff
@@ -61,6 +61,10 @@ export class AzExtTreeDataProvider implements IAzExtTreeDataProviderInternal, ty
             ti.tooltip = await treeItem.resolveTooltip();
         }
 
+        if (treeItem.resolve && treeItem.resolveResult === undefined) {
+            await treeItem.resolve(false);
```
bwateratmsft, line 65: This approach isn't going to do what we want. Resolving a tree item (in the sense of "resolve" that we created, not the VSCode sense) needs to be able to change the label, description, etc. In the spec for VSCode's `resolveTreeItem`, it states that the tree items are already shown in the UI (so label, description, etc. can't be changed), _and_ also says that `onDidChangeTreeData` should _not_ be triggered (which eliminates it as a workaround for the lack of label etc. change).

bwateratmsft:
Closing in favor of #1150

PR #1150, opened 2022-05-06
Add option to `pickAppResource` for resolving quick picks ahead of time

bwateratmsft:
Alternative to #1148

### DIFF
--- a/utils/package.json
+++ b/utils/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@microsoft/vscode-azext-utils",
     "author": "Microsoft Corporation",
-    "version": "0.2.5",
+    "version": "0.2.6",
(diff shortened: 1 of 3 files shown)


alexweininger (dismissed):

utils/package.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "@microsoft/vscode-azext-utils",
     "author": "Microsoft Corporation",
-    "version": "0.2.5",
+    "version": "0.2.6",
```
bwateratmsft, line 4: Technically a semver violation, but otherwise we have to update all the upstream packages to avoid version conflicts, and I don't really feel like doing that right now...

bwateratmsft: Thanks to CG alerts we no longer have a choice.

alexweininger (approved):

PR #1158, opened 2022-05-16
Remove instanceof in getDeployNode

alexweininger:
Fixes https://github.com/microsoft/vscode-azurefunctions/issues/3169 and https://github.com/microsoft/vscode-azureappservice/issues/2247

### DIFF
--- a/appservice/src/deploy/getDeployNode.ts
+++ b/appservice/src/deploy/getDeployNode.ts
@@ -11,7 +11,7 @@ import { getWorkspaceSetting } from '../utils/settings';
 import { AppSource, IDeployContext } from './IDeployContext';
 
 function isAzExtTreeItem(ti: unknown): ti is AzExtTreeItem {
-    return (ti as AzExtTreeItem).fullId !== undefined;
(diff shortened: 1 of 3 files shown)


appservice/src/deploy/getDeployNode.ts
```diff
@@ -10,6 +10,10 @@ import { localize } from '../localize';
 import { getWorkspaceSetting } from '../utils/settings';
 import { AppSource, IDeployContext } from './IDeployContext';
 
+function isAzExtTreeItem(ti: unknown): ti is AzExtTreeItem {
+    return (ti as AzExtTreeItem).fullId !== undefined;
```
bwateratmsft, line 14: It's probably not really necessary in practice, but in principle this might be slightly more resilient:
```suggestion
    return (ti as AzExtTreeItem).fullId !== undefined && (ti as AzExtTreeItem).fullId !== null;
```

bwateratmsft (dismissed):

bwateratmsft (approved):

PR #1162, opened 2022-05-19
Add 307 to getItem responses

alexweininger:
Fixes https://github.com/microsoft/vscode-azureappservice/issues/2254


nturinski (approved):

nturinski:
For the record, 307 is a redirect.  Rather than throwing an error, it's going to follow the redirect now (which fixes the behavior)

PR #1164, opened 2022-05-20
Add 307 response to the correct endpoint

alexweininger:
Added it to `getItem2` in https://github.com/microsoft/vscode-azuretools/pull/1162, but it should've been added to `getItem` 🤦

### DIFF
--- a/kudu/package-lock.json
+++ b/kudu/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "vscode-azurekudu",
-  "version": "0.2.1",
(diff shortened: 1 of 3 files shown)


nturinski (approved):
You sure we shouldn't leave it in both?

alexweininger:
> You sure we shouldn't leave it in both?

I tested Logs and Files with it just in `getFiles`. But we could leave it in both if that's safer.

nturinski (approved):

PR #1165, opened 2022-05-24
Added new QP Separators for groupings to replace custom dropdowns

MicroFish91:
QP separators were introduced in a newer version of VS Code.  Since the repo is still running off of ```"@types/vscode": "1.57.0"```, I made the following changes/notes.  "kind" is currently considered an optional field so I believe typecasting like the following would be an okay workaround until we start referencing the newer types.

```javascript
// TODO: Switch kind from -1 to QuickPickItemKind.Separator when VS Code type files are updated          
picks.push({
    label: group.name,
    kind: -1,
    data: group
} as types.IAzureQuickPickItem<unknown>);
```

Here are some rough side-by-side comparisons of before and after the change (left - OG, right - new):
locations-side-by-side

runtime-side-by-side


alexweininger (commented):
> until we start referencing the newer types.

We should update the `@types/vscode` package.

alexweininger:
> runtime-side-by-side

I do actually prefer the new look (right image), it has less clutter. Did we ever talk to @fiveisprime about making this change?

MicroFish91:
I originally planned to do that, but the update to the newest version caused a bunch of other type mismatch issues in un-related files that I thought might make more sense to be captured in a different pull request.  Let me know what you think, thanks!

utils/src/userInput/showQuickPick.ts
```diff
@@ -236,13 +227,13 @@ function getGroupedPicks<TPick extends types.IAzureQuickPickItem<unknown>>(group
             if (!group.name) {
                 picks.push(...group.picks);
             } else {
+                // TODO: Switch kind from -1 to QuickPickItemKind.Separator when VS Code type files are updated
```
alexweininger, line 230: We should update the `@types/vscode` package to solve this. Most of our extensions require a minimum VS Code version of `1.62.0` ([example](https://github.com/microsoft/vscode-azurestaticwebapps/blob/main/package.json#L10)). Do you know which version of the types package we need?

We'll want to make sure every extension requires a minimum version of VS Code above the types package we install in the tools package. And we can increase the min version requirement of our extensions if needed.

MicroFish91:
> > runtime-side-by-side
> 
> I do actually prefer the new look (right image), it has less clutter. Did we ever talk to @fiveisprime about making this change?

Yup, I brought it up in our previous stand-up.  I believe you were out that day.  I got a thumbs up emoji from him so I'm assuming he liked the look of it.  I also found an old post from Eric mentioning that the custom dropdown implementation was originally intended to be migrated to this new VS Code QP separator feature

utils/src/userInput/showQuickPick.ts
alexweininger: Replying here to make it easier

> I originally planned to do that, but the update to the newest version caused a bunch of other type mismatch issues in un-related files that I thought might make more sense to be captured in a different pull request. Let me know what you think, thanks!

That's odd. Usually not too many things should be changing between versions on the `@types/vscode` package. I'll try it out myself and see what's up.

MicroFish91: Looks like the feature was added in v1.64.0

alexweininger: > Looks like the feature was added in v1.64.0

I think it's worth the effort to do the upgrade to 1.64.0 before merging this PR. We'll also have to bump the minimum versions of VS Code in the extensions to be at least 1.64.0.

bwateratmsft: Agreed. The most current VSCode version is 1.67.2; users generally update quite quickly. By now I'm guessing 99% or more are running 1.64+.

bwateratmsft (commented):
I think we should take off the `-0` from the package versions.

dev/index.d.ts
```diff
@@ -35,7 +35,8 @@ export class TestOutputChannel implements OutputChannel {
     public name: string;
     public append(value: string): void;
     public appendLine(value: string): void;
-    public appendLog(value: string, options?: { resourceName?: string, date?: Date }): void
+    public appendLog(value: string, options?: { resourceName?: string, date?: Date }): void;
+    public replace(value): void;
```
alexweininger, line 39: These changes should be added in a separate PR so we can release them before merging this.

bwateratmsft (approved):
Looks good to me!

PR #1168, opened 2022-05-25
Add Zone Redundancy Step when creating an App Service Plan

ccastrotrejo:
Add zone redundancy step when creating an App Service Plan


appservice/src/createAppService/AppServicePlanRedundancyStep.ts
```diff
@@ -0,0 +1,58 @@
...
+            'australiaeast',
+            'eastus2euap',
+        ];
+
+        return zoneRedundancySupportedLocations.includes(location);
+    }
+
+    public shouldPrompt(context: AppServiceWizardContext): boolean {
+        const { customLocation, _location, plan } = context;
+        if (plan === undefined && customLocation === undefined && _location && _location.name) {
```
alexweininger, line 53: According to the [requirements in the docs](https://docs.microsoft.com/en-us/azure/app-service/how-to-zone-redundancy#requirements), zone redundancy:
> Requires either Premium v2 or Premium v3 App Service plans

This step should not be shown unless the user selects a compatible App Service plan. If the ASP is not compatible, then zone redundancy should be disabled.

ccastrotrejo: Yes, you are right. Missing the decision over there. Thanks!

alexweininger: Testing it again now

PR #1172, opened 2022-06-02
Update to latest html-to-text package.

### LINKED ISSUE #1171: Eliminate string_decoder from vscode-azext-utils
The corporate governance tools don't like inclusion of `string_decoder` (used by a number of common NPM packages due to use by all but the most recent versions of `readable-stream`).  The `vscode-azext-utils` pulls this in (at runtime) via its older `html-to-text` dependency.  (It has other impacted build-time dependencies that can be dealt with separately.) The good news is that the latest version of `html-to-text` no longer has it in its dependency chain.  The bad news is that it's a major version change, with deprecations of APIs used by `vscode-azext-utils`, so it's not a simple pull-forward update.

```
├─┬ @microsoft/vscode-azext-utils@0.3.2
│ └─┬ html-to-text@5.1.1
│   └─┬ htmlparser2@3.10.1
│     └─┬ readable-stream@3.4.0
│       └── string_decoder@1.3.0
```

philliphoff:
Update `vscode-azext-utils` to the latest version of `html-to-text` to avoid inclusion of `string_decoder`, which triggers corporate governance alerts.

There are minor changes in the way the latest `html-to-text` version formats HTML compared to the old version we were using, but the updated formatting options gets (IMO) acceptably close to the prior behavior.

Resolves #1171


bwateratmsft (dismissed):
The `string_decoder` CG alert is a "low" priority one so AFAIK it's not strictly needed to fix it at this time, although there's also no reason we can't fix it anyway.

bwateratmsft:
May as well rev the version of this package for release while we're here.

philliphoff:
> The string_decoder CG alert is a "low" priority one so AFAIK it's not strictly needed to fix it at this time, although there's also no reason we can't fix it anyway.

There are apparently new rules that say "low" priority alerts are now to be fixed (albeit with a different timeline).

philliphoff:
> May as well rev the version of this package for release while we're here.

I'll be looking into a similar change in the "dev" package, which this package will then need to pick up, so I left that for next time.

bwateratmsft (dismissed):

philliphoff:
Since it looks like I can't update the "dev" package (and it's a dev dependency, which shouldn't count towards CG alerts), I'll update the package version here after all.

bwateratmsft (approved):

PR #1179, opened 2022-06-15
Suppress incorrect credscan alert

bwateratmsft:
There is an incorrect credscan alert on `.vscode-test\...\emmetNodeMain.js`. That directory does not need to be scanned. As far as I can tell, the alert is incorrect anyway.


bwateratmsft:
I'm going to just merge. We need a build.

PR #1180, opened 2022-06-15
Update `@azure/arm-resources-subscriptions` to v2

bwateratmsft:
Couple things in this PR:

- Updating to non-deprecated version of `@azure/arm-resources-subscriptions`
- Changing a few `import`s to be `import type`s

### DIFF
--- a/azure/src/azure-account.api.d.ts
+++ b/azure/src/azure-account.api.d.ts
@@ -13,35 +13,35 @@ import { CancellationToken, Event, Progress, Terminal } from 'vscode';
 export type AzureLoginStatus = 'Initializing' | 'LoggingIn' | 'LoggedIn' | 'LoggedOut';
 
 export interface AzureAccountExtensionApi {
-    readonly apiVersion: string;
-    readonly status: AzureLoginStatus;
-    readonly filters: AzureResourceFilter[];
-    readonly sessions: AzureSession[];
-    readonly subscriptions: AzureSubscription[];
-    readonly onStatusChanged: Event<AzureLoginStatus>;
-    readonly onFiltersChanged: Event<void>;
-    readonly onSessionsChanged: Event<void>;
-    readonly onSubscriptionsChanged: Event<void>;
-    readonly waitForFilters: () => Promise<boolean>;
-    readonly waitForLogin: () => Promise<boolean>;
-    readonly waitForSubscriptions: () => Promise<boolean>;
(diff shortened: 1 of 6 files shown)


azure/src/azure-account.api.d.ts
```diff
@@ -41,7 +41,7 @@ export interface AzureSession {
 
 export interface AzureSubscription {
 	readonly session: AzureSession;
-	readonly subscription: SubscriptionModels.Subscription;
+	readonly subscription: Subscription;
```
bwateratmsft, line 44: I'm less sure about this change. We're going from the type in `@azure/arm-resources-subscriptions@1.0.0` (which I think is track 1) to `@azure/arm-resources-subscriptions@2.0.0` (which I think is track 2). But neither of those match what's actually exported by the Azure Account extension, which is from `@azure/arm-subscriptions@3.1.2` (which I think is track 1). In any case, the objects _look_ identical but...I dunno.

nturinski (dismissed):
A partner team implemented this in storage here: https://github.com/microsoft/vscode-azurestorage/pull/905/files and here https://github.com/microsoft/vscode-azuretools/pull/844/files

We might be able to get them to test this for us to make sure it doesn't break them.  We would probably want to update the -arm-storage-profile package in Azure Storage as well though.

alexweininger:
What's the status on this? Are we still trying to have the partner team test the changes for us?

bwateratmsft:
> What's the status on this? Are we still trying to have the partner team test the changes for us?

That would be good, since I have no clue how to test it...

azure/src/azure-account.api.d.ts
alexweininger: Interesting... I need to look into this more before stamping a "should work just fine" on it 😄 

> But neither of those match what's actually exported

What do you mean by "match"? The types are incompatible? 

I wonder why haven't we moved to the arm-subscriptions T2 SDK yet? 🤔

bwateratmsft: By match, I just mean, is not from the same package. Azure Account is using `@azure/arm-subscriptions@3.1.2`.

alexweininger: Seems ok since that's how it was before. That said, I would like to know why we're using `@azure/arm-subscriptions@3.1.2` in Azure Account, but `@azure/arm-resources-subscriptions` here.

bwateratmsft: 🤷

PR #1182, opened 2022-06-15
Add basic interfaces to describe a friendlier tree API

bwateratmsft:


utils/mvtree.ts
```diff
@@ -0,0 +1,62 @@
...
+
+import * as vscode from 'vscode';
+import { AppResource } from './hostapi';
+
+/**
+ * This interface describes required properties in order to support the quick pick-through-the-tree
+ * logic. It is not required to be implemented, but the quick pick behavior will ignore the
+ * branch of the tree if these properties are not present.
+ */
+export interface QuickPickableModelNode {
```
philliphoff, line 14: Because it's several related properties, to avoid a chance collision with the implementor's own properties, I'd suggest something like:

```typescript

export interface QuickPickableModelOptions {
    readonly contextValues: string[];
    readonly isLeafNode: boolean;
}

export interface QuickPickableModelNode {
    readonly quickPickOptions: QuickPickableModelOptions;
}

```

utils/mvtree.ts
```diff
@@ -0,0 +1,62 @@
...
+    /**
+     * An array of context values.
+     */
+    readonly contextValuesArray: string[];
```
philliphoff, line 18: Suggest just `contextValues`; the array part seems obvious enough.

utils/mvtree.ts
```diff
@@ -0,0 +1,62 @@
...
+
+    /**
+     * Get the children of `element`. `element` will never be falsy.
+     *
+     * This is otherwise identical to {@link vscode.TreeDataProvider.getChildren}, but redefined
+     * to reflect that `element` will always have a value.
+     * @param element The element from which the provider gets children.
+     * @return Children of `element`.
+     */
+    getChildren(element: T): vscode.ProviderResult<T[]>;
```
philliphoff, line 49: It will be interesting to see how the TS compiler deals with the overlapping definitions.  If it works, great, seems like a good idea.  If it proves problematic, it's probably just as easy to note to implementors the slight changes in semantics; thorough implementors should/would be guarding against and handling the `undefined` case anyway.  In particular, for the `onDidChangeTreeData` event, we can easily treat being passed `undefined` as the equivalent to being passed the trunk element (i.e. to indicate that *everything* changed, which is still pretty much the same idea).

bwateratmsft: IIRC it basically treats it like `override`, i.e. the nearest interface defining a method "wins". So if they say `Foo implements BranchDataProvider<Bar>`, it will treat `element` as a mandatory argument to `getChildren`, and a mandatory parameter to the event.

PR #1187, opened 2022-06-29
Add deployment id to zipdeploy event

nturinski:
Fixes https://github.com/microsoft/vscode-azureappservice/issues/2153

After messing with the kudu API endpoints, I realized that the response was coming back with the requested `scm-deployment-id` with our Kudu API call.  I couldn't tell you why-- it clearly states in the code that if the query parameter is false, to not add it, but it definitely is...

https://github.com/projectkudu/kudu/blob/2da32e6538ba90fa6bc1687dd2a31ca458027e2c/Kudu.Services/Deployment/PushDeploymentController.cs#L80

https://github.com/projectkudu/kudu/blob/2da32e6538ba90fa6bc1687dd2a31ca458027e2c/Kudu.Services/Deployment/PushDeploymentController.cs#L476-L480
image

The other change might help some of our timeouts.  The old logic works the same, but if we get the location url, it'll use that instead to ping the status of the deployment. There's a comment on their code that indicates not using it can cause the deployment to appear as if it timed-out so seems better to use the location url instead.

> `Deploy from zip url.   This may timeout depending on the size of zip file and load on the VM.`

https://github.com/projectkudu/kudu/wiki/REST-API#zip-deployment
image


appservice/src/deploy/deployZip.ts
```diff
@@ -14,14 +14,19 @@ import { waitForDeploymentToComplete } from './waitForDeploymentToComplete';
...
-    await runWithZipStream(context, {
+    const response = await runWithZipStream(context, {
         fsPath, site, pathFileMap,
         callback: async zipStream => {
-            await kuduClient.pushDeployment.zipPushDeploy(() => zipStream, { isAsync: true, author: 'VS Code' });
+            return await kuduClient.pushDeployment.zipPushDeploy(() => zipStream, { isAsync: true, author: 'VS Code' });
         }
     });
+    let locationUrl: string | undefined;
+    if (response) {
```
bwateratmsft, line 24: How badly might things go if the `response` or `locationUrl` don't come through?

nturinski: Shouldn't break anything. Both of these properties are optional-- the `scm-deployment-id` is just for partner telemetry and the location url is the preferred way to track deployment progress, but our old method will still work if location url isn't defined.

nturinski: But I'll wrap this in a try catch just in-case response._response is undefined for some reason so it doesn't throw an errors

nturinski:
Depends on https://github.com/microsoft/vscode-azuretools/pull/1188

bwateratmsft (approved):

nturinski:
Gonna go ahead and merge this since stale approval dismissal sucks

PR #1189, opened 2022-06-29
Use node 16 in pipelines

alexweininger:
Just testing out the macOS pipeline with node 16. 

Not meant to be merged.

Note: even if this doesn't fix the macOS pipeline, it's probably a good idea to use Node 16 in the pipelines since that's what version of Node VS Code uses (and we use).

### DIFF
--- a/.azure-pipelines/before-all.yml
+++ b/.azure-pipelines/before-all.yml
@@ -1,8 +1,8 @@
 steps:
 - task: NodeTool@0
-  displayName: 'Use Node 14.x'
+  displayName: 'Use Node 16.x'
   inputs:
-    versionSpec: 14.x
+    versionSpec: 16.x
 
 - script: |
     sudo cp .azure-pipelines/xvfb.init /etc/init.d/xvfb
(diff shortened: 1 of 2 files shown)


bwateratmsft:
Docker uses Node v16 and it's still been failing like crazy, so I doubt that will fix it.

philliphoff:
@alexweininger I wonder if you could try using `macOS-12` as the host, just for kicks, as `macOS-latest` actually means MacOS 11 Big Sur, which is, confusingly, not the latest.

bwateratmsft (approved):

nturinski:
Was this just a fluke, or did upgrading to macOS-12 actually fix this?

alexweininger:
> Was this just a fluke, or did upgrading to macOS-12 actually fix this?

Since we check directories for changes before running certain checks, the failing checks weren't run for this PR. Should I merge this so we can find out if it really works? Can't make it worse...

image

philliphoff:
@alexweininger Should we try moving to `@vscode/test-electron`, which seems to have a fix for a similar issue: [vscode-test#153](https://github.com/microsoft/vscode-test/issues/153).

PR #1196, opened 2022-07-14
Add parseAzureResourceId util

alexweininger:
We have this code copy pasted in a bunch of extensions.

### DIFF
--- a/azure/src/utils/parseAzureResourceId.ts
+++ b/azure/src/utils/parseAzureResourceId.ts
@@ -16,7 +16,7 @@ export function parseAzureResourceId(id: string): types.ParsedAzureResourceId {
     return {
         rawId: id,
         subscriptionId: matches[1],
-        resourceGroupId: matches[2],
+        resourceGroupName: matches[2],
         provider: matches[3],
         resourceName: matches[4]
     }
--- a/azure/src/utils/parseAzureResourceId.ts
+++ b/azure/src/utils/parseAzureResourceId.ts
@@ -21,3 +21,7 @@ export function parseAzureResourceId(id: string): types.ParsedAzureResourceId {
         resourceName: matches[4]
     }
 }
+
+export function getResourceGroupFromId(id: string): string {
+    return parseAzureResourceId(id).resourceGroup;
+}
--- a/azure/src/utils/parseAzureResourceId.ts
+++ b/azure/src/utils/parseAzureResourceId.ts
@@ -1,0 +1,23 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.txt in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import * as types from '../../index';
+import { localize } from "../localize";
+
+export function parseAzureResourceId(id: string): types.ParsedAzureResourceId {
+    const matches: RegExpMatchArray | null = id.match(/\/subscriptions\/(.*)\/resourceGroups\/(.*)\/providers\/(.*)\/(.*)/i);
+
+    if (matches === null || matches.length < 3) {
+        throw new Error(localize('InvalidResourceId', 'Invalid Azure Resource Id'));
+    }
+
+    return {
+        rawId: id,
+        subscriptionId: matches[1],
+        resourceGroupId: matches[2],
+        provider: matches[3],
+        resourceName: matches[4]
+    }
+}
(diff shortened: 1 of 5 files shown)


nturinski (approved):

nturinski:
Actually, seeing it be used in https://github.com/microsoft/vscode-azureappservice/pull/2290/files, that looks awful lol. Maybe we should still have the `getResourceGroupFromId` function? I think that's why we use it like 99% of the time.

alexweininger:
> Actually, seeing it be used in [microsoft/vscode-azureappservice#2290 (files)](https://github.com/microsoft/vscode-azureappservice/pull/2290/files), that looks awful lol. Maybe we should still have the `getResourceGroupFromId` function? I think that's why we use it like 99% of the time.

If you like having it split into multiple function that's how we'll keep it. I thought we'd like this one more 😄

Trying to think of why I liked my way more... I like that it's one function that can give me any value from the resource id?

I experimented extending the string class too, kinda cool... [Playground Link](https://www.typescriptlang.org/play?#code/MYGwhgzhAECCBeBXATgUwEqogexcVAkgCbSoAeALqgHZEwDKFyAltQObQDeAsAFB-RB0AA6IARiGbBobVBWgRxEYC2EVm2asQAUASgBcCpqw49eQi9DQUU1aBQAWzCADphYZBFR6A2gEYAXQBuASEAXz5QwVEJKRk5KyxcZHwAcWRcYR0DIxZ2LijLa1t7J1d3T29dHwAmYMKI-nMhGMlpWXlhDIA3ZiJUZD1DCGN8s0shYuQ7R2c3Dy9fAGZ65sFGwta4jsScPFQAOTAAWyrh0dNCiymZsvnK3wAWVYsNtZEWbrAqEQWz6EwbAAomRhABZb7ABywZDIMAATwK7wswE0I2gx0hDiwhkBIPBWJhcMRAB9oNRECAQNAALylOaYihQ7QAegAOizFGJlKp1GiOdoXAAqXQctB7FKodKZCAC4WilldbC9fqeOUi9W6FnMXQhJoTQTMABm0G0jKhWFpNLpFKp0BJZPN2NcIBobEc0AAPNAlrokQbLI4MgB3cmoUNA2HYQYAcgI1C+khICBQqABSX20GIMd1V3CkWRkzkJSdWD1rz4b1R1HRI2QtLDoZTaEwEvwOhjnKUKmYag0NZZAFYAOwABj8fmHYFHAFojaglsOZ48iP0Z2BUI8ABwzsQANgAnDVgH4xDVB0axyzxck0hlEMIICzg6hWCYBjO0ERnDPx4qen0AxPmCUgZDgRoUC4ADCYAWiymDfk+L5vuwH5fj+445nqfDVjgrouCA2BsNodYuFyPK9nyWhELmvC4dg+GEcRpE3vs0oPsQtH0YxREkUwbgAaqXFogxqAEbxLEZpKRynLqQA)

azure/src/utils/parseAzureResourceId.ts
```diff
@@ -0,0 +1,23 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.txt in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import * as types from '../../index';
+import { localize } from "../localize";
+
+export function parseAzureResourceId(id: string): types.ParsedAzureResourceId {
+    const matches: RegExpMatchArray | null = id.match(/\/subscriptions\/(.*)\/resourceGroups\/(.*)\/providers\/(.*)\/(.*)/i);
```
bwateratmsft, line 10: I think we should try to make this regex at least a little more precise. For example, subscriptions are always guids, so we could use `([a-f0-9-]+)`. Providers always(ish?) have something like `word.word/word`.

I'm sure we also have regexs somewhere for resource group names, though I wouldn't bother with resource names because those rules are [all over the place](https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules).

alexweininger: Making it stricter could cause issues when there are non-english characters in the id. For example, not even the officially documented resource group regex `[-\w\._\(\)]+` supports Chinese characters. However, the portal lets me create resource groups with Chinese characters.

My stricter regex: https://regexr.com/6prg2

bwateratmsft: Yeah, that's fair. I don't want something excessively-strict for this but I can also see risks if the individual regex groups can take slashes (e.g. the original `(.*)` group for regexs matches slashes too). If something had an unexpected number of slashes for whatever reason, things could break down.

nturinski (approved):
I'll let Brandon request changes if he strongly disagrees, but I am okay with the current regex for Azure resource ids. We have been using it for quite a while now and it seems pretty stable especially for what we are using it (namely, to extract rg names from resources).

I share the same concerns as Alex-- that by making it stricter, we will accidentally block and isolate users with international characters.

bwateratmsft (approved):
I can't argue with a history of stability. :smile:

PR #1206, opened 2022-07-21
Include a settings.json in the root of the repo

nturinski:
I've been noticing that tslint/eslint isn't auto-fixing anything on save. I don't know how everyone else opens the azuretools repo, but I usually just open the whole folder rather than individual folders since there's usually a lot of intertwined code.

If nobody opposes, I'd love to just have a settings.json in the root. I'm pretty sure that if you were to open it by package folder, this wouldn't affect you either.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,19 @@
+{
+    "editor.codeActionsOnSave": {
+        "source.fixAll.eslint": true,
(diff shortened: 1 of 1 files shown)


alexweininger (approved):
Nice. Been wanting to be able to do this for a while and now I can!

PR #1207, opened 2022-07-22
Add util function for reading/writing JSON files

nturinski:
Mostly to be used in Azure Functions because it relies a lot of `fse.readJSON` and `fse.writeJSON`. I believe the behavior should be the same-- if the unknown data type is not JSON parsable, throw an error.

### DIFF
--- a/utils/index.d.ts
+++ b/utils/index.d.ts
@@ -1516,8 +1516,8 @@ export declare namespace AzExtFsExtra {
     export function readFile(resource: Uri | string): Promise<string>;
     export function writeFile(resource: Uri | string, contents: string): Promise<void>;
     export function pathExists(resource: Uri | string): Promise<boolean>;
-    export function readJSON(resource: Uri | string): Promise<unknown>
-    export function writeJSON(resource: Uri | string, json: unknown): Promise<void>
+    export function readJSON<T>(resource: Uri | string): Promise<T>
+    export function writeJSON<T>(resource: Uri | string, contents: string | T): Promise<void>
 }
 
 export declare function maskValue(data: string, valueToMask: string | undefined): string;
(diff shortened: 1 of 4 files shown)


utils/src/utils/AzExtFsExtra.ts
```diff
@@ -76,6 +76,16 @@ export namespace AzExtFsExtra {
         return !!stats;
     }
 
+    export async function readJSON(resource: Uri | string): Promise<unknown> {
```
alexweininger, line 79: nit: Would you want to use a generic instead of `unknown` here?

alexweininger (approved):

utils/src/utils/AzExtFsExtra.ts
bwateratmsft: Agreed, I like generic better.

utils/index.d.ts
```diff
@@ -1516,6 +1516,8 @@ export declare namespace AzExtFsExtra {
     export function readFile(resource: Uri | string): Promise<string>;
     export function writeFile(resource: Uri | string, contents: string): Promise<void>;
     export function pathExists(resource: Uri | string): Promise<boolean>;
+    export function readJSON(resource: Uri | string): Promise<unknown>
+    export function writeJSON(resource: Uri | string, json: unknown): Promise<void>
```
bwateratmsft, line 1520: Up to you if you think this is needed, but you could have it accept either `string` (for already-JSON) or `unknown` (for objects).

alexweininger (dismissed):

utils/src/utils/AzExtFsExtra.ts
```diff
@@ -76,6 +76,24 @@ export namespace AzExtFsExtra {
         return !!stats;
     }
 
+    export async function readJSON<T>(resource: Uri | string): Promise<T> {
+        const file = await readFile(resource);
+        return JSON.parse(file) as T;
```
alexweininger, line 81: I was wondering if that would cause issues :D

alexweininger (approved):

PR #1208, opened 2022-07-23
Customize Wizard-integrated Activity tree items

alexweininger:
Goal: Make it easy to customize the wizard-integrated Activity tree items

I recommend looking at what these changes accomplish in https://github.com/microsoft/vscode-azurefunctions/pull/3255 before getting into the code changes here. 

To summarize: 
- Added a new `displayOptions?: ActivityTreeItemOptionsFactory` property to `ExecuteActivityContext`. Which if set controls how the activity is displayed in the Activity Log.
- If left unset, the Wizard will use the new `CreateOrDeleteActivityOptions` (what used to be ExecuteActivity) by default. 
- Added an `ActivityOptionsFactoryBase` util class to remove boilerplate from creating a custom `ActivityTreeItemOptionsFactory`

### DIFF
--- a/utils/hostapi.d.ts
+++ b/utils/hostapi.d.ts
@@ -264,11 +264,6 @@ export interface Activity {
     cancellationTokenSource?: vscode.CancellationTokenSource;
 
     /**
-     * Options for how the activity is displayed as a tree item
-     */
-    options: ActivityTreeItemOptions;
-
-    /**
      * Fire this event to start the activity
      */
     onStart: vscode.Event<OnStartActivityData>;
--- a/utils/hostapi.d.ts
+++ b/utils/hostapi.d.ts
@@ -4,7 +4,7 @@
 *--------------------------------------------------------------------------------------------*/
 
 import type * as vscode from 'vscode';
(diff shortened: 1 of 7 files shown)


bwateratmsft:
The way the activity API is meant to be implemented, in this case, would be for the Functions extension to implement its own inheritor of the `ActivityBase` class, or even an entirely separate implementation of the `Activity` interface. The four events (start, progress, success, error) give control over all aspects of the display that the client extensions are meant to control (context values, label, children). I don't see the need to add a fifth options set.

alexweininger:
> The way the activity API is meant to be implemented, in this case, would be for the Functions extension to implement its own inheritor of the `ActivityBase` class, or even an entirely separate implementation of the `Activity` interface. The four events (start, progress, success, error) give control over all aspects of the display that the client extensions are meant to control (context values, label, children). I don't see the need to add a fifth options set.

Got it. I think I can refactor what I've done here so that I don't need to add the options property to the `Activity` interface.

alexweininger:
@bwateratmsft 

I've removed the `options` property from `Activity`. This means https://github.com/microsoft/vscode-azureresourcegroups/pull/342 is no longer needed.

With this change, everything shown in https://github.com/microsoft/vscode-azurefunctions/pull/3255 still works. Adding the `options` property to `Activity` made my life unnecessarily hard...

alexweininger:
Closing in favor of #1212

PR #1210, opened 2022-07-25
Let writeJson accept spaces parameter

nturinski:
(Annoyingly) I didn't realize, but Functions sometimes uses the space parameter in `writeJSON` to format the jsons to 2 spaces. This is in `JSON.stringify` natively, so just leveraging that property now.

image

### DIFF
--- a/utils/index.d.ts
+++ b/utils/index.d.ts
@@ -1517,6 +1517,9 @@ export declare namespace AzExtFsExtra {
     export function writeFile(resource: Uri | string, contents: string): Promise<void>;
     export function pathExists(resource: Uri | string): Promise<boolean>;
     export function readJSON<T>(resource: Uri | string): Promise<T>
+    /**
+     * @param spaces Defaults to 2 spaces. If the default JSON.stringify behavior is required, input 0
+     */
     export function writeJSON(resource: Uri | string, contents: string | unknown, spaces?: string | number): Promise<void>
     export function readDirectory(resource: Uri | string): Promise<{ fsPath: string, name: string, type: FileType }[]>;
     export function emptyDir(resource: Uri | string): Promise<void>;
--- a/utils/index.d.ts
+++ b/utils/index.d.ts
@@ -1518,7 +1518,10 @@ export declare namespace AzExtFsExtra {
     export function pathExists(resource: Uri | string): Promise<boolean>;
     export function readJSON<T>(resource: Uri | string): Promise<T>
     export function writeJSON(resource: Uri | string, contents: string | unknown, spaces?: string | number): Promise<void>
+    export function readDirectory(resource: Uri | string): Promise<{ fsPath: string, name: string, type: FileType }[]>;
     export function emptyDir(resource: Uri | string): Promise<void>;
+    export function copy(src: Uri | string, dest: Uri | string, options?: { overwrite?: boolean }): Promise<void>;
+    export function deleteResource(resource: Uri | string, options?: { recursive?: boolean, useTrash?: boolean }): Promise<void>
 }
 
 export declare function maskValue(data: string, valueToMask: string | undefined): string;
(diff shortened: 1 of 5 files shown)


alexweininger (dismissed):

utils/src/utils/AzExtFsExtra.ts
```diff
@@ -81,14 +81,14 @@ export namespace AzExtFsExtra {
         return JSON.parse(file) as T;
     }
 
-    export async function writeJSON(resource: Uri | string, contents: string | unknown): Promise<void> {
+    export async function writeJSON(resource: Uri | string, contents: string | unknown, spaces?: string | number): Promise<void> {
```
bwateratmsft, line 84: Should we define a default value (e.g. 2 or 4)? If the purpose is for the JSON to be written to a file it seems like it would almost always (or at least, usually) be desirable to have whitespace be used. I think that _not_ wanting whitespace would be the exception, and for that we could require the caller to pass in a [zero value](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#parameters):
> Values less than 1 indicate that no space should be used.

nturinski: Yeah, that makes sense to me.

nturinski:
emptyDir was used in functions test code to reset the workspace so I made a util function for that. 

Added some unit tests as well as fixed the index.d.ts. I never noticed, but they didn't accept parameters in the index file.

utils/src/utils/AzExtFsExtra.ts
```diff
@@ -4,6 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { dirname } from 'path';
+import path = require('path');
```
bwateratmsft, line 7: I'd merge this with the above and do `import * as path from 'path';`, and update where `dirname` gets used.

utils/src/utils/AzExtFsExtra.ts
```diff
@@ -81,19 +81,48 @@ export namespace AzExtFsExtra {
...
             stringified = contents;
         } else {
-            stringified = JSON.stringify(contents);
+            stringified = JSON.stringify(contents, undefined, spaces);
         }
 
         await writeFile(resource, stringified);
     }
 
+    export async function readDirectory(resource: Uri | string): Promise<{ fsPath: string, name: string, type: FileType }[]> {
```
bwateratmsft, line 97: These didn't get added to the index.d.ts file

nturinski: This is why I don't push code during meetings 😅

utils/index.d.ts
```diff
@@ -1509,15 +1509,19 @@ export declare function registerReportIssueCommand(commandId: string): void;
...
+    export function isDirectory(resource: Uri | string): Promise<boolean>;
+    export function isFile(resource: Uri | string): Promise<boolean>;
     export function ensureDir(resource: Uri | string): Promise<void>;
     export function ensureFile(resource: Uri | string): Promise<void>;
     export function readFile(resource: Uri | string): Promise<string>;
     export function writeFile(resource: Uri | string, contents: string): Promise<void>;
     export function pathExists(resource: Uri | string): Promise<boolean>;
     export function readJSON<T>(resource: Uri | string): Promise<T>
-    export function writeJSON(resource: Uri | string, contents: string | unknown): Promise<void>
+    export function writeJSON(resource: Uri | string, contents: string | unknown, spaces?: string | number): Promise<void>
```
alexweininger, line 1520: Since defaulting to 2 spaces is not the same behavior as [`JSON.stringify`](https://github.com/microsoft/vscode-azuretools/blob/main/azure/src/createAzureClient.ts#L197), it would be good to add a doc comment that at least makes note of that.

alexweininger (approved):

PR #1211, opened 2022-07-25
Let activityResult be a resource Id

alexweininger:
[Context for this change](https://github.com/microsoft/vscode-azuretools/pull/1191/files#r927708816)

Made possible by https://github.com/microsoft/vscode-azureresourcegroups/pull/341

This will make the package incompatible with older Resources extension versions, but I don't think that's a problem.

### DIFF
--- a/utils/index.d.ts
+++ b/utils/index.d.ts
@@ -1118,9 +1118,11 @@ export declare interface ExecuteActivityContext {
      */
     activityTitle?: string;
     /**
+     * Resource or resourceId
+     *
      * Set to show a "Click to view resource" child on success.
      */
-    activityResult?: AppResource;
+    activityResult?: AppResource | string;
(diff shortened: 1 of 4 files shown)


utils/index.d.ts
```diff
@@ -1118,9 +1118,11 @@ export declare interface ExecuteActivityContext {
      */
     activityTitle?: string;
     /**
+     * Resource or resourceId
+     *
      * Set to show a "Click to view resource" child on success.
      */
-    activityResult?: AppResource;
+    activityResult?: AppResource | string;
```
alexweininger, line 1125: In the future, we want to make this only accept a `string`. But for compatibility we'll take `AppResource` too for now.

bwateratmsft (approved):

PR #1212, opened 2022-07-25
Allow passing an ExecuteActivity implementation

alexweininger:
I think this is a much better alternative to https://github.com/microsoft/vscode-azuretools/pull/1208. Instead of passing an options interface that then gets passed to the activity, this allows extensions to just extend `ExecuteActivity` and pass it to the wizard.

See https://github.com/microsoft/vscode-azurefunctions/pull/3258 for a usage example

* Export `ExecuteActivity` so that extensions can extend it
* Add `wizardActivity` property to `ExecuteActivityContext`, which allows passing an implementation of `ExecuteActivity` to be passed in and used by the wizard

### DIFF
--- a/utils/index.d.ts
+++ b/utils/index.d.ts
@@ -1111,7 +1111,7 @@ export declare class AzureWizard<T extends IActionContext & Partial<ExecuteActiv
     public execute(): Promise<void>;
 }
 
-export class ExecuteActivity<C extends ExecuteActivityContext> extends ActivityBase<void> {
+export class ExecuteActivity<C extends ExecuteActivityContext = ExecuteActivityContext> extends ActivityBase<void> {
     protected readonly data: ExecuteActivityData<C>;
     public constructor(data: ExecuteActivityData<C>, task: ActivityTask<void>);
     public initialState(): ActivityTreeItemOptions;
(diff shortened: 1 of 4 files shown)


nturinski (dismissed):
Though if you want to go ahead and include your [refactor](https://github.com/microsoft/vscode-azuretools/compare/alex/executeActivity2...alex/activity-refactor), we can just merge it all at once in here.

nturinski (approved):

utils/src/activityLog/activities/ExecuteActivity.ts
```diff
@@ -3,20 +3,16 @@
...
 import { AzExtParentTreeItem } from "../../tree/AzExtParentTreeItem";
 import { GenericTreeItem } from "../../tree/GenericTreeItem";
 import { ActivityBase } from "../Activity";
 
-interface ExecuteActivityData<C extends types.ExecuteActivityContext> {
-    context: C;
-}
-
-export class ExecuteActivity<C extends types.ExecuteActivityContext> extends ActivityBase<void> {
+export class ExecuteActivity<C extends types.ExecuteActivityContext = types.ExecuteActivityContext> extends ActivityBase<void> {
```
bwateratmsft, line 13: Maybe `TContext` instead of `C`?

utils/index.d.ts
```diff
@@ -1127,6 +1136,11 @@ export declare interface ExecuteActivityContext {
      * Hide activity notifications
      */
     suppressNotification?: boolean;
+
+    /**
+     * The activity implementation to use, defaults to ExecuteActivity
+     */
+    wizardActivity?: new <C extends ExecuteActivityContext>(context: C, task: ActivityTask<void>) => ExecuteActivity;
```
bwateratmsft, line 1143: Maybe `TContext` instead of `C`?

bwateratmsft (approved):

PR #1214, opened 2022-07-27
Start of removal of T1 stuff

bwateratmsft:


bwateratmsft:
#1255 is the start but this is so out of date at this point I thought it wasn't worth keeping...

PR #1218, opened 2022-08-02
Add Azure Stacks WebSiteManagementClient

nturinski:
I tried replacing the `WebSiteManagementClient` type with the new hybrid version, and besides some mismatch typing in Site response, it seems like it has all of the same namespaces and operations.

I bumped a minor version though because I think that this could sort of be considered a breaking change though realistically, shouldn't impact 99% of people since they won't have `isCustomCloud` set.

### DIFF
--- a/appservice/package.json
+++ b/appservice/package.json
@@ -34,6 +34,7 @@
     "dependencies": {
         "@azure/abort-controller": "^1.0.4",
         "@azure/arm-appinsights": "^5.0.0-beta.4",
+        "@azure/arm-appservice-profile-2020-09-01-hybrid": "^2.0.0",
         "@azure/arm-appservice": "^11.0.0",
         "@azure/arm-operationalinsights": "^8.0.1",
         "@azure/arm-resourcegraph": "^4.0.0",
--- a/appservice/package.json
+++ b/appservice/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@microsoft/vscode-azext-azureappservice",
     "author": "Microsoft Corporation",
-    "version": "0.7.2",
+    "version": "0.8.0",
     "description": "Common tools for developing Azure App Service extensions for VS Code",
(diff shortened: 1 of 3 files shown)


nturinski:
Actually, I can just pack and install this to Functions and make a custom VSIX to test Azure Stacks, so I'll hold off on merging until then.

appservice/package.json
```diff
@@ -34,6 +34,7 @@
     "dependencies": {
         "@azure/abort-controller": "^1.0.4",
         "@azure/arm-appinsights": "^5.0.0-beta.4",
+        "@azure/arm-appservice-profile-2020-09-01-hybrid": "^2.0.0",
```
bwateratmsft, line 37: Can this replace `@azure/arm-appservice` entirely?

bwateratmsft: Especially until I can get to https://github.com/microsoft/vscode-docker/issues/3594

nturinski: We definitely can't replace the `@azure/arm-appservice` package with this one as the api-versions are different. I'm pretty sure that this only works with custom cloud. That being said, I understand the apprehension to adding more dependencies.

bwateratmsft: What is it about the primary `@azure/arm-appservice` package that makes it incompatible with Azure Stack?

nturinski: https://github.com/microsoft/vscode-azurefunctions/issues/3242 The api-version itself which we can't dynamically change unfortunately.

bwateratmsft: This seems like an easy thing for the SDK to handle...should we file an issue for it?

nturinski: Yeah or ideally they adopt the hybrid SDKs with the new REST sdks that are much smaller.

alexweininger:
Closing as stale.

PR #1222, opened 2022-08-11
Fix AzExtResourceType

alexweininger:
I wasn't able to import `getAzExtResourceType` using 
```import { getAzExtResourceType } from '@microsoft/vscode-azext-utils/getAzExtResourceType``` so I'm exporting it from index.ts now.

Also fixed `getAzExtResourceType`

### CODE REVIEW

File: utils/hostapi.d.ts [hunk 1]
```diff
@@ -4,8 +4,7 @@
 *--------------------------------------------------------------------------------------------*/
 
 import type * as vscode from 'vscode';
-import type { AzExtResourceType } from './azExtResourceType';
-import type { AbstractAzExtTreeItem, AzExtParentTreeItem, AzExtTreeDataProvider, AzExtTreeItem, IActionContext, IAzureQuickPickOptions, ISubscriptionContext, ITreeItemPickerContext, SealedAzExtTreeItem } from './index'; // This must remain `import type` or else a circular reference will result
+import type { AbstractAzExtTreeItem, AzExtParentTreeItem, AzExtTreeDataProvider, AzExtTreeItem, IAzureQuickPickOptions, ISubscriptionContext, ITreeItemPickerContext, SealedAzExtTreeItem, AzExtResourceType } from './index'; // This must remain `import type` or else a circular reference will result
```
bwateratmsft, line 7: nit: Sort

utils/index.d.ts
```diff
@@ -1599,3 +1599,78 @@ export declare class DeleteConfirmationStep extends AzureWizardPromptStep<IActio
...
+ * @param resource The resource to check the {@link AzExtResourceType} for
+ * @returns The normalized Azure resource type
+ */
+export declare function getAzExtResourceType(resource: { type: string; kind?: string }): AzExtResourceType | undefined;
+
+/**
+ * Normalized type for Azure resources that uniquely identifies resource type for the purposes
+ * of the Azure extensions
+ */
+export declare enum AzExtResourceType {
```
alexweininger, line 1615: Do we really have to have this enum written out twice? There has to be a better way.

utils/src/AzExtResourceType.ts
```diff
@@ -1,79 +1,67 @@
...
- * @param resource The resource to check the {@link AzExtResourceType} for
- * @returns The normalized Azure resource type
- */
-export declare function getAzExtResourceType(resource: { type: string; kind?: string }): AzExtResourceType | undefined;
+
+/**
+ * Normalized type for Azure resources that uniquely identifies resource type for the purposes
+ * of the Azure extensions
+ */
+export enum AzExtResourceType {
```
bwateratmsft, line 6: This file shouldn't still be needed since the index contains the declarations

alexweininger: I think it's still needed because the [enum exists at runtime](https://www.typescriptlang.org/docs/handbook/enums.html#enums-at-runtime).

bwateratmsft: Ugh...this is why we should be generating our declarations files, not drafting them by hand...

alexweininger: > Are there any other enums in index.d.ts that get used elsewhere?

No, this is the first one.

alexweininger: > Ugh...this is why we should be generating our declarations files, not drafting them by hand...

I agree. Do you happen to know why we don't generate them? Before I joined, I always thought they were auto generated.

utils/src/AzExtResourceType.ts
```diff
@@ -1,79 +1,67 @@
-/*---------------------------------------------------------------------------------------------
```
bwateratmsft, line 1: This file lost its header

utils/index.d.ts
bwateratmsft, line 1615: While these remain duplicated, we should make sure there are comments pointing one to the other.

utils/src/AzExtResourceType.ts
nturinski: Yeah, same here. I think it was just the easiest way to do at the time and we never revisited actually automating it.

bwateratmsft (approved):

PR #1225, opened 2022-08-19
Adding generateRelatedName for Storage Account Name

ccastrotrejo:
### Main Code Changes
- Add generateRelatedName in case the name of the storageAcount isn't valid (took the implementation from SiteNameStep)
- Update package version

### Facing issues on the Logic Apps (Standard) extension
- When user types a valid name for a LA and not valid name for Storage account (i.e. refortie-test-la2), an error popped up, not been able to continue with the creation of the LA. This is happening due that the LA name has dashes (invalid chars for Storage account names) and/or name is too long.

f3779de5-d16e-4d08-a082-48aca78823ec

eacd1d35-3b48-4ac4-a8ef-c76b4d24fb8d

### Implementation
- I added the generateRelatedName as I noticed the creation of other resources uses it and the portal uses this kind of behavior as well


azure/package.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "@microsoft/vscode-azext-azureutils",
     "author": "Microsoft Corporation",
-    "version": "0.3.4",
+    "version": "0.3.5",
```
bwateratmsft, line 4: This version number needs to be updated in `package-lock.json` as well. It appears in two places near the top of the document.

azure/src/wizard/StorageAccountCreateStep.ts
```diff
@@ -4,14 +4,18 @@
  *--------------------------------------------------------------------------------------------*/
 
 import type { SkuName, StorageManagementClient } from '@azure/arm-storage';
+import type { IAzureNamingRules } from '@microsoft/vscode-azext-utils';
+import { AzureWizardExecuteStep } from '@microsoft/vscode-azext-utils';
```
bwateratmsft, line 8: I think it's fine to include both of these in the same `import` (not `import type`), as the utils package is designed to not need to be lazy-loaded.

azure/src/wizard/StorageAccountCreateStep.ts
```diff
@@ -47,6 +51,57 @@ export class StorageAccountCreateStep<T extends types.IStorageAccountWizardConte
...
+        for (let invalidCharsRegExp of namingRules.map((n: IAzureNamingRules) => n.invalidCharsRegExp)) {
+            // Ensure the regExp uses the 'g' flag to replace _all_ invalid characters
+            invalidCharsRegExp = new RegExp(invalidCharsRegExp, 'g');
+            preferredName = preferredName.replace(invalidCharsRegExp, '');
+        }
+
+        const minLength: number = Math.max(...namingRules.map((n: IAzureNamingRules) => n.minLength));
+        const maxLength: number = Math.min(...namingRules.map((n: IAzureNamingRules) => n.maxLength));
+
+        const maxTries: number = 100;
```
bwateratmsft, line 70: 100 attempts (which will include a web request to check the name availability) is an awful lot of tries. I think 3-5 is a more reasonable maximum.

You could use a sort of exponential backoff on the `i` value, or choose a random number between 1-X, and try that, rather than sequentially checking each `i`.

azure/package.json
ccastrotrejo: I will update it over there as well. Thanks!

azure/src/wizard/StorageAccountCreateStep.ts
ccastrotrejo: Ohh ok ok. Thanks!

azure/package.json
nturinski: You should just run `npm install` as it will handle that for you.

azure/src/wizard/StorageAccountCreateStep.ts
ccastrotrejo: I reduced the amount of trying times to 5. But I couldn't understand quite well your comment on checking each `i`. Could you please elaborate on it?

bwateratmsft: What I mean is that as you have it implemented, you are checking 0, 1, 2, 3..., i.e. mystorageaccount2, mystorageaccount3, mystorageaccount4, etc. I think there are three reasonable approaches to this:

1. Sequential checking, like you are doing. As an advantage, it is "predictable", and behaves just like duplicate files are handled in OSs.
2. Exponential checking, e.g. you check 2^i (1, 2, 4, 8, 16, ...) or something similar
3. Random checking, e.g. pick a random number 1 to 100, check if it exists, if so just try again

I think it's sort of a PM question as to whether approach 1 is important enough to be desirable.

ccastrotrejo: Ohhh I see your point of view now. However I feel that sequential, int this case is not that bad as we are just doing it 5 times, random and exponential will be the same. Do you know who I can reach out to? or how can i solve this issue?

fiveisprime: I like the idea and I can't think of any immediately confusing or bad side-effects. I say go for it.

bwateratmsft: Which one?

bwateratmsft: I talked to @fiveisprime offline, the suggestion was to pick a random one 1-100, with a few retries (e.g. 5).

bwateratmsft (approved):
Looks good to me. @nturinski can you also take a look?

azure/src/wizard/StorageAccountCreateStep.ts
```diff
@@ -26,7 +29,7 @@ export class StorageAccountCreateStep<T extends types.IStorageAccountWizardConte
     public async execute(wizardContext: T, progress: Progress<{ message?: string; increment?: number }>): Promise<void> {
         const newLocation: string = (await LocationListStep.getLocation(wizardContext, storageProvider)).name;
         // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-        const newName: string = wizardContext.newStorageAccountName!;
+        const newName = await this.generateRelatedName(wizardContext, wizardContext.newStorageAccountName!, storageAccountNamingRules) as string;
```
nturinski, line 32: It seems unlikely, but `newName` can potentially be `undefined` by `generateRelatedName`. Maybe we should handle the error just in case? Since it would error out the entire create process (ie creating a Function App with a storage account). I think we could just output a message saying that we're skipping storage account creation and just return undefined.

azure/src/wizard/StorageAccountCreateStep.ts
```diff
@@ -47,6 +50,57 @@ export class StorageAccountCreateStep<T extends types.IStorageAccountWizardConte
...
+
+        return undefined;
+    }
+
+    protected async isRelatedNameAvailable(wizardContext: T, name: string): Promise<boolean> {
+        return await ResourceGroupListStep.isNameAvailable(wizardContext, name);
+    }
+
+    private generateSuffixedName(preferredName: string, minLength: number, maxLength: number): string {
+        const suffix: string = (Math.floor(Math.random() * (100 - 1)) + 1).toString();
```
nturinski, line 88: I don't see it in our azure package, but our other packages use a `getRandomHexString` to generate a suffix. I think it's more guaranteed to not have conflicts rather than a random number between 1-100.

```
export function getRandomHexString(length: number = 6): string {
        const buffer: Buffer = crypto.randomBytes(Math.ceil(length / 2));
        return buffer.toString('hex').slice(0, length);
    }
```

ccastrotrejo: Thanks Nathan! Gonna implement that one as the other packages

azure/src/wizard/StorageAccountCreateStep.ts
```diff
@@ -47,6 +50,59 @@ export class StorageAccountCreateStep<T extends types.IStorageAccountWizardConte
...
+        let newName: string;
+        while (count < maxTries) {
+            newName = this.generateSuffixedName(preferredName, minLength, maxLength);
+            if (await this.isRelatedNameAvailable(wizardContext, newName)) {
+                return newName;
+            }
+            count += 1;
+        }
+
+        const createdStorageAccount: string = localize('FailStorageAccount', 'Something went wrong in the creation of the storage account');
```
nturinski, line 80: Sorry, but could you remove this line? I know I mentioned that it'd be good to catch the error, but with this implementation, they'll see this error message in output, but when this line runs:
```
wizardContext.storageAccount = await storageClient.storageAccounts.beginCreateAndWait(
            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
            wizardContext.resourceGroup!.name!,
            newName,
            {
                sku: { name: newSkuName },
                kind: this._defaults.kind,
                location: newLocation,
                enableHttpsTrafficOnly: true
            }
        );
```
it'll throw another error, probably something about how the name is invalid or something (since it'll be undefined).

I'm not super worried about this actually happening with a randomHexString as a suffix, so we can probably just go ahead and remove these lines.

nturinski (approved):

PR #1229, opened 2022-08-30
Implement the ultimate quick pick experience

bwateratmsft:


alexweininger:
### Tree Item Pick scenarios

These scenarios cover all of the current uses of the Tree Item Picker API (TIP). More common scenarios are sorted to the top.

**A. Pick an Azure resource of a specific type (optional create)**
Example: "Azure Functions: Deploy to Function App" command must pick a deploy target

**B. Pick a child of an Azure resource (optional create)**
Example: "Azure Functions: Deploy to Slot" command must pick a slot of a Function App. If one doesn't exist, user should be presented with the option to create one.

**C. Pick an Azure resource or Workspace resource**
Example: "Azure Storage: Create Queue" command must pick a Storage account in Azure, or an attached (emulated) one

**D. Pick a child of an Azure resource or Workspace resource**
Example: "Azure Functions: Execute Function Now" command must pick a Function running in Azure, or a Function from a Functions Project in the workspace being run locally

**E. Pick an Azure resource (of ANY type) or a child of this resource**
Example: "Azure: View in Portal" command must pick any Azure resource or child resource that can be opened in the portal.

utils/src/treev2/quickPickWizard/ContextValueQuickPickStep.ts
```diff
@@ -0,0 +1,89 @@
...
+/**
+ * Describes filtering based on context value. Items that pass the filter will
+ * match at least one of the `include` filters, but none of the `exclude` filters.
+ */
+export interface ContextValueFilter {
+    /**
+     * This filter will include items that match *any* of the values in the array.
+     * When a string is used, exact value comparison is done.
+     */
+    include: string | RegExp | (string | RegExp)[];
```
bwateratmsft, line 20: Should we allow strings at all, or just `RegExp`? I'm worried the behavior could get unintuitive / confusing quickly when you consider that context values are now being treated as an array (sometimes), and are being sourced from multiple extensions.

alexweininger:
So far in my initial testing it works well. I had to make a few small changes but the concept is proven for sure. I want to go deeper and make sure we can cover all scenarios I outlined so I will report back next week with those results.

utils/src/treev2/quickPickWizard/ContextValueQuickPickStep.ts
```diff
@@ -0,0 +1,89 @@
...
+        const excludeOption = this.pickOptions.contextValueFilter.exclude;
+
+        const includeArray: (string | RegExp)[] = Array.isArray(includeOption) ? includeOption : [includeOption];
+        const excludeArray: (string | RegExp)[] = excludeOption ?
+            (Array.isArray(excludeOption) ? excludeOption : [excludeOption]) :
+            [];
+
+        const nodeContextValues: string[] = isV2TreeNode(node) ?
+            node.quickPickOptions.contextValues :
+            [node.contextValue];
```
alexweininger, line 54: Why are we nesting `contextValues` inside of `quickPickOptions`? If there isn't a strong reason then I would vote to not do it since we could avoid having to do this check.

utils/src/treev2/quickPickWizard/ContextValueQuickPickStep.ts
```diff
@@ -82,7 +82,7 @@ export class ContextValueQuickPickStep<TNode extends ContextValueFilterableTreeN
 function isV2TreeNode(maybeNode: unknown): maybeNode is ContextValueFilterableTreeNodeV2 {
     if (typeof maybeNode === 'object') {
         return Array.isArray((maybeNode as ContextValueFilterableTreeNodeV2).quickPickOptions?.contextValues) &&
-            (maybeNode as ContextValueFilterableTreeNodeV2).quickPickOptions?.isLeaf !== undefined;
+            (maybeNode as ContextValueFilterableTreeNodeV2).quickPickOptions?.isLeaf !== null;
```
bwateratmsft, line 85: We should probably check both `null` and `undefined`.

bwateratmsft:
Changed the base to the `api-v2` branch.

utils/src/treev2/quickPickWizard/quickPickAzureResource/tempTypes.ts
```diff
@@ -0,0 +1,40 @@
...
+export interface ApplicationResource extends ResourceBase {
+    readonly subscription: ApplicationSubscription;
+    readonly azExtResourceType?: types.AzExtResourceType;
+    readonly resourceGroup?: string;
+}
+
+export type ApplicationSubscription = unknown;
+
+export interface Box<T> {
+    unwrap(): T;
```
alexweininger, line 39: We originally had it as [`unwrap<T>(): Promise<T>`](https://github.com/microsoft/vscode-azuretools/blob/e7e66a25e7f3f21b2e08cbfbb196a90c76ca65fb/utils/hostapi.v2.d.ts#L15) but I think it makes sense to not make it a `Promise`.

bwateratmsft: I feel like I had a reason for doing `Promise<T>` but I can't remember what it is.

alexweininger: We need to use unwrap in a lot of non-async places (ex: in `ContextValueQuickPickStep#isDirectPick` so I will make it not a Promise for now.

utils/src/treev2/quickPickWizard/ContextValueQuickPickStep.ts
```diff
@@ -0,0 +1,90 @@
...
+        const excludeOption = this.pickOptions.contextValueFilter.exclude;
+
+        const includeArray: (string | RegExp)[] = Array.isArray(includeOption) ? includeOption : [includeOption];
+        const excludeArray: (string | RegExp)[] = excludeOption ?
+            (Array.isArray(excludeOption) ? excludeOption : [excludeOption]) :
+            [];
+
+        const nodeContextValues: string[] = isContextValueFilterableTreeNodeV2(node) ?
+            node.quickPickOptions.contextValues :
+            [node.contextValue];
```
alexweininger, line 54: note to self: need to do `node = isBox(node) ? node.unwrap() : node` here

alexweininger:
Just a note: we keep calling this the quick pick experience, but I think "pick tree item wizard" or "tree item picker wizard" is a better name. 😄

utils/src/treev2/quickPickWizard/GenericQuickPickStep.ts
```diff
@@ -0,0 +1,112 @@
...
+    public async prompt(wizardContext: TContext): Promise<void> {
+        try {
+            const pick = await this.promptInternal(wizardContext);
+            wizardContext.pickedNodes.push(pick);
+        } catch (err) {
+            const error = parseError(err);
+            if (error.errorType === 'GoBackError') {
+                // Instead of wiping out a property value, which is the default wizard behavior for `GoBackError`, pop the most recent
+                // value off from the provenance of the picks
+                wizardContext.pickedNodes.pop();
```
alexweininger, line 40: Going back currently doesn't work if a previous step was skipped. Skipped here meaning there was only one pick which was auto-selected and pushed into `pickedNodes` and never shows a prompt.

If a step never shows a prompt, the wizard will pop it from the finished steps array. Which in our case is bad because then the steps and pickedNodes become out of sync.

That happens here: https://github.com/microsoft/vscode-azuretools/blob/4f1923e6cff24fc23987f4b139cdf23f5aa4d716/utils/src/wizard/AzureWizard.ts#L232

Example:

Before hitting the back button:

```
currentStep: PickAppResourceStep
finishedSteps: [PickSubscriptionStep, PickGroupStep]
pickedNodes: [SubscriptionItem, ResourceTypeItem]
```

After hitting the back button:

```
currentStep: PickSubscriptionStep
finishedSteps: [PickSubscriptionStep]
pickedNodes: [SubscriptionItem]
```

Now the picks shown in the PickSubscriptionStep are children of the SubscriptionItem, which is no good.

utils/index.d.ts
```diff
@@ -1678,3 +1682,99 @@ export declare enum AzExtResourceType {
...
+/**
+ * Describes filtering based on context value. Items that pass the filter will
+ * match at least one of the `include` filters, but none of the `exclude` filters.
+ */
+export declare interface ContextValueFilter {
+    /**
+     * This filter will include items that match *any* of the values in the array.
+     * When a string is used, exact value comparison is done.
+     */
+    include: string | RegExp | (string | RegExp)[];
```
nturinski, line 1733: I think @bwateratmsft's comment kind of got buried, but supporting just RegExp/RegExp[] feels easier to manage for us. String comparisons feel less useful now with how context values are appended to the array now. 

For backwards compat, we can just change any string context values to RegExps.

bwateratmsft: We can get that change in after #1239.

utils/src/treev2/quickPickWizard/ContextValueQuickPickStep.ts
bwateratmsft: Duplicated here: https://github.com/microsoft/vscode-azuretools/pull/1229#discussion_r975785370

alexweininger:
@bot agree

alexweininger:
Probably time to mark this as ready for review and work towards merging.

utils/src/treev2/quickPickWizard/GenericQuickPickStep.ts
```diff
@@ -0,0 +1,99 @@
...
+
+        const picks: types.IAzureQuickPickItem<unknown>[] = [];
+        for (const choice of promptChoices) {
+            picks.push(await this.getQuickPickItem(...choice));
+        }
+
+        return picks;
+    }
+
+    private async getQuickPickItem(node: unknown, item: vscode.TreeItem): Promise<types.IAzureQuickPickItem<unknown>> {
```
nturinski, line 92: Since this is based on the `vscode.TreeDataProvider`, I feel like it might offer more clarity to name `node` as `element`. Are we forced to type `node` as `unknown`? Otherwise it seems like `T` would make more sense.

utils/src/treev2/quickPickWizard/QuickPickWithCreateStep.ts
```diff
@@ -0,0 +1,53 @@
...
+            }
+        }
+
+        picks.push(this.getCreatePick());
+        return picks as types.IAzureQuickPickItem<unknown>[];
+    }
+
+    private getCreatePick(): types.IAzureQuickPickItem<CreateCallback> {
+        return {
+            label: this.pickOptions.createLabel || localize('createQuickPickLabel', '$(add) Create...'),
```
nturinski, line 49: For us to keep consistency amongst our team, I think it might easier if there is were `resourceLabel` property that automatically generates `'$(add) Create ${this.pickOptions.resourceLabel}...`

But still allow a fully customizable `createLabel`.

utils/src/treev2/quickPickWizard/QuickPickWithCreateStep.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export class CreateQuickPickStep<TContext extends types.QuickPickWizardContext> extends ContextValueQuickPickStep<TContext, CreateQuickPickOptions> {
+    public override async prompt(wizardContext: TContext): Promise<void> {
+        await super.prompt(wizardContext);
+
+        const lastNode = getLastNode<unknown | CreateCallback>(wizardContext);
+        if (typeof lastNode === 'function') {
+            // If the last node is a function, pop it off the list and execute it
+            const callback = wizardContext.pickedNodes.pop() as unknown as CreateCallback;
+            wizardContext.pickedNodes.push(await callback());
```
nturinski, line 27: Definitely out of the scope of this PR, but I think ideally, we wouldn't await the callback here, but would instead only prompt, store those values in context, and then execute and set the remaining contexts as needed during execution.

utils/src/treev2/quickPickWizard/QuickPickWizardContext.ts
```diff
@@ -0,0 +1,14 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+import * as types from '../../../index';
+
+export function getLastNode<TNode = unknown>(context: types.QuickPickWizardContext): TNode | undefined {
+    if (context.pickedNodes.length) {
+        return context.pickedNodes[context.pickedNodes.length - 1] as TNode;
+    }
+
+    return undefined;
```
nturinski, line 13: I feel like that this should just throw an error if it was `undefined`. I pretty much see every instance of this throwing some variation of the `NoResourceFoundError` whenever it's invoked anyway.

If we need it for when `getLastNode` is the root (which I don't think we should allow users to go back to anyway), then maybe have a flag to stifle error throwing?

utils/src/treev2/quickPickWizard/RecursiveQuickPickStep.ts
```diff
@@ -0,0 +1,33 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as types from '../../../index';
+import { ContextValueFilterQuickPickOptions, ContextValueQuickPickStep } from './ContextValueQuickPickStep';
+import { getLastNode } from './QuickPickWizardContext';
+
+export class RecursiveQuickPickStep<TContext extends types.QuickPickWizardContext> extends ContextValueQuickPickStep<TContext, ContextValueFilterQuickPickOptions> {
```
nturinski, line 10: Even though you return a new step with `hideStepCount` set to `true`, we should just define it as its default property. Otherwise wouldn't the first step has a step count and then it just disappears?

utils/src/treev2/quickPickWizard/common/PickFilter.ts
```diff
@@ -0,0 +1,20 @@
...
+*--------------------------------------------------------------------------------------------*/
+
+import * as vscode from "vscode";
+
+export interface PickFilter<TPick = vscode.TreeItem> {
+    /**
+     * Filters for nodes that match the final target.
+     * @param node The node to apply the filter to
+     */
+    isDirectPick(node: TPick): boolean;
```
nturinski, line 13: I wasn't clear what exactly these filters meant by their name alone. The descriptions are perfect, but what if we named them something more like `isFinalPick` and `isAncestorPick`? 

I also like the use of ancestor because it implies that is has to be a parent node.

utils/src/treev2/quickPickWizard/compatibility/CompatibilityContextValueQuickPickStep.ts
```diff
@@ -0,0 +1,54 @@
...
+                if (customPick) {
+                    wizardContext.pickedNodes.push(customPick);
+                    return true;
+                }
+            }
+        }
+        return false;
+    }
+
+    private async getCustomChildren(context: TContext, node: AzExtParentTreeItem): Promise<AzExtTreeItem | undefined> {
```
nturinski, line 51: Shouldn't we also filter based on `exclude`?

utils/src/treev2/quickPickWizard/compatibility/CompatibilityRecursiveQuickPickStep.ts
```diff
@@ -0,0 +1,115 @@
...
+import { isWrapper } from "../../../registerCommandWithTreeNodeUnwrapping";
+
+export interface CompatibilityRecursiveQuickPickOptions extends ContextValueFilterQuickPickOptions {
+    create?: types.CreateOptions;
+}
+
+/**
+ * Recursive step which is compatible which adds create picks based if the node has {@link types.CompatibleQuickPickOptions.createChild quickPickOptions.createChild} defined.
+ */
+export class CompatibilityRecursiveQuickPickStep<TContext extends types.QuickPickWizardContext> extends CompatibilityContextValueQuickPickStep<TContext, CompatibilityRecursiveQuickPickOptions> {
```
nturinski, line 23: Woof.

Why does the recursive step also have to support createChild? Shouldn't we just have a `CompatibilityCreateQuickPickStep` instead of trying to force both of these implementations into one?

utils/src/treev2/quickPickWizard/experiences/appResourceExperience.ts
```diff
@@ -0,0 +1,63 @@
...
+import { getLastNode } from '../QuickPickWizardContext';
+import { NoResourceFoundError } from '../../../errors';
+import { AzureWizardPromptStep } from '../../../wizard/AzureWizardPromptStep';
+import { AzExtResourceType } from '../../../AzExtResourceType';
+import { AzureWizard } from '../../../wizard/AzureWizard';
+import { AzureResourceQuickPickWizardContext } from '../../../../hostapi.v2';
+import { isWrapper } from '../../../registerCommandWithTreeNodeUnwrapping';
+import { ResourceGroupsItem } from '../quickPickAzureResource/tempTypes';
+
+export async function appResourceExperience<TPick>(context: types.IActionContext, tdp: vscode.TreeDataProvider<ResourceGroupsItem>, resourceTypes?: AzExtResourceType | AzExtResourceType[], childItemFilter?: types.ContextValueFilter): Promise<TPick> {
```
nturinski, line 21: Should this support creating resources as well?

utils/src/treev2/quickPickWizard/experiences/contextValueExperience.ts
```diff
@@ -0,0 +1,42 @@
...
+import { RecursiveQuickPickStep } from '../RecursiveQuickPickStep';
+import { getLastNode } from '../QuickPickWizardContext';
+import { NoResourceFoundError } from '../../../errors';
+import { AzureWizardPromptStep } from '../../../wizard/AzureWizardPromptStep';
+import { AzureWizard } from '../../../wizard/AzureWizard';
+import { isWrapper } from '../../../registerCommandWithTreeNodeUnwrapping';
+import { ResourceGroupsItem } from '../quickPickAzureResource/tempTypes';
+
+export async function contextValueExperience<TPick extends unknown>(context: types.IActionContext, tdp: vscode.TreeDataProvider<ResourceGroupsItem>, contextValueFilter: types.ContextValueFilter): Promise<TPick> {
+    const promptSteps: AzureWizardPromptStep<types.QuickPickWizardContext>[] = [
```
nturinski, line 17: I'm a little bit confused here. Do we not have to still do the first three steps of `appResourceExperience` to bypass the subscription, group, and app resource?

Will `RecursiveQuickPickStep` be able to do that on its own?

utils/src/treev2/quickPickWizard/experiences/contextValueExperience.ts
```diff
@@ -0,0 +1,42 @@
...
+        new RecursiveQuickPickStep(tdp, {
+            contextValueFilter: contextValueFilter,
+            skipIfOne: false,
```
nturinski, line 20: I feel like there are many times where we would want this to be true, such as only one sub, only one group, etc. It's only the last pick where this would be unwanted behavior, right?

utils/src/treev2/quickPickWizard/experiences/subscriptionExperience.ts
```diff
@@ -0,0 +1,32 @@
...
+
+import * as types from '../../../../index';
+import * as vscode from 'vscode';
+import { AzureWizard } from '../../../wizard/AzureWizard';
+import { QuickPickAzureSubscriptionStep } from '../quickPickAzureResource/QuickPickAzureSubscriptionStep';
+import { ApplicationSubscription, PickSubscriptionWizardContext } from '../../../../hostapi.v2';
+import { ResourceGroupsItem } from '../quickPickAzureResource/tempTypes';
+import { NoResourceFoundError } from '../../../errors';
+
+export async function subscriptionExperience(context: types.IActionContext, tdp: vscode.TreeDataProvider<ResourceGroupsItem>): Promise<ApplicationSubscription> {
```
nturinski, line 14: Should the experience just have an ExperienceBase class that does all this boilerplate? You have to declare the wizard, wizard context, prompt every experience.

utils/src/treev2/quickPickWizard/quickPickAzureResource/QuickPickAzureSubscriptionStep.ts
```diff
@@ -0,0 +1,42 @@
...
+
+class AzureSubscriptionPickFilter implements PickFilter {
+    isDirectPick(_node: vscode.TreeItem): boolean {
+        // Subscription is never a direct pick
+        return false;
+    }
+
+    isIndirectPick(_node: vscode.TreeItem): boolean {
+        // All nodes at this level are always subscription nodes
+        return true;
```
nturinski, line 40: Should we have a different step for handling workspace _and_ Subscription nodes?

nturinski (commented):
There seems to still be a lot of TODO comments. Wasn't sure if we should just approve and merge so we know what's reviewed already since this isn't going into main anyway.

utils/src/registerCommandWithTreeNodeUnwrapping.ts
```diff
@@ -0,0 +1,54 @@
+/*---------------------------------------------------------------------------------------------
```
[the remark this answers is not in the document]
alexweininger: Yup

utils/src/tree/isAzExtParentTreeItem.ts
```diff
@@ -0,0 +1,14 @@
...
+*  Licensed under the MIT License. See License.txt in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import type { AzExtParentTreeItem } from "./AzExtParentTreeItem";
+import type { IAzExtParentTreeItemInternal } from "./InternalInterfaces";
+
+/**
+ * Using instanceof AzExtParentTreeItem causes issues since each extension has their own version of the utils. Instead, check _isAzExtParentTreeItem
+ */
+export function isAzExtParentTreeItem(maybeParentTreeItem: unknown): maybeParentTreeItem is AzExtParentTreeItem {
```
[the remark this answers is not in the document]
alexweininger: It breaks stuff if this isn't in it's own file. I'll move `isAzExtTreeItem` in here if I can.

utils/src/treev2/quickPickWizard/ContextValueQuickPickStep.ts
```diff
@@ -0,0 +1,53 @@
...
+
+        const nodeContextValues: string[] = parseContextValue(node.contextValue);
+
+        return includeArray.some(i => this.matchesSingleFilter(i, nodeContextValues)) &&
+            !excludeArray.some(e => this.matchesSingleFilter(e, nodeContextValues));
+    }
+
+    isIndirectPick(node: TreeItem): boolean {
+        // `TreeItemCollapsibleState.None` and `undefined` are both falsy, and indicate that a `TreeItem` cannot have children--and therefore, cannot be an indirect pick
+        return !node.collapsibleState;
```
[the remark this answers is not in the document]
alexweininger: This step is used by both the compatibility experience, and the non-compatibility experiences. So `TreeItem` won't necessarily be an `AzExtTreeItem`.

> I know that we want to use vscode's implementation of TreeItem but I feel uneasy relying on a property that vscode controls and changes whenever it sees fit.

I don't think VS Code changes this property. For example, we define it only as a `getter` in `AzExtParentTreeItem`.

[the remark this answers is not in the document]
alexweininger: > Does it ever come in as undefined because the collapsibleState hadn't been set yet? I remember you were playing with this quite a bit.

If you don't set `collapsibleState`, then VS Code treats it as `TreeItemCollapsibleState.None` which has a value of 0. It can do this since `undefined` and `0` are falsy.

utils/src/treev2/quickPickWizard/experiences/appResourceExperience.ts
alexweininger: Eventually yes. I have an API in the works to allow for it. Need to dedicate some more time to it.

utils/src/treev2/quickPickWizard/experiences/subscriptionExperience.ts
alexweininger: The API I keep mentioning that I'm working on fixes this problem. It's an "experience builder" tool. You'll use it to build an experience by adding the steps you want with the settings you want. Still WIP

utils/src/treev2/quickPickWizard/quickPickAzureResource/QuickPickAzureSubscriptionStep.ts
alexweininger: We still have to work out how we want selecting from either a subscription or a workspace item to work. Once I'm further along with the "experience builder" we can go from there on how to tackle this.

utils/src/treev2/quickPickWizard/QuickPickWithCreateStep.ts
alexweininger: Yeah that's exactly the experience we want.

utils/src/treev2/quickPickWizard/compatibility/CompatibilityContextValueQuickPickStep.ts
alexweininger: We're only passing `include` because that's what `AzExtParentTreeItem.pickTreeItemImpl` is expecting. We've never had an `exclude` option before, so it will work as expected by ignoring it. Only v2 extensions will use the `exclude` property, and they won't be using the compat steps.

utils/src/treev2/quickPickWizard/common/PickFilter.ts
alexweininger: I agree. I like `isFinalPick`, and `isAncestorPick` (I thought about `isIntermediatePick`, but I like `isAncestorPick` better).

utils/src/treev2/quickPickWizard/RecursiveQuickPickStep.ts
alexweininger: Makes sense. Although, all the wizards using this step have `hideStepCount` set to `true`.

utils/src/treev2/quickPickWizard/ContextValueQuickPickStep.ts
```diff
@@ -0,0 +1,53 @@
...
+import * as types from '../../../index';
+import { parseContextValue } from '../../utils/contextUtils';
+import { PickFilter } from './common/PickFilter';
+import { GenericQuickPickOptions, GenericQuickPickStep } from './GenericQuickPickStep';
+
+export interface ContextValueFilterQuickPickOptions extends GenericQuickPickOptions {
+    contextValueFilter: types.ContextValueFilter;
+}
+
+export class ContextValueQuickPickStep<TContext extends types.QuickPickWizardContext, TOptions extends ContextValueFilterQuickPickOptions> extends GenericQuickPickStep<TContext, TOptions> {
```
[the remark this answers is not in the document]
alexweininger: I think the naming difference is OK since the `FindByIdQuickPickStep` is not used for "pick" experiences. It doesn't show any UI and is only used to "find" items.

Also I'm not sure the find by id stuff is going to stick around. It may end up being implemented directly on the TDP in resource groups.

utils/hostapi.v2.d.ts
```diff
@@ -1,39 +1,72 @@
...
 /**
- * Tests to see if something is a box, by ensuring it is an object
- * and has an "unwrap" function
- * @param maybeBox An object to test if it is a box
- * @returns True if a box, false otherwise
+ * Represents an individual resource in Azure.
+ * @remarks The `id` property is expected to be the Azure resource ID.
  */
-export declare function isBox(maybeBox: unknown): maybeBox is Box;
+export declare interface ApplicationResource extends ResourceBase {
```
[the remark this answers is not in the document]
alexweininger: That's valid, @philliphoff @bwateratmsft thoughts?

utils/src/treev2/quickPickWizard/quickPickAzureResource/QuickPickAzureSubscriptionStep.ts
nturinski: It's currently broken for all of the extensions that used to: Functions, Storage, and Databases.

utils/hostapi.v2.d.ts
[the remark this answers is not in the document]
bwateratmsft: IIRC I just copied what was up in the api-v2 branch from microsoft/vscode-azureresourcegroups. We probably need to refresh these interfaces from upstream, and move it into a separate PR.

utils/src/treev2/quickPickWizard/FindByIdQuickPickStep.ts
```diff
@@ -0,0 +1,72 @@
...
+
+import * as types from '../../../index';
+import * as vscode from 'vscode';
+import { getLastNode } from './QuickPickWizardContext';
+import { GenericQuickPickStep, SkipIfOneQuickPickOptions } from './GenericQuickPickStep';
+import { PickFilter } from './common/PickFilter';
+
+interface FindByIdQuickPickOptions extends SkipIfOneQuickPickOptions {
+    id: string;
+    skipIfOne?: true;
```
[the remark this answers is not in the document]
bwateratmsft: You're correct. I think I created `FindByIdQuickPickOptions` before `SkipIfOneQuickPickOptions` and this carried over unnecessarily.

utils/src/treev2/quickPickWizard/GenericQuickPickStep.ts
```diff
@@ -0,0 +1,99 @@
...
+import { getLastNode } from './QuickPickWizardContext';
+import { AzureWizardPromptStep } from '../../wizard/AzureWizardPromptStep';
+import { PickFilter } from './common/PickFilter';
+import { localize } from '../../localize';
+
+export interface GenericQuickPickOptions {
+    skipIfOne?: boolean;
+}
+
+export interface SkipIfOneQuickPickOptions extends GenericQuickPickOptions {
```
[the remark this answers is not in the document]
bwateratmsft: In my mind that's basically the value--certain steps _always_ have `skipIfOne` as `true`, so I wanted to be maximally explicit--so that users of those steps understand they cannot set it to `false` (and, TS will complain if they try).

utils/src/treev2/quickPickWizard/GenericQuickPickStep.ts
bwateratmsft: > Are we forced to type `node as unknown`

I don't think so, but I could be wrong. We could add a generic argument of `TElement extends unknown` and apply that as the type of the `TreeDataProvider` given in the constructor and the argument here.

utils/src/treev2/quickPickWizard/QuickPickWithCreateStep.ts
bwateratmsft: IIRC Docker extension does something along those lines.

utils/src/treev2/quickPickWizard/compatibility/CompatibilityRecursiveQuickPickStep.ts
bwateratmsft: That's fair. I drew it up this way because we can't have multiple inheritance :sob:

utils/src/treev2/quickPickWizard/experiences/contextValueExperience.ts
bwateratmsft: `contextValueExperience` was designed with not-necessarily-the-Azure-views in mind (e.g., I wanna use it in the Docker extension). As long as we have properly defined `isAncestorPick` / `isFinalPick`, then it will work in the Azure Resource views _and_ elsewhere.

utils/src/treev2/quickPickWizard/RecursiveQuickPickStep.ts
bwateratmsft: I didn't realize that `hideStepCount` could be a property of the steps themselves, else I'd have done it that way from the start. :smile:

utils/hostapi.v2.d.ts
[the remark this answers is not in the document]
philliphoff: I don't have any more insight on the original naming, though I'll totally buy @alexweininger's explanation.

utils/src/tree/isAzExtParentTreeItem.ts
```diff
@@ -0,0 +1,19 @@
...
+*--------------------------------------------------------------------------------------------*/
+
+import type { AzExtParentTreeItem } from "./AzExtParentTreeItem";
+import { AzExtTreeItem } from "./AzExtTreeItem";
+import type { IAzExtParentTreeItemInternal } from "./InternalInterfaces";
+
+/**
+ * Using instanceof AzExtParentTreeItem causes issues since each extension has their own version of the utils. Instead, check _isAzExtParentTreeItem
+ */
+export function isAzExtParentTreeItem(maybeParentTreeItem: unknown): maybeParentTreeItem is AzExtParentTreeItem {
```
philliphoff, line 13: If we end up with any more of these I'd suggest switching to something like `_azExtParentTreeItemType` with an enum/flags type.

utils/hostapi.v2.d.ts
```diff
@@ -0,0 +1,67 @@
...
+    subscription?: ApplicationSubscription;
+}
+
+export declare interface AzureResourceQuickPickWizardContext extends QuickPickWizardContext, PickSubscriptionWizardContext {
+    resource?: ApplicationResource;
+    resourceGroup?: string;
+}
+
+export interface V2AzureResourcesApi extends AzureExtensionApi {
+    readonly applicationResourceTreeDataProvider: vscode.TreeDataProvider<unknown>;
```
philliphoff, line 65: It looks like only the `getTreeItem()` and `getChildren()` functions of the TDPs are actually used; would it make sense to expose only those (i.e. just a subset of the TDP interface), to help ensure extensions don't take dependencies on other aspects of the TDP?  Maybe something like:

```typescript
export interface V2AzureResourcesApi extends AzureExtensionApi {
  readonly applicationResources: Pick<vscode.TreeDataProvider<unknown>, 'getChildren' | 'getTreeItem'>;
  readonly workspaceResources: Pick<vscode.TreeDataProvider<unknown>, 'getChildren' | 'getTreeItem'>;
}
```

utils/hostapi.v2.d.ts
```diff
@@ -0,0 +1,67 @@
...
+    readonly location?: string;
+    readonly resourceGroup?: string;
+    /** Resource tags */
+    readonly tags?: {
+        [propertyName: string]: string;
+    };
+    /* add more properties from GenericResource if needed */
+}
+
+export declare interface PickSubscriptionWizardContext extends QuickPickWizardContext {
```
philliphoff, line 55: Should these context types belong here vs. exposed more directly as part of the utils library?

utils/src/treev2/quickPickWizard/common/PickFilter.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export interface PickFilter<TPick = vscode.TreeItem> {
+    /**
+     * Filters for nodes that match the final target.
+     * @param node The node to apply the filter to
+     */
+    isFinalPick(node: TPick): boolean;
+
+    /**
+     * Filters for nodes that could have a descendant matching the final target.
```
philliphoff, line 16: "have a descendant matching..." or "have an ancestor matching..."

utils/hostapi.v2.d.ts
bwateratmsft: That seems reasonable to me.

utils/hostapi.v2.d.ts
bwateratmsft: That's a good question. Some of them it definitely makes sense to have on the "basic" parts of the Utils library--e.g. those related to generic context-based picking, not anything specific to RGs. But some IMO kinda belong on the RG interfaces, e.g. those related to picking Azure subscriptions, resources, etc.

utils/src/treev2/quickPickWizard/common/PickFilter.ts
bwateratmsft: To match the function name I think it could be something like this:
"Filters for nodes that could be an ancestor of a node matching the final target."

utils/src/utils/apiUtils.ts
```diff
@@ -0,0 +1,32 @@
...
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { Extension, extensions } from "vscode";
+import { localize } from "vscode-nls";
+import type { AzureExtensionApi, AzureExtensionApiProvider } from "../../api";
+import { AzureHostExtensionApi } from "../../hostapi";
+
+export async function getResourceGroupsApi(apiVersionRange: AzureExtensionApi['apiVersion']): Promise<AzureHostExtensionApi>;
+export async function getResourceGroupsApi<T extends AzureExtensionApi>(apiVersionRange: string): Promise<T> {
```
philliphoff, line 12: The RG `getApi()` API has an options argument where extensions can pass in their ID for telemetry purposes (which we should encourage).  We should mirror that here.

utils/src/treev2/quickPickWizard/quickPickAzureResource/tempTypes.ts
```diff
@@ -0,0 +1,26 @@
...
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { ApplicationResource, ApplicationSubscription } from '../../../../hostapi.v2';
+import * as types from '../../../../index';
+
+// TODO: THIS FILE IS TEMPORARY //
+// It needs to be replaced by real Resources extension interfaces //
+
+// These are assumptions made about the nodes in the tree
```
philliphoff, line 12: Is there a way we could formalize this in the V2 API definitions without necessarily having them be blind casts?  If we strongly control the TDP functions we expose, we could do something like expose the `ResourceGroupsItem` type with a `type` property of `'Subscription' | 'Grouping' | 'Resource' | 'BranchData'` or some such thing.

utils/src/treev2/quickPickWizard/quickPickAzureResource/QuickPickAppResourceStep.ts
```diff
@@ -0,0 +1,62 @@
...
+        // If childItemFilter is undefined, this cannot be an indirect pick
+        if (!this.pickOptions.childItemFilter) {
+            return false;
+        }
+
+        return this.matchesResourceType(parseContextValue(node.contextValue));
+    }
+
+    private matchesResourceType(contextValues: string[]): boolean {
+        if (!contextValues.includes('azureResource')) {
```
philliphoff, line 56: Just checking: we've made sure that `BranchDataProviderItem` instances that represent application resources generates tree items with the `'azureResource'` context value (in addition to whatever the BDP generates)?

utils/src/utils/apiUtils.ts
alexweininger: Planning on doing that in a follow up PR. I probably shouldn't have included these api utils in this PR to begin with.

utils/src/treev2/quickPickWizard/quickPickAzureResource/QuickPickAppResourceStep.ts
alexweininger: Yes. That's definitely happening in the compatibility branch. I'm not sure if it's currently happening in the api-v2 branch yet.

alexweininger: The `'azureResource'` context is put into the `options.contextValues` in `GroupingItem.getChildren`.

utils/src/tree/isAzExtParentTreeItem.ts
alexweininger: I don't think we're planning on adding more of these.

utils/src/treev2/quickPickWizard/experiences/subscriptionExperience.ts
philliphoff, line 14: Should `context` be of a type that folds in `PickSubscriptionWizardContext`? Or, if the intent is that the caller not provide any "pre-filled" context that would otherwise bypass wizard steps, then it seems like the wizard context should be an object that's independent of the passed context.  Otherwise, the context will end up with a `subscription` property set as a side-effect, which may not be expected.

This might also work:

```TypeScript
const wizardContext = { ...context } as PickSubscriptionWizardContext;
```

philliphoff: (And the question of the other contexts in the "experience" functions.)

utils/src/treev2/quickPickWizard/experiences/compatibility/compatibilityPickSubscriptionExperience.ts
```diff
@@ -0,0 +1,25 @@
...
+import { ISubscriptionContext } from '@microsoft/vscode-azext-dev';
+import { subscriptionExperience } from '../subscriptionExperience';
+import { createSubscriptionContext } from '../../../../utils/credentialUtils';
+import { ResourceGroupsItem } from '../../quickPickAzureResource/tempTypes';
+import { isAzExtTreeItem } from '../../../../tree/isAzExtParentTreeItem';
+
+/**
+ * Returns `ISubscriptionContext` instead of `ApplicationSubscription` for compatibility.
+ */
+export async function compatibilitySubscriptionExperience(context: types.IActionContext, tdp: vscode.TreeDataProvider<ResourceGroupsItem>): Promise<ISubscriptionContext> {
```
philliphoff, line 17: Nit: I feel like we need a name that's a little more verb-y than `compatibilityXxxExperience()`, maybe something plain and boring like `pickSubscriptionWithCompatibility()`?

utils/src/treev2/quickPickWizard/RecursiveQuickPickStep.ts
```diff
@@ -0,0 +1,34 @@
...
+
+    public async getSubWizard(wizardContext: TContext): Promise<types.IWizardOptions<TContext> | undefined> {
+        const lastPickedItem = getLastNode(wizardContext);
+
+        if (!lastPickedItem) {
+            // Something went wrong, no node was chosen
+            throw new Error('No node was set after prompt step.');
+        }
+
+        if (this.pickFilter.isFinalPick(await this.treeDataProvider.getTreeItem((lastPickedItem)))) {
```
philliphoff, line 21: Nit: are the double parens needed? (i.e. `((lastPickedItem))`)

utils/src/treev2/quickPickWizard/QuickPickWizardContext.ts
```diff
@@ -0,0 +1,14 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as types from '../../../index';
+
+export function getLastNode<TNode = unknown>(context: types.QuickPickWizardContext): TNode | undefined {
+    if (context.pickedNodes.length) {
```
philliphoff, line 9: Maybe: `return context.pickedNodes.at(-1) as TNode;` ?

utils/index.d.ts
```diff
@@ -1711,10 +1714,71 @@ export declare interface Wrapper {
...
+
 /**
  * Tests to see if something is a wrapper, by ensuring it is an object
  * and has an "unwrap" function
  * @param maybeWrapper An object to test if it is a wrapper
  * @returns True if a wrapper, false otherwise
  */
 export declare function isWrapper(maybeWrapper: unknown): maybeWrapper is Wrapper;
+
+export declare function appResourceExperience<TPick extends unknown>(context: IActionContext, tdp: TreeDataProvider<ResourceGroupsItem>, resourceTypes?: AzExtResourceType | AzExtResourceType[], childItemFilter?: ContextValueFilter): Promise<TPick>;
```
philliphoff, line 1728: Maybe namespacing could be helpful? E.g.:

```TypeScript
picks.byResourceType(context, tdp, resourceTypes, ....);
```
Not sure what wording would work best, but an idea.

philliphoff (commented):
Seems like a good direction, though I'm not as familiar with the wizarding/step framework as the others, so I can't speak as much to how those fit together.  I think my main feedback is trying to limit direct access of the RG's TDPs as much as reasonably possible, both to minimize the surface area of the RG API as well as discourage other users of the API from taking dependencies on functions we intent to be used only in very specific scenarios.

utils/src/treev2/quickPickWizard/QuickPickWizardContext.ts
alexweininger: Nice, forgot we could use that now. Note: had to update `@types/node` to v16.

utils/src/tree/isAzExtParentTreeItem.ts
```diff
@@ -0,0 +1,19 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.txt in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import type { AzExtParentTreeItem } from "./AzExtParentTreeItem";
+import { AzExtTreeItem } from "./AzExtTreeItem";
```
bwateratmsft, line 7: Can this also be `import type`?

utils/src/treev2/quickPickWizard/FindByIdQuickPickStep.ts
```diff
@@ -0,0 +1,71 @@
...
+        // TODO: this code is nearly identical to `RecursiveQuickPickStep`, but this class can't inherit from it because it's
+        // not at all based on context value for filtering
+        const lastPickedItem = getLastNode(wizardContext);
+
+        if (!lastPickedItem) {
+            // Something went wrong, no node was chosen
+            throw new Error('No node was set after prompt step.');
+        }
+
+        if (this.pickFilter.isFinalPick(await this.treeDataProvider.getTreeItem(lastPickedItem))) {
```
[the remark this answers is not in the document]
bwateratmsft: We could cache but I would hope that `getTreeItem` should be so fast it doesn't matter.

utils/src/treev2/quickPickWizard/GenericQuickPickStep.ts
```diff
@@ -0,0 +1,99 @@
...
+            });
+
+            return selected.data;
+        }
+    }
+
+    protected async getPicks(wizardContext: TContext): Promise<types.IAzureQuickPickItem<unknown>[]> {
+        const lastPickedItem: unknown | undefined = getLastNode(wizardContext);
+
+        // TODO: if `lastPickedItem` is an `AzExtParentTreeItem`, should we clear its cache?
```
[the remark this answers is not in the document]
bwateratmsft: Yes. we want the quick pick experience to always fetch the latest data.

utils/src/tree/isAzExtParentTreeItem.ts
alexweininger: Yup

utils/src/treev2/quickPickWizard/experiences/compatibility/compatibilityPickSubscriptionExperience.ts
alexweininger: I organized it into a namespace so it's `PickTreeItemWithCompatibility.subscription`

utils/src/treev2/quickPickWizard/experiences/findByIdExperience.ts
```diff
@@ -0,0 +1,40 @@
...
+import * as types from '../../../../index';
+import * as vscode from 'vscode';
+import { getLastNode } from '../QuickPickWizardContext';
+import { NoResourceFoundError } from '../../../errors';
+import { FindByIdQuickPickStep } from '../FindByIdQuickPickStep';
+import { isWrapper } from '../../../registerCommandWithTreeNodeUnwrapping';
+import { AzureWizard } from '../../../wizard/AzureWizard';
+import { ResourceGroupsItem } from '../quickPickAzureResource/tempTypes';
+
+export async function findByIdExperience<TPick extends types.FindableByIdTreeNode>(context: types.IActionContext, tdp: vscode.TreeDataProvider<ResourceGroupsItem>, id: string): Promise<TPick> {
```
alexweininger, line 15: I'm thinking about removing the `findByIdExperience` altogether for a couple of reasons:

1. We aren't accounting for the custom `isAncestorOf` logic that needs to take place to find items under grouping items in the application resource tree.
2. We only need this for compatibility with the v1.5 extensions, so it makes sense for find to be a custom method inside RGs. Fully converted extensions shouldn't use find to reach into other extensions anymore.

utils/hostapi.v2.d.ts
```diff
@@ -0,0 +1,69 @@
...
+import { AzureExtensionApi } from "./api";
+
+export declare interface ApplicationAuthentication {
+    getSession(scopes?: string[]): vscode.ProviderResult<vscode.AuthenticationSession>;
+}
+
+/**
+ * Information specific to the Subscription
+ */
+export interface ApplicationSubscription {
```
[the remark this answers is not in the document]
alexweininger: Waiting on #1289

utils/src/treev2/quickPickWizard/ContextValueQuickPickStep.ts
```diff
@@ -0,0 +1,53 @@
...
+}
+
+class ContextValuePickFilter implements PickFilter {
+    constructor(private readonly pickOptions: ContextValueFilterQuickPickOptions) { }
+
+    isDirectPick(node: TreeItem): boolean {
+        const includeOption = this.pickOptions.contextValueFilter.include;
+        const excludeOption = this.pickOptions.contextValueFilter.exclude;
+
+        const includeArray: (string | RegExp)[] = Array.isArray(includeOption) ? includeOption : [includeOption];
```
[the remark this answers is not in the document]
alexweininger: Ehhh idk how I feel about replacing a one liner with a one liner util

utils/index.d.ts
```diff
@@ -995,7 +997,7 @@ export interface IAzureQuickPickItem<T = undefined> extends QuickPickItem {
 /**
  * Provides additional options for QuickPicks used in Azure Extensions
  */
-export interface IAzureQuickPickOptions extends QuickPickOptions, AzExtUserInputOptions {
+export interface IAzureQuickPickOptions extends VSCodeQuickPickOptions, AzExtUserInputOptions {
```
nturinski, line 1000: I think in the past, we used to prefix all of our interfaces/classes/etc with `Azure`, but decided to change it to `AzExt` due to there being confusion about using the `Azure` interfaces on non-Azure resources.

This came about specifically when we started working with local function projects and the name `AzureParentTreeItem` didn't really make sense on it.

So unless something actually has a hard dependency on Azure (as in, it requires an Azure subscription/resource), we have been opting to prefix it with `AzExt`. I personally prefer this naming scheme, but no strong opinions either way.

utils/src/treev2/quickPickWizard/ContextValueQuickPickStep.ts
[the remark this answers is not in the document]
nturinski: Fair enough, it's just ugly and a pain to write or copy and paste every time 😋

utils/index.d.ts
bwateratmsft: I don't wanna change this now, though. We aren't adding the interface, it's already in.

nturinski: Oh, whoops. Okay, probably back from when we were still using `Azure` as the prefix then 😬

alexweininger: I agree that we should change it. But also agree we should do it later.

nturinski (approved):
I think I've resolved any open comments that I had. Anyway, looks good to me 👍

PR #1230, opened 2022-09-02
Fix request aborted error on zip deploy

alexweininger:
Fixes https://github.com/microsoft/vscode-azureappservice/issues/2305

I'm able to reproduce https://github.com/microsoft/vscode-azureappservice/issues/2305 consistently by using a large project (over 100 MB).

Request sometimes takes longer than the 5s timeout, which caused "The request was aborted" errors.

In my testing, the request usually took less than 2s. Sometimes took more than 5s to complete. And rarely took more than 10s.

Changes made to prevent the error:
* Increase request timeout from 5 to 10 seconds
* Retry request up to 4 times

### DIFF
--- a/appservice/package-lock.json
+++ b/appservice/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "@microsoft/vscode-azext-azureappservice",
-    "version": "0.7.2",
+    "version": "0.7.3",
     "lockfileVersion": 2,
     "requires": true,
     "packages": {
         "": {
             "name": "@microsoft/vscode-azext-azureappservice",
-            "version": "0.7.2",
+            "version": "0.7.3",
             "license": "MIT",
             "dependencies": {
                 "@azure/abort-controller": "^1.0.4",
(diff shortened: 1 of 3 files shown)


bwateratmsft (dismissed):

nturinski (commented):
I'm not sure that you need to use the `retryKuduCall` because that was kind of the point of having the `sendRequestWithTimeout` call within a while loop.  We already have a `maxTimeToWaitForExpectedId` of a minute, so it was retrying 12 times already.  If we increase the timeout to 10 seconds, then it'll change it to 6 times (so we may want to adjust the `maxTimeToWaitForExpectedId`) but I'm worried using the `retryKuduCall` will make the cancellationToken to stop working if the deployment is canceled while it's in the `retryKuduCall` loop.

For reference, deployments are only canceled if you try to local git deploy something that is already being deployed.

alexweininger:
> it was retrying 12 times already

I don't think it was, we're awaiting the call and it throws an exception, exiting the while loop immediately.

nturinski:
> I don't think it was, we're awaiting the call and it throws an exception, exiting the while loop immediately.

Ah you're right, if only retries if if the get request returns nothing. Yeah, I think it makes sense to just try/catch it rather than using the `retryKuduCall`. Though we should probably only swallow connection timeout errors.

nturinski (approved):

PR #1235, opened 2022-09-14
Add compatibility steps

alexweininger:
* `CompatibilityRecursiveQuickPickStep` - provides compatibility with the `AzExtParentTreeItem.pickTreeItemImpl` method.
* `CompatibilityContextValueQuickPickStep` - provides compatibility with `AzExtParentTreeItem.createChildImpl` and related properties for showing a "+ Create..." pick.
* Added `_isAzExtTreeItem: boolean` property to `AzExtTreeItem`
	* clients will need to update to the latest utils to pick up this change
* Removed compatibility logic from the existing steps, since the dedicated compatibility steps should be used whenever compatibility with the old `AzExtTreeItems` is needed.

Associated Resource Groups branch: https://github.com/microsoft/vscode-azureresourcegroups/tree/alex/compat-shim-with-pick-app-resource

Azure Functions branch: https://github.com/microsoft/vscode-azurefunctions/tree/alex/api-v2-compat


utils/src/treev2/quickPickWizard/experiences/findByIdExperience.ts
```diff
@@ -21,7 +22,7 @@ export async function findByIdExperience<TPick extends types.FindableByIdTreeNod
     const wizardContext = context as types.QuickPickWizardContext<TPick>;
     wizardContext.pickedNodes = [];
 
-    const wizard = new types.AzureWizard(context, {
+    const wizard = new AzureWizard(context, {
```
alexweininger, line 25: explanation: can't import a class as a type and "new" it.

utils/src/treev2/quickPickWizard/compatibility/CompatibilityRecursiveQuickPickStep.ts
```diff
@@ -0,0 +1,120 @@
...
+            } else {
+                promptChoices = indirectChoices;
+            }
+        } else {
+            promptChoices = directChoices;
+        }
+
+        const picks: types.IAzureQuickPickItem<TNode | types.CreateCallback>[] = [];
+        for (const choice of promptChoices) {
+            picks.push(await this.getQuickPickItem(choice as TNode));
```
bwateratmsft, line 103: This function is nearly identical to `GenericQuickPickStep.getPicks`. Instead of making `getQuickPickItem` `protected`, and duplicating this logic, can you instead do `super.getPicks()` and then push the create pick? Same as what `QuickPickWithCreateStep.ts` does.

utils/index.d.ts
```diff
@@ -360,6 +360,7 @@ export interface IAzExtParentTreeItem extends IAzExtTreeItem {
  * NOTE: *Impl methods are not meant to be called directly - just implemented.
  */
 export declare abstract class AzExtTreeItem implements IAzExtTreeItem {
+    public readonly _isAzExtTreeItem = true;
```
bwateratmsft, line 363: I don't think this should be public. Instead we should declare and provide an `isAzExtTreeItem(maybeTreeItem: unknown): maybeTreeItem is AzExtTreeItem`.

utils/src/treev2/quickPickWizard/compatibility/CompatibilityContextValueQuickPickStep.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * Mimics how the legacy {@link AzExtParentTreeItem.pickChildTreeItem}
+     * uses {@link AzExtParentTreeItem.pickTreeItemImpl} to customize the tree item picker.
+     *
+     * An example customization is skipping having to pick a UI-only node (ex: App Settings parent node)
+     */
+    private async provideCompatabilityWithPickTreeItemImpl(wizardContext: TContext): Promise<boolean> {
+        const lastPickedItem = getLastNode(wizardContext);
+        const lastPickedItemUnwrapped = isWrapper(lastPickedItem) ? lastPickedItem.unwrap() : lastPickedItem
```
bwateratmsft, line 31: nit
```suggestion
        const lastPickedItemUnwrapped = isWrapper(lastPickedItem) ? lastPickedItem.unwrap() : lastPickedItem;
```

utils/index.d.ts
```diff
@@ -1729,14 +1737,31 @@ export declare interface ContextValueFilter {
...
-    }
+    readonly quickPickOptions: QuickPickOptions;
+}
+
+export declare interface CompatibleContextValueFilterableTreeNode {
+    readonly quickPickOptions: CompatibleQuickPickOptions;
 }
 
-export declare type ContextValueFilterableTreeNode = ContextValueFilterableTreeNodeV2 | AzExtTreeItem;
+export declare type ContextValueFilterableTreeNode = ContextValueFilterableTreeNodeV2;
```
bwateratmsft, line 1764: Seems like we may as well get rid of the `V2` copy of the interface and keep only `ContextValueFilterableTreeNode`.

bwateratmsft: Additionally that probably means changing a few things in `FindByIdQuickPickStep.ts`.

utils/index.d.ts
```diff
@@ -1703,10 +1704,17 @@ export declare interface Wrapper {
  */
 export declare function isWrapper(maybeWrapper: unknown): maybeWrapper is Wrapper;
 
-export declare function appResourceExperience<TPick extends ContextValueFilterableTreeNode>(context: IActionContext, tdp: TreeDataProvider<ResourceGroupsItem>, resourceType: AzExtResourceType, childItemFilter?: ContextValueFilter): Promise<TPick>;
+export declare function appResourceExperience<TPick extends ContextValueFilterableTreeNode>(context: IActionContext, tdp: TreeDataProvider<ResourceGroupsItem>, resourceTypes?: AzExtResourceType | AzExtResourceType[], childItemFilter?: ContextValueFilter): Promise<TPick>;
 export declare function contextValueExperience<TPick extends ContextValueFilterableTreeNode>(context: IActionContext, tdp: TreeDataProvider<ResourceGroupsItem>, contextValueFilter: ContextValueFilter): Promise<TPick>;
```
bwateratmsft, line 1708: A typo in my original code.
```suggestion
export declare function contextValueExperience<TPick extends ContextValueFilterableTreeNode>(context: IActionContext, tdp: TreeDataProvider<TPick>, contextValueFilter: ContextValueFilter): Promise<TPick>;
```

utils/index.d.ts
```diff
@@ -1703,10 +1704,17 @@ export declare interface Wrapper {
...
 export declare function findByIdExperience<TPick extends FindableByIdTreeNode>(context: IActionContext, tdp: TreeDataProvider<TPick>, id: string | Uri): Promise<TPick>;
 
+interface CompatibilityPickResourceExperienceOptions {
+    resourceTypes?: AzExtResourceType | AzExtResourceType[];
+    childItemFilter?: ContextValueFilter
+}
+
+export declare function compatibilityPickAppResourceExperience<TPick>(context: IActionContext, tdp: TreeDataProvider<ResourceGroupsItem>, options: CompatibilityPickResourceExperienceOptions): Promise<TPick>;
```
bwateratmsft, line 1716: Should `TPick` have an `extends AzExtTreeItem` constraint?

utils/index.d.ts
alexweininger: Yeah I left it to get feedback before removing it all.

To clarify, the nodes here should never be an `AzExtTreeItem` since they will be wrapped by a `CompatibleBranchDataItem` in Resource Groups.

utils/src/treev2/quickPickWizard/compatibility/CompatibilityRecursiveQuickPickStep.ts
alexweininger: I had to override it because we currently [throw a `NoResourceFoundError` when there are no picks](https://github.com/microsoft/vscode-azuretools/blob/alex/tip-changes3/utils/src/treev2/quickPickWizard/GenericQuickPickStep.ts#L79).

```ts
if (directChoices.length === 0) {
    if (indirectChoices.length === 0) {
        throw new NoResourceFoundError();
    } else {
        promptChoices = indirectChoices;
    }
} else {
    promptChoices = directChoices;
}
```

In the override method I changed it to

```ts
if (directChoices.length === 0) {
    if (indirectChoices.length === 0 && !this.pickOptions.create) {
        throw new NoResourceFoundError();
    } else {
        promptChoices = indirectChoices;
    }
} else {
    promptChoices = directChoices;
}
```

utils/src/treev2/quickPickWizard/experiences/appResourceExperience.ts
```diff
@@ -17,14 +17,14 @@ import { AzureWizard } from '../../../wizard/AzureWizard';
 import { AzureResourceQuickPickWizardContext, ResourceGroupsItem } from '../../../../hostapi.v2';
 import { isWrapper } from '../../../registerCommandWithTreeNodeUnwrapping';
 
-export async function appResourceExperience<TPick>(context: types.IActionContext, tdp: vscode.TreeDataProvider<ResourceGroupsItem>, resourceType: AzExtResourceType, childItemFilter?: types.ContextValueFilter): Promise<TPick> {
+export async function appResourceExperience<TPick>(context: types.IActionContext, tdp: vscode.TreeDataProvider<ResourceGroupsItem>, resourceTypes?: AzExtResourceType | AzExtResourceType[], childItemFilter?: types.ContextValueFilter): Promise<TPick> {
     const promptSteps: AzureWizardPromptStep<AzureResourceQuickPickWizardContext>[] = [
         new QuickPickAzureSubscriptionStep(tdp),
         new QuickPickGroupStep(tdp, {
-            groupType: resourceType,
+            groupType: resourceTypes ? Array.isArray(resourceTypes) ? resourceTypes : [resourceTypes] : undefined,
```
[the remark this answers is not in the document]
alexweininger: What you don't like nested ternaries? 😂

utils/index.d.ts
alexweininger: Yes, since v1 pickAppResource also has this constraint.

utils/src/treev2/quickPickWizard/compatibility/CompatibilityRecursiveQuickPickStep.ts
bwateratmsft: Could we catch and eat that `NoResourceFoundError` instead, if `this.pickOptions.create` is set? It'd be a bit messy but IMO better than duplicate (or near-duplicate) code.

alexweininger: Good idea

bwateratmsft: I think this goes without saying, but some comments to explain why it's catching and eating that error would be good too. I think the standard `QuickPickWithCreateStep` needs this same logic; however I don't feel too badly about duplicating a couple lines vs. a whole function.

utils/index.d.ts
alexweininger: Removed it from the declaration, and added `isAzExtTreeItem(maybeTreeItem: unknown): maybeTreeItem is AzExtTreeItem`.

utils/src/tree/AzExtTreeItem.ts
```diff
@@ -216,3 +215,7 @@ export abstract class AzExtTreeItem implements types.AzExtTreeItem {
         }
     }
 }
+
+export function isAzExtTreeItem(maybeTreeItem: unknown): maybeTreeItem is types.AzExtTreeItem {
+    return (maybeTreeItem as AzExtTreeItem)._isAzExtTreeItem === true;
```
bwateratmsft, line 220: Might wanna test if it's an object first--`5._isAzExtTreeItem` would throw (I think?). In practice I think it's highly unlikely someone would use it in such a way, but better safe than sorry.

bwateratmsft:
```suggestion
    return typeof maybeTreeItem === 'object' && (maybeTreeItem as AzExtTreeItem)._isAzExtTreeItem === true;
```

utils/src/treev2/quickPickWizard/compatibility/CompatibilityRecursiveQuickPickStep.ts
alexweininger: Ok, added the logic to `QuickPickWithCreateStep`. FYI I haven't tested/used `QuickPickWithCreateStep` yet, so there may need to be changes once I get to adding that to the experience builder API.

bwateratmsft (approved):

PR #1237, opened 2022-09-16
Require `ServiceClient` request attempts during warmup

MicroFish91:
For [#3337](https://github.com/microsoft/vscode-azurefunctions/issues/3337)

See comment in the file changed for additional info.

Note 1:  I mostly run into `BadGateway` errors, but occasionally `ServiceUnavailable` will still surface.

Note 2 :  From my testing, I've always gotten the information on attempt 2.  I personally felt that going up to 5 would take a really long time and I think the user experience would feel bad if we increased it more.

### DIFF
--- a/appservice/src/siteFiles.ts
+++ b/appservice/src/siteFiles.ts
@@ -66,9 +66,9 @@ async function getFsResponse(context: IActionContext, site: ParsedSite, filePath
             }
 
             /*
-                * Related to issue: https://github.com/microsoft/vscode-azurefunctions/issues/3337
+                Related to issue: https://github.com/microsoft/vscode-azurefunctions/issues/3337
                 Sometimes receive a 'BadGateway' or 'ServiceUnavailable' error on initial fetch, but consecutive re-fetching usually fixes the issue.
-                Under these circumstances, we will attempt to do the call 3 times during warmup before throwing the error
+                Under these circumstances, we will attempt to do the call 3 times during warmup before throwing the error.
             */
             const retries = 3;
             const badGateway: RegExp = /BadGateway/i;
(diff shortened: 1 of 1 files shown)


appservice/src/siteFiles.ts
```diff
@@ -64,11 +65,31 @@ async function getFsResponse(context: IActionContext, site: ParsedSite, filePath
...
-            return await client.sendRequest({
-                method: 'GET',
-                url: `${site.id}/hostruntime/admin/vfs/${filePath}/?api-version=2018-11-01`
-            });
+            /*
+                * Related to issue: https://github.com/microsoft/vscode-azurefunctions/issues/3337
+                Sometimes get a 'BadGateway' or 'ServiceUnavailable' error on initial fetch, but consecutive re-fetching usually fixes the issue.
+                Under these circumstances, we will attempt to do the call 3 times during warmup before throwing the error
+            */
+            let attempts: number = 1;
```
bwateratmsft, line 73: Do we not have a common retry class/function? If not we should make one.

MicroFish91: Oh actually I think I see where we've used the `p-retry` library, should I just re-implement it with that instead?

Example:
image

MicroFish91: Hmm, actually, I don't know if I like using the `p-retry` library's pattern here.  I'm currently retrying only on certain error instances, and then throwing the error else... 
using `p-retry`, I would have to wrap the error as a special `AbortError` to get it to cancel early... not sure if we want to mess with wrapping our errors like that just to get this to work.

I could explore creating our own custom utility version of this if needed, what do you think is the right move here?

appservice/src/siteFiles.ts
```diff
@@ -5,11 +5,11 @@
 
 import { HttpOperationResponse, RestError, ServiceClient } from '@azure/ms-rest-js';
 import { createGenericClient } from '@microsoft/vscode-azext-azureutils';
-import { IActionContext } from '@microsoft/vscode-azext-utils';
+import { IActionContext, IParsedError, parseError } from '@microsoft/vscode-azext-utils';
+import * as retry from 'p-retry';
 import * as path from 'path';
 import { createKuduClient } from './createKuduClient';
 import { ParsedSite } from './SiteClient';
-
```
bwateratmsft, line 12: Nit: this newline should stay

MicroFish91: Oops

bwateratmsft (approved):

PR #1245, opened 2022-10-12
Add compatibility pick subscription experience

alexweininger:
I'm purposely not exposing `subscriptionExperience` in the type definitions because I want to include this behavior in the experience builder later on.

### DIFF
--- a/utils/src/utils/credentialUtils.ts
+++ b/utils/src/utils/credentialUtils.ts
@@ -1,0 +1,47 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as vscode from 'vscode';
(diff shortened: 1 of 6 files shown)


utils/src/utils/credentialUtils.ts
```diff
@@ -0,0 +1,47 @@
...
+
+import * as vscode from 'vscode';
+import { ApplicationSubscription } from '../../hostapi.v2';
+import { AzExtServiceClientCredentials, ISubscriptionContext } from '../../index';
+import { localize } from '../localize';
+
+/**
+ * Converts a VS Code authentication session to an Azure Track 1 & 2 compatible compatible credential.
+ */
+export function createCredential(getSession: (scopes?: string[]) => vscode.ProviderResult<vscode.AuthenticationSession>): AzExtServiceClientCredentials {
```
bwateratmsft, line 14: We should note that this is more-or-less copied from Resource Groups. Is there an opportunity for code sharing or is this going to be too temporary to justify it?

bwateratmsft (approved):

PR #1253, opened 2022-10-18
Add `VSCodeTokenCredential`

bwateratmsft:


alexweininger (commented):
Do you need to export the types in `index.d.ts` too?

bwateratmsft:
> Do you need to export the types in `index.d.ts` too?

Yup. I always forget that's not generated here.

alexweininger (approved):

PR #1255, opened 2022-10-20
Take all T1 packages (and indeed, all Azure packages) out of dev package

bwateratmsft:


nturinski (commented):
Where should we migrate our Azure testing stuff, if at all?

bwateratmsft:
The `@azure/identity` package contains `EnvironmentCredential`, which will wholly replace all the `*Credential` stuff. As for subscription listing and that, @philliphoff and I are still discussing. In any case, that subscription listing bit will not be in the dev package--it doesn't belong there--it'll be either in its own library (probably) or in the RG extension.

bwateratmsft:
Closing, this is more or less done

PR #1257, opened 2022-10-25
Don't use `findTreeItem` in `AzExtTreeFileSystem`

alexweininger:
We haven't gotten `findTreeItem` to work for v2 yet since tree item IDs haven't been ironed out. However, I'm proposing that we move away from using `findTreeItem` in `AzExtTreeFileSystem`.  `AzExtTreeFileSystem` is one of the main places where `findTreeItem` is used.

`vscode.FileSystemProvider` relies on handling `Uri`s within VS Code. Within the file system provider methods we need access to the tree item in order to read/write the file content. Currently, we're adding the tree item ID to the `Uri` query params, and then using `findTreeItem` to retrieve the tree item within the handler.

Although that works great, I found it sorta wasteful to use `findTreeItem` to retrieve a reference to an object that we already have when forming the `Uri`. Instead, I just store the tree item in a `Map` with a random key and pass the key in the `Uri`, eliminating the need for `findTreeItem`. This method is probably a bit faster too.

Drawbacks of this method include:
* File system provider will not be able to handle URIs that have originated from this method and have tree items in the Map. This is fine since we currently don't support this anyway.

### DIFF
--- a/utils/src/AzExtTreeFileSystem.ts
+++ b/utils/src/AzExtTreeFileSystem.ts
@@ -3,29 +3,25 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import { randomUUID } from "crypto";
 import { parse as parseQuery, ParsedUrlQuery, stringify as stringifyQuery } from "querystring";
 import { Disposable, Event, EventEmitter, FileChangeEvent, FileStat, FileSystemError, FileSystemProvider, FileType, TextDocumentShowOptions, Uri, window } from "vscode";
 import * as types from '../index';
 import { callWithTelemetryAndErrorHandling } from "./callWithTelemetryAndErrorHandling";
 import { localize } from "./localize";
-import { AzExtTreeDataProvider } from "./tree/AzExtTreeDataProvider";
-import { AzExtTreeItem } from "./tree/AzExtTreeItem";
 import { nonNullProp } from "./utils/nonNull";
 
 const unsupportedError: Error = new Error(localize('notSupported', 'This operation is not supported.'));
 
-export abstract class AzExtTreeFileSystem<TItem extends AzExtTreeItem> implements FileSystemProvider {
+export abstract class AzExtTreeFileSystem<TItem> implements FileSystemProvider {
+
+    private readonly itemCache: Map<string, TItem> = new Map<string, TItem>();
+
     public abstract scheme: string;
 
     private readonly _emitter: EventEmitter<FileChangeEvent[]> = new EventEmitter<FileChangeEvent[]>();
     private readonly _bufferedEvents: FileChangeEvent[] = [];
     private _fireSoonHandle?: NodeJS.Timer;
-
-    private _tree: AzExtTreeDataProvider;
-
-    public constructor(tree: AzExtTreeDataProvider) {
-        this._tree = tree;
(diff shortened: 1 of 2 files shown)


bwateratmsft (approved):

utils/src/AzExtTreeFileSystem.ts
```diff
@@ -70,7 +69,7 @@ export abstract class AzExtTreeFileSystem<TItem extends AzExtTreeItem> implement
         await callWithTelemetryAndErrorHandling('writeFile', async (context) => {
             const item: TItem = await this.lookup(context, uri);
             await this.writeFileImpl(context, item, content, uri);
-            await item.refresh(context);
+            // await item.refresh(context);
```
alexweininger, line 72: I will have to do more investigation on what this was doing and if we need to replicate this behavior still.

utils/src/AzExtTreeFileSystem.ts
```diff
@@ -8,25 +8,20 @@ import { Disposable, Event, EventEmitter, FileChangeEvent, FileStat, FileSystemE
...
-import { AzExtTreeDataProvider } from "./tree/AzExtTreeDataProvider";
-import { AzExtTreeItem } from "./tree/AzExtTreeItem";
 import { nonNullProp } from "./utils/nonNull";
 
 const unsupportedError: Error = new Error(localize('notSupported', 'This operation is not supported.'));
 
-export abstract class AzExtTreeFileSystem<TItem extends AzExtTreeItem> implements FileSystemProvider {
+export abstract class AzExtTreeFileSystem<TItem extends types.AzExtTreeFileSystemItem> implements FileSystemProvider {
+
+    private readonly itemCache: Map<string, TItem> = new Map<string, TItem>();
```
nturinski, line 17: Should we clear and repopulate this every time a refresh happens? Seems like it could cause stale tree item ids if we're not careful.

nturinski:
Oh, this is _only_ for the FileSystem? So `findTreeItem` will still be used in normal `TreeDataProviders`?

alexweininger:
> Oh, this is _only_ for the FileSystem? So `findTreeItem` will still be used in normal `TreeDataProviders`?

Yes this PR is only removing the usage of `findTreeItem` (and anything AzExtTree* related for that matter) from the file system. Side note: maybe we should rename this class...

However, `findTreeItem` is still used in other places in the v1.5 client extensions. But we are **not** exposing a find API in v2. At least not yet. There's a chance we absolutely must add it when upgrading v1.5 extensions to v2. However we want to avoid this as much as we can because the whole find tree item thing is a bit problematic.

nturinski (approved):

utils/src/AzExtTreeFileSystem.ts
alexweininger: I haven't run into that in my testing, but I am worried about the possibility of something like that happening. The tree item ids correspond to Azure resource ids so unless something is deleted I think it's fine. I will test that now.

alexweininger: Also whenever I tried clearing this map, it resulted in errors since we can't really know if the items in the map are still in use easily.

alexweininger: A potential fix for all of this would be to really remove the reliance on items at all, and base if fully on URIs.

PR #1270, opened 2022-11-16
Fix `findByIdExperience` for children of `GroupingItem`s

alexweininger:
`GroupingItem`s are special and we can't determine if an item is a descendant of a `GroupingItem` based solely on the id. Instead, we have to check `GroupingItem.resources` to see if any of the resources match the id we are searching for.

I was hesitant about this since it's making a lot of assumptions about the underlying tree data provider, but out of all the alternatives, this may be the simplest.

### DIFF
--- a/utils/src/treev2/quickPickWizard/FindByIdQuickPickStep.ts
+++ b/utils/src/treev2/quickPickWizard/FindByIdQuickPickStep.ts
@@ -8,6 +8,7 @@ import * as vscode from 'vscode';
 import { getLastNode } from './QuickPickWizardContext';
 import { GenericQuickPickStep, SkipIfOneQuickPickOptions } from './GenericQuickPickStep';
 import { PickFilter } from './common/PickFilter';
+import { ApplicationResource } from '../../../hostapi.v2';
 
 interface FindByIdQuickPickOptions extends SkipIfOneQuickPickOptions {
     id: string;
@@ -21,7 +22,7 @@ export class FindByIdQuickPickStep<TContext extends types.QuickPickWizardContext
         });
     }
 
-    protected readonly pickFilter: PickFilter = new FindByIdPickFilter(this.pickOptions);
+    protected readonly pickFilter = new FindByIdPickFilter(this.pickOptions);
 
     public async getSubWizard(wizardContext: TContext): Promise<types.IWizardOptions<TContext> | undefined> {
         // TODO: this code is nearly identical to `RecursiveQuickPickStep`, but this class can't inherit from it because it's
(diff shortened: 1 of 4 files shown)


utils/src/treev2/quickPickWizard/FindByIdQuickPickStep.ts
```diff
@@ -52,12 +53,16 @@ export class FindByIdQuickPickStep<TContext extends types.QuickPickWizardContext
...
 
-    isAncestorPick(node: vscode.TreeItem): boolean {
+    isAncestorPick(node: vscode.TreeItem, item?: { resources?: ApplicationResource[] }): boolean {
         if (!node.collapsibleState) {
             // can't be an indirect pick if it doesn't have children
             return false;
         }
 
+        if (item?.resources) {
+            return item.resources.some((resource) => this.pickOptions.id.startsWith(resource.id));
```
alexweininger, line 63: We probably want to guard this better. Check if `item.resources` is an array first. And maybe rename the `resources` property to something more obscure so we don't mistakenly do this check on not-`GroupingItem`s.

I considered making `find` a RGs internal method because then we could do it in a safer way. We could do `instanceof GroupingItem`.

bwateratmsft:
Do we still plan to move the `findByIdExperience` into the RG extension as special code for the V1.5 extensions to consume? If so I don't feel bad about making assumptions about the structure of the tree.

alexweininger:
> Do we still plan to move the `findByIdExperience` into the RG extension as special code for the V1.5 extensions to consume? If so I don't feel bad about making assumptions about the structure of the tree.

Oh in that case, I have a better proposal. I wasn't aware we planned on it being internal and only for V1.5 extensions.

[Please see how I've implemented find here.](https://github.com/microsoft/vscode-azureresourcegroups/compare/api-v2...alex/v2-findItemById) We can remove the `findByIdExperience` and only use the internal find for V1.5 compatibility.

bwateratmsft:
I saw you'd left a comment that only the V1.5 extensions needed it and you were thinking to move it to RGs, but now I can't find that comment.

alexweininger:
> I saw you'd left a comment that only the V1.5 extensions needed it and you were thinking to move it to RGs, but now I can't find that comment.

I deleted it because I thought it was a bad idea 😞. I was thinking that the V2 extensions will need find, but I think we should try to implement the V2 clients without find, and if we **really** do need it, then we can expose it for V2 extensions later.

alexweininger:
Ok, I proposed some changes that can replace these changes and `findByIdExperience` in https://github.com/microsoft/vscode-azureresourcegroups/pull/431.

PR #1278, opened 2022-11-28
Fix an issue where quick pick items were losing selection focus

### LINKED ISSUE #1276: Quick pick prompting loses the selection focus
When we go into quick pick prompting mode, we now have to tab to get back into selection mode:

image

I tested this with `a11y-kit` and it seems like it has no issue (meaning I couldn't repro).  I'm seeing this behavior across multiple Azure extensions, so I'm assuming that we will need to investigate a fix for our `showQuickPick` method.

MicroFish91:
Fixes #1276 

Seems like now we need to be calling `show` only after we have the `items` populated or we lose selection focus.

### DIFF
--- a/utils/src/userInput/showQuickPick.ts
+++ b/utils/src/userInput/showQuickPick.ts
@@ -62,7 +62,6 @@ export async function showQuickPick<TPick extends types.IAzureQuickPickItem<unkn
             // Show progress bar while loading quick picks
             quickPick.busy = true;
             quickPick.enabled = false;
-            quickPick.show();
             try {
                 quickPick.items = await createQuickPickItems<TPick>(picks, options, groups, recentlyUsedKey);
 
@@ -81,6 +80,7 @@ export async function showQuickPick<TPick extends types.IAzureQuickPickItem<unkn
             } catch (err) {
                 reject(err);
             }
+            quickPick.show();
         });
 
         if (recentlyUsedKey && !Array.isArray(result) && !result.suppressPersistence) {


utils/src/userInput/showQuickPick.ts
```diff
@@ -62,7 +62,6 @@ export async function showQuickPick<TPick extends types.IAzureQuickPickItem<unkn
             // Show progress bar while loading quick picks
             quickPick.busy = true;
             quickPick.enabled = false;
-            quickPick.show();
```
alexweininger, line 65: We were calling `show()` here so that while quick pick items are being created an empty quick pick with a progress bar is shown. We may need to call `show()` in both places.

MicroFish91: That solution doesn't seem to work unfortunately, I'll mess with it some more and see if there's another way we can handle this

alexweininger: Seems like the issue is that if you call `show()` on a quick pick with no items, then items are added, focus is lost.

MicroFish91: I am also able to repro by just doing this:

image

alexweininger: Nice 👍

MicroFish91:
Filing an issue with VS Code instead

alexweininger:
https://github.com/microsoft/vscode/issues/167478

PR #1280, opened 2022-11-30
Add `isAzExtTreeItem` util

alexweininger:


utils/src/tree/isAzExtParentTreeItem.ts
```diff
@@ -0,0 +1,19 @@
...
+
+import type { AzExtParentTreeItem } from "./AzExtParentTreeItem";
+import type { AzExtTreeItem } from "./AzExtTreeItem";
+import type { IAzExtParentTreeItemInternal } from "./InternalInterfaces";
+
+/**
+ * Using instanceof AzExtParentTreeItem causes issues since each extension has their own version of the utils. Instead, check _isAzExtParentTreeItem
+ */
+export function isAzExtParentTreeItem(maybeParentTreeItem: unknown): maybeParentTreeItem is AzExtParentTreeItem {
+    return typeof maybeParentTreeItem === 'object' && (maybeParentTreeItem as IAzExtParentTreeItemInternal)._isAzExtParentTreeItem === true;
```
bwateratmsft, line 14: An alternative that's about the same length:
```suggestion
    return isAzExtTreeItem(maybeParentTreeItem) && (maybeParentTreeItem as IAzExtParentTreeItemInternal)._isAzExtParentTreeItem === true;
```
Not critical one way or the other but I like taking advantage of the other `isBlah` methods when dealing with inherited classes.

utils/src/tree/isAzExtParentTreeItem.ts
```diff
@@ -0,0 +1,19 @@
...
+*--------------------------------------------------------------------------------------------*/
+
+import type { AzExtParentTreeItem } from "./AzExtParentTreeItem";
+import type { AzExtTreeItem } from "./AzExtTreeItem";
+import type { IAzExtParentTreeItemInternal } from "./InternalInterfaces";
+
+/**
+ * Using instanceof AzExtParentTreeItem causes issues since each extension has their own version of the utils. Instead, check _isAzExtParentTreeItem
+ */
+export function isAzExtParentTreeItem(maybeParentTreeItem: unknown): maybeParentTreeItem is AzExtParentTreeItem {
```
bwateratmsft, line 13: I'd actually prefer renaming the file to `isAzExtTreeItem` and flipping the order around, since `AzExtTreeItem` is the base class, and `AzExtParentTreeItem` is the inheriting class.

bwateratmsft:
IMO we should just merge this straight into `main`.

utils/src/tree/isAzExtParentTreeItem.ts
alexweininger: Good suggestion, thanks

bwateratmsft (approved):

PR #1292, opened 2022-12-06
Organize api interface

alexweininger:
Open to feedback here. I separated the interfaces of the activity related methods and the host methods. I also scoped the properties on the API.


alexweininger:
I'm going to make a release with the current api definitions so we can move forward with the other PRs. We can make a new release later if we want to change them.

PR #1294, opened 2022-12-06
Add v2 api types to main

alexweininger:
Needed in order to release. Maybe we should stop using the api-v2 branch and just use main in this repo.


bwateratmsft (approved):

bwateratmsft:
> Maybe we should stop using the api-v2 branch and just use main in this repo.

Once the quick pick changes get merged into api-v2, let's do that.

PR #1307, opened 2022-12-14
Edit createApiProvider to support `GetApiOptions`

alexweininger:
Now when creating an API provider, you must pass in an array of API factories instead of API objects. That way the provider can "stamp" new APIs with options passed in from the consumer.

For now the only option is the consumer extension id, which can be attached to API telemetry to track which extensions are using the API and how.


utils/src/createApiProvider.ts
```diff
@@ -31,20 +30,22 @@ class ApiVersionError extends Error {
...
         context.telemetry.properties.apiVersionRange = apiVersionRange;
+        context.telemetry.properties.callingExtensionId = options?.extensionId;
 
         const apiVersions: string[] = azExts.map((a: AzureExtensionApi) => a.apiVersion);
         context.telemetry.properties.apiVersions = apiVersions.join(', ');
 
         const matchedApiVersion: string | null = semver.maxSatisfying(apiVersions, apiVersionRange);
         if (matchedApiVersion) {
-            return <T>(azExts.find((a: AzureExtensionApi) => a.apiVersion === matchedApiVersion));
+            const apiFactory = azExts.find(a => a.apiVersion === matchedApiVersion);
```
bwateratmsft, line 47: We should use semver. I'd worry about exact matching versions.

utils/index.d.ts
```diff
@@ -1685,8 +1690,15 @@ export declare enum AzExtResourceType {
...
 /**
- * Describes command callbacks for tree node context menu commands
+ * Gets the exported API from the given extension id and version range.
+ * Automatically populates the consumer extension id to the created API.
+ *
+ * @param extensionId The extension id to get the API from
+ * @param apiVersionRange The version range of the API you need. Any semver syntax is allowed. For example "1" will return any "1.x.x" version or "1.2" will return any "1.2.x" version
+ * @throws Error if extension with id is not installed.
  */
+export function getExtensionApi<T extends AzureExtensionApi>(extensionId: string, apiVersionRange: string): Promise<T>;
```
bwateratmsft, line 1700: Is this meant to be this?
```suggestion
export declare function getAzureExtensionApi<T extends AzureExtensionApi>(extensionId: string, apiVersionRange: string): Promise<T>;
```

utils/src/utils/apiUtils.ts
```diff
@@ -1,5 +1,30 @@
...
+import { AzureExtensionApi, AzureExtensionApiProvider } from "../../api";
+import { getPackageInfo } from "../getPackageInfo";
+
+class ExtensionNotFoundError extends Error {
+    constructor(extensionId: string) {
+        super(`Extension with id ${extensionId} not found.`);
+    }
+}
+
+export async function getAzureExtensionApi<T extends AzureExtensionApi>(extensionId: string, apiVersionRange: string): Promise<T> {
```
bwateratmsft, line 17: If there are other things added to the `GetApiOptions` aside from extension id, it's entirely possible the caller would need to be able to specify them. I think we should go ahead and accept it, setting `extensionId` by default if it isn't present.

alexweininger: Good idea

utils/index.d.ts
alexweininger: Does adding the `declare` inside a `.d.ts` file make a difference?

bwateratmsft: I don't think it makes a significant difference, but `declare` is more "correct".

utils/src/utils/apiUtils.ts
```diff
@@ -14,12 +14,12 @@ class ExtensionNotFoundError extends Error {
...
 }
 
-export async function getAzureExtensionApi<T extends AzureExtensionApi>(extensionId: string, apiVersionRange: string): Promise<T> {
+export async function getAzureExtensionApi<T extends AzureExtensionApi>(extensionId: string, apiVersionRange: string, options?: GetApiOptions): Promise<T> {
     const apiProvider: AzureExtensionApiProvider | undefined = await getExtensionExports(extensionId);
 
     if (apiProvider) {
         return apiProvider.getApi<T>(apiVersionRange, {
-            extensionId: getPackageInfo().extensionId
+            extensionId: options?.extensionId ?? getPackageInfo().extensionId
```
bwateratmsft, line 22: Don't forget the options mixin
```typescript
{
    ...options,
    extensionId: options?.extensionId || getPackageInfo().extensionId
}
```

utils/index.d.ts
```diff
@@ -1351,11 +1351,16 @@ export interface IAddUserAgent {
...
+    apiVersion: string,
+    createApi: (options?: GetApiOptions) => T
+};
+
 /**
  * Wraps an Azure Extension's API in a very basic provider that adds versioning.
  * Multiple APIs with different versions can be supplied, but ideally a single backwards-compatible API is all that's necessary.
  */
-export declare function createApiProvider(azExts: AzureExtensionApi[]): AzureExtensionApiProvider;
+export declare function createApiProvider(azExts: AzureExtensionApiFactory[]): AzureExtensionApiProvider;
```
bwateratmsft, line 1363: Is this a breaking change, i.e. would require bumping the major version?

bwateratmsft (dismissed):

utils/index.d.ts
alexweininger: Yes it is. I considered trying to make it backwards compatible by accepting `(AzureExtensionApi | AzureExtensionApiFactory)[]`, but decided against it.

A lot of the extensions use `createApiProvider`, but only to expose an API with reveal. Maybe we should leave this `createApiProvider` function alone, and add a new method that we can use going forward.

The breaking change is pretty easy to adapt when needed, however I think it impacts all extensions.

alexweininger: Ok I went ahead and made it backwards compatible. This will make our lives easier for a while until we can convert all APIs over to the factory pattern.

bwateratmsft (approved):

PR #1312, opened 2022-12-19
Add `ConfirmPreviousInputStep`

MicroFish91:
For validating your previous input (e.g. if you need to revalidate a password during account creation)


utils/src/wizard/InputRevalidationStep.ts
```diff
@@ -0,0 +1,30 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { AzureWizardPromptStep, IActionContext } from "../..";
+import { localize } from "../localize";
+
+export class InputRevalidationStep<T extends IActionContext> extends AzureWizardPromptStep<T> {
```
alexweininger, line 9: This wasn't what I was expecting a step named `InputRevalidationStep` to do. Maybe `ConfirmPreviousInputStep` (or similar) would be a better name?

utils/src/wizard/InputRevalidationStep.ts
```diff
@@ -0,0 +1,30 @@
...
+    constructor(private readonly _key: string, private readonly _isPassword?: boolean) {
+        super();
+    }
+
+    public async prompt(context: T): Promise<void> {
+        await context.ui.showInputBox({
+            prompt: localize('verifyPreviousInput', 'Please confirm by re-entering the previous value.'),
```
alexweininger, line 16: Should the prompt value be customizable? I could see us wanting to make it say "Re-enter password to confirm" or something.

utils/src/wizard/InputRevalidationStep.ts
```diff
@@ -0,0 +1,30 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { AzureWizardPromptStep, IActionContext } from "../..";
+import { localize } from "../localize";
+
+export class InputRevalidationStep<T extends IActionContext> extends AzureWizardPromptStep<T> {
+    constructor(private readonly _key: string, private readonly _isPassword?: boolean) {
```
alexweininger, line 10: nit: remove underscores

MicroFish91: I thought we were using underscores to represent private properties but excluding them on private methods.  What exactly is the standard convention again so I don't keep messing this up 😅

alexweininger: I thought it was for private properties that are exposed via a public getter.

Like if the property represents an internal value for something that is also exposed publicly, then add an underscore to avoid a naming conflict.

utils/src/wizard/ConfirmPreviousInputStep.ts
```diff
@@ -0,0 +1,31 @@
...
+import { AzureWizardPromptStep } from "./AzureWizardPromptStep";
+
+export class ConfirmPreviousInputStep extends AzureWizardPromptStep<types.IActionContext> {
+    constructor(private readonly key: string, private readonly options?: types.IConfirmInputOptions) {
+        super();
+    }
+
+    public async prompt(context: types.IActionContext): Promise<void> {
+        await context.ui.showInputBox({
+            prompt: this.options?.prompt || localize('verifyPreviousInput', 'Please confirm by re-entering the previous value.'),
```
alexweininger, line 17:
```suggestion
            prompt: this.options?.prompt ?? localize('verifyPreviousInput', 'Please confirm by re-entering the previous value.'),
```

utils/src/wizard/ConfirmPreviousInputStep.ts
```diff
@@ -0,0 +1,31 @@
...
+            password: this.options?.isPassword,
+            validateInput: (value: string | undefined) => this.validateInput(context, value)
```
alexweininger, line 19:
```suggestion
            validateInput: (value?: string) => this.validateInput(context, value)
```

utils/src/wizard/ConfirmPreviousInputStep.ts
```diff
@@ -0,0 +1,31 @@
...
+        });
+    }
+
+    public shouldPrompt(): boolean {
+        return true;
+    }
+
+    private validateInput(context: types.IActionContext, value: string | undefined): string | undefined {
```
alexweininger, line 27:
```suggestion
    private validateInput(context: types.IActionContext, value?: string): string | undefined {
```

alexweininger (dismissed):

alexweininger:
Bad merge broke the build, should be easy fix. It can't find the file since it was renamed.

MicroFish91:
Thanks for catching that!  Import sorting + the following merge conflict was not fun to look at 😅

alexweininger (approved):

PR #1313, opened 2022-12-19
Add a conditional `setLocation` in `ResourceGroupListStep` 

### LINKED ISSUE #1290: `ResourceGroupListStep` should call `LocationListStep.setLocation()` after a resource group is selected
In `ResourceGroupListStep`....

A `location` is set whenever a new resource group is created:

image

but not when it is selected from an existing list:

image

I'm thinking we should probably be consistent and call `setLocation` after the user has made a choice, in a similar way to how it's done in `StorageAccountListStep`.

Let me know if anyone can think of a reason that we shouldn't do this.

MicroFish91:
Closes #1290 

Similar to how it's done in `StorageAccountListStep`.

### DIFF
--- a/azure/src/wizard/ResourceGroupListStep.ts
+++ b/azure/src/wizard/ResourceGroupListStep.ts
@@ -45,7 +45,7 @@ export class ResourceGroupListStep<T extends types.IResourceGroupWizardContext> 
         // Cache resource group separately per subscription
         const options: IAzureQuickPickOptions = { placeHolder: 'Select a resource group for new resources.', id: `ResourceGroupListStep/${wizardContext.subscriptionId}` };
         wizardContext.resourceGroup = (await wizardContext.ui.showQuickPick(this.getQuickPicks(wizardContext), options)).data;
-        if (wizardContext.resourceGroup) {
+        if (wizardContext.resourceGroup && !LocationListStep.hasLocation(wizardContext)) {
             await LocationListStep.setLocation(wizardContext, nonNullProp(wizardContext.resourceGroup, 'location'));
         }
     }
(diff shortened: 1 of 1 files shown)


azure/src/wizard/ResourceGroupListStep.ts
```diff
@@ -46,6 +45,9 @@ export class ResourceGroupListStep<T extends types.IResourceGroupWizardContext>
         // Cache resource group separately per subscription
         const options: IAzureQuickPickOptions = { placeHolder: 'Select a resource group for new resources.', id: `ResourceGroupListStep/${wizardContext.subscriptionId}` };
         wizardContext.resourceGroup = (await wizardContext.ui.showQuickPick(this.getQuickPicks(wizardContext), options)).data;
+        if (wizardContext.resourceGroup) {
+            await LocationListStep.setLocation(wizardContext, nonNullProp(wizardContext.resourceGroup, 'location'));
```
MicroFish91, line 49: Discussion question...

Should we always set location whenever we pick a resource group, or should it only be set if `_location` is undefined?  Whenever we go through the flow to create a resource group, we automatically select and set the location, so I'm kind of leaning towards `option A` but could see both ways being feasible but having different side effects.

A)
``` javascript
if (wizardContext.resourceGroup) {
  ...
}
``` 
OR

B)
```javascript
if (wizardContext.resourceGroup && !LocationListStep.hasLocation(wizardContext)) {
  ...
}
```

nturinski: I think I prefer option B a little bit better. If we set the location ahead of time for whatever reason, I'd prefer to respect those settings.

nturinski (approved):

PR #1314, opened 2022-12-20
Add option to not unwrap result of appResourceExperience

alexweininger:
I need to use appResourceExperience in resource groups, but I don't want the result to be unwrapped. This option is intended to only be used by the resources extension.

further explanation:
In the host, we wrap any items provided by the client extensions in host-internal classes and when handing items back to client extensions we have to unwrap them (else we'll be giving clients different objects than they expect). 

But when doing things in the host, I want to work with the host-internal class, not the client-class. Thus I need to leave it wrapped.

Basically there was no way to pick an Azure resource and keep the result as a host-internal class. It only returned the underlying client class. But since edit tags is a resources command, I want to work with the internal class.

### DIFF
--- a/utils/index.d.ts
+++ b/utils/index.d.ts
@@ -1740,7 +1740,14 @@ type ResourceGroupsItem = unknown;
  */
 export declare function isWrapper(maybeWrapper: unknown): maybeWrapper is Wrapper;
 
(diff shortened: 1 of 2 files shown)


nturinski (approved):
Discussed offline the trade off between exposing dontUnwrap via the context and alternatives.

PR #1330, opened 2023-01-03
Add `asyncValidationTask` to input box options and verify during `onDidAccept`

### LINKED ISSUE #1275: Revisit `showInputBox` final validation logic to ensure no false positives
Resume discussion of [this](https://github.com/microsoft/vscode-azurefunctions/pull/3390#discussion_r1018328292) here.

In the process of writing generic debounce logic for `showInputBox` validation, I'm running into false positives being thrown for `validateInput` when processes take too long.  You can verify this by simply going into any `validateInput` and setting a long delay like 2000 ms.  `validateInput` will start to resolve to `undefined` even if the true return is supposed to be a string.

The current validation logic we have in `onDidAccept` seems to yield the same false positive when it reruns on final validation:

image

MicroFish91:
Closes #1275 

This PR is based on/inspired by our discussions during stand-up right before New Years...

I added a new optional method to `showInputBox` called `asyncValidationTask`.

Let me know what you guys like/dislike and I can make any changes as necessary.  Thanks for reviewing and hope you all had a great holiday break :)


utils/src/userInput/showInputBox.ts
```diff
@@ -18,29 +18,40 @@ export async function showInputBox(context: IInternalActionContext, options: typ
...
+                        validateInputTask = Promise.resolve(options.validateInput(text));
+                        latestValidateInputTask = validateInputTask;
+                        const message: string | undefined | null = await validateInputTask;
+                        if (validateInputTask === latestValidateInputTask) {
                             inputBox.validationMessage = message || '';
                         }
                     }
+                    if (options.asyncValidationTask && !inputBox.validationMessage && validateInputTask === latestValidateInputTask) {
+                        // Start loading the promise, but wrap with a short debounce to avoid performing server calls too frequently
+                        latestAsyncValidationTask = inputBoxDebounce(options.asyncValidationTask, text);
```
nturinski, line 37: I may be wrong, but I thought we were okay to only call the async validation task `onDidAccept`. That way we won't have to worry about multiple API calls being made repeatedly.

The example was whenever you register a new account, they don't check if the username is available until you hit submit, then it will come back and say it's unavailable.

@alexweininger Thoughts?

MicroFish91: Yeah actually I think I'll change it back to that.  I ended up changing it to the internal debounce because I thought it might have some performance benefits, but now that I'm thinking about it, you're always going to be introducing a delay on the last keystroke now which isn't ideal.  I'll just remove the async queue and do it all during `onDidAccept`. 🤔

nturinski (approved):
So much cleaner too 😊

nturinski:
Don't forget to bump the version. It's not really a breaking change, but probably still deserves a minor update since it changes the way we validate input a decent amount.

nturinski (approved):

alexweininger (approved):

PR #1332, opened 2023-01-04
Add compat helper for showTreeItemPicker

alexweininger:
There are a few places where we still use `AzExtParentTreeItem.showTreeItemPicker` in the v1.5 client extensions. This is a helper used by the v2 Resources API to provide compatibility with `showTreeItemPicker`.

See usage in Resources: https://github.com/microsoft/vscode-azureresourcegroups/pull/478

### DIFF
--- a/utils/src/pickTreeItem/experiences/compatibility/PickTreeItemWithCompatibility.ts
+++ b/utils/src/pickTreeItem/experiences/compatibility/PickTreeItemWithCompatibility.ts
@@ -6,6 +6,13 @@ import { subscriptionExperience } from '../subscriptionExperience';
 import { isAzExtTreeItem } from '../../../tree/isAzExtTreeItem';
 import { createSubscriptionContext } from '../../../utils/credentialUtils';
 import { ISubscriptionContext } from '@microsoft/vscode-azext-dev';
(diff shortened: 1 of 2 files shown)


nturinski (approved):

utils/src/pickTreeItem/experiences/compatibility/PickTreeItemWithCompatibility.ts
```diff
@@ -28,4 +35,37 @@ export namespace PickTreeItemWithCompatibility {
...
+            new CompatibilityRecursiveQuickPickStep(tdp, {
+                contextValueFilter: {
+                    include: expectedContextValues,
+                },
+                skipIfOne: false,
+            }),
+        ];
+
+        // Fill in the `pickedNodes` property
+        const wizardContext = { ...context } as types.QuickPickWizardContext;
```
nturinski, line 53: Should this code not be moved into its own experience?

alexweininger: Though initially I had the same thought as you, there is no v2 equivalent of showTreeItemPicker so there doesn't need to be an experience for it. The other functions in here are wrapping experiences with some compatibility stuff. I think it's fine to keep the code here.

PR #1333, opened 2023-01-05
Add `revealAzureResource` to v2 api interface

alexweininger:


utils/hostapi.v2.d.ts
```diff
@@ -275,6 +275,8 @@ export type WorkspaceResourceBranchDataProvider<TModel extends WorkspaceResource
 // scope down vscode.TreeDataProvider to exactly what's allowed to be used
 type ResourceGroupsTreeDataProvider = Pick<vscode.TreeDataProvider<unknown>, 'getChildren' | 'getTreeItem'>;
 
+export type VSCodeRevealOptions = Parameters<vscode.TreeView<unknown>['reveal']>['1'];
```
bwateratmsft, line 278: TIL! This is pretty cool. Helpful for these cases where there isn't an interface defined.

Does this need to be exported though?

bwateratmsft (approved):

utils/hostapi.v2.d.ts
alexweininger: I'm using it [here](https://github.com/microsoft/vscode-azureresourcegroups/pull/483/files#diff-2305203227d9502195174e3a2af2439170cad801d91e2249f557e3eff58f04feR47).

PR #1346, opened 2023-01-12
Add `configureBeforePrompt` to `AzureWizardPromptStep`

### LINKED ISSUE #1279: Add new function to `AzureWizardPromptStep` to configure context
Today we have a lot of wizard steps using `prompt` or `shouldPrompt` to configure things on the context without actually prompting users. This is good since we can make smart choices based on previous context values, however it's a bit awkward to use either of these functions.

`prompt` only gets called once, but because `shouldPrompt` has to be `true` for this behavior, it messes with the step count. It can also be a hard bug to discern because a context value might be set to skip the prompt, but we actually need to configure a couple more properties.

`shouldPrompt` is constantly getting called throughout the wizard, so it's not ideal to put it in here either. Getting a context value set repeatedly could have some negative impact and may also cause bugs where we inadvertently overwrite some properties.

We should have a `configure` or `configureBeforePrompt` function that handles this scenario. See this thread for more details: https://github.com/microsoft/vscode-azurefunctions/pull/3390/files#r1009498719

MicroFish91:
Closes #1279 

I think the current `goBack` functionality is pretty good at preserving values that originally started as null or undefined.  For the more rare/complex/exotic context changes, I'm thinking the optional `undo` method can be used to cover any remaining gaps/edge-cases (all already implemented).

I tried using this in a couple different scenarios so far, and it seems to work fine; however, if you can think of any interesting configurations to try (especially involving the `goBack` functionality) definitely do so and let me know if it leads to any additional considerations I may have missed! 🙂

### DIFF
--- a/utils/index.d.ts
+++ b/utils/index.d.ts
@@ -1209,6 +1209,12 @@ export declare abstract class AzureWizardPromptStep<T extends IActionContext> {
     public getSubWizard?(wizardContext: T): Promise<IWizardOptions<T> | undefined>;
 
     /**
+     * Can be used to optionally configure the wizard context before determining if prompting is required
+     * This method will be called before `shouldPrompt`
+     */
+    public configureBeforePrompt?(wizardContext: T): Promise<void>;
+
+    /**
      * Return true if this step should prompt based on the current state of the wizardContext
      * Used to prevent duplicate prompts from sub wizards, unnecessary prompts for values that had a default, and to accurately describe the number of steps
      */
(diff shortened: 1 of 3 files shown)


MicroFish91:
Related to: https://github.com/microsoft/vscode-azurefunctions/issues/3465

nturinski (approved):
I tested this out a bit myself and it seems like it works pretty well. I'm not really sure why `step.propertiesBeforePrompt` was within the `if (step.shouldPrompt())` statement before, but `goBack()` seems to be still working fine.

PR #1348, opened 2023-01-13
Start host api v2 documentation

alexweininger:
This is just a starting point. There will certainly be updates and additions over the next few weeks. The process of writing the docs has helped me notice some changes I'd like to make to the client extension developer experience that I'll try to make in the next week or so.

[Preview it here](https://github.com/microsoft/vscode-azuretools/blob/alex/v2-api-docs/utils/docs/host-api.md)

The old docs lived in the resource groups repo, however at least for now I think it's best to keep the docs in the same repo as the declaration. That way we can easily update them together.


bwateratmsft (dismissed):

bwateratmsft (approved):

alexweininger:
Going to merge for now, but will be moved soon to the new api package in the resource groups repo. See https://github.com/microsoft/vscode-azureresourcegroups/pull/530

Issue #1353, opened 2023-01-18
Configure webpack config to handle "webworker" target


nturinski:
I'm using [Gitlens](https://github.com/gitkraken/vscode-gitlens/blob/8a712cb4314973644ae3e09073222387baa3681c/webpack.config.js) as an example of how we could set up webpack config. They currently return an array of configs which I'm not entirely sure if that's what we want to adopt as well.

alexweininger:
Check out what vscode-python does as well. Looks like they just have separate webpack config files for each target. Their webpack configs might be easier to copy since they don't have webviews like GitLens does (that I know of).

https://github.com/microsoft/vscode-python/tree/main/build/webpack

PR #1354, opened 2023-01-18
Edit config file to target webworker

nturinski:
Resolves https://github.com/microsoft/vscode-azuretools/issues/1353

Still a WIP-- wanna look at the Python extension as an example as well.


dev/index.d.ts
```diff
@@ -109,42 +114,42 @@ export declare function gulp_webpack(mode: string): cp.ChildProcess;
  * Loose type to use for T1 and T2 versions of "@azure/ms-rest-js".  The Azure Account extension returns
  * credentials that will satisfy both T1 and T2 requirements
  */
- export type AzExtServiceClientCredentials = AzExtServiceClientCredentialsT1 & AzExtServiceClientCredentialsT2;
```
alexweininger, line 112: what happened here

dev/src/webpack/getDefaultWebpackConfig.ts
```diff
@@ -45,11 +45,87 @@ export function getDefaultWebpackConfig(options: DefaultWebpackOptions): webpack
...
+        plugins.push(new webpack.ProvidePlugin({
+            process: 'process/browser'
+        }));
+    }
+
     const config: webpack.Configuration = {
+        mode: 'none',
         context: options.projectRoot,
 
         // vscode extensions run in a Node.js context, see https://webpack.js.org/configuration/node/
```
alexweininger, line 127: Comment could use some updating now that target is configurable.

dev/package.json
```diff
@@ -47,22 +47,33 @@
...
         "@azure/ms-rest-nodeauth": "^3.0.5",
         "clean-webpack-plugin": "^3.0.0",
-        "copy-webpack-plugin": "^6.0.0",
+        "copy-webpack-plugin": "6.0.0",
         "fs-extra": "^8.0.0",
         "terser-webpack-plugin": "^5.0.0",
         "ts-loader": "^9.4.2",
-        "webpack": "5.28.0"
+        "webpack": "^5.74.0",
+        "console-browserify": "^1.2.0",
```
alexweininger, line 75: Where are we using console?

dev/index.d.ts
nturinski: Dude... I have no idea =_=

dev/package.json
nturinski: All those browserify packages were required to get Resource Groups to work right now. I'm debating if I should put all the fallbacks in the dev extension or just let the extensions that need it to import it.

Probably something in between where the most common ones are in dev, and very specific ones are in client extensions.

alexweininger:
Also - to what extent have you tested this config with the extensions?

nturinski:
Tested it with the Resources extension. Seems like webpack still works for both targets and can be debugged. It's not super easy to test on other extensions due to having to set up `package.json` and some other stuff.

dev/index.d.ts
nturinski: I think that whenever this file was edited (most likely copied/pasted) a stray prefixed space snuck in. I think how it is now is properly formatted.

image

alexweininger:
There's a lint error in the build, otherwise looks good to me.

alexweininger (dismissed):

alexweininger (approved):

alexweininger:
😡

dev/src/webpack/getDefaultWebpackConfig.ts
```diff
@@ -45,11 +45,78 @@ export function getDefaultWebpackConfig(options: DefaultWebpackOptions): webpack
...
+        new CopyWebpackPlugin({
+            patterns: [
+                // Test files -> dist/test (these files are ignored during packaging)
+                {
+                    from: path.posix.join(options.projectRoot.replace(/\\/g, '/'), 'out', 'test', '*.*'),
+                    to: path.posix.join(options.projectRoot.replace(/\\/g, '/'), 'dist', 'test'),
+                    noErrorOnMissing: true
+                },
+                {
+                    from: path.join(options.projectRoot, 'node_modules', '@microsoft', 'vscode-azext-azureutils', 'resources', '**', '*.svg'),
```
nturinski, line 60: Removed some of the old copy paths because they were old azuretools packages that we weren't using anymore.

PR #1366, opened 2023-02-02
Support Premium V3 sku in vscode app service extensions

Flanker32:
- Support Premium V3 sku in vscode app service extensions
- For Jboss EAP, only provide V3 sku when create service plan (As others sku are not supported)

### DIFF
--- a/appservice/src/createAppService/AppServicePlanSkuStep.ts
+++ b/appservice/src/createAppService/AppServicePlanSkuStep.ts
@@ -22,7 +22,7 @@ export class AppServicePlanSkuStep extends AzureWizardPromptStep<IAppServiceWiza
             skus = this.getWorkflowStandardSkus();
             // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
         } else if (context['newSiteJavaStack']?.majorVersion?.value === 'jbosseap') {
-            // for jboss eap, only Pv3 plan] is supported
+            // for jboss eap, only Pv3 plan is supported
             skus = this.getAdvancedSkus().filter(s => s.family === 'Pv3');
         }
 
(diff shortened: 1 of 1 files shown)


appservice/src/createAppService/AppServicePlanSkuStep.ts
```diff
@@ -20,6 +20,10 @@ export class AppServicePlanSkuStep extends AzureWizardPromptStep<IAppServiceWiza
             skus.push(...this.getElasticPremiumSkus());
         } else if (context.newSiteKind?.includes(AppKind.workflowapp)) {
             skus = this.getWorkflowStandardSkus();
+            // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+        } else if (context['newSiteJavaStack']?.majorVersion?.value === 'jbosseap') {
+            // for jboss eap, only Pv3 plan] is supported
```
nturinski, line 25:
```suggestion
            // for jboss eap, only Pv3 plan is supported
```

appservice/src/createAppService/AppServicePlanSkuStep.ts
```diff
@@ -144,6 +148,27 @@ export class AppServicePlanSkuStep extends AzureWizardPromptStep<IAppServiceWiza
                 size: 'P3v2',
                 family: 'Pv2',
                 capacity: 1
+            },
+            {
```
nturinski, line 152: Actually, sorry, but I just realized that we may not want to show all of these Pxv3 plans in advanced creation. Could you move these plans into another function that is used specifically for jboss eap apps? Something like `getPremiumV3Skus`?

Flanker32: @nturinski thanks for your message and really sorry for the late response, i've updated the pr that only show v3 pricing for jboss eap

nturinski (approved):

PR #1368, opened 2023-02-03
v1.5 shim: Add `isAncestorOfImpl` support

alexweininger:
I was naive thinking I could avoid implementing this, but it's actually needed to ensure the compat shim fully supports all palette experiences. Luckily it wasn't actually that hard to implement.

Basically ancestor picks will be selected based on the `isAncesotOfImpl` method, if present. Just like it worked before. And we only enable skip if one, if there is no `GenericTreeItem` in the pick list, since we don't want to auto select commands.

Needed for https://github.com/microsoft/vscode-azurestorage/issues/1198

### DIFF
--- a/utils/src/tree/GenericTreeItem.ts
+++ b/utils/src/tree/GenericTreeItem.ts
@@ -8,6 +8,7 @@ import { AzExtTreeItem } from './AzExtTreeItem';
 import { IAzExtParentTreeItemInternal } from "./InternalInterfaces";
 
 export class GenericTreeItem extends AzExtTreeItem implements types.GenericTreeItem {
+    public readonly _isGenericTreeItem = true;
     public label: string;
     public contextValue: string;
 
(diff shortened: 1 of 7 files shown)


utils/src/tree/GenericTreeItem.ts
```diff
@@ -28,3 +29,7 @@ export class GenericTreeItem extends AzExtTreeItem implements types.GenericTreeI
         return this._includeInTreeItemPicker;
     }
 }
+
+export function isGenericTreeItem(item: unknown): item is GenericTreeItem {
+    return (item as GenericTreeItem)._isGenericTreeItem;
```
bwateratmsft, line 34: For maximum runtime safety?
```suggestion
    return typeof item === 'object' && (item as GenericTreeItem)._isGenericTreeItem;
```
(e.g. as written, if I passed in `undefined`, it would cause an error)

bwateratmsft (approved):

bwateratmsft (approved):

PR #1376, opened 2023-02-07
Add `createPortalUri` util

alexweininger:


azure/src/utils/createPortalUri.ts
```diff
@@ -0,0 +1,15 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.txt in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import { AzureSubscription } from "@microsoft/vscode-azureresources-api";
```
bwateratmsft, line 6: Is `import type` sufficient here? I think that would also leave it as a dev dependency.

bwateratmsft (approved):

PR #1379, opened 2023-02-10
Faster webpacking by doing type checking on separate process

alexweininger:
See https://webpack.js.org/guides/build-performance/#typescript-loader

Testing on Resource Groups, running `webpack build --mode production` and the results were great.
Before: ~30s
After: ~15s

@nturinski verify that this doesn't mess with any of the recent work you've been doing for .dev. But it might make it easier.

Also, our webpack gulp tasks run tsc before webpacking. Ideally we would only rely on this plugin for doing type checking instead of running tsc first, then running webpack. 

If for some reason we can't use this plugin, at the very least we should only type check once. We're currently doing it twice, once with tsc, and once with ts-loader while webpacking.

### DIFF
--- a/dev/package-lock.json
+++ b/dev/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "@microsoft/vscode-azext-dev",
-    "version": "0.2.0",
+    "version": "0.2.1",
     "lockfileVersion": 2,
     "requires": true,
     "packages": {
         "": {
             "name": "@microsoft/vscode-azext-dev",
-            "version": "0.2.0",
+            "version": "0.2.1",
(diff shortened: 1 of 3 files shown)


bwateratmsft:
Is the goal to use webpacking on inner-loop or just shave 15 seconds off our CI builds? If it's the former, I'd use `webpack --watch` instead.

alexweininger:
> Is the goal to use webpacking on inner-loop or just shave 15 seconds off our CI builds? If it's the former, I'd use `webpack --watch` instead.

Good point. It still might be worth adding since it should improve watch performance too. Which for me right now averages 2-3s.

PR #1382, opened 2023-02-16
Support command tree items in quick pick wizard

alexweininger:
Fixes https://github.com/microsoft/vscode-azureresourcegroups/issues/578

When a command is picked, the current step is rerun as a sub wizard after the command is executed.


utils/src/pickTreeItem/GenericQuickPickStep.ts
```diff
@@ -33,6 +33,21 @@ export abstract class GenericQuickPickStep<TContext extends types.QuickPickWizar
         wizardContext.pickedNodes.push(pick);
     }
 
+    // provide support for picking command tree items, and running the command within the wizard
+    public async getSubWizard(wizardContext: TContext): Promise<types.IWizardOptions<TContext> | undefined> {
```
bwateratmsft, line 37: I'd lean toward putting this in a subclass of `GenericQuickPickStep`. Also, if at all possible, deferring the "doing" part to execution after the wizard input...

alexweininger: > I'd lean toward putting this in a subclass of `GenericQuickPickStep`.

Makes sense. In that case, I think I'll only need to change `QuickPickAzureSubscriptionStep` to extend the new subclass instead of extending `GenericQuickPickStep`.

> Also, if at all possible, deferring the "doing" part to execution after the wizard input...

For scenarios like https://github.com/microsoft/vscode-azureresourcegroups/issues/578 we can't defer execution 
since the commands need to execute in order to finish the "input" if that makes sense.

bwateratmsft: Yeah, makes sense. Could we split it into two separate wizards? Signin is chosen, then executed, then "another" wizard pops up to proceed?

It's certainly not an easy change but might be a better way to do it.

alexweininger: Yes, I've experimented with that in the past and will revisit. Just to clarify, the idea is to have an "ensure user is signed in and has subscriptions selected" wizard that runs before the pick subscription wizard?

bwateratmsft: I was thinking the "Pick subscription" ends in executing the sign-in command but I actually like your idea much better.

alexweininger: Unless I access the Azure Account API directly in utils rather than only using the tree data provider. I feel like the current solution works best. Making it a subclass means this logic is only applied to the `QuickPickAzureSubscriptionStep`. That way we can easily change it later if we want.

bwateratmsft (approved):

PR #1386, opened 2023-02-21
utils: Fix create pick not showing on step directly after resource step

alexweininger:
Currently the logic to add a create pick is in `getSubWizard`. That means it only runs after the step is run, which means the create pick is never added to first recursive compat step.

Moving the logic to prompt from getSubWizard fixes this.

Fixes https://github.com/microsoft/vscode-azureappservice/issues/2384

### DIFF
--- a/utils/src/pickTreeItem/contextValue/compatibility/CompatibilityRecursiveQuickPickStep.ts
+++ b/utils/src/pickTreeItem/contextValue/compatibility/CompatibilityRecursiveQuickPickStep.ts
@@ -33,13 +33,12 @@ export class CompatibilityRecursiveQuickPickStep<TContext extends types.QuickPic
         const lastPickedItem = getLastNode(wizardContext);
         const lastPickedItemTi = isWrapper(lastPickedItem) ? lastPickedItem.unwrap<AzExtTreeItem>() : lastPickedItem;
 
-        this.promptOptions = isAzExtParentTreeItem(lastPickedItemTi) ? {
-            ...this.promptOptions,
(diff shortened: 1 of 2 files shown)


utils/src/pickTreeItem/contextValue/compatibility/CompatibilityRecursiveQuickPickStep.ts
```diff
@@ -29,6 +29,25 @@ interface CompatibilityRecursiveQuickPickOptions extends types.ContextValueFilte
 export class CompatibilityRecursiveQuickPickStep<TContext extends types.QuickPickWizardContext & types.ITreeItemPickerContext> extends CompatibilityContextValueQuickPickStep<TContext, CompatibilityRecursiveQuickPickOptions> {
 
     protected override async promptInternal(wizardContext: TContext): Promise<unknown> {
+
+        const lastPickedItem = getLastNode(wizardContext);
+        const lastPickedItemTi = isWrapper(lastPickedItem) ? lastPickedItem.unwrap<AzExtTreeItem>() : lastPickedItem;
+
+        this.promptOptions = isAzExtParentTreeItem(lastPickedItemTi) ? {
```
bwateratmsft, line 36: Can't we just edit the properties rather than recreating the object? Then `promptOptions` can stay `readonly` in `GenericQuickPickStep`.

alexweininger: Fixed. That's why we do code reviews! 😄

bwateratmsft (approved):

PR #1390, opened 2023-02-22
azure: Revert changes to StorageAccountCreateStep

motm32:
Fixes https://github.com/microsoft/vscode-azurefunctions/issues/3562

Seems like [this PR](https://github.com/microsoft/vscode-azuretools/pull/1225/files) introduced changes that caused a suffix to always be added to Storage Account names. Reverting the changes as the original issue only seems to be with Logic Apps.

### DIFF
--- a/azure/src/wizard/StorageAccountCreateStep.ts
+++ b/azure/src/wizard/StorageAccountCreateStep.ts
@@ -4,17 +4,14 @@
  *--------------------------------------------------------------------------------------------*/
 
 import type { SkuName, StorageManagementClient } from '@azure/arm-storage';
-import { AzureWizardExecuteStep, IAzureNamingRules } from '@microsoft/vscode-azext-utils';
+import { AzureWizardExecuteStep } from '@microsoft/vscode-azext-utils';
 import { Progress } from 'vscode';
 import * as types from '../../index';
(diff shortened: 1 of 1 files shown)


bwateratmsft (approved):

ccastrotrejo:
Hi @motm32 & @bwateratmsft, since the issue got reverted, we are experiencing the issue again in logic apps extension. Maybe a way to fix it is to check first if the name is a valid name and is available and if not we can add the suffix to it. What do you think about it?
I first followed this implementation as utils/src/wizard/AzureNameStep.ts follows it

motm32:
> Hi @motm32 & @bwateratmsft, since the issue got reverted, we are experiencing the issue again in logic apps extension. Maybe a way to fix it is to check first if the name is a valid name and is available and if not we can add the suffix to it. What do you think about it? I first followed this implementation as utils/src/wizard/AzureNameStep.ts follows it

Hi @ccastrotrejo does this issue occur with all scenarios? This doesn't seem to be happening in functions. It possibly could be because of this added [step](https://github.com/microsoft/vscode-azurefunctions-clone/blob/94f5eee2f87522c3d1ec996ab35e268857b9764b/src/commands/deploy/storageAccountSteps/CustomLocationStorageAccountStep.ts).

PR #1397, opened 2023-02-23
utils: Add more descriptive error message for "Unexpected token in JSON..." error

motm32:
Fixes https://github.com/microsoft/vscode-azurefunctions/issues/3572

### DIFF
--- a/utils/src/utils/AzExtFsExtra.ts
+++ b/utils/src/utils/AzExtFsExtra.ts
@@ -79,14 +79,16 @@ export namespace AzExtFsExtra {
     export async function readJSON<T>(resource: Uri | string): Promise<T> {
         const file = await readFile(resource);
         try {
-            JSON.parse(file)
+            return JSON.parse(file);
         } catch (err) {
             const pError = parseError(err);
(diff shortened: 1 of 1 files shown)


utils/src/utils/AzExtFsExtra.ts
```diff
@@ -78,7 +78,15 @@ export namespace AzExtFsExtra {
 
     export async function readJSON<T>(resource: Uri | string): Promise<T> {
         const file = await readFile(resource);
-        return JSON.parse(file) as T;
+        try {
+            JSON.parse(file)
```
bwateratmsft, line 82: This is parsing twice but doesn't need to. You can do this instead:

```typescript
try {
    return JSON.parse(file) as T;
} catch (err) {
    const pError = parseError(err);
    if (pError.errorType === 'SyntaxError') {
        // If it's a syntax error, wrap it in a better error message
        throw new Error(`Error parsing JSON file: ${resource}. ${pError.message}`);
    } else {
        // Otherwise rethrow the original error
        throw err;
    }
}

bwateratmsft (approved):

PR #1399, opened 2023-02-24
all: Switch to `l10n` API

### LINKED ISSUE #1359: Use VSCode's new localization API

bwateratmsft:
Closes #1359.

We'll also need to cascade up the `utils` and `azure` packages to fully get rid of `vscode-nls`.


appservice/src/swapSlot.ts
```diff
@@ -31,12 +30,13 @@ export async function swapSlot(context: IActionContext, sourceSlot: ParsedSite,
         }
     }
 
-    const placeHolder: string = localize('selectSlotToSwap', 'Select which slot to swap with "{0}".', sourceSlot.slotName);
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const placeHolder: string = l10n.t('Select which slot to swap with "{0}".', sourceSlot.slotName!);
```
nturinski, line 34: Normally I'd say we should probably use nonNullValue(), but throwing an error because of a string error seems excessive.

nturinski (approved):

appservice/src/swapSlot.ts
bwateratmsft: I think it's also partly because the Swagger definitions for these APIs suggest that some things can be optional, that never actually can be optional. The `!` "operator" seems fitting in that scenario.

PR #1401, opened 2023-03-13
utils + azure: vscode.dev migration work

nturinski:
TODO: There's still some commented out code that I need to fix before we can merge this, just getting the PR up


utils/src/utils/openUrl.ts
```diff
@@ -3,11 +3,9 @@
...
 
-import open = require("open");
+import * as vscode from 'vscode';
 
 export async function openUrl(url: string): Promise<void> {
     // Using this functionality is blocked by https://github.com/Microsoft/vscode/issues/85930
-    // await vscode.commands.executeCommand('vscode.open', vscode.Uri.parse(url));
-
-    await open(url);
+    await vscode.commands.executeCommand('vscode.open', vscode.Uri.parse(url));
```
bwateratmsft, line 10: Might need to make changes in the default webpack config in the dev package too, with `open` removed

azure/src/index.ts
```diff
@@ -4,6 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 export * from './createAzureClient';
+export { registerAzureUtilsExtensionVariables } from './extensionVariables';
```
alexweininger, line 7: nit: Undo this change since it's just reordering the imports?

azure/src/azure-account.api.d.ts
```diff
@@ -4,61 +4,61 @@
  *--------------------------------------------------------------------------------------------*/
```
alexweininger, line 4: nit: revert the whitespace changes in this file.

azure/package.json
```diff
@@ -35,19 +35,23 @@
...
+        "@azure/arm-storage": "^18.0.0",
         "@azure/arm-storage-profile-2020-09-01-hybrid": "^2.0.0",
-        "@azure/ms-rest-js": "^2.2.1",
-        "@microsoft/vscode-azext-utils": "^0.4.3",
+        "@azure/core-client": "^1.6.0",
+        "@azure/core-rest-pipeline": "^1.9.0",
+        "@microsoft/vscode-azext-utils": "0.4.3-alpha",
         "semver": "^7.3.7",
+        "uuid": "^9.0.0",
+        "uuidv4": "^6.2.13",
```
alexweininger, line 45: Why do we need both of these?

azure/package.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "@microsoft/vscode-azext-azureutils",
     "author": "Microsoft Corporation",
-    "version": "0.3.8",
+    "version": "0.3.9",
```
alexweininger, line 4: These changes may warrant a minor version bump. Specifically around the client/pipeline stuff.

Part of me wants to release a 1.0.0 of the tools package so we can actually use major/minor/patch instead of just minor/patch.

utils/package.json
```diff
@@ -34,18 +34,20 @@
...
         "@microsoft/vscode-azureresources-api": "^2.0.2",
         "@vscode/extension-telemetry": "^0.6.2",
+        "crypto-randomuuid": "^1.0.0",
         "dayjs": "^1.11.2",
         "escape-string-regexp": "^2.0.0",
-        "html-to-text": "^8.2.0",
-        "open": "^8.0.4",
         "semver": "^7.3.7",
+        "text-encoding": "^0.7.0",
+        "uuid": "^9.0.0",
```
alexweininger, line 42: We need both `crypto-randomuuid` and `uuid`?

utils/src/createExperimentationService.ts
```diff
@@ -1,135 +1,135 @@
...
-
-    /**
-     * Implements a telemetry handler that adds the shared properties to the event
-     * @param actionContext The action context
-     */
-    public handleTelemetry(actionContext: IActionContext): void {
-        Object.assign(actionContext.telemetry.properties, this.sharedProperties);
-    }
-}
+// /*---------------------------------------------------------------------------------------------
```
alexweininger, line 1: What's the plan for this?

utils/src/createTelemetryReporter.ts
```diff
@@ -3,9 +3,9 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import TelemetryReporter from '@vscode/extension-telemetry';
```
alexweininger, line 6: nit: undo reordering

utils/src/registerReportIssueCommand.ts
```diff
@@ -3,16 +3,16 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import * as dayjs from 'dayjs';
-// eslint-disable-next-line import/no-internal-modules
-import * as relativeTime from 'dayjs/plugin/relativeTime';
+// import * as dayjs from 'dayjs';
```
alexweininger, line 6: We can't get `dayjs` to work? 😢

utils/src/utils/openUrl.ts
alexweininger, line 10: I'm sorry I think this has been discussed already, but why can't we use `vscode.env.openExternal` here?

bwateratmsft: VSCode is over-encoding some parts of a URI: https://github.com/Microsoft/vscode/issues/85930

However, the workaround suggested--pass a `string` and use `as unknown as vscode.Uri` to placate Typescript--should work. The code definitely very deliberately supports strings being passed in, even if the type declaration doesn't say so.

I think we should change this `openUrl` utility method to accept either `string` or `vscode.Uri`, and pass unchanged.

nturinski: I made this change in resources, but I think that I'll just do it here and delete it everywhere else?

azure/src/index.ts
nturinski: I mean, it was in the wrong spot before anyway. It's in alphabetical order now.

azure/src/azure-account.api.d.ts
nturinski: I think the spaces were wrong originally. Everything is indented twice for some reason.

azure/package.json
nturinski: I wouldn't mind. We don't really have a "GA" process for these packages so we could just go ahead and do it.

azure/src/index.ts
alexweininger: Ok fine. I'm just being nit-picky about this PR touching all these extra files. Kinda messes with the git blame. I'd prefer if we fixed all the imports in a separate PR but it's fine as is.

utils/src/createExperimentationService.ts
nturinski: Looks like I was able to just uncomment it and it's working. I can't remember what was broken before.

alexweininger: 😄

utils/src/registerReportIssueCommand.ts
nturinski: It didn't have a browser entry point in their `package.json` so I may have just assumed. I'll see if it works-- the browser entry point may only specifically be a VS Code for the Web thing,

utils/src/createTelemetryReporter.ts
alexweininger: Ok, then just ignore my nits about undoing the import reordering.

azure/src/azure-account.api.d.ts
bwateratmsft: It was tabs before, IIRC, and every time we get near it, it's trying to change it to spaces. Not even worth fighting with it...

azure/test/request.test.ts
```diff
@@ -115,38 +115,22 @@ suite('request', () => {
...
-
-        const request = new WebResource(url);
-        request.operationSpec = { httpMethod: "GET", responses: { 200: {}, default: {} }, serializer: new Serializer() };
-        const client = await createGenericClient(await createTestActionContext(), undefined);
-        await assertThrowsAsync(async () => await client.sendRequest(request), /oops/);
-    });
-
-    test('operationSpec with expected error', async () => {
-        testResponses = [{ status: 404, body: 'oops' }];
+    //test('operationSpec with unexpected  error', async () => {
```
nturinski, line 118: I'm not sure if we're still using operationSpec errors because it was our StatusCodePolicy pipeline controlling that. There are still some pipelines that I need to investigate into adding back in.

utils/src/registerReportIssueCommand.ts
alexweininger: `browser` is a standardized property, however it's not required and I think a lot of packages haven't switched over to it yet.

dev/src/webpack/getDefaultWebpackConfig.ts
```diff
@@ -211,6 +209,10 @@ export function getDefaultWebpackConfig(options: DefaultWebpackOptions): webpack
                     "async_hooks": false,
                     "child_process": false,
                     "fs": false,
+                    'html-to-text': false,
```
nturinski, line 212: This package is used to parse errors. I'm not sure how to really test it (because I don't see html errors often) but it's wrapped in a try/catch so I don't think that this should cause any exceptions to be thrown in a vscode.dev environment.

bwateratmsft: Hasn't this one given us CG alerts before? Just to parse errors feels like a pretty flimsy reason to keep it...

alexweininger: It's only used to parse error messages in case the error message is an html document, which seems somewhat common for Azure errors. See https://github.com/microsoft/vscode-azuretools/pull/360

alexweininger: Also, I couldn't quickly find any alternatives that are browser compatible, so 🤷

bwateratmsft: Oh the irony...can't find a browser-compatible package to parse HTML

utils/src/registerReportIssueCommand.ts
nturinski: Yeah, good news, I'm pretty sure that it does work.

azure/package.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "@microsoft/vscode-azext-azureutils",
     "author": "Microsoft Corporation",
-    "version": "0.3.8",
+    "version": "1.0.0",
```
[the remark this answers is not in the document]
nturinski: lol I thought we agreed to this?!?!

azure/test/request.test.ts
```diff
@@ -115,38 +115,22 @@ suite('request', () => {
...
 
-    test('operationSpec with unexpected  error', async () => {
-        testResponses = [{ status: 404, body: 'oops' }];
-
-        const request = new WebResource(url);
-        request.operationSpec = { httpMethod: "GET", responses: { 200: {}, default: {} }, serializer: new Serializer() };
-        const client = await createGenericClient(await createTestActionContext(), undefined);
-        await assertThrowsAsync(async () => await client.sendRequest(request), /oops/);
-    });
+    // test('operationSpec with unexpected  error', async () => {
```
alexweininger, line 118: Can we uncomment these tests?

azure/package.json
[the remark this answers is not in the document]
alexweininger: The gif was supposed to relay celebration, but looks like I picked the wrong gif. I really wanted a Kramer gif. I was lazy and didn't take the time to pick the right gif. That's on me.

azure/index.d.ts
```diff
@@ -377,18 +362,19 @@ export type AzExtGenericClientInfo = AzExtGenericCredentials | { credentials: Az
  * 2. Uses resourceManagerEndpointUrl to support sovereigns (if clientInfo corresponds to an Azure environment)
  * @param clientInfo The client/credentials info or `undefined` if no credentials are needed
  */
-export function createGenericClient(context: IActionContext, clientInfo: AzExtGenericClientInfo): Promise<ServiceClient>;
+export declare function createGenericClient(context: IActionContext, clientInfo: AzExtGenericClientInfo | undefined): Promise<ServiceClient>;
```
alexweininger, line 365:
```suggestion
export declare function createGenericClient(context: IActionContext, clientInfo?: AzExtGenericClientInfo): Promise<ServiceClient>;
```

azure/src/createAzureClient.ts
```diff
@@ -35,131 +33,121 @@ export function parseClientContext(clientContext: InternalAzExtClientContext): I
...
+    const client = await createGenericClient(context, clientInfo, { noRetryPolicy: true });
     return await client.sendRequest(request);
 }
 
 interface IGenericClientOptions {
     noRetryPolicy?: boolean;
 }
 
-export async function createGenericClient(context: IActionContext, clientInfo: types.AzExtGenericClientInfo, options?: IGenericClientOptions): Promise<ServiceClient> {
+export async function createGenericClient(context: IActionContext, clientInfo: types.AzExtGenericClientInfo | undefined, options?: IGenericClientOptions): Promise<ServiceClient> {
```
alexweininger, line 74:
```suggestion
export async function createGenericClient(context: IActionContext, clientInfo?: types.AzExtGenericClientInfo, options?: IGenericClientOptions): Promise<ServiceClient> {
```

azure/test/request.test.ts
nturinski: Not yet-- I need to either edit them or manipulate with the pipelines a bit more to see why they are still failing. 

But the way that the error is shaped in the tests don't really make sense to me and I don't see any pipeline that would fix this.

azure/index.d.ts
nturinski: I think the idea for it to be undefined was so that you could create a generic client to make a one-off request to an endpoint without having to specify a client sdk info.

alexweininger: Yeah but isn't my suggestion just changing the syntax and not the result?

nturinski: Oh yeah, I didn't see the ?.

Which leads me to why I'd prefer not to do your suggestion. When you mark it as optional, it makes omitting `clientInfo` less intentional. 

Whereas with the old syntax, I believe that you _must_ type `undefined` in the parameter list. It kind of enforces whoever is using the command to acknowledge that they are not passing in anything on purpose.

alexweininger: Nice, I guess that's a perfect scenario to not use `?`. 👍

azure/src/createAzureClient.ts
```diff
@@ -208,3 +210,29 @@ class AddEndpointPolicy implements PipelinePolicy {
         return await next(request);
     }
 }
+
+/**
+ * The Azure SDK will only throw errors for bad status codes if it has an "operationSpec", but none of our "generic" requests will have that
+ */
+class StatusCodePolicy implements PipelinePolicy {
```
nturinski, line 217: This is basically what we had before for the following reason:
> The Azure SDK will only throw errors for bad status codes if it has an "operationSpec", but none of our "generic" requests will have that

OperationSpec isn't something that you can pass into the request anymore, but I believe it's built internally into the SDK. If the SDK call errors out, it throws the error during the pipeline. However, with generic requests, it will accept any status code as a resolved response, so we need to be checking the error and throwing it ourselves.

azure/test/request.test.ts
```diff
@@ -52,85 +52,66 @@ suite('request', () => {
...
     test('400 with json error message', async () => {
-        await assertThrowsAsync(async () => await sendTestRequest({ status: 400, body: '{ "message": "oops" }' }), (err: Error) => err.message.includes('oops') && !err.message.includes('message'));
+        await assertThrowsAsync(async () => await sendTestRequest({ statusCode: 400, body: '{ "message": "oops" }' }), (err: Error) => err.message.includes('oops') && !err.message.includes('message'));
     });
 
     test('ECONNRESET', async () => {
         await assertThrowsAsync(async () => await sendTestRequest(res => res.destroy()), /socket hang up/i);
     });
-
-    // test('operationSpec with unexpected  error', async () => {
```
nturinski: As I mentioned, we don't have `operationSpec` as a property anymore, so I removed these tests. The errors above basically test the same thing, which was the `StatusCodePolicy`

nturinski (commented):
Okay, I think I am done with all the PR feedback and other things that I wanted to test. So shouldn't be adding anymore changes to this one (unless there ends up being more feedback)

alexweininger:
Merge conflicts?

nturinski:
Okaaay, _that_ is my last commit. Stupid version changes-- why can't git figure it out?

alexweininger (dismissed):

alexweininger (approved):

nturinski:
Need to release dev -> utils -> azure in that order to make fix the linting errors

PR #1409, opened 2023-03-17
utils: Allow using `InputBoxValidationMessage` introduced in @types/vscode@1.74

alexweininger:
It's in @types/vscode@1.74.0 but was introduced in VS Code 1.76.0.

See https://code.visualstudio.com/api/references/vscode-api#InputBoxValidationMessage

### DIFF
--- a/utils/package.json
+++ b/utils/package.json
@@ -50,7 +50,7 @@
         "@types/mocha": "^7.0.2",
         "@types/node": "^16.0.0",
         "@types/semver": "^7.3.9",
-        "@types/vscode": "1.64.0",
+        "@types/vscode": "^1.74.0",
         "@typescript-eslint/eslint-plugin": "^5.53.0",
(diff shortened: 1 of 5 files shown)


utils/package.json
```diff
@@ -50,7 +50,7 @@
         "@types/mocha": "^7.0.2",
         "@types/node": "^16.0.0",
         "@types/semver": "^7.3.9",
-        "@types/vscode": "1.64.0",
+        "@types/vscode": "^1.74.0",
```
bwateratmsft, line 53:
```suggestion
        "@types/vscode": "1.74.0",
```

bwateratmsft, line 53: Should we just do 1.76 instead, so that we don't run into problems, if it was introduced in 1.74 but not available until 1.76?

bwateratmsft (approved):

PR #1413, opened 2023-03-20
utils: add `startingNode` parameter to `runQuickPickWizard`

### LINKED ISSUE #1411: Add an easy way to start picker experience from a specific (non-root) node
We used to have a way to easily specify a starting node for the tree item picker by passing it as an argument to `showTreeItemPicker`.

We need to add a similar feature for the new v2 picker utilities.

alexweininger:
Fixes #1411

### DIFF
--- a/utils/src/pickTreeItem/runQuickPickWizard.ts
+++ b/utils/src/pickTreeItem/runQuickPickWizard.ts
@@ -9,10 +9,10 @@ import { NoResourceFoundError } from '../errors';
 import { AzureWizard } from '../wizard/AzureWizard';
 import { getLastNode } from './getLastNode';
 
-export async function runQuickPickWizard<TPick>(context: types.PickExperienceContext, wizardOptions?: types.IWizardOptions<types.AzureResourceQuickPickWizardContext>): Promise<TPick> {
+export async function runQuickPickWizard<TPick>(context: types.PickExperienceContext, wizardOptions?: types.IWizardOptions<types.AzureResourceQuickPickWizardContext>, startingNode?: unknown): Promise<TPick> {
     // Fill in the `pickedNodes` property
     const wizardContext = { ...context } as types.AzureResourceQuickPickWizardContext;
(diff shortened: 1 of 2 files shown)


utils/src/pickTreeItem/runQuickPickWizard.ts
```diff
@@ -9,10 +9,10 @@ import { NoResourceFoundError } from '../errors';
 import { AzureWizard } from '../wizard/AzureWizard';
 import { getLastNode } from './getLastNode';
 
-export async function runQuickPickWizard<TPick>(context: types.PickExperienceContext, wizardOptions?: types.IWizardOptions<types.AzureResourceQuickPickWizardContext>): Promise<TPick> {
+export async function runQuickPickWizard<TPick>(context: types.PickExperienceContext, wizardOptions?: types.IWizardOptions<types.AzureResourceQuickPickWizardContext>, startingNode?: unknown): Promise<TPick> {
     // Fill in the `pickedNodes` property
     const wizardContext = { ...context } as types.AzureResourceQuickPickWizardContext;
-    wizardContext.pickedNodes = [];
+    wizardContext.pickedNodes = startingNode ? [startingNode] : [];
```
nturinski, line 15: It doesn't need the whole list of nodes? Like, starting from the subscription -> resource, etc?

alexweininger: It doesn't need that whole list. Any information that has been skipped by starting at a specific node needs to be filled out in the wizard context. That info will be unique to what wizard steps you're running and you can fill that out on your own.

We've been using this exact pattern with the compatibility picker util https://github.com/microsoft/vscode-azuretools/blob/e47bec01a4eca5a61c7fbf71c0fb85f90dc156a0/utils/src/pickTreeItem/experiences/compatibility/PickTreeItemWithCompatibility.ts#L54

This is kinda meant for running context value based steps anyway.

nturinski (approved):

PR #1416, opened 2023-03-22
[azure] - Set defaultToOAuthAuthentication to true when creating storage account 

ccastrotrejo:
When create a storage account property `defaultToOAuthAuthentication` is set to "true"

This is necessary for a security response, and it represents a defense-in-depth measure. This property does not change how the function app interacts with the storage account, but it does change how Storage client experiences work for users.
https://dev.azure.com/devdiv/DevDiv/_workitems/edit/1775987

This property just controls what the storage clients do when someone attempts to use them to browse contents. Customers can always switch back over on their own, so this isn’t a hard block or anything, but it does change how permission requirements are presented to them


alexweininger:
I tested with Azure Storage and Azure Functions extensions. Tested creating and browsing Blobs/File Shares. As well as connecting a Function to a Storage Account using a blob binding. Was able to debug and trigger Blob function by adding items to the container.

alexweininger:
@ccastrotrejo changes released in https://github.com/microsoft/vscode-azuretools/releases/tag/azext-v0.3.9-azure

ccastrotrejo:
Thank you so much! @alexweininger

PR #1421, opened 2023-03-27
dev: Update the webpack config to alias crypto

nturinski:
Also removes clean-webpack because all it has done it cause me pain


dev/src/webpack/getDefaultWebpackConfig.ts
```diff
@@ -254,17 +262,6 @@ export function getDefaultWebpackConfig(options: DefaultWebpackOptions): webpack
         ]
     };
 
-    // Clean the dist folder before webpacking
-    if (!options.suppressCleanDistFolder) {
```
alexweininger, line 258: There's an option to suppress it, should we just use that?

alexweininger (commented):
Did you mean to include some utils changes in this PR?

nturinski:
Crap. I did not.

dev/src/webpack/getDefaultWebpackConfig.ts
nturinski: Uggggh I guess we could. It was nice to delete something that hurt me though.

nturinski:
Closing in favor of https://github.com/microsoft/vscode-azuretools/compare/dev/nat/webpackConfig?expand=1

PR #1422, opened 2023-03-27
Export randomUtils. Use webCrypto for browser environment

nturinski:
Relies on https://github.com/microsoft/vscode-azuretools/pull/1421

`getPseudononymousStringHash` now async so I bumped up a minor version since it's a breaking change (though should only be within its own package, but sitll)


alexweininger (approved):

nturinski:
Alex already said yes sooooo 😗

PR #1423, opened 2023-03-27
azure: Check if theres a parsedBody before activating statusCodePipeline

nturinski:
So this one is a little complicated...

I ran into an error where `resources.resourceGroup.checkExistence` was throwing an error because the status code is 404 due to our new pipeline. However, a 404 just means it doesn't exist in this case, so it shouldn't be throwing.

The statusCodePolicy used to only check requests that didn't have an `operationSpec` attached to them. This would usually indicate that they were a generic request being made by us. But I can't find an equivalent field to check in the current pipelines. 

It doesn't seem like the options parameter is passed into the request or response. I guess we could set up a pipelines to do that?

What I have now works for the scenario I described because all of our azure requests have a `parsedBody`. If not, then it was probably a generic request or an error that wasn't handled.  But I don't exactly love this approach either.

### DIFF
--- a/azure/package-lock.json
+++ b/azure/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "@microsoft/vscode-azext-azureutils",
-    "version": "1.0.0",
+    "version": "1.0.1",
     "lockfileVersion": 2,
     "requires": true,
     "packages": {
         "": {
             "name": "@microsoft/vscode-azext-azureutils",
-            "version": "1.0.0",
+            "version": "1.0.1",
             "license": "MIT",
             "dependencies": {
                 "@azure/arm-resources": "^5.0.0",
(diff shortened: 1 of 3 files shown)


alexweininger (approved):
Fine for now. Could we add a custom setting to control if the pipeline is applied or not?

nturinski:
That may be the best alternative. The tests failed because an error message was included as a JSON and so it was the parsedBody, so looking at that property might not work anyway.

nturinski:
I figured it made sense for `sendRequestWithTimeout` to just default to using the `statusCodePolicy` since it was a quick, generic call. But if you create a `genericServiceClient`, then we'd allow the users to set whether or not they want the `statusCodePolicy`

alexweininger:
Do you need to update `index.d.ts`?

alexweininger (approved):

PR #1429, opened 2023-03-31
dev: Fix CopyWebpackPlugin test file paths

alexweininger:
Test files are expected to be in this location. [example](https://github.com/microsoft/vscode-azureresourcegroups/blob/b4c3927beaf9046d59f4c517540d64b1e6505436/test/runTest.ts#L35)

### DIFF
--- a/dev/package-lock.json
+++ b/dev/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "@microsoft/vscode-azext-dev",
-    "version": "1.0.4",
+    "version": "1.0.5",
     "lockfileVersion": 2,
     "requires": true,
     "packages": {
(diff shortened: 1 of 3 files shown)


nturinski:
Weird, I'm pretty sure that this PR should have fixed it: https://github.com/microsoft/vscode-azuretools/pull/1420/files

I don't think it was the vscode.dev stuff specifically, but because I updated webpack to ^5.7

https://github.com/microsoft/vscode-azuretools/commit/10fb87f2a443228f2cce96ea47ede5ab9a84a513

alexweininger:
I kept running into errors because the file path was wrong, but now I can't repro. 😕

PR #1437, opened 2023-04-04
dev: Support LogOutputChannel

alexweininger:
This is a breaking change since I updated the VS Code types, but it feels wrong to do a major version update.

### DIFF
--- a/dev/package.json
+++ b/dev/package.json
@@ -42,7 +42,7 @@
         "@types/mocha": "^7.0.2",
         "@types/node": "^16.0.0",
         "@types/terser-webpack-plugin": "^5.0.0",
-        "@types/vscode": "^1.77.0",
+        "@types/vscode": "^1.76.0",
         "@typescript-eslint/eslint-plugin": "^5.53.0",
         "@vscode/test-electron": "^2.1.5",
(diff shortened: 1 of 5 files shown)


dev/package.json
```diff
@@ -42,7 +42,7 @@
         "@types/mocha": "^7.0.2",
         "@types/node": "^16.0.0",
         "@types/terser-webpack-plugin": "^5.0.0",
-        "@types/vscode": "1.64.0",
+        "@types/vscode": "^1.77.0",
```
bwateratmsft, line 45: Does 1.76 have the necessary types? I think most of our other packages have been bumped to that version at this point. Also, I'm not sure it actually is breaking, I think `LogOutputChannel` is just an interface so this would be strictly a dev-time import.

alexweininger: 1.76 does have the necessary types. 

And I thought it was a breaking change since updating to this version of the dev package will could create compile-time TS errors since the VS Code typings have breaking changes.

bwateratmsft (dismissed):

motm32 (approved):

MicroFish91 (approved):

PR #1439, opened 2023-04-06
utils: Add `TreeElementStateManager`

alexweininger:
Also added `createGenericElement` util. Along with tests for each.

The implementation is copied from [TreeItemState.ts in Container Apps](https://github.com/microsoft/vscode-azurecontainerapps/blob/main/src/tree/TreeItemState.ts) with minor modifications which mostly consist of renaming things.

See how this is consumed in https://github.com/microsoft/vscode-azurecontainerapps/pull/341

---

A little note on "elements" vs "items"...

I'm referring to what we have been calling "items" as "elements" since that's what [VS Code calls them](https://code.visualstudio.com/api/references/vscode-api#TreeDataProvider&lt;T&gt;). I find calling them items is confusing since `vscode.TreeItem` is not the same as what we say is an "item". At the very least I think it's important to make this distinction here since `TreeElementStateManager` is dealing with elements and tree items.

We are still calling elements items all over the place in the codebase, but I hope we can slowly transition to calling them elements.

### DIFF
--- a/utils/index.d.ts
+++ b/utils/index.d.ts
@@ -1944,10 +1944,15 @@ export declare namespace randomUtils {
     export function getRandomInteger(minimumInclusive: number, maximumExclusive: number): number;
 }
 
+/**
+ * Base element for a tree view (v2)
+ */
 export declare interface TreeElementBase extends ResourceModelBase {
     getChildren?(): ProviderResult<TreeElementBase[]>;
     getTreeItem(): TreeItem | Thenable<TreeItem>;
 }
+
+export type TreeElementWithId = TreeElementBase & { id: string };
 
 export interface GenericElementOptions extends IGenericTreeItemOptions {
     commandArgs?: unknown[];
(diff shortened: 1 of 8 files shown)


utils/index.d.ts
```diff
@@ -1943,3 +1943,95 @@ export declare namespace randomUtils {
...
+
+/**
+ * Base element for a tree view (v2)
+ */
+export declare interface TreeElementBase extends ResourceModelBase {
+    getChildren?(): ProviderResult<TreeElementBase[]>;
+    getTreeItem(): TreeItem | Thenable<TreeItem>;
+}
+
+export type TreeElementWithId = TreeElementBase & { id: string };
```
alexweininger, line 1955: This is the first time `TreeElementBase` is being introduced into utils. Client extensions using the v2 API are not forced to use this model. However, we will be writing all of our own utils based on this model. Thus, our own extensions don't _have to_, but _should_ use this model when they consume the v2 api.

model model model

utils/index.d.ts
```diff
@@ -1950,3 +1950,95 @@ export declare namespace randomUtils {
...
+
+    /**
+     * Notify a resource that its children have changed.
+     * Calls the refresh callback with the given id.
+     * @param id - The id of the element for which the children have changed
+     */
+    notifyChildrenChanged(id: string): void;
+
+    /**
+     * Wraps an element in a state model that can be used to update the tree item. An element should only be wrapped once.
```
bwateratmsft, line 2034: > An element should only be wrapped once.

Is there a convenient way to throw an error if they try to double-wrap an object?

bwateratmsft (approved):

utils/index.d.ts
alexweininger: We could set a property on the tree element object to mark it as wrapped and throw an error if we see that property.

alexweininger: I implemented this in https://github.com/microsoft/vscode-azuretools/pull/1448, but I'm not sure we want to do it. Open to opinions.

bwateratmsft: I don't feel strongly, but it might save us from a very hard-to-debug issue in the future.

bwateratmsft (approved):

PR #1447, opened 2023-04-10
azure: Update return type for `sendRequestWithTimeout`

nturinski:


alexweininger (approved):

alexweininger:
That's probably a breaking change. At least will break compilation.

nturinski:
I think it should be okay since `AzExtPipelineResponse` extends `PipelineResponse`.

PR #1461, opened 2023-04-19
auth: Create a package for handling Azure auth and subscription listing

bwateratmsft:
Eventually this package can replace the Azure Account extension

Still TODO:
- [x] Add this package to build pipelines
- [x] Support custom clouds--probably a setting in the Resources extension that overrides what is present in the `microsoft-sovereign-cloud.endpoint` setting
- [x] Support configuring tenant and cloud (https://github.com/microsoft/vscode-azure-account/issues/805)
- [x] ~~Maybe like, actually do some testing IDK~~
  > "Nah" - Nathan
- [x] Work on the readme
- [ ] Add unit tests
- [ ] Add E2E tests


.azure-pipelines/job-steps.yml
bwateratmsft, line 1: I took Kudu out and put this one in

auth/src/AzureSubscription.ts
```diff
@@ -0,0 +1,58 @@
...
+
+    /**
+     * The display name of this subscription.
+     */
+    readonly name: string;
+
+    /**
+     * The ID of this subscription.
+     */
+    readonly subscriptionId: SubscriptionId;
```
alexweininger, line 47: This property currently is only the subscription id uuid. Are we intentionally changing it to include the tenant id?

bwateratmsft: I think we should make that change only for the filter setting. I don't know for sure but I think the code here gets just the subscription ID and not the tenant ID.

bwateratmsft: Still, a comment clarifying exactly what is meant would be good.

alexweininger: I guess there's conflicting types or comments here. Since `SubscriptionId` type says it includes the tenant id

```ts
/**
 * A type representing an Azure subscription ID, in the form `${tenantId}/${subscriptionId}`.
 */
export type SubscriptionId = string;
```

bwateratmsft: Comment is just plain wrong. I've fixed it.

auth/src/VSCodeAzureSubscriptionProvider.ts
```diff
@@ -0,0 +1,241 @@
...
+    }
+
+    /**
+     * Gets a fully-configured subscription client for a given tenant ID
+     *
+     * @param tenantId The tenant ID to get a client for
+     *
+     * @returns A client, the credential used by the client, and the authentication function
+     */
+    private async getSubscriptionClient(tenantId: 'organizations' | string): Promise<{ client: SubscriptionClient, credential: TokenCredential, authentication: AzureAuthentication }> {
```
alexweininger, line 178: Is `'organizations'` representing the default tenant?

auth/src/VSCodeAzureSubscriptionProvider.ts
```diff
@@ -0,0 +1,241 @@
...
+
+    /**
+     * Gets a normalized list of scopes
+     *
+     * @param scopes An input scope string, list, or undefined
+     * @param tenantId (Optional) The tenant ID, will be added to the scopes
+     *
+     * @returns A list of scopes, with the default scope and (optionally) the tenant scope added
+     */
+    private getScopes(scopes: string | string[] | undefined, tenantId?: string): string[] {
```
alexweininger, line 216: Tenant id will either be `'organizations'` or an actual tenant id. Never undefined.

```suggestion
    private getScopes(scopes: string | string[] | undefined, tenantId: string): string[] {
```

or 

```suggestion
    private getScopes(scopes: string | string[] | undefined, tenantId: 'organizations' | string): string[] {
```

If we go with this one I suggest extracting `'organizations'` as a new type.

auth/src/VSCodeAzureSubscriptionProvider.ts
alexweininger: If so, I'm not sure having a tenant id that represents the default tenant is really needed. If I understand what the default tenant is, it's the tenant associated with the current session, whatever tenant the user used when logging in.

The default tenant id can be grabbed from the session id. Can we just use that instead of `'organizations'`?

bwateratmsft: `organizations` is the scope needed to get a list of _tenants_ that a user has access to. It can be more than one.

bwateratmsft: > The default tenant id can be grabbed from the session id

We can but didn't we want to list all subs across all tenants?

alexweininger (approved):

PR #1465, opened 2023-04-19
Fix another packaging bug

bwateratmsft:


nturinski:
Should I do the change we discussed offline instead so we don't need to add this dependency?

bwateratmsft:
> Should I do the change we discussed offline instead so we don't need to add this dependency?

Yeah, I think that's totally fine.

alexyaang (approved):

Issue #1468, opened 2023-04-20
`@microsoft/vscode-azureresources-api`, `@microsoft/vscode-azext-utils`, `@microsoft/vscode-azext-azureutils`, `@microsoft/vscode-azext-azureappservice` fail TypeScript lib check


bwateratmsft:
It doesn't prevent compiling, but these packages fail the TypeScript library check, requiring consumers to either do `"skipLibCheck": true` in the TSConfig, or adding `@azure/ms-rest-azure-env` as a dev dependency. This can be fixed by making `@azure/ms-rest-azure-env` a **peer** dependency of the following packages:

- [x] `@microsoft/vscode-azureresources-api` https://github.com/microsoft/vscode-azureresourcegroups/commit/b52f3bc0de1d881d2cd619eb2f4e13fe5eeefdd7
- [x] `@microsoft/vscode-azext-utils` https://github.com/microsoft/vscode-azuretools/pull/1469
- [x] `@microsoft/vscode-azext-azureutils` https://github.com/microsoft/vscode-azuretools/pull/1470
- [x] `@microsoft/vscode-azext-azureappservice` https://github.com/microsoft/vscode-azuretools/pull/1471

`@microsoft/vscode-azext-azureappservice` also fails due to the `KuduModels` being imported from the old `vscode-azurekudu` package. These models should just be moved up into `@microsoft/vscode-azext-azureappservice`.

- [x] `@microsoft/vscode-azext-azureappservice` https://github.com/microsoft/vscode-azuretools/pull/1471

Lastly, `@microsoft/vscode-azext-azureappservice` fails due to an `import` instead of `import type` on `@microsoft/vscode-azext-azureappsettings` in `SiteClient.ts`

- [x] `@microsoft/vscode-azext-azureappservice` https://github.com/microsoft/vscode-azuretools/pull/1471

PR #1469, opened 2023-04-20
utils: Fix dependency

bwateratmsft:
Part of https://github.com/microsoft/vscode-azuretools/issues/1468

### DIFF
--- a/utils/package.json
+++ b/utils/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@microsoft/vscode-azext-utils",
     "author": "Microsoft Corporation",
-    "version": "1.2.1",
+    "version": "1.2.2",
     "description": "Common UI tools for developing Azure extensions for VS Code",
     "tags": [
         "azure",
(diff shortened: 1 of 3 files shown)


nturinski (approved):

utils/package.json
```diff
@@ -32,7 +32,7 @@
         "test": "node ./out/test/runTest.js"
     },
     "dependencies": {
-        "@microsoft/vscode-azureresources-api": "^2.0.2",
+        "@microsoft/vscode-azureresources-api": "^2.0.4",
```
bwateratmsft, line 35: Ideally we'd be using `@microsoft/vscode-azureresources-api@2.0.5` but I don't actually know how Alex releases it. However, an `npm update` will pick it up since the versions are compatible, and ultimately it doesn't actually matter because this package is declaring `@azure/ms-rest-azure-env` as a peer dependency too, so it will transit to consumers of this package.

PR #1457, opened 2023-04-13
Add ADO pipelines that can be shared across all our repos

bwateratmsft:
(title)

PR #1474, opened 2023-05-09
Add reusable GitHub Actions pipeline for lint/build/test

bwateratmsft:
@alexweininger if you want to add some stuff for the test case reporting please do.

Related to #1457

### DIFF
--- a/azure-pipelines/templates/setup.yml
+++ b/azure-pipelines/templates/setup.yml
@@ -1,15 +1,15 @@
 steps:
   - task: NodeTool@0
-    displayName: "Using Node.js"
+    displayName: Using Node.js
     inputs:
-    versionSource: "fromFile"
-    versionFilePath: .nvmrc
+      versionSource: fromFile
+      versionFilePath: .nvmrc
 
   - task: Npm@1
     inputs:
       command: custom
       customCommand: ci --no-optional
-    displayName: "Install Dependencies"
+    displayName: Install Dependencies
 
   - bash: |
       /usr/bin/Xvfb :99 -screen 0 1024x768x24 > /dev/null 2>&1 &


.github/workflows/main.yml
bwateratmsft, line 1: I'd have rather put this in `github-actions/workflows/main.yml`, however, GitHub Actions cannot be reused unless they are rooted in `.github/workflows`, which is a bummer.

alexweininger:
> @alexweininger if you want to add some stuff for the test case reporting please do.

It requires npm package updates and some other changes so I'll do that in a separate PR.

bwateratmsft:
> > @alexweininger if you want to add some stuff for the test case reporting please do.
> 
> It requires npm package updates and some other changes so I'll do that in a separate PR.

Sounds good. Plus the NPM stuff would have to be in the respective `package.json` files.

.github/workflows/main.yml
```diff
@@ -0,0 +1,37 @@
+on:
+  workflow_call:
+
+jobs:
+  Build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [16.x]
```
alexweininger, line 9: I'd like to fetch the node version from the `.nvm` file so we can update that in one place only. The setup node action has built-in support for doing that, but I'm not sure how to do that while still using the matrix.

https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#node-version-file

```
- uses: actions/setup-node@v3
  with:
    node-version-file: '.nvmrc'
```

bwateratmsft: Can we configure it to use a default if the `.nvmrc` file isn't present?

alexweininger: I can't find an example that does this, but maybe it'll work if you specify multiple.

```yaml
- uses: actions/setup-node@v3
  with:
    node-version-file: '.nvmrc'
    node-version: 16
```

bwateratmsft: image

bwateratmsft: Looks like both ADO and GH Actions equally support using `.nvmrc`, so I'll change the tasks to use that. And the readme to reflect that that's the requirement, instead of Node 16.

Issue #1419, opened 2023-03-23
Service Connector Integration


nturinski:
### What is it?
Service Connector helps you connect Azure compute services to other backing services. This service configures the network settings and connection information (for example, generating environment variables) between compute services and target backing services in management plane.

We can leverage this to connect web and container apps to backends such as Storage or Cosmos DB. It will do the connection, can use managed identities (user and system).

### Implementation
I propose that we make this a new shared package. The reason being, it seems like the only `App Service,` `Container Apps,` and `Spring Apps` support Service Connector.

We can use this Azure SDK to do the implementation: https://azuresdkdocs.blob.core.windows.net/$web/javascript/azure-arm-servicelinker/2.1.0/interfaces/Linker.html#beginCreateOrUpdate
https://www.npmjs.com/package/@azure/arm-servicelinker

For the UI, it could be another node under either an app service or container app nodes. Underneath that node, we could show something similar to the Portal:
![image](https://user-images.githubusercontent.com/5290572/227294632-75f34896-93e3-48ac-9935-f0c431c8dfb8.png)

### Documentation:
https://learn.microsoft.com/en-us/azure/service-connector/overview
https://learn.microsoft.com/en-us/azure/service-connector/quickstart-cli-app-service-connection?tabs=Using-access-key
https://learn.microsoft.com/en-us/azure/service-connector/quickstart-portal-app-service-connection?tabs=SMI

### Supported Regions
Service Connector doesn't support every region. We should have some smart checks to either 
  a.) only allow creating apps in these compatible locations or 
  b.) throw an error/disable SC for apps in unsupported locations
https://learn.microsoft.com/en-us/azure/service-connector/concept-region-support

PR #1476, opened 2023-05-25
serviceconnector: Add Service Connector 🔗

motm32:
Part of #1419. 

This will currently be used with the Container Apps and App Service extensions. The current implementation has picks for only storage connections.

Still to do: (these will go in seperate PR's)

- Add UI
- Add database connections


serviceconnector/src/createServiceConnector/AuthenticationTypeStep.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export class AuthenticationTypeStep extends AzureWizardPromptStep<ICreateLinkerContext>{
+    public async prompt(context: ICreateLinkerContext): Promise<void> {
+        const placeHolder = 'Select Authentication Type';
+        const picks: IAzureQuickPickItem<AuthInfoBase>[] = [
+            { label: 'System assigned managed identity', data: { authType: 'systemAssignedIdentity' } },
+            { label: 'User assigned managed identity', data: { authType: 'userAssignedIdentity' } },
+            { label: 'Service principal', data: { authType: 'servicePrincipalSecret' } },
+            { label: 'Connection string', data: { authType: 'secret' } },
+        ];
```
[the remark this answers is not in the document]
bwateratmsft: Related to above, would recommend `vscode.l10n.t` for that

serviceconnector/src/deleteServiceConnector/ServiceConnectorsListStep.ts
```diff
@@ -0,0 +1,35 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.md in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import { ServiceLinkerManagementClient } from "@azure/arm-servicelinker";
```
bwateratmsft, line 6: Consider lazy loading

serviceconnector/src/validateServiceConnector/ValidateLinkerStep.ts
```diff
@@ -0,0 +1,21 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.md in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import { ServiceLinkerManagementClient } from "@azure/arm-servicelinker";
```
bwateratmsft, line 6: Consider lazy loading

serviceconnector/src/createServiceConnector/ICreateLinkerContext.ts
```diff
@@ -0,0 +1,24 @@
...
+import { AuthInfoBase, LinkerResource } from "@azure/arm-servicelinker";
+import { IStorageAccountWizardContext } from "@microsoft/vscode-azext-azureutils";
+import { ISubscriptionActionContext } from "@microsoft/vscode-azext-utils";
+import { ServiceType } from "../../constants";
+
+
+export interface ICreateLinkerContext extends ISubscriptionActionContext, IStorageAccountWizardContext {
+    //Source resource
+    resourceUri?: string;
+    scope?: string;
```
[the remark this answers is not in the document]
motm32: Scope is used specifically for container apps so the service connector knows which container it's connecting to. I can add a comment to clarify this.

serviceconnector/src/deleteServiceConnector/LinkerListStep.ts
```diff
@@ -0,0 +1,35 @@
...
+
+    public async configureBeforePrompt(context: IPickLinkerContext): Promise<void> {
+        const picks = await this.getPicks(context);
+        if (picks.length === 1) {
+            context.linkerName = picks[0].data;
+        }
+    }
+
+    private async getPicks(context: IPickLinkerContext): Promise<IAzureQuickPickItem<string>[]> {
+        const client = new (await import('@azure/arm-servicelinker')).ServiceLinkerManagementClient(context.credentials);
```
alexweininger, line 29: We lazy load this client in a few places. I'd create a util to do it like we do in other packages/extensions.

serviceconnector/src/deleteServiceConnector/deleteLinker.ts
```diff
@@ -0,0 +1,40 @@
...
+export async function deleteLinker(context: IActionContext, item: LinkerItem | AzExtTreeItem): Promise<void> {
+    const subscription = item instanceof AzExtTreeItem ? item.subscription : createSubscriptionContext(item.subscription);
+
+    const wizardContext: IPickLinkerContext = {
+        ...context,
+        ...subscription,
+        sourceResourceUri: item?.id,
+    }
+
+    const title: string = vscode.l10n.t('Delete service connector');
```
alexweininger, line 22: `title` is only used when you create the wizard down below so I'd just eliminate the variable.

serviceconnector/src/deleteServiceConnector/deleteLinker.ts
```diff
@@ -0,0 +1,40 @@
...
+
+import { AzExtTreeItem, AzureWizard, AzureWizardExecuteStep, AzureWizardPromptStep, IActionContext, createSubscriptionContext } from "@microsoft/vscode-azext-utils";
+import * as vscode from 'vscode';
+import { LinkerItem } from "../createServiceConnector/createLinker";
+import { DeleteLinkerStep } from "./DeleteLinkerStep";
+import { IPickLinkerContext } from "./IPickLinkerContext";
+import { LinkerListStep } from "./LinkerListStep";
+
+export async function deleteLinker(context: IActionContext, item: LinkerItem | AzExtTreeItem): Promise<void> {
+    const subscription = item instanceof AzExtTreeItem ? item.subscription : createSubscriptionContext(item.subscription);
```
alexweininger, line 14: Although it's somewhat obvious, add a comment that this is for v1.5 compatibility. It'll make it easier when we go and remove v1.5 compatibility someday. That'll be a great day.

serviceconnector/src/validateServiceConnector/ValidateLinkerStep.ts
```diff
@@ -0,0 +1,20 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.md in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import { AzureWizardExecuteStep, nonNullValue } from "@microsoft/vscode-azext-utils";
+import { IPickLinkerContext } from "../deleteServiceConnector/IPickLinkerContext";
+
+export class ValidateLinkerStep extends AzureWizardExecuteStep<IPickLinkerContext> {
+    public priority: number = 200;
```
alexweininger, line 10: Add comment to explain why this is set to 200. What is it meant to come before or after?

serviceconnector/src/createServiceConnector/createLinker.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export async function createLinker(context: IActionContext, item: LinkerItem | AzExtTreeItem): Promise<void> {
+    const subscription = item instanceof AzExtTreeItem ? item.subscription : createSubscriptionContext(item.subscription);
+    const wizardContext: ICreateLinkerContext = {
+        ...context,
+        ...subscription,
+        sourceResourceUri: nonNullValue(item?.id),
+    }
+
+    const title: string = vscode.l10n.t('Create Service Connector');
```
alexweininger, line 29: Remove variable and put value directly where it's used below

serviceconnector/src/createServiceConnector/TargetServiceTypeStep.ts
```diff
@@ -0,0 +1,58 @@
...
+            //case for each database type
+        }
+        return { promptSteps };
+    }
+}
+
+
+
+
+
```
alexweininger, line 58: Nit: extra newlines here

serviceconnector/src/createServiceConnector/AuthenticationTypeStep.ts
```diff
@@ -0,0 +1,33 @@
...
+*  Licensed under the MIT License. See License.md in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+import { AuthInfoBase, KnownAuthType } from "@azure/arm-servicelinker";
+import { AzureWizardPromptStep, IAzureQuickPickItem } from "@microsoft/vscode-azext-utils";
+import * as vscode from 'vscode';
+import { ICreateLinkerContext } from "./ICreateLinkerContext";
+
+export class AuthenticationTypeStep extends AzureWizardPromptStep<ICreateLinkerContext>{
+    public async prompt(context: ICreateLinkerContext): Promise<void> {
+        const authInfoLabels: string[] = [
```
alexweininger, line 12: Why are these in an array?

I think it'd be better to just put the values directly into the `picks` array below.

serviceconnector/src/validateServiceConnector/ValidateLinkerStep.ts
motm32: In all cases (create, delete, and validate) only one execute step happens, so 200 was mostly an arbitrary number. I'll set it to a smaller number in case we add more execute steps in the future.

serviceconnector/src/createServiceConnector/ICreateLinkerContext.ts
```diff
@@ -0,0 +1,24 @@
...
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.md in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+import { AuthInfoBase, KnownClientType, LinkerResource } from "@azure/arm-servicelinker";
+import { IStorageAccountWizardContext } from "@microsoft/vscode-azext-azureutils";
+import { ISubscriptionActionContext } from "@microsoft/vscode-azext-utils";
+import { TargetServiceType } from "../../constants";
+
+
+export interface ICreateLinkerContext extends ISubscriptionActionContext, IStorageAccountWizardContext {
```
MicroFish91, line 11: Do we need to add any support for `AzureSubscription`?

@alexweininger 
I noticed in V2 migration of Container Apps that we introduced an `AzureSubscription` to the context.  Is this eventually supposed to replace `ISubscriptionActionContext`?  Do we need to do add additional support here?

serviceconnector/src/createServiceConnector/ICreateLinkerContext.ts
```diff
@@ -0,0 +1,24 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.md in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+import { AuthInfoBase, KnownClientType, LinkerResource } from "@azure/arm-servicelinker";
+import { IStorageAccountWizardContext } from "@microsoft/vscode-azext-azureutils";
+import { ISubscriptionActionContext } from "@microsoft/vscode-azext-utils";
+import { TargetServiceType } from "../../constants";
+
+
```
MicroFish91, line 10: Nit: extra new line

serviceconnector/src/createServiceConnector/TargetServiceTypeStep.ts
```diff
@@ -0,0 +1,58 @@
...
+            vscode.l10n.t('Azure Storage - Table'),
+            vscode.l10n.t('Azure Storage - File'),
+        ];
+
+        const placeHolder = vscode.l10n.t('Select Target Service Type');
+        const picks: IAzureQuickPickItem<TargetServiceType>[] = [
+            { label: targetServiceLabels[0], data: { name: "storageBlob", type: TargetServiceTypeName.storage, id: '/blobServices/default' } },
+            { label: targetServiceLabels[1], data: { name: "storageQueue", type: TargetServiceTypeName.storage, id: '/queueServices/default' } },
+            { label: targetServiceLabels[2], data: { name: "storageTable", type: TargetServiceTypeName.storage, id: '/tableServices/default' } },
+            { label: targetServiceLabels[3], data: { name: "storageFile", type: TargetServiceTypeName.storage, id: '/fileServices/default' } },
```
MicroFish91, line 26: Nit, but the other enums like `StorageAccountKind` use properties that start with a capital letter, so I think it would look better to do the same, i.e. `TargetServiceTypeName.Storage`

MicroFish91:
Also... would it make sense to add activity log support for createLinker/deleteLinker/validateLinker?

serviceconnector/src/createServiceConnector/AuthenticationTypeStep.ts
nturinski: I think I've seen Matt use this sort of design pattern as well. But I'm also in favor of just putting it in the picks array. It seems like it would be easier to have a mismatch of label/data pairs.

serviceconnector/src/createServiceConnector/ClientTypeStep.ts
```diff
@@ -0,0 +1,33 @@
...
+
+import { KnownClientType } from "@azure/arm-servicelinker";
+import { AzureWizardPromptStep, IAzureQuickPickItem } from "@microsoft/vscode-azext-utils";
+import * as vscode from 'vscode';
+import { ICreateLinkerContext } from "./ICreateLinkerContext";
+
+export class ClientTypeStep extends AzureWizardPromptStep<ICreateLinkerContext>{
+    public async prompt(context: ICreateLinkerContext): Promise<void> {
+        const placeHolder = vscode.l10n.t('Select Language/Framework');
+        const picks: IAzureQuickPickItem<KnownClientType>[] = [
```
nturinski, line 14: Is this ordering the same as the portal? It feels a little random/arbitrary.

I'd think we should either do it alphabetically or by what we expect to be the most common choices. That being said, I think it'd be cool if client extensions have a way to recommend/organize these picks.

For example, for a Function App, it would be very easy for us to see what runtime their app is running and have it be at the top.

serviceconnector/src/createServiceConnector/ICreateLinkerContext.ts
nturinski: TLDR I think that this is fine.

As far as I know, they're different things and an `ISubscriptionActionContext` can be created from an `AzureSubscription`, but not vice-versa.  `ISubscriptionActionContext` basically takes a subscription object (I can't remember what the T1 version was called, but it was the equivalent of `AzureSubscription` and combines it with an `IActionContext`.

serviceconnector/src/createServiceConnector/TargetServiceTypeStep.ts
```diff
@@ -0,0 +1,58 @@
...
+*  Licensed under the MIT License. See License.md in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import { INewStorageAccountDefaults, StorageAccountKind, StorageAccountListStep, StorageAccountPerformance, StorageAccountReplication } from "@microsoft/vscode-azext-azureutils";
+import { AzureWizardPromptStep, IAzureQuickPickItem, IWizardOptions } from "@microsoft/vscode-azext-utils";
+import * as vscode from 'vscode';
+import { TargetServiceType, TargetServiceTypeName } from "../../constants";
+import { ICreateLinkerContext } from "./ICreateLinkerContext";
+
+export class TargetServiceTypeStep extends AzureWizardPromptStep<ICreateLinkerContext>{
```
nturinski, line 12: I think that this should be `StorageServiceTypeStep` or something similar. I think ideally, `TargetServiceTypeStep` would be broader categories, like `Azure Storage`, `Databases`, `Key Vault`.

Then we would have the specific service type steps for each target service. I know right now that we are only supporting storage, but we should design it with expanding in mind.

alexweininger: I think I would still push to have `target` in the name still. I think we could remove type and make it `TargetServiceStep`, or `TargetStorageServiceStep`. Having target in there just adds a lot of much-needed context for what's the user is picking in the step.

nturinski: `StorageTargetServiceTypeStep`?

serviceconnector/src/createServiceConnector/ICreateLinkerContext.ts
alexweininger: There is quite a bit of context needed here...

Brandon and Nathan have been working over the past year (maybe two years) or so to migrate from Track 1 Azure SDKs to Track 2 Azure SDKs. One of the significant differences between T1 and T2 is how they handle authentication (switching to @azure/identity from @azure/ms-rest-nodeauth).

`AzureSubscription` is the new gold standard for representing a subscription within the extensions, and is aligned with the Track 2 authentication formats. It just so happened that it was convenient to introduced this new subscription format as part of the Azure Resources API v2 rewrite.

For `ISubscriptionActionContext`, I think we can change it out later on for a simple interface that just has an `AzureSubscription` on it. I believe this will need to happen when the extensions upgrade to use the v2 shared packages which are fully switched to Track 2 SDKs. It _should_ be a fairly easy thing.

serviceconnector/src/createServiceConnector/ClientTypeStep.ts
motm32: This is the order the portal does except they exclude PHP and Ruby. I do think that the portal has some smart detection where they will pre fill in a client type for you based on the runtime of your app which would be helpful to add.

serviceconnector/src/createServiceConnector/TargetServiceTypeStep.ts
motm32: I had originally thought TargetServiceTypeStep would branch out in the subwizard to database types (Cosmos DB, PostGreSql, etc.). 

Currently the portal lists out each service type like so: 
image
So I was following that schema but based on what you proposed Nathan we would split this into multiple steps?

motm32: Based on offline conversation will be keeping this as one step and adding groupings.

serviceconnector/src/createServiceConnector/CreateLinkerStep.ts
```diff
@@ -0,0 +1,32 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.md in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import { AzureWizardExecuteStep, nonNullValue } from "@microsoft/vscode-azext-utils";
+import { createServiceConnectorClient } from "../serviceConnectorClient";
+import { ICreateLinkerContext } from "./ICreateLinkerContext";
+
+export class CreateLinkerStep extends AzureWizardExecuteStep<ICreateLinkerContext>{
```
nturinski, line 10: nit: Generally, we call things `<noun>CreateStep`. Just makes it easier to organize/read (otherwise they would all start with Create.

serviceconnector/package.json
```diff
@@ -0,0 +1,51 @@
...
+        "prepack": "tsc -p ./",
+        "compile": "tsc -watch -p ./",
+        "lint": "eslint --ext .ts .",
+        "lint-fix": "eslint --ext .ts . --fix",
+        "pretest": "npm run build",
+        "test": "node ./out/test/runTest.js"
+    },
+    "dependencies": {
+        "@microsoft/vscode-azext-utils": "^1.2.2",
+        "@microsoft/vscode-azext-azureutils": "^1.0.0",
```
alexweininger, line 35: These should be updated to ^2

serviceconnector/src/createServiceConnector/AuthenticationTypeStep.ts
MicroFish91: Yeah I've done something similar for `ImageSourceListStep`.  My reason for doing so was also readability, because we were conditionally pushing in separate picks across multiple lines, I thought it would be easier to be able to look at the list of labels at the top and quickly get an idea of what all the prompts being used were.  Not super opinionated about it though, would have been fine to add it directly into each pick.

In this case the picks are all constructed together anyway though, so it doesn't really benefit in the same way.

serviceconnector/src/createServiceConnector/ICreateLinkerContext.ts
```diff
@@ -0,0 +1,24 @@
...
+*--------------------------------------------------------------------------------------------*/
+import { AuthInfoBase, KnownClientType, LinkerResource } from "@azure/arm-servicelinker";
+import { IStorageAccountWizardContext } from "@microsoft/vscode-azext-azureutils";
+import { ISubscriptionActionContext } from "@microsoft/vscode-azext-utils";
+import { TargetServiceType } from "../../constants";
+
+export interface ICreateLinkerContext extends ISubscriptionActionContext, IStorageAccountWizardContext {
+    //Source resource
+    sourceResourceUri?: string;
+    scope?: string; // This is only assigned when the source resource is a container app to indicate which container is being connected
```
alexweininger, line 13: If you use a multi-line comment (`/** */` kind) then VS Code will pick it up as documentation. I think you should change this to a multi-line comment to leverage this.

image

serviceconnector/src/createServiceConnector/ICreateLinkerContext.ts
```diff
@@ -0,0 +1,24 @@
...
+    runtime?: KnownClientType[];
+    clientType?: KnownClientType;
```
alexweininger, line 15: Are these properties specific to app service? If so, add a comment to add some context

serviceconnector/src/createServiceConnector/createLinker.ts
```diff
@@ -0,0 +1,53 @@
...
+import { LinkerNameStep } from "./LinkerNameStep";
+import { TargetServiceListStep } from "./TargetServiceListStep";
+
+export interface LinkerItem {
+    subscription: AzureSubscription,
+    id: string,
+}
+
+export async function createLinker(context: IActionContext & ExecuteActivityContext, item: LinkerItem | AzExtTreeItem, preSteps: AzureWizardPromptStep<ICreateLinkerContext>[] = [], runtime?: KnownClientType[]): Promise<void> {
+    const subscription = item instanceof AzExtTreeItem ? item.subscription : createSubscriptionContext(item.subscription); // For v1.5 compatibility
```
alexweininger, line 23: Should be using `isAzExtTreeItem()` here instead of `instanceof`. Classes are unique per instance of the utils package.

serviceconnector/src/createServiceConnector/createLinker.ts
```diff
@@ -0,0 +1,53 @@
...
+import { LinkerCreateStep } from "./LinkerCreateStep";
+import { LinkerNameStep } from "./LinkerNameStep";
+import { TargetServiceListStep } from "./TargetServiceListStep";
+
+export interface LinkerItem {
+    subscription: AzureSubscription,
+    id: string,
+}
+
+export async function createLinker(context: IActionContext & ExecuteActivityContext, item: LinkerItem | AzExtTreeItem, preSteps: AzureWizardPromptStep<ICreateLinkerContext>[] = [], runtime?: KnownClientType[]): Promise<void> {
```
alexweininger, line 22: `preSteps` isn't optional, but it's not clear to me (just from looking at this method) what the preSteps need to accomplish in order to work properly. Are there specific properties on `ICreateLinkerContext` the preSteps should define?

If so, a comment would go a long way here.

serviceconnector/src/deleteServiceConnector/LinkerListStep.ts
```diff
@@ -0,0 +1,29 @@
...
+
+import { uiUtils } from "@microsoft/vscode-azext-azureutils";
+import { AzureWizardPromptStep, IAzureQuickPickItem, nonNullValue } from "@microsoft/vscode-azext-utils";
+import * as vscode from 'vscode';
+import { createServiceConnectorClient } from "../serviceConnectorClient";
+import { IPickLinkerContext } from "./IPickLinkerContext";
+
+export class LinkerListStep extends AzureWizardPromptStep<IPickLinkerContext>{
+    public async prompt(context: IPickLinkerContext): Promise<void> {
+        const placeHolder = vscode.l10n.t("Select a service connector");
```
alexweininger, line 14: nit: create the value inline where it's used below.

serviceconnector/src/deleteServiceConnector/deleteLinker.ts
```diff
@@ -0,0 +1,43 @@
...
+
+import { AzExtTreeItem, AzureWizard, AzureWizardExecuteStep, AzureWizardPromptStep, DeleteConfirmationStep, ExecuteActivityContext, IActionContext, createSubscriptionContext } from "@microsoft/vscode-azext-utils";
+import * as vscode from 'vscode';
+import { LinkerItem } from "../createServiceConnector/createLinker";
+import { DeleteLinkerStep } from "./DeleteLinkerStep";
+import { IPickLinkerContext } from "./IPickLinkerContext";
+import { LinkerListStep } from "./LinkerListStep";
+
+export async function deleteLinker(context: IActionContext & ExecuteActivityContext, item: LinkerItem | AzExtTreeItem, preSteps: AzureWizardPromptStep<IPickLinkerContext>[] = []): Promise<void> {
+    const subscription = item instanceof AzExtTreeItem ? item.subscription : createSubscriptionContext(item.subscription); // For v1.5 compatibility
```
alexweininger, line 14: Use `isAzExtTreeItem()` instead of `instanceof`.

serviceconnector/src/validateServiceConnector/ValidateLinkerStep.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export class ValidateLinkerStep extends AzureWizardExecuteStep<IPickLinkerContext> {
+    public priority: number = 10;
+
+    public async execute(context: IPickLinkerContext): Promise<void> {
+        const client = await createServiceConnectorClient(context.credentials);
+        const response = await client.linker.beginValidateAndWait(nonNullValue(context.sourceResourceUri), nonNullValue(context.linkerName));
+        for (const detail of nonNullValue(response.validationDetail)) {
+            if (detail.result === "failure") {
+                throw new Error(detail.description);
```
alexweininger, line 18: Something we can deal with later: throwing an error here will tell our telemetry that this wizard actually failed to execute. But in reality the step succeeded in doing it's job, the linker is just invalid.

serviceconnector/src/validateServiceConnector/validateLinker.ts
```diff
@@ -0,0 +1,42 @@
...
+export async function validateLinker(context: IActionContext & ExecuteActivityContext, item: LinkerItem | AzExtTreeItem, preSteps: AzureWizardPromptStep<IPickLinkerContext>[] = []): Promise<void> {
+    const subscription = item instanceof AzExtTreeItem ? item.subscription : createSubscriptionContext(item.subscription);
+
+    const wizardContext: IPickLinkerContext = {
+        ...context,
+        ...subscription,
+        sourceResourceUri: item?.id,
+    }
+
+    const title: string = vscode.l10n.t('Validate service connector');
```
alexweininger, line 22: nit: create the value where it's used instead of making a variable up here

serviceconnector/test/.eslintrc.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
+    "extends": [
+        "@microsoft/eslint-config-azuretools",
+        "@microsoft/eslint-config-azuretools/test"
+    ],
+    "rules": {
+        "no-restricted-imports": "off"
```
alexweininger, line 7: Curious, where/what are we importing that we needed to turn this rule off?

alexweininger:
Looking good. Just some final touches and we should be good to move forward and publish 🎸

MicroFish91:
Are you planning to add a README added for this package?

serviceconnector/src/validateServiceConnector/ValidateLinkerStep.ts
bwateratmsft: Maybe a `UserCancelledError` with a good cancellation step would be more accurate?

serviceconnector/src/createServiceConnector/ICreateLinkerContext.ts
motm32: `runtime` is so I will add a comment about that being specific to app service but `clientType` will always be set. In the case of container apps it is set to None but is still being set.

serviceconnector/test/.eslintrc.js
motm32: I was mostly basing this off the other packages since they all turn this rule off. Based on some more investigation it seems like the base rule needs to be turned off since it can return incorrect errors. For more info here's the [link](https://typescript-eslint.io/rules/no-restricted-imports/).

serviceconnector/src/createServiceConnector/createLinker.ts
motm32: `preSteps` is referring to steps that are specific to the client extension that should to be set before any of the general service connector steps. For example in container apps we need to set the container. I had originally had this as an optional parameter but ran into issues with that since I was using nonNullValue which caused an error when testing with app service since there are no pre steps for that service. For more context here is the comment Nathan left: 

>You shouldn't assume that this is a nonNullValue. It'll cause problems in the event no prompt steps are passed through.

>You should could probably just make preSteps default to an empty array:

>`export async function createLinker(context: IActionContext & ExecuteActivityContext, item: LinkerItem | >AzExtTreeItem, preSteps: AzureWizardPromptStep<ICreateLinkerContext>[] = []): Promise<void>`

alexweininger: Oh I'm sorry, I totally missed that it was defaulting to an empty array. That makes more sense.

serviceconnector/src/serviceConnectorClient.ts
```diff
@@ -0,0 +1,10 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.md in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import { AzExtServiceClientCredentials } from "@microsoft/vscode-azext-utils";
+
+export async function createServiceConnectorClient(credentials: AzExtServiceClientCredentials) {
```
alexweininger, line 8: Ok so if we are going full LINKER then we should rename this file and method

alexweininger (approved):
🔗

PR #1486, opened 2023-06-05
auth: Fire events for session changes

bwateratmsft:
Based on https://github.com/microsoft/vscode-azureresourcegroups/pull/707/files#r1214680575. This PR is clearer if [whitespace is hidden](https://github.com/microsoft/vscode-azuretools/pull/1486/files?diff=unified&w=1).

### DIFF
--- a/auth/src/VSCodeAzureSubscriptionProvider.ts
+++ b/auth/src/VSCodeAzureSubscriptionProvider.ts
@@ -6,17 +6,48 @@
 import type { SubscriptionClient, TenantIdDescription } from '@azure/arm-subscriptions'; // Keep this as `import type` to avoid actually loading the package before necessary
 import type { TokenCredential } from '@azure/core-auth'; // Keep this as `import type` to avoid actually loading the package (at all, this one is dev-only)
 import * as vscode from 'vscode';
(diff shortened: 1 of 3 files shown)


auth/src/VSCodeAzureSubscriptionProvider.ts
```diff
@@ -6,17 +6,48 @@
...
-export class VSCodeAzureSubscriptionProvider implements AzureSubscriptionProvider {
+export class VSCodeAzureSubscriptionProvider extends vscode.Disposable implements AzureSubscriptionProvider {
+    private readonly onDidSignInEmitter = new vscode.EventEmitter<void>();
+    private lastSignInEventFired: number = 0;
+
+    private readonly onDidSignOutEmitter = new vscode.EventEmitter<void>();
+    private lastSignOutEventFired: number = 0;
+
+    public constructor() {
+        const disposable = vscode.authentication.onDidChangeSessions(async e => {
```
bwateratmsft, line 29: TODO: I don't know for sure, but I think this may fire when `getSubscriptions` is called and new sessions are being created for each tenant. If so we need to suppress firing the event while `getSubscriptions` is running.

alexweininger (approved):

PR #1489, opened 2023-06-08
utils: Fix command source telemetry

### LINKED ISSUE #1481: Update telemetry properties to accurately determine when users are using the command palette vs. tree item
Our telemetry for usage of command palette/context menu/explorer seems to be corrupted. We need to fix this.

alexweininger:
Fixes #1481 

We may want to back port this fix to v1 of utils since upgrading to v2 in some of the extensions might not happen for a while.

### DIFF
--- a/utils/src/registerCommand.ts
+++ b/utils/src/registerCommand.ts
@@ -9,6 +9,11 @@ import { callWithTelemetryAndErrorHandling } from './callWithTelemetryAndErrorHa
 import { ext } from './extensionVariables';
 import { addTreeItemValuesToMask } from './tree/addTreeItemValuesToMask';
 import { AzExtTreeItem } from './tree/AzExtTreeItem';
+import { isAzExtTreeItem } from './tree/isAzExtTreeItem';
+
+function isTreeElementBase(object?: unknown): object is types.TreeElementBase {
+    return typeof object === 'object' && object !== null && 'getTreeItem' in object;
+}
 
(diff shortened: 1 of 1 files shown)


utils/src/registerCommand.ts
```diff
@@ -19,17 +24,18 @@ export function registerCommand(commandId: string, callback: (context: types.IAc
...
                 if (args.length > 0) {
                     const firstArg: unknown = args[0];
 
-                    if (firstArg instanceof AzExtTreeItem) {
-                        context.telemetry.properties.contextValue = firstArg.contextValue;
-                    } else if (firstArg instanceof Uri) {
+                    if (firstArg instanceof Uri) {
                         context.telemetry.properties.contextValue = 'Uri';
+                    } else if (isTreeElementBase(firstArg)) {
+                        context.telemetry.properties.contextValue = (await firstArg.getTreeItem()).contextValue;
```
alexweininger, line 36: If we still want to base the command source from if the item has a contextValue, then we need to do it like this. 

I'm not opposed to changing this, but it'd mean we'd have to change a bunch of queries in the dashboard.

bwateratmsft: Can we flip the order around, i.e. do `else if (isAzExtTreeItem(firstArg))` _before_ doing this check involving `getTreeItem()`?

bwateratmsft: In fact, could we cheat a little and just see if the element has a `contextValue` property that is a `string`, and just use that if present, otherwise do the `getTreeItem()`-based test? Theoretically that could be less accurate but for the sake of telemetry I think it's good enough.

bwateratmsft (approved):

PR #1501, opened 2023-06-23
github: Init the new GitHub shared package 🐙

MicroFish91:
🐙🐙

**Example of use**:
https://github.com/microsoft/vscode-azurecontainerapps/pull/399

**Todo at some point**:
- [ ] Add unit tests


github/package.json
```diff
@@ -0,0 +1,53 @@
...
+        "eslint": "^8.34.0",
+        "eslint-plugin-import": "^2.22.1",
+        "glob": "^7.1.6",
+        "mocha": "^9.1.3",
+        "mocha-junit-reporter": "^2.0.2",
+        "mocha-multi-reporters": "^1.1.7",
+        "typescript": "^5.1.3"
+    },
+    "dependencies": {
+        "@microsoft/vscode-azext-utils": "^1.2.0",
```
alexweininger, line 50: Lets update to ^2.0.0 if possible

github/src/utils/getGitHubAccessToken.ts
alexweininger, line 1: We will want to add the same workaround for CodeSpaces as I did in SWA here https://github.com/microsoft/vscode-azurestaticwebapps/pull/841

github/src/wizard/GitHubBranchListStep.ts
```diff
@@ -0,0 +1,59 @@
...
+import { Branches, GetBranchesParams, getBranches } from "../wrappers/getBranches";
+
+export class GitHubBranchListStep extends AzureWizardPromptStep<GitHubContext> {
+    private picks: IAzureQuickPickItem<string>[];
+
+    public async prompt(context: GitHubContext): Promise<void> {
+        // We always want fresh picks before prompting in case the user has pressed the back button
+        this.picks = [];
+
+        const placeHolder: string = l10n.t('Select a branch');
```
alexweininger, line 19: I don't love seeing `l10n.t` everywhere. I think I'd prefer if we made an alias function named `localize` that called `l10n.t`. I think `localize` looks better. Not something we have to change here and now though.

github/src/tree/StepItem.ts
```diff
@@ -0,0 +1,39 @@
...
+
+    constructor(
+        readonly parentResourceId: string,
+        readonly contextValueExtensionPrefix: string,
+        readonly step: JobStep) { }
+
+    id: string = `${this.parentResourceId}/steps/${this.step.number}`;
+    label: string = this.step.name;
+
+    contextValue: string = `${this.contextValueExtensionPrefix}${StepItem.contextValueSuffix}`;
```
alexweininger, line 23: If context value isn't used outside of `getTreeItem()`, then create this string inline in that method.

github/src/utils/actionUtils.ts
```diff
@@ -0,0 +1,110 @@
...
+export enum Status {
+    Queued = 'queued',
+    InProgress = 'in_progress',
+    Completed = 'completed'
+}
+
+// Description
+export function getJobBasedDescription(data: Job | JobStep): string {
+    if (data.conclusion !== null) {
+        return l10n.t('{0} {1}', convertConclusionToVerb(<Conclusion>nonNullProp(data, 'conclusion')), dayjs(data.completed_at).fromNow());
```
nturinski, line 33: convertConclusionToVerb is already localized below. I'm not sure how embedded localized strings work.

github/src/utils/gitHubUrlParse.ts
```diff
@@ -0,0 +1,23 @@
...
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+export interface ParsedGitHubUrl {
+    urlReference?: string;
+    ownerOrOrganization?: string;
+    repositoryName?: string;
+}
+
+export function gitHubUrlParse(url?: string): ParsedGitHubUrl {
```
nturinski, line 12: Is there any reason to make `url` an optional argument? I feel like there's no point in calling it if you don't have a url to begin with.

Then `ParsedGitHubUrl` could have every property be required. It feels like it'd be annoying to have to type check that any time that you wanted to use `gitHubUrlParse`

github/src/wizard/GitHubBranchListStep.ts
nturinski: image

github/src/wizard/GitHubRepositoryListStep.ts
```diff
@@ -0,0 +1,74 @@
...
+            // Doesn't matter if context.gitHubOrg is undefined; that just means we will end up calling 'getRepositoriesByUser' instead
+            org: context.gitHubOrg ?? '',
+            type: 'all',
+            per_page: perPage,
+            page
+        };
+
+        this.picks.push(
+            ...(context.gitHubOrg ?
+                (await getRepositoriesByOrg(context, orgRepoParams)).map((repo) => { return { label: repo.name, data: { owner: repo.owner.login, repo: repo.name, url: repo.html_url } } }) :
```
nturinski, line 65: minor nit: Kind of bothers me that this is copy pasta. Maybe we can just do something like:

`const repos = context.gitHubOrg ? ...`

And then do:

`this.picks.push(...repos.map(repo => ))...`

github/src/utils/actionUtils.ts
MicroFish91: Yeah, this was a product of the original I pulled it from, but I kind of liked how it accidentally made the returned string a lot more readable.  Would you prefer I change this?

github/src/wizard/GitHubRepositoryListStep.ts
MicroFish91: What do you mean by copy pasta? And sure, I'll make that change

github/package.json
MicroFish91: Sure, I can bump the version before we merge.  Container Apps is still on 1.2.0 so using this on 2.0.0 was causing some issues and I was trying to test it quickly.

github/src/wizard/GitHubRepositoryListStep.ts
nturinski: Copy pasta = copy and paste = `.map((repo) => { return { label: repo.name, data: { owner: repo.owner.login, repo: repo.name, url: repo.html_url } } })` being used twice.

github/src/utils/actionUtils.ts
nturinski: Yeah, I say we just change it if it doesn't make any sense.

github/src/tree/ActionsItemBase.ts
```diff
@@ -0,0 +1,106 @@
...
+                    id: `${this.parentId}/connectToGitHub`,
+                    label: l10n.t('Connect to a GitHub Repository...'),
+                    commandId: command?.commandId,
+                    commandArgs: command?.commandArgs
+                })
+            ];
+        }
+    }
+
+    getConnectToGitHubCommand?(): Promise<ConnectToGitHubCommand>;
```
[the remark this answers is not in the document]
MicroFish91: It doesn't need to be, the generic item just won't issue a command when clicked.  I'm happy to make it required if we want to always enforce a command to be hooked up though, I think that might be a reasonable expectation

github/src/wizard/GitHubRepositoryListStep.ts
MicroFish91: Oooooh, I see what you mean now.

MicroFish91: I think I remember why I had it that way, the type inference when you do the map with the interchangeable repo sources seems to stump Typescript into typecasting to an `any`.  I don't know the actual array element types to manually cast it to, so it was a lot easier to just keep the maps strict like that to force the proper type inference

MicroFish91: Added a comment to better explain the old way of doing it.  Let me know if you prefer a different way still

MicroFish91:
Should be ready for review again - made suggested review changes and tested on Container Apps! ✅

alexweininger:
Since this package uses utils it needs to register the utils extension variables like Megan did for the service connector package [here](https://github.com/microsoft/vscode-azuretools/blob/main/serviceconnector/src/tree/extensionVariables.ts).

image

github/tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
+    "extends": "../tsconfig.base.json",
+    "compilerOptions": {
+        "outDir": "out",
+        "rootDir": ".",
```
alexweininger, line 5: We should turn on the declaration option here unless we have a good reason not to. That way we can remove `index.d.ts` and we don't have to maintain the types later on.

MicroFish91: Noticed you made carried these changes over already, thanks for doing that!

alexweininger (approved):

github/src/utils/getGitHubAccessToken.ts
```diff
@@ -0,0 +1,26 @@
...
+        const token = (await authentication.getSession(gitHubAuthProviderId, gitHubScopes, { createIfNone: true })).accessToken;
+        // Workaround for VS Code returning a different token when connected to a CodeSpace in a browser
+        // see https://github.com/microsoft/vscode-azurestaticwebapps/issues/827#issuecomment-1597881084 for details
+        if (token.startsWith('ghu_')) {
+            // Request a fake scope to force VS Code to give us a token of the right type
+            return (await authentication.getSession(gitHubAuthProviderId, [...gitHubScopes, 'x-AzToolsScope'], { createIfNone: true })).accessToken;
+        }
+        return token;
+    } catch (error) {
+        if (parseError(error).message === 'User did not consent to login.') {
```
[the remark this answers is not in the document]
nturinski: This is a good point. Is there an error type that we can use instead?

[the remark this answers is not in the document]
MicroFish91: Yeah good call, let me get back to you on that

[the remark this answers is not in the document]
MicroFish91: Not too many options, no type is parsed that we can use:

image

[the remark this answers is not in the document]
bwateratmsft: You could make it slightly more robust by changing to a regex covering a critical part of the string--e.g.,

```typescript
// The error message is "User did not consent to login"
if (/did not consent/i.test(parseError(error).message)) {
    // ...
}
```

PR #1516, opened 2023-07-12
utils: Export new `delay` and `isDeepEqual` utility functions

MicroFish91:
Technically the delay util already existed, but we weren't exporting it from utils yet 🙂

### DIFF
--- a/utils/src/utils/isDeepEqual.ts
+++ b/utils/src/utils/isDeepEqual.ts
@@ -1,0 +1,30 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+export function isDeepEqual(obj1: object, obj2: object): boolean {
+    const objKeys1: string[] = Object.keys(obj1);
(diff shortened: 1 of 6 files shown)


utils/src/utils/isDeepEqual.ts
```diff
@@ -0,0 +1,30 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+export function isDeepEqual(obj1: object, obj2: object): boolean {
```
alexweininger, line 6: Currently this is only used by Container Apps to compare CA objects to check if there are unsaved changes.

There are many nuances to a correct deep equal function, and I'm worried we'll try to use this function somewhere else and it'll cause bugs. Could we instead either keep this specific to Container Apps and name is as such, OR use a library like https://github.com/chaijs/deep-eql that is a near perfect deep equal implementation? Ex: current implementation breaks if symbols are present as keys on the object.

Also, this is something that needs tests so we know what cases it does work for and what cases it doesn't work for.

bwateratmsft: Agreed. I think we should expose it but I like the idea of using an underlying library.

PR #1525, opened 2023-07-18
utils: add AzFsExtraFile.appendFile

nturinski:


utils/src/utils/AzExtFsExtra.ts
```diff
@@ -73,6 +73,12 @@ export namespace AzExtFsExtra {
         await workspace.fs.writeFile(uri, Buffer.from(contents));
     }
 
+    export async function appendFile(resource: Uri | string, contents: string): Promise<void> {
```
alexweininger, line 76: maybe a better name is `appendToFile`?

nturinski: Mimicing this https://www.geeksforgeeks.org/node-js-fs-appendfile-function/ so we can have a 1:1 conversion.

alexweininger (approved):
Do you need to add the method to the declaration file?

nturinski:
Ah crap, I was thinking of the exports file. Yeah, I need to do that (and will do that with the bump)

utils/src/utils/AzExtFsExtra.ts
```diff
@@ -73,6 +73,12 @@ export namespace AzExtFsExtra {
...
+        const uri = convertToUri(resource);
+        const existingContent = await AzExtFsExtra.readFile(uri);
+        await AzExtFsExtra.writeFile(uri, existingContent + '\r\n\r\n' + contents);
```
bwateratmsft, line 79: Do we need to add the newlines though? Seems like separator should be the decision of the caller.

nturinski: We don't have to, I was emulating what `fs.appendFile` does though. We could have an option for "separator" though,  I'd like to default to `\r\n\r\n` since that is the default behavior of `fs.appendFile`.

bwateratmsft: I think that's a good middle ground.

PR #1450, opened 2023-04-11
appservice: Convert appservice requests to use pipelines ServiceClient

nturinski:
Relies on https://github.com/microsoft/vscode-azuretools/pull/1449/

I tested this on App Service and everything seems to be working. 

Well, except the GitHub stuff-- I purposely commented that all out intending that we replace all of that implementation with the shared GitHub package. I'm not sure if it makes more sense for the `appservice` package to adopt the `github` package or if the extensions themselves should directly consume it.

PR #1537, opened 2023-07-20
appservice: Fix site files by avoiding kudu redirect

alexweininger:
Browsing site files was broken during the migration to Track 2 SDK in https://github.com/microsoft/vscode-azuretools/pull/1450. Detailed explanation below.

Related extension issue: https://github.com/microsoft/vscode-azureappservice/issues/2544

### Problem

The App Service VFS API endpoints are in the format `/api/vfs/${path}?api-version=2022=03-01` where `path` is a path to a file or folder. If you make a request for a folder, but omit the trailing slash the API will correct this by redirecting the request and adding the slash onto the redirect location. 

So `/api/vfs/thisIsAFolder?api-version=2022=03-01` will be redirected to `/api/vfs/thisIsAFolder/?api-version=2022=03-01` (see the added slash).

Sounds great right? Wrong! The redirection causes a 401 error. We recently migrated to the Track 2 client. The Track 2 client [redirection policy purposefully removes the authentication header](https://github.com/timovv/azure-sdk-for-js/blob/ebe34430af273a9e5535534fc3875abae95b5c5e/sdk/core/core-rest-pipeline/src/policies/redirectPolicy.ts#L73) from the redirected request. This is a security fix and not a bug [according to the SDK team](https://github.com/Azure/azure-sdk-for-js/issues/20957).

I haven't checked, but I'm assuming Track 1 client didn't have this security patch so it just worked even with the redirections.

### Solution

Technically we could circumvent the default redirect policy, however A. it's a security risk, and B. a lot of work.

Instead, we now form requests to the App Service VFS API such that no redirect is needed. We can do so by ensuring we include or exclude trailing slashes based on if the path points to a folder or file. I also found that we were including an extra leading slash which also resulted in a redirect.

To do this, I refactored how the site files and site file tree items worked. Instead of tracking file paths to make requests, I'm just storing the `href` property that the API returns which in my testing is always correct. This makes it really easy for us and means I was able to delete some code. I don't know why we didn't use this property originally, maybe it's new.

I tested this extensively in both App Service and Functions extensions. This impacts the Logs feature too which I also tested. Afaik Logs and Files are the only thing these changes impact.

It'd be awesome if we could mock the Kudu APIs or maybe just test against a live App Service resource. I'm super open to brainstorming ways we can refactor this stuff to make it more testable to prevent this sort of bug in the future.

### DIFF
--- a/appservice/src/siteFiles.ts
+++ b/appservice/src/siteFiles.ts
@@ -8,7 +8,6 @@ import { RestError, createPipelineRequest } from '@azure/core-rest-pipeline';
 import { AzExtPipelineResponse, createGenericClient } from '@microsoft/vscode-azext-azureutils';
 import { IActionContext, IParsedError, parseError } from '@microsoft/vscode-azext-utils';
 import * as retry from 'p-retry';
-import * as path from 'path';
 import { ParsedSite } from './SiteClient';
 
 export interface ISiteFile {
@@ -20,12 +19,20 @@ export interface ISiteFileMetadata {
     mime: string;
     name: string;
     path: string;
+    href: string;
 }
 
-export async function getFile(context: IActionContext, site: ParsedSite, filePath: string): Promise<ISiteFile> {
+/**
+ * @param path - Do not include leading slash. Include trailing slash if path represents a folder.
+ */
+export function createSiteFilesHref(site: ParsedSite, path: string): string {
+    return `${site.kuduUrl}/api/vfs/${path}?api-version=2022=03-01`
+}
+
+export async function getFile(context: IActionContext, site: ParsedSite, href: string): Promise<ISiteFile> {
     let response: AzExtPipelineResponse;
     try {
-        response = await getFsResponse(context, site, filePath);
+        response = await getFsResponse(context, site, href);
     } catch (error) {
         if (error instanceof RestError && error.code === 'PARSE_ERROR' && error.response?.status === 200) {
             // Some files incorrectly list the content-type as json and fail to parse, but we always just want the text itself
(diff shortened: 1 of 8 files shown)


bwateratmsft (dismissed):

appservice/src/siteFiles.ts
```diff
@@ -20,12 +19,20 @@ export interface ISiteFileMetadata {
...
     path: string;
+    href: string;
 }
 
-export async function getFile(context: IActionContext, site: ParsedSite, filePath: string): Promise<ISiteFile> {
+/**
+ * @param path - Do not include leading slash. Include trailing slash if path represents a folder.
+ */
+export function createSiteFilesHref(site: ParsedSite, path: string): string {
+    return `${site.kuduUrl}/api/vfs/${path}?api-version=2022=03-01`
```
nturinski, line 29: I'm pretty sure that Kudu doesn't have an api-version. That was my mistake when migrating over Kudu calls (because I was thinking it was an ARM call)

nturinski:
> I tested this extensively in both App Service and Functions extensions. This impacts the Logs feature too which I also tested. Afaik Logs and Files are the only thing these changes impact.

This should be correct.

appservice/src/siteFiles.ts
```diff
@@ -80,7 +82,7 @@ async function getFsResponse(context: IActionContext, site: ParsedSite, filePath
                     try {
                         return await client.sendRequest(createPipelineRequest({
                             method: 'GET',
-                            url: `${site.id}/hostruntime/admin/vfs/${filePath}/?api-version=2022-03-01`
+                            url: href,
```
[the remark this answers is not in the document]
alexweininger: Ok, I'm assigning the href property to url when creating the folder/file tree items. So now it's known as url everywhere.

appservice/src/siteFiles.ts
alexweininger: Removed it.

nturinski (approved):
🤩

appservice/src/siteFiles.ts
```diff
@@ -47,24 +54,19 @@ export async function listFiles(context: IActionContext, site: ParsedSite, fileP
...
 
 /**
  * Kudu APIs don't work for Linux consumption function apps and ARM APIs don't seem to work for web apps. We'll just have to use both
  */
-async function getFsResponse(context: IActionContext, site: ParsedSite, filePath: string): Promise<AzExtPipelineResponse> {
+async function getFsResponse(context: IActionContext, site: ParsedSite, url: string): Promise<AzExtPipelineResponse> {
     try {
         if (site.isFunctionApp) {
-            const linuxHome: string = '/home';
-            if (site.isLinux && !filePath.startsWith(linuxHome)) {
```
nturinski, line 64: Oh yeah, are you still considering this special case?

alexweininger: We aren't using the path anymore so this isn't needed. I'm assuming the href returned by the API is always correct. Do you know how to test this case?

alexweininger: I verified it works with a Linux Function App. Not sure how else to test it.

nturinski: That works for me. I honestly don't remember the full context of why we had this in the first place.

PR #1539, opened 2023-07-20
serviceconnector: Fix ServiceConnector issues with UI

motm32:
When testing the UI with the App Service extension I ran into some issues since the item being passed is the ServiceConnectorGroupTreeItem and not the SiteTreeItem. This caused create/delete/validate to fail since the id was incorrect. After some discussion it seemed like the best thing to do was to just pass in the subscription and id instead of the AzExtTreeItem. 

Refer to this [PR](https://github.com/microsoft/vscode-azureappservice/pull/2511) to see how it is hooked up with the extension.

### DIFF
--- a/serviceconnector/src/deleteLinker/deleteLinker.ts
+++ b/serviceconnector/src/deleteLinker/deleteLinker.ts
@@ -27,9 +27,7 @@ export async function deleteLinker(context: IActionContext & ExecuteActivityCont
         new DeleteConfirmationStep(confirmMessage)
     ];
 
-    if (serviceConnectorName) {
-        wizardContext.linkerName = serviceConnectorName;
-    }
+    wizardContext.linkerName = serviceConnectorName;
 
     preSteps.forEach(step => promptSteps.unshift(step));
 
--- a/serviceconnector/src/deleteLinker/deleteLinker.ts
+++ b/serviceconnector/src/deleteLinker/deleteLinker.ts
@@ -3,20 +3,21 @@
 *  Licensed under the MIT License. See License.md in the project root for license information.
 *--------------------------------------------------------------------------------------------*/
 
-import { AzExtTreeItem, AzureWizard, AzureWizardExecuteStep, AzureWizardPromptStep, DeleteConfirmationStep, ExecuteActivityContext, IActionContext, createSubscriptionContext, isAzExtTreeItem } from "@microsoft/vscode-azext-utils";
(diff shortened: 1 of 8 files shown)


alexweininger (dismissed):
This is a great improvement. 👍

serviceconnector/src/deleteLinker/deleteLinker.ts
```diff
@@ -3,20 +3,21 @@
...
 import * as vscode from 'vscode';
-import { LinkerItem } from "../createLinker/createLinker";
+import { isAzureSubscription } from "../createLinker/createLinker";
 import { DeleteLinkerStep } from "./DeleteLinkerStep";
 import { IPickLinkerContext } from "./IPickLinkerContext";
 import { LinkerListStep } from "./LinkerListStep";
 
-export async function deleteLinker(context: IActionContext & ExecuteActivityContext, item: LinkerItem | AzExtTreeItem, preSteps: AzureWizardPromptStep<IPickLinkerContext>[] = []): Promise<void> {
-    const subscription = isAzExtTreeItem(item) ? item.subscription : createSubscriptionContext(item.subscription); // For v1.5 compatibility
+export async function deleteLinker(context: IActionContext & ExecuteActivityContext, id: string, subscription: ISubscriptionContext | AzureSubscription, serviceConnectorName?: string, preSteps: AzureWizardPromptStep<IPickLinkerContext>[] = []): Promise<void> {
```
nturinski, line 14: Is there any reason for serviceConnectorName to be optional? It's a bit weird, but it seems like this would have a no-op if for some reason it was undefined. If that happens, we should probably throw an error otherwise the user wouldn't know that nothing happened.

nturinski (commented):
I would change the ServiceConnector items to also have a "resourceId" property that gets passed into their constructor. It'll make the code for the client extensions a bit easier to use create/deleteLinker.

serviceconnector/src/deleteLinker/deleteLinker.ts
motm32: It is optional because when the command palette is used you have to choose a specific service connection (LinkerListStep) and in this case serviceConnectorName wouldn't be passed.

nturinski: I see.  Hmm, though you could get rid of the if conditional. If serviceConnectorName is undefined, you'd just be setting linkerName to undefined, which it probably already was that.

bwateratmsft (approved):

alexweininger (approved):

PR #1540, opened 2023-07-21
auth: Ignore `.default` if it is passed as a scope

### LINKED ISSUE #1538: [@microsoft/vscode-azext-azureauth] AzureSubscription.credential.getToken() returns an error with '.default' scope
When calling ,
 ```typescript
AzureSubscription.credential.getToken('.default')
```
or 
 ```typescript
AzureSubscription.credential.getToken(['.default'])
```
I am getting this error:
_You are not signed in to an Azure account. Please sign in._

However, this doesn't happen if you pass an empty array `[]` inside the `getToken` function.

bwateratmsft:
Fixes #1538

### DIFF
--- a/auth/package-lock.json
+++ b/auth/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "@microsoft/vscode-azext-azureauth",
-    "version": "1.1.0",
+    "version": "1.1.1",
     "lockfileVersion": 3,
     "requires": true,
     "packages": {
         "": {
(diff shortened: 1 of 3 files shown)


nturinski (approved):
Are there any concerns about other scopes not being formatted properly? Or is it only `.default` where it screws up?

bwateratmsft:
> Are there any concerns about other scopes not being formatted properly? Or is it only `.default` where it screws up?

We'll find out when we find out. :smile:

PR #1548, opened 2023-08-03
appservice: Add streamResponseStatusCodes

nturinski:
Log streaming can't use basic auth credentials anymore, so attaching access token on header.
Also, the new pipelines requires the `streamResponseStatusCodes` otherwise it'll just hang whenever we do a stream.

### DIFF
--- a/appservice/package-lock.json
+++ b/appservice/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "@microsoft/vscode-azext-azureappservice",
-    "version": "2.2.2",
+    "version": "2.2.3",
     "lockfileVersion": 2,
     "requires": true,
     "packages": {
         "": {
             "name": "@microsoft/vscode-azext-azureappservice",
-            "version": "2.2.2",
+            "version": "2.2.3",
             "license": "MIT",
             "dependencies": {
                 "@azure/abort-controller": "^1.0.4",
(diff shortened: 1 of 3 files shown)


alexweininger (approved):

JonathanNathanson:
How can I find out when this fix will be released in to the current version of Azure Tools extension? I've been suffering from "Unable to connect to log stream" for my App Services for a couple of months now, but I can see this PR potentially fixes that issue. I can't however work out when this PR will be released into the version of the extension available from the VS Code marketplace.

Thanks

nturinski:
@JonathanNathanson, thanks for bringing this to our attention. We're actually releasing this today among some other bug fixes!

JonathanNathanson:
Hi Nturnski 

I haven't seen the update to the extension released - where can I find information about releases and what's included in them? The tags on this repo don't seem to be updated recently:

https://github.com/microsoft/vscode-node-azure-pack

> @JonathanNathanson, thanks for bringing this to our attention. We're actually releasing this today among some other bug fixes!

jinglouMSFT:
@JonathanNathanson I believe the fixes were released in the AppService extension last Wed, 10/11.  If you didn't have the update automagically, try to download the AppService extension to see if the fixes resolved the issues you've experienced.  Thanks.

JonathanNathanson:
Yep I was looking in the wrong place, the App Service extension has been updated and the issue I had has been fixed in that update.  Many thanks for all your hard work 😄

PR #1557, opened 2023-08-14
serviceconnector: Add children to validate activity log

motm32:
Here is how the validation success state looks: 
image

Here is how the error state looks: 
image

### DIFF
--- a/serviceconnector/src/deleteLinker/IPickLinkerContext.ts
+++ b/serviceconnector/src/deleteLinker/IPickLinkerContext.ts
@@ -3,9 +3,14 @@
 *  Licensed under the MIT License. See License.md in the project root for license information.
 *--------------------------------------------------------------------------------------------*/
 
+import { ValidationResultItem } from "@azure/arm-servicelinker";
+import { ExecuteActivityContext } from "@microsoft/vscode-azext-utils";
 import { ICreateLinkerContext } from "../createLinker/ICreateLinkerContext";
 
-export interface IPickLinkerContext extends ICreateLinkerContext {
+export interface IPickLinkerContext extends ICreateLinkerContext, ExecuteActivityContext {
     linkerName?: string;
+
+    //Validation
+    validationResult?: ValidationResultItem[]
 }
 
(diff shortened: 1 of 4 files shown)


serviceconnector/src/validateLinker/ValidateLinkerStep.ts
```diff
@@ -3,20 +3,34 @@
...
+
         for (const detail of nonNullValue(response.validationDetail)) {
-            if (detail.result === KnownValidationResultStatus.Failure) {
+            if (detail.result === "failure") {
+                context.activityChildren.push(new GenericTreeItem(undefined, {
+                    contextValue: `validateResult-${detail.name}`,
+                    label: nonNullValue(detail.name),
+                    iconPath: new ThemeIcon('error', new ThemeColor('testing.iconFailed'))
+                }));
                 throw new Error(detail.description);
```
alexweininger, line 27: Although one validation failed, I think it's best to still show all other failed/passed validations as activity children. Maybe throw the error after all the children are added?

serviceconnector/src/deleteLinker/IPickLinkerContext.ts
```diff
@@ -3,9 +3,14 @@
...
+import { ValidationResultItem } from "@azure/arm-servicelinker";
+import { ExecuteActivityContext } from "@microsoft/vscode-azext-utils";
 import { ICreateLinkerContext } from "../createLinker/ICreateLinkerContext";
 
-export interface IPickLinkerContext extends ICreateLinkerContext {
+export interface IPickLinkerContext extends ICreateLinkerContext, ExecuteActivityContext {
     linkerName?: string;
+
+    //Validation
+    validationResult?: ValidationResultItem[]
```
alexweininger, line 14: Where is this set or used?

serviceconnector/src/validateLinker/ValidateLinkerStep.ts
motm32: Actually tested this out and for some reason wen there is an error the validation result only contains the failed check. So moving the error after all the children doesn't change anything.

alexweininger: I'm not exactly sure why that happens. But that's fine

alexweininger (approved):

PR #1570, opened 2023-08-24
serviceconnector: Fix icon path for webpack config

motm32:
Need to add webpack config in order for the icons to show up correctly in the extension, so fixing the file path in order to add that.

### DIFF
--- a/serviceconnector/src/tree/IconPath.ts
+++ b/serviceconnector/src/tree/IconPath.ts
@@ -14,6 +14,6 @@ export function getIconPath(iconName: string): TreeItemIconPath {
 function getResourcesPath(): string {
     return ext.ignoreBundle ?
         path.join(__dirname, '..', '..', '..', 'resources') :
-        path.join(__dirname, 'node_modules', '@microsoft', 'vscode-azext-azureutils', 'resources');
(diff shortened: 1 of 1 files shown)


serviceconnector/src/tree/IconPath.ts
```diff
@@ -14,6 +14,6 @@ export function getIconPath(iconName: string): TreeItemIconPath {
 function getResourcesPath(): string {
     return ext.ignoreBundle ?
         path.join(__dirname, '..', '..', '..', 'resources') :
-        path.join(__dirname, 'node_modules', '@microsoft', 'vscode-azext-azureutils', 'resources');
+        path.join(__dirname, 'node_modules', '@microsoft', 'vscode-azext-serviceconnector', 'resources');
```
bwateratmsft, line 17: Shouldn't this be an additional line, not a replacement?

motm32: This getResourcesPath is specifically for service connector resources. The azure package has a seperate IconPath.ts file that handles its resources, so AFAIK it shouldn't need the additional line.

bwateratmsft (approved):

alexweininger (approved):
Tested locally and it works 🥳

PR #1575, opened 2023-08-29
appservice: Don't use .gitignore for ignoring zip files

alexweininger:
https://github.com/microsoft/vscode-azurefunctions/issues/3817

Regression caused by https://github.com/microsoft/vscode-azuretools/pull/1436. We aren't supposed to ignore files specified in the .gitignore. Only the .funcignore should be used.


motm32 (approved):

nturinski:
Do you know what we use in the case of App Service?

PR #1577, opened 2023-09-01
serviceconnector: Add "Create connection" and "Delete connection" activity children

motm32:
Add activity children that displays the name of the application settings added: 
image


alexweininger (approved):
We can do "App setting" like the portal does instead of "application setting"
image

motm32:
> We can do "App setting" like the portal does instead of "application setting" image

I don't mind that just named it application settings since thats what the tree node is called 
image

alexweininger:
> > We can do "App setting" like the portal does instead of "application setting" image
> 
> I don't mind that just named it application settings since thats what the tree node is called image

Ya ignore me. Keep it as it is.

PR #1579, opened 2023-09-07
Use Node 18.15

alexweininger:
VS Code has updated to Node 18.15: https://code.visualstudio.com/updates/v1_82#_electron-25-update


bwateratmsft:
We'll want to make a note on subsequent [releases](https://github.com/microsoft/vscode-azuretools/releases) that the new package versions require a minimum VSCode version of 1.82

alexweininger:
I believe the build failed because no packages were changed.

bwateratmsft:
> I believe the build failed because no packages were changed.

That's correct. We still need to fix that...

bwateratmsft (approved):

PR #1580, opened 2023-09-08
serviceconnector: Add user agent

motm32:
Using the `createAzureSubscriptionClient` so a user agent is added.


nturinski (approved):
Don't forget to bump the version 👊

PR #1581, opened 2023-09-08
serviceconnector: Add database subtypes

motm32:
Add in database subtypes and filter the database accounts based on the chosen type. Here is how the new list looks: 
image

### DIFF
--- a/serviceconnector/src/createLinker/TargetServiceListStep.ts
+++ b/serviceconnector/src/createLinker/TargetServiceListStep.ts
@@ -15,16 +15,16 @@ export class TargetServiceListStep extends AzureWizardPromptStep<ICreateLinkerCo
     public async prompt(context: ICreateLinkerContext): Promise<void> {
         const placeHolder = vscode.l10n.t('Select Target Service Type');
         const picks: IAzureQuickPickItem<TargetServiceType>[] = [
-            { label: vscode.l10n.t('Blob'), data: { name: "storageBlob", type: TargetServiceTypeName.Storage, id: '/blobServices/default' }, group: TargetServiceTypeName.Storage },
-            { label: vscode.l10n.t('Queue'), data: { name: "storageQueue", type: TargetServiceTypeName.Storage, id: '/queueServices/default' }, group: TargetServiceTypeName.Storage },
-            { label: vscode.l10n.t('Table'), data: { name: "storageTable", type: TargetServiceTypeName.Storage, id: '/tableServices/default' }, group: TargetServiceTypeName.Storage },
-            { label: vscode.l10n.t('File'), data: { name: "storageFile", type: TargetServiceTypeName.Storage, id: '/fileServices/default' }, group: TargetServiceTypeName.Storage },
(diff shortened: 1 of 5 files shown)


serviceconnector/src/createLinker/TargetServiceListStep.ts
```diff
@@ -43,7 +47,19 @@ export class TargetServiceListStep extends AzureWizardPromptStep<ICreateLinkerCo
             case TargetServiceTypeName.Storage:
                 promptSteps.push(new StorageAccountListStep(storageAccountCreateOptions));
                 break;
-            case TargetServiceTypeName.CosmosDB:
+            case TargetServiceTypeName.Cassandra:
```
nturinski, line 50: It might make more sense for the switch case to check the group. That shouldn't change anything for Storage (since all of the groups are also `TargetServiceTypeName.Storage`).

But it'd make this a lot cleaner since all are part of `TargetServiceTypeName.CosmosDB` and are all using the same promptStep as far as I can tell.

motm32: Very true I will do that!

serviceconnector/src/createLinker/TargetServiceListStep.ts
```diff
@@ -15,16 +15,16 @@ export class TargetServiceListStep extends AzureWizardPromptStep<ICreateLinkerCo
...
-            { label: vscode.l10n.t('Cassandra'), data: { name: "cassandra", type: TargetServiceTypeName.Cassandra, id: '/databases' }, group: TargetServiceTypeName.CosmosDB },
-            { label: vscode.l10n.t('Apache Gremlin'), data: { name: "gremlin", type: TargetServiceTypeName.Gremlin, id: '/databases' }, group: TargetServiceTypeName.CosmosDB },
-            { label: vscode.l10n.t('NoSQL'), data: { name: "sql", type: TargetServiceTypeName.NoSQL, id: '/databases' }, group: TargetServiceTypeName.CosmosDB },
-            { label: vscode.l10n.t('Table'), data: { name: "table", type: TargetServiceTypeName.Table, id: '/databases' }, group: TargetServiceTypeName.CosmosDB },
-            { label: vscode.l10n.t('Key Vault'), data: { name: "keyVault", type: TargetServiceTypeName.KeyVault, id: '/keyVault' }, group: TargetServiceTypeName.KeyVault },
+            { label: vscode.l10n.t('Blob'), data: { name: "storageBlob", type: TargetServiceTypeName.Storage, id: '/blobServices/default', group: TargetServiceTypeName.Storage }, group: TargetServiceTypeName.Storage },
```
nturinski, line 18: Heh, this is a little bit goofy.  It definitely works, but the repeat code hurts my eyes 😭

Could I suggest that rather than duplicating the group property in `data`, that you instead save the result of the quickPick item instead of just the `data`?

Something like this

```
const result = await context.ui.showQuickPick(picks, { placeHolder, enableGrouping: true };
context.targetServiceType = result.data;

switch (result.group) {
...
```

nturinski (approved):

PR #1583, opened 2023-09-14
utils: Treat VSCode's `CancellationError` as a user-cancelled error

bwateratmsft:


alexweininger (approved):
Looks good. Curious, is this change to fix behavior or for telemetry or something else?

bwateratmsft:
> Looks good. Curious, is this change to fix behavior or for telemetry or something else?

@alexweininger we're doing a registry extensibility package that uses `@types/vscode` but _not_ `@microsoft/vscode-azext-utils`. It throws `vscode.CancellationError` since it doesn't have access to `UserCancelledError`. But the error was being caught in a `callWithTelemetryAndErrorHandling` block, and we wanted to have it treated the same as `UserCancelledError`.

PR #1585, opened 2023-09-14
auth: Check if tenant is signed in before listing subscriptions

alexweininger:
Fixes an issue where if you have tenants that aren't authenticated it currently throws an error and returns no subscriptions.

Instead of just try/catching and ignoring the `NotSignedIn` error, I added a way to check before we throw the error at all. I did this because I have some work that builds on top of this which allows users to select an unauthenticated tenant and sign in to that tenant.

### DIFF
--- a/auth/package-lock.json
+++ b/auth/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "@microsoft/vscode-azext-azureauth",
-    "version": "1.1.2",
+    "version": "1.1.3",
     "lockfileVersion": 3,
     "requires": true,
     "packages": {
         "": {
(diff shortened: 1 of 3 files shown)


bwateratmsft (commented):
Can you also update the README, e.g. line https://github.com/microsoft/vscode-azuretools/blob/main/auth/README.md?plain=1#L35

alexweininger:
> Can you also update the README, e.g. line https://github.com/microsoft/vscode-azuretools/blob/main/auth/README.md?plain=1#L35

I didn't alter the public api. At least not yet. I will do that in a future PR when I have the pick tenant logic finished.

bwateratmsft (approved):

PR #1590, opened 2023-09-19
auth: Add enhanced tenant support

alexweininger:
API changes:
* Make the `getTenants` function public
* Add `tenantId` parameter to the `signIn` function to enable signing in to a specific tenant
* Expose the `tenantId` parameter to the `isSignedIn` function (was already implemented, just not exposed)

I also am using cross-fetch to list the tenants because the SDK uses a very old API version that doesn't retrieve the full tenant object.

I will update the readme once the API changes are reviewed.

### DIFF
--- a/auth/src/AzureSubscriptionProvider.ts
+++ b/auth/src/AzureSubscriptionProvider.ts
@@ -5,11 +5,20 @@
 
 import type * as vscode from 'vscode';
 import type { AzureSubscription } from './AzureSubscription';
+import { TenantIdDescription } from '@azure/arm-subscriptions';
 
 /**
  * An interface for obtaining Azure subscription information
  */
 export interface AzureSubscriptionProvider {
+    /**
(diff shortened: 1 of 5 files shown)


auth/src/AzureSubscriptionProvider.ts
```diff
@@ -5,11 +5,20 @@
...
  * An interface for obtaining Azure subscription information
  */
 export interface AzureSubscriptionProvider {
+    /**
+     * Gets a list of tenants available to the user.
+     * Use {@link isSignedIn} to check if the user is signed in to a particular tenant.
+     *
+     * @returns A list of tenants.
+     */
+    getTenants(): Promise<TenantIdDescription[]>;
```
bwateratmsft, line 20: Don't forget to update the README as well

bwateratmsft (dismissed):

bwateratmsft (approved):

PR #1594, opened 2023-09-21
auth: Fix getScopes always injecting the management scope, even if a scope for a different resource is specified

MRayermannMSFT:
When trying to get a credential with scope `http://storage.azure.com`, the management scope was being injected. This caused `vscode.authentication.getSession` to return an `undefined` session.


alexweininger (approved):

MRayermannMSFT:
Note for future people looking back at this PR. I just noticed my dumb code `new Set<string>([...arrScopes]);`. I inteionally choose not to fix to avoid a re-review. :)

PR #1602, opened 2023-09-29
Testing - this should fail

alexweininger:


bwateratmsft:
Looks like CheckResults still skipped even though it wasn't supposed to.

PR #1605, opened 2023-10-16
utils: Fix sorting of activity children

MicroFish91:
So turns out we never really needed to reverse the tree items, it only appeared reversed by coincidence because of the alphabetical sorting.  This should fix that.


motm32 (approved):

MicroFish91:
Going to keep this open for a little bit in case I need to bundle an extra change in a separate PR before the version bump :)

PR #1610, opened 2023-10-20
auth: Add `signInToTenant` command and `getUnauthenticatedTenants` util

alexweininger:
Works towards resolving https://github.com/microsoft/vscode-azureresourcegroups/issues/750

### DIFF
--- a/auth/src/signInToTenant.ts
+++ b/auth/src/signInToTenant.ts
@@ -1,0 +1,45 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.txt in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import * as vscode from "vscode";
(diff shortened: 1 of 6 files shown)


auth/src/signInToTenant.ts
```diff
@@ -0,0 +1,45 @@
...
+import * as vscode from "vscode";
+import type { TenantIdDescription } from "@azure/arm-subscriptions";
+import type { AzureSubscriptionProvider } from "./AzureSubscriptionProvider";
+import { getUnauthenticatedTenants } from "./utils/getUnauthenticatedTenants";
+
+/**
+ * Prompts user to select from a list of unauthenticated tenants.
+ * Once selected, requests a new session from VS Code specifially for this tenant.
+ */
+export async function signInToTenant(subscriptionProvider: AzureSubscriptionProvider): Promise<void> {
```
bwateratmsft, line 15: How would e.g. the Docker extension use these new functions? Would it need to also implement a command to sign in to a specific tenant?

Alternatively, would signing in via Azure Resources allow the same tenant sign in to be used by the Docker extension?

PR #1615, opened 2023-10-26
auth: Hard-code node-fetch version

nturinski:
Fixes https://github.com/microsoft/vscode-azurefunctions/issues/3878

There's a lot of weirdness going on here, but the gist of it is whenever fetch is called _before_ `@azure/client-core`, it is overwriting the `node-fetch` package that is used for most of our requests (because we generate our ServiceClients with core-client).

For whatever reason, >= v3.7.0 are causing zipdeploy to fail in App Service/Functions, but _only_ on Windows machines. I've investigated a lot and can't pinpoint the exact reason, but my guess is that the socket isn't closing at the same time as it was on earlier versions.

I've spent a lot of time trying to figure this one out, but have thrown in the hat in terms of understanding exactly why/how this is happening. 😐

### DIFF
--- a/auth/package-lock.json
+++ b/auth/package-lock.json
@@ -1,16 +1,17 @@
 {
     "name": "@microsoft/vscode-azext-azureauth",
-    "version": "1.3.1",
+    "version": "1.3.1-beta",
     "lockfileVersion": 3,
     "requires": true,
     "packages": {
         "": {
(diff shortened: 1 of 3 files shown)


motm32:
I can confim from my stinkpad that this fixes the issue for me🥳

alexweininger (approved):
🥴

bwateratmsft (approved):
How strange...

PR #1616, opened 2023-10-27
Use vscode.workspace.findFiles rather than globby to get files to zip

nturinski:
In order to support the weird blob patterns array that we have in App Service, I had to do some weird stuff because it's not trivial (at least, I couldn't easily figure it out) how to combine multiple blob patterns that have embedded patterns in them. 
See here for an example of what I mean: https://github.com/microsoft/vscode-azureappservice/wiki/Configuring-Zip-Deployment

globby would return everything in terms of relative paths, but vscode.workspace.findFiles gave the absolute, which is why I'm returning everything as a relative path so no changes would need to be made in runWithZipStream

### DIFF
--- a/appservice/package.json
+++ b/appservice/package.json
@@ -64,7 +64,7 @@
         "@types/mocha": "^9.0.0",
         "@types/node": "^16.0.0",
         "@types/p-retry": "^2.0.0",
-        "@types/vscode": "1.76.0",
+        "@types/vscode": "1.82.0",
         "@types/ws": "^8.5.3",
         "@types/yazl": "^2.4.2",
         "@typescript-eslint/eslint-plugin": "^5.53.0",
--- a/appservice/src/deploy/runWithZipStream.ts
+++ b/appservice/src/deploy/runWithZipStream.ts
@@ -4,16 +4,15 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { AzExtPipelineResponse } from '@microsoft/vscode-azext-azureutils';
-import { IActionContext } from '@microsoft/vscode-azext-utils';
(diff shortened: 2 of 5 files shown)


appservice/package.json
bwateratmsft, line 1: Can we remove globby as a dependency?

nturinski: Actually, we have to keep it as a devDependency though. VS Code's `workspace.findFiles` only works for workspaces that are active-- I can't just search a random folder. To grab the test files, `globby` is going into the test folder which is not one the active workspaces at test runtime.

appservice/src/deploy/runWithZipStream.ts
```diff
@@ -83,60 +82,52 @@ function getPathFromMap(realPath: string, pathfileMap?: Map<string, string>): st
...
-    followSymbolicLinks: true, // Follow symlinks to get all sub folders https://github.com/microsoft/vscode-azurefunctions/issues/1289
-};
-
 /**
  * Adds files using glob filtering
  */
-async function getFilesFromGlob(folderPath: string, site: ParsedSite): Promise<string[]> {
-    const zipDeployConfig: vscode.WorkspaceConfiguration = vscode.workspace.getConfiguration(ext.prefix, vscode.Uri.file(folderPath));
-    const globOptions = { cwd: folderPath, followSymbolicLinks: true, dot: true };
+export async function getFilesFromGlob(folderPath: string, resourceName: string): Promise<string[]> {
```
nturinski, line 88: A lot of this weird new logics is because a lot of these previous assumptions will break the `appservice` package when running the tests.

bwateratmsft (approved):

alexweininger (approved):

bwateratmsft (approved):

PR #1619, opened 2023-11-02
auth: Make `getSession` synchronous

nturinski:
Actually fixes: https://github.com/microsoft/vscode-azurefunctions/issues/3878

This [PR](https://github.com/microsoft/vscode-azuretools/commit/4ae935426a2ff4bc191eca248cd3a258568f741c) broke everything. 😡 

Apparently `AzureAuthentication` doesn't like `getSession` being async. So we need to be able to get the session and pass that along before we create the authentication object.

The async call must have somehow caused some parts of the zipfile not to get deployed since the stream was constantly making requests. Not sure why it only reproduces on Windows either. I'm assuming the auth provider was fast enough on a Mac to never actually be an issue.

Anyway, I'm making scopes an optional parameter on `createSubscriptionClient`. I think that should resolve the problem of not being able to pass scopes.

Also could someone double check to make sure that I reverted everything? I think that this is all, but my sanity bar is nearly depleted atm.


motm32:
Can confirm (after multiple tries) this does in fact fix.

auth/package.json
```diff
@@ -54,6 +54,6 @@
     "dependencies": {
         "@azure/arm-subscriptions": "^5.1.0",
         "@azure/ms-rest-azure-env": "^2.0.0",
-        "node-fetch": "2.6.7"
+        "cross-fetch": "^4.0.0"
```
bwateratmsft, line 57: Since we're on node 18 now do we even need cross-fetch?

bwateratmsft (approved):

bwateratmsft:
> Also could someone double check to make sure that I reverted everything? I think that this is all, but my sanity bar is nearly depleted atm.

_Don't Starve intensifies_

alexweininger (dismissed):
awkward


But also yeah we shouldn't need cross-fetch or node-fetch now that we can use the build-in fetch. I did that in https://github.com/microsoft/vscode-azuretools/pull/1614, so just copy over those changes to here.

bwateratmsft (approved):

PR #1621, opened 2023-11-06
auth: Refactor `getSubscriptions` by filtering subs later

alexweininger:
Filtering the subscriptions at the end to make it possible to know all the discovered subscriptions. Working towards https://github.com/microsoft/azcode-internal/issues/85


bwateratmsft (approved):
The code looks good but did the old telemetry event send all subs or just the ones matching the filters? If the latter IMO we don't need this.

alexweininger:
> The code looks good but did the old telemetry event send all subs or just the ones matching the filters? If the latter IMO we don't need this.

We are currently sending all discovered subscriptions. 

See relevant code

https://github.com/microsoft/vscode-azure-account/blob/1d42902d8bbce0f909c5c17572410b1e970b12c7/src/login/updateSubscriptions.ts#L27

https://github.com/microsoft/vscode-azure-account/blob/1d42902d8bbce0f909c5c17572410b1e970b12c7/src/login/updateSubscriptions.ts#L117-L132

PR #1623, opened 2023-11-09
eslint-config: Enforce import types with `@typescript-eslint/consistent-type-imports`

MicroFish91:
How does everyone feel about just adding this as an eslint rule so we don't have to think about whether something should be imported as a type or a value for optimization?

This should make it easier to enforce import types where we can essentially just run `eslint --fix`

Related: https://github.com/microsoft/vscode-azurecontainerapps/pull/568

### DIFF
--- a/eslint-config-azuretools/index.js
+++ b/eslint-config-azuretools/index.js
@@ -29,6 +29,7 @@ module.exports = {
         "@typescript-eslint/restrict-template-expressions": "off",
         "@typescript-eslint/unbound-method": "off",
         "eqeqeq": ["error", "always"],
+        "import/consistent-type-specifier-style": ["error", "prefer-inline"],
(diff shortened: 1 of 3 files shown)


bwateratmsft:
Can it be configured so it applies not to internal imports but only to package imports?

MicroFish91:
> Can it be configured so it applies not to internal imports but only to package imports?
I haven't been able to find a way to have it make this distinction, but if anyone else knows of a way please let me know

bwateratmsft:
It looks like as long as TS 4.5+ is in used, you can have it fix imports to look like this:
```typescript
import { type Foo } from 'bar';
```

https://typescript-eslint.io/rules/consistent-type-imports/

bwateratmsft (approved):

PR #1624, opened 2023-11-13
Fix `eslint-config` `packageToPublish` name

MicroFish91:
Our `eslint-config` package deviates from the normal `vscode-azext` prefix, so when we go to publish the package, the tgz name isn't detected properly unless we fix the name to match like so.


MicroFish91:
Not sure how to actually test this, but I assume it would fix this issue we're running into:

image
image

@bwateratmsft @alexweininger

bwateratmsft (approved):

PR #1626, opened 2023-11-20
auth: Unverified fixes

alexweininger:
A set of attempts/fixes for Brandon to use


bwateratmsft:
I think I know what's remaining that isn't right--we never set the `endpoint` property on the subscription client, so it would only ever list subscriptions from the public cloud--but we're getting a token for the sovereign cloud, so it throws an error that the audience of the token is wrong.

We need to configure the subscription client's `endpoint` property.

PR #1627, opened 2023-11-20
auth: Fix sovereign clouds not working

bwateratmsft:
I've tested with AzureChina but I'd like to do one more sanity pass on:

- [x] Azure public cloud
- [x] Azure China cloud
- [x] Azure US Government cloud

Some notes about the PR:
- We were never passing the `endpoint` value to the subscription client, so it was only ever making requests against the public cloud. This explains the error message @alexweininger sees in #1625
- It seems that `@azure/arm-subscriptions` was an older package and we should have used `@azure/arm-resources-subscriptions` instead. One issue from #1590 that @alexweininger found was that the `@azure/arm-subscriptions` used a very old API version which did not include much tenant information; doing a direct HTTP request with a newer API version worked. However, it looks like `@azure/arm-resources-subscriptions` uses a newer API version and would no longer need the direct HTTP request.
- Marking it as a 2.0.0 since the underlying Azure package--whose types we effectively reexport--changed

### DIFF
--- a/auth/package-lock.json
+++ b/auth/package-lock.json
@@ -1,15 +1,15 @@
 {
     "name": "@microsoft/vscode-azext-azureauth",
-    "version": "1.4.0",
+    "version": "2.0.0",
     "lockfileVersion": 3,
     "requires": true,
     "packages": {
         "": {
             "name": "@microsoft/vscode-azext-azureauth",
-            "version": "1.4.0",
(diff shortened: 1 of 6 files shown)


nturinski (approved):

alexweininger:
Fixes https://github.com/microsoft/vscode-azureresourcegroups/issues/768

PR #1635, opened 2023-11-27
utils: Add `GenericParentTreeItem`

MicroFish91:
This addition combined with a soon-to-follow activity log update will eventually be used to support this: https://github.com/microsoft/vscode-azurecontainerapps/pull/576

### DIFF
--- a/utils/src/tree/AzExtTreeItem.ts
+++ b/utils/src/tree/AzExtTreeItem.ts
@@ -52,7 +52,7 @@ export abstract class AzExtTreeItem implements types.AzExtTreeItem {
             return undefined;
         }
 
-        if (this.treeDataProvider.collapsibleStateTracker) {
+        if (this.parent && this.treeDataProvider.collapsibleStateTracker) {
             return this.treeDataProvider.collapsibleStateTracker.getCollapsibleState(this);
         }
 
--- a/utils/src/tree/AzExtTreeItem.ts
+++ b/utils/src/tree/AzExtTreeItem.ts
@@ -52,7 +52,14 @@ export abstract class AzExtTreeItem implements types.AzExtTreeItem {
             return undefined;
         }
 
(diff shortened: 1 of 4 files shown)


utils/src/tree/GenericParentTreeItem.ts
```diff
@@ -0,0 +1,54 @@
...
+
+export class GenericParentTreeItem extends AzExtParentTreeItem implements types.GenericParentTreeItem {
+    public childTypeLabel?: string;
+    public contextValue: string;
+    public label: string;
+    public suppressMaskLabel?: boolean;
+
+    public readonly initialCollapsibleState: TreeItemCollapsibleState;
+
+    constructor(parent: IAzExtParentTreeItemInternal | undefined, readonly options: GenericParentTreeItemOptions) {
```
MicroFish91, line 32: Question: Should we allow parent to be passed as `undefined`?

There is one known issue with calling `this.treeDataProvider` if passed as undefined as `treeDataProvider` is not set and `nonNullProp` will throw an error.

image

image

utils/src/tree/AzExtTreeItem.ts
```diff
@@ -52,7 +52,7 @@ export abstract class AzExtTreeItem implements types.AzExtTreeItem {
             return undefined;
         }
 
-        if (this.treeDataProvider.collapsibleStateTracker) {
+        if (this.parent && this.treeDataProvider.collapsibleStateTracker) {
```
MicroFish91, line 55: Temporary work-in that I had for testing while we discuss... not meant to be the final state.

utils/src/tree/GenericParentTreeItem.ts
nturinski: Discussed in engineering office hours

nturinski (approved):

PR #1637, opened 2023-11-30
auth: Use management endpoint for scope by default

alexweininger:
Fixes https://github.com/microsoft/vscode-azurefunctions/issues/3913

### DIFF
--- a/auth/package-lock.json
+++ b/auth/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "@microsoft/vscode-azext-azureauth",
-    "version": "2.0.0",
+    "version": "2.1.0",
     "lockfileVersion": 3,
     "requires": true,
     "packages": {
         "": {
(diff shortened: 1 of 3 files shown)


bwateratmsft (approved):

alexweininger:
I'm going to have CTI test Azure Resources with this update before merging this and releasing.

bwateratmsft:
> > Maybe leave a comment about how we've tested that `resourceManagerEndpointUrl`'s audience token seems to a subset of `managementEndpointUrl`? Otherwise people might get really confused as to why we use `managementEndpointUrl` here, but everywhere else we create the clients, we use `resourceManagerEndpointUrl` as the endpoint.
> 
> I want to add a comment but I want it to be accurate so we should consult someone who actually knows.

inb4 comment: We have no idea why this works but it does so don't change it

nturinski:
image
image

🤷

Issue #1639, opened 2023-12-04
[feature request] Allow auth package to acquire access token with customized audience endpoint


JasonYeMSFT:
Currently it only allows acquiring access tokens with management.azure.com as the audience.

PR #1640, opened 2023-12-05
auth: Allow getting access token with custom scope

JasonYeMSFT:
resolves: https://github.com/microsoft/vscode-azuretools/issues/1639

### DIFF
--- a/auth/src/VSCodeAzureSubscriptionProvider.ts
+++ b/auth/src/VSCodeAzureSubscriptionProvider.ts
@@ -237,30 +237,37 @@ export class VSCodeAzureSubscriptionProvider extends vscode.Disposable implement
      *
      * @returns A client, the credential used by the client, and the authentication function
      */
-    private async getSubscriptionClient(tenantId?: string, scopes?: string[]): Promise<{ client: SubscriptionClient, credential: TokenCredential, authentication: AzureAuthentication }> {
-        const armSubs = await import('@azure/arm-resources-subscriptions');
-        const session = await getSessionFromVSCode(scopes, tenantId, { createIfNone: false, silent: true });
-        if (!session) {
+    private async getSubscriptionClient(tenantId?: string): Promise<{ client: SubscriptionClient, credential: TokenCredential, authentication: AzureAuthentication }> {
+        const initialSession = await getSessionFromVSCode(undefined, tenantId, { createIfNone: false, silent: true });
+        if (!initialSession) {
             throw new NotSignedInError();
         }
 
         const credential: TokenCredential = {
-            getToken: async () => {
-                return {
-                    token: session.accessToken,
-                    expiresOnTimestamp: 0
-                };
+            getToken: async (scopes, _options) => {
+                const session = await getSessionFromVSCode(scopes, tenantId, { createIfNone: false, silent: true });
+                if (session?.accessToken) {
+                    return {
+                        token: session.accessToken,
+                        expiresOnTimestamp: 0
+                    };
+                } else {
+                    return null;
+                }
             }
         }
 
         const configuredAzureEnv = getConfiguredAzureEnv();
         const endpoint = configuredAzureEnv.resourceManagerEndpointUrl;
 
+        const armSubs = await import('@azure/arm-resources-subscriptions');
         return {
             client: new armSubs.SubscriptionClient(credential, { endpoint }),
             credential: credential,
             authentication: {
-                getSession: () => session
+                getSession: async (scopes) => {
+                    return await getSessionFromVSCode(scopes, tenantId, { createIfNone: false, silent: true });
+                }
             }
         };
     }
--- a/auth/package-lock.json
+++ b/auth/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "@microsoft/vscode-azext-azureauth",
-    "version": "2.0.0",
+    "version": "2.0.1",
     "lockfileVersion": 3,
     "requires": true,
     "packages": {
         "": {
             "name": "@microsoft/vscode-azext-azureauth",
-            "version": "2.0.0",
+            "version": "2.0.1",
             "license": "MIT",
             "dependencies": {
                 "@azure/arm-resources-subscriptions": "^2.1.0",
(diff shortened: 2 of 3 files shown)


bwateratmsft (commented):
Isn't this undoing much of what #1619 did?

auth/src/VSCodeAzureSubscriptionProvider.ts
```diff
@@ -237,30 +237,37 @@ export class VSCodeAzureSubscriptionProvider extends vscode.Disposable implement
...
         const configuredAzureEnv = getConfiguredAzureEnv();
         const endpoint = configuredAzureEnv.resourceManagerEndpointUrl;
 
+        const armSubs = await import('@azure/arm-resources-subscriptions');
         return {
             client: new armSubs.SubscriptionClient(credential, { endpoint }),
             credential: credential,
             authentication: {
-                getSession: () => session
+                getSession: async (scopes) => {
```
alexweininger, line 268: This function needs to stay non-async because of https://github.com/microsoft/vscode-azuretools/pull/1619

JasonYeMSFT: OK, I'll look at 1619 and see why.

nturinski: But can you have async code in a function who's return type isn't a Thenable/Promise?

auth/src/VSCodeAzureSubscriptionProvider.ts
```diff
@@ -237,30 +237,37 @@ export class VSCodeAzureSubscriptionProvider extends vscode.Disposable implement
      *
      * @returns A client, the credential used by the client, and the authentication function
      */
-    private async getSubscriptionClient(tenantId?: string, scopes?: string[]): Promise<{ client: SubscriptionClient, credential: TokenCredential, authentication: AzureAuthentication }> {
-        const armSubs = await import('@azure/arm-resources-subscriptions');
-        const session = await getSessionFromVSCode(scopes, tenantId, { createIfNone: false, silent: true });
-        if (!session) {
+    private async getSubscriptionClient(tenantId?: string): Promise<{ client: SubscriptionClient, credential: TokenCredential, authentication: AzureAuthentication }> {
+        const initialSession = await getSessionFromVSCode(undefined, tenantId, { createIfNone: false, silent: true });
```
nturinski, line 241: Can you not call `getSessionFromVSCode` here and just have the scopes be a parameter that gets passed in? That way the async work will be done at this point.

auth/src/VSCodeAzureSubscriptionProvider.ts
alexweininger: > I am wondering what will happen if I make the getSession function synchronous but let it return a Promise.

Tested and it makes the deploy fail just like https://github.com/microsoft/vscode-azurefunctions/issues/3878 with an "Number of entries expected in End Of Central Directory does not correspond to number of entries in Central Directory." error.

auth/src/VSCodeAzureSubscriptionProvider.ts
bwateratmsft: Should the MS auth provider be used directly instead of trying to make this subscription-listing package do it?

alexweininger: I tested this locally and it works. I think it would work for Jason's use case and also keep things working on our end. However, I don't love it because any of our existing requests that pass scopes here would now be async. I'm not sure if that's an issue, but it could be since it was an issue for the zip deploy requests.
```ts
            authentication: {
                getSession: async (scopes) => {
                    if (scopes) {
                        return await getSessionFromVSCode(scopes, tenantId, { createIfNone: false, silent: true });
                    } else {
                        return initialSession;
                    }
                }
            }
```

bwateratmsft: The SDKs aren't using `getSession`, that's just meant to mirror the VSCode API for getting auth sessions. The SDKs will use `getToken`. In fact I don't think anything should be directly using `getSession`.

appservice/src/deploy/runWithZipStream.ts
```diff
@@ -43,15 +43,9 @@ export async function runWithZipStream(context: IActionContext, options: {
         ext.outputChannel.appendLog(vscode.l10n.t('Creating zip package...'), { resourceName: site.fullName });
         const zipFile: yazl.ZipFile = new yazl.ZipFile();
         let filesToZip: string[] = [];
-        let sizeOfZipFile: number = 0;
 
-        zipFile.outputStream.on('data', (chunk) => {
```
JasonYeMSFT, line 48: I believe this 'data' event listener is the root cause of the deployment issue. During zip deployment we pass `zipFile.outputStream` to the request body and later the SDK calls pipe() on it to read the data and write it to the request body. Before that the SDK needs to acquire the access token. Once acquire token become async, there is chance that this event listener is called before pipe has been called towards the request body. When that happens, the data consumed by this event listener will be removed from the stream and the request body will never be able to receive them.

`yazl` internally maintains a cursor which gives the number of bytes it has written to the output stream. We can use that to report the number of bytes instead of using our own 'data' event listener.

appservice/package.json
```diff
@@ -65,7 +65,6 @@
         "@types/p-retry": "^2.0.0",
         "@types/vscode": "1.82.0",
         "@types/ws": "^8.5.3",
-        "@types/yazl": "^2.4.2",
```
JasonYeMSFT, line 68: This package doesn't give the correct typing of the package.

auth/src/VSCodeAzureSubscriptionProvider.ts
JasonYeMSFT: A fix for the deploy issue has been merged. The getSession function no longer needs to be synchronous.

auth/src/VSCodeAzureSubscriptionProvider.ts
JasonYeMSFT: A fix to the deployment issue has been merged.

bwateratmsft: Great find @JasonYeMSFT!

bwateratmsft (dismissed):

alexweininger (approved):

PR #1641, opened 2023-12-08
Fix race condition in zip deploy

JasonYeMSFT:
The 'data' event listener is the root cause of the deployment issue https://github.com/microsoft/vscode-azurefunctions/issues/3878. During zip deployment we pass zipFile.outputStream to the request body and later the SDK calls pipe() on it to read the data and write it to the request body. Before that the SDK needs to acquire the access token. Once acquire token becomes async (and takes more event loop cycles), there is a chance that this event listener is called before pipe has been called towards the request body. When that happens, the data consumed by this event listener will be removed from the stream and the request body will never be able to receive them.


alexweininger (approved):

nturinski:
Make sure you bump the package.json version so that we can release the changes.

PR #1648, opened 2023-12-18
utils:  Add an option to automatically select an `AzureSubscription` by `subscriptionId`

MicroFish91:
Would like to eventually use here: https://github.com/microsoft/vscode-azurecontainerapps/pull/578

Wanted to add the ability to pass an option to `subscriptionExperience` to automatically select by `subscriptionId` without having to re-prompt the user.

Example:

```
const subscription: AzureSubscription = await subscriptionExperience(context, ext.rgApiV2.resources.azureResourceTreeDataProvider, {
        selectSubscriptionById: '<insert subscriptionId>'
});
```

### DIFF
--- a/utils/src/pickTreeItem/quickPickAzureResource/QuickPickAzureSubscriptionStep.ts
+++ b/utils/src/pickTreeItem/quickPickAzureResource/QuickPickAzureSubscriptionStep.ts
@@ -4,13 +4,13 @@
  *--------------------------------------------------------------------------------------------*/
 
 import * as vscode from 'vscode';
-import { AzureResourceQuickPickWizardContext, AzureSubscriptionQuickPickOptions, IAzureQuickPickItem, SkipIfOneQuickPickOptions } from '../../../index';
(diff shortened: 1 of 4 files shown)


utils/src/pickTreeItem/quickPickAzureResource/QuickPickAzureSubscriptionStep.ts
```diff
@@ -30,6 +34,14 @@ export class QuickPickAzureSubscriptionStep extends GenericQuickPickStepWithComm
 
         return pickedSubscription;
     }
+
+    protected override getPickWithoutPrompt(picks: IAzureQuickPickItem<SubscriptionItem>[]): IAzureQuickPickItem<SubscriptionItem> | undefined {
```
bwateratmsft, line 38: Instead of this implementation I'd try to do a skip-if-one step with pick filters that select only the chosen subscription.

bwateratmsft (approved):

PR #1656, opened 2024-01-02
Introduce `validationUtils` namespace & character length validation logic

MicroFish91:
Beginning of a series of PRs to add a validation utility to simplify input string validation in our Azure Wizard `inputBox` steps.

Example of use:
image

See: https://github.com/microsoft/vscode-azurecontainerapps/pull/582

Let me know if you think something like this would be useful :)

### DIFF
--- a/utils/index.d.ts
+++ b/utils/index.d.ts
@@ -2076,24 +2076,27 @@ export declare class TreeElementStateManager<TElement extends TreeElementWithId 
  */
 export declare namespace validationUtils {
     /**
-     * A generic set of validation utilities that can be used for most input strings
+     *  Defines the constraints for a valid size range
+     *  @property `lowerLimitIncl`: The minimum size of the range (inclusive)
+     *  @property `upperLimitIncl`: The maximum size of the range (inclusive)
      */
-    export namespace General {
-        /**
-         * Checks if the given input string has a valid length as determined by the optional lower and upper limit parameters
-         * @param value The input string to validate
-         * @param lowerLimitIncl The minimum string length requirement (inclusive)
-         * @param upperLimitIncl The maximum string length requirement (inclusive)
-         * @related getInvalidCharLengthMessage
-         */
-        export function hasValidCharLength(value: string, lowerLimitIncl?: number, upperLimitIncl?: number): boolean;
-
-        /**
-         * Provides a message that can be used to inform the user of invalid string length as determined by the optional lower and upper limit parameters
-         * @param lowerLimitIncl The minimum string length requirement (inclusive)
-         * @param upperLimitIncl The maximum string length requirement (inclusive)
-         * @related hasValidCharLength
-         */
-        export function getInvalidCharLengthMessage(lowerLimitIncl?: number, upperLimitIncl?: number): string;
+    export interface RangeConstraints {
+        lowerLimitIncl?: number;
+        upperLimitIncl?: number;
     }
+
+    /**
+     * Checks if the given input string has a valid character length as determined by the optional lower and upper limit parameters
+     * @param value The input string to validate
+     * @param constraints An object defining the range of character lengths that are considered valid
+     * @related getInvalidCharLengthMessage
+     */
+    export function hasValidCharLength(value: string, constraints?: RangeConstraints): boolean;
+
+    /**
+     * Provides a message that can be used to inform the user of invalid character length as determined by the optional lower and upper limit parameters
+     * @param constraints An object defining the range of character lengths that are considered valid
+     * @related hasValidCharLength
+     */
+    export function getInvalidCharLengthMessage(constraints?: RangeConstraints): string;
 }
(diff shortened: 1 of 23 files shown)


utils/index.d.ts
```diff
@@ -2070,3 +2070,33 @@ export declare class TreeElementStateManager<TElement extends TreeElementWithId
...
+/**
+ * Input string validation utilities for Azure Tools
+ */
+export declare namespace validationUtils {
+    /**
+     *  Defines the constraints for a valid size range
+     *  @property `lowerLimitIncl`: The minimum size of the range (inclusive)
+     *  @property `upperLimitIncl`: The maximum size of the range (inclusive)
+     */
+    export interface RangeConstraints {
```
MicroFish91, line 2083: Kept this named very generically because I was planning to reuse this interface again later for another validation method

utils/src/utils/validationUtils.ts
```diff
@@ -0,0 +1,31 @@
...
+export namespace validationUtils {
+    export interface RangeConstraints {
+        lowerLimitIncl?: number;
+        upperLimitIncl?: number;
+    }
+
+    export function hasValidCharLength(value: string, rc?: RangeConstraints): boolean {
+        const lowerLimitIncl = (!rc?.lowerLimitIncl || rc.lowerLimitIncl < 1) ? 1 : rc.lowerLimitIncl;
+        const upperLimitIncl = (!rc?.upperLimitIncl || rc.upperLimitIncl > Number.MAX_SAFE_INTEGER) ? Number.MAX_SAFE_INTEGER : rc.upperLimitIncl;
+        return lowerLimitIncl <= upperLimitIncl && value.length >= lowerLimitIncl && value.length <= upperLimitIncl;
```
nturinski, line 17: I feel like that this should just throw an error: `lowerLimitIncl <= upperLimitIncl`. If the upperLimit is lower than the lowerLimit, somebody created the RangeConstraints backwards and that'd be kind of difficult to understand what's going wrong.

utils/src/utils/validationUtils.ts
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    export function getInvalidCharLengthMessage(rc?: RangeConstraints): string {
```
nturinski, line 20: Could you explain how this is working? I'm not really sure how you could figure out which message to output without having the `value.length` as part of the function.

I'm not understanding how the `lowerLimitIncl` existing but not the `upperLimitIncl` can be extrapolated to mean that the value needs to be greater than the lower limit.

Oh actually, I think I understand now. The message isn't reactive to the value-- so if the error is too few characters, it will still output `The value must be between {0} and {1} characters long.'` if both limits exist.

I don't know about other people, but my assumption made this very confusing to me, but I'm not sure if my expectation that it would output a message relevant to the error is realistic.

Also-- I feel like the lower limit should default to 1. I can't think of any case where we wouldn't have at least a requirement of 1 character.

nturinski: Actually disregard this:

> I don't know about other people, but my assumption made this very confusing to me, but I'm not sure if my expectation that it would output a message relevant to the error is realistic.

It's easier just to maintain the messages with how it is now. I think it's a pretty common error message pattern anyway.

MicroFish91: > Could you explain how this is working? I'm not really sure how you could figure out which message to output without having the value.length as part of the function.

In this pattern, you would need to determine whether the message needs to be output before calling getMessage, then you pass it your constraints, which defines what the error message needs to say.  The error message wording in this case is always constant, and tied to the constraints which are known ahead of time.  

> I'm not understanding how the lowerLimitIncl existing but not the upperLimitIncl can be extrapolated to mean that the value needs to be greater than the lower limit.

To me a boundary not existing means there is not limit in that direction.  I hope that's a safe assumption, I feel like it is for this.

I'm thinking of it like: 
[lowerLimit, upperLimit]: lowerLimit <= x <= upperLimit (same as `x >= lowerLimit` && `x <= upperLimit`)
[2, 5]: 2 <= x <= 5 (same as `x >= 2 && x <= 5`)
[5, _]: 5 <= x (same as `x >= 5`)

> I don't know about other people, but my assumption made this very confusing to me, but I'm not sure if my expectation that it would output a message relevant to the error is realistic.

I based it around the common pattern I observed in the extensions, which was something like:

```
if (!valid) {
  return "not valid for x reason";
}
```

We could always change this pattern, there's other validation libraries which we could take inspiration from instead.  

> Also-- I feel like the lower limit should default to 1. I can't think of any case where we wouldn't have at least a requirement of 1 character.

For the actual validation logic, it does default to 1.  For the message portion, it _kind of_ basically does this, but is only explicit about the number in the message if you supply the exact value.  I guess the only case that I think might look weird is if you put in a lower limit of `1`.  I think it would look better to show the first message in that case, but that's just a preference and not sure how much we care about that edge case (let me know if you have a preference)

MicroFish91:
Added a message condition for when the lower and upper limits are equal*

nturinski (approved):

PR #1665, opened 2024-01-23
Improve azure pipelines yaml

MRayermannMSFT:
- Allow only building one azuretools package at a time
- Allow choosing what branch to publish from during release


MRayermannMSFT:
Example build: https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=67660&view=results

What queueing a build looks like now:
image

PR #1669, opened 2024-01-25
auth: Allow getting session and access token with custom scope

alexweininger:
Changes are pulled from @JasonYeMSFT's original PR https://github.com/microsoft/vscode-azuretools/pull/1643. Jason needs this change to support a feature in the Databases extension.

Creating as a draft because this code change still breaks some things.

Known breakages:
1. ~Breaks zip deploy.~ Fixed in the `appservice` shared package by https://github.com/microsoft/vscode-azuretools/commit/751ee12d693a4cd62944a53794bbf4ca1d09ec6e
1. ~Zip deploying to a function/app service app in a sovereign cloud. (along with all other kudu operations like viewing deployments and logs)~ Fixed in the `azure` shared package by https://github.com/microsoft/vscode-azuretools/pull/1677

### DIFF
--- a/auth/src/VSCodeAzureSubscriptionProvider.ts
+++ b/auth/src/VSCodeAzureSubscriptionProvider.ts
@@ -237,30 +237,37 @@ export class VSCodeAzureSubscriptionProvider extends vscode.Disposable implement
      *
      * @returns A client, the credential used by the client, and the authentication function
      */
-    private async getSubscriptionClient(tenantId?: string, scopes?: string[]): Promise<{ client: SubscriptionClient, credential: TokenCredential, authentication: AzureAuthentication }> {
-        const armSubs = await import('@azure/arm-resources-subscriptions');
-        const session = await getSessionFromVSCode(scopes, tenantId, { createIfNone: false, silent: true });
-        if (!session) {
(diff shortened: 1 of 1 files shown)


alexweininger:
@bwateratmsft I'm investigating this, but would appreciate some help figuring out why this would break zip deploy for sovereign clouds.

alexweininger:
I found that the access token audiences are different between the working and broken access tokens.

Broken access token audience: “https://management.chinacloudapi.cn\” (resource manager endpoint)
Working access token audience: “https://management.core.chinacloudapi.cn\” (management endpoint)

alexweininger:
Looks like I already investigated and found out about the endpoint difference in December https://github.com/microsoft/vscode-azuretools/pull/1637

However, that was changing the default if no scopes were passed. This change makes it so scopes will be respected, but now the wrong scopes are being passed in 🙃

alexweininger:
Once https://github.com/microsoft/vscode-azuretools/pull/1670 is merged and released, we'll be good to merge this too.

alexweininger:
Ok so #1670 broke a bunch of other stuff so https://github.com/microsoft/vscode-azuretools/pull/1677 is the actual fix that doesn't break everything! 😄

PR #1670, opened 2024-01-26
azure: use management endpoint in `createGenericClient`

alexweininger:
Use `managementEndpointUrl` instead of `resourceManagerEndpointUrl` in `createGenericClient`.

Fixes an issue that surfaces with the changes in https://github.com/microsoft/vscode-azuretools/pull/1669 with sovereign clouds. I'll explain more below.

`createGenericClient` is used to make calls against the kudu service, which were failing with 403 errors because the access token audience was incorrectly set to the resource manager endpoint and not the management endpoint. Sovereign clouds must be more sensitive about which audience is used because it works fine for normal Azure.

We still use `resourceManagerEndpointUrl` two other times in this file, for `createAzureSubscriptionClient` and `createAzureClient`. I'm going to leave these as-is for now because as far as I know, only kudu calls were breaking.

### DIFF
--- a/azure/package-lock.json
+++ b/azure/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "@microsoft/vscode-azext-azureutils",
-    "version": "2.0.4",
+    "version": "2.0.5",
     "lockfileVersion": 3,
     "requires": true,
     "packages": {
         "": {
             "name": "@microsoft/vscode-azext-azureutils",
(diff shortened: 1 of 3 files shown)


nturinski (approved):
Seems like we would want to include this in both the Functions and App Service releases?

alexweininger:
> Seems like we would want to include this in both the Functions and App Service releases?

Yes

PR #1678, opened 2024-01-31
appservice: Force azext-azureutils to be v2.0.2 to roll back token bug

nturinski:


alexweininger:
For future Alex: We're reverting the version of azureutils to 2.0.2 to avoid consuming e1cd099 which breaks stuff. e1cd099 will be rolled back and fixed soon properly in https://github.com/microsoft/vscode-azuretools/pull/1676.

alexweininger (approved):

PR #1680, opened 2024-02-01
Changes for azure agent integration

MRayermannMSFT:
This [doc](https://github.com/microsoft/vscode-azure-agent/blob/main/docs/Architecture.md) is recommended reading before reviewing this PR. It gives a high level overview of how the Azure agent works, and most importantly, how Azure extensions and their existing functionality will integrate with it.

The main thing this change accomplishes is the adding of types/classes needed for integrating extensions with the Azure agent extension. This includes:

# Integration
- `ExtensionAgentMetadata`: the metadata an extension adds to its `package.json` which provides integration information for the agent extension.
- `WizardBasedCommandConfig`: the config object extensions use to expose existing wizard command based functionality to the agent extension.
- `AgentBenchmarkConfig`: the config object extensions use to expose benchmarks (similar to tests) to the agent extension.

# Runtime
- `IAzureAgentInput`: the azure input class the the agent extension gives to service extensions when running command wizards in the background.
- `AzureUserInputQueue`: the queue of azure inputs that the agent extensions gives to service extensions when a user chooses to run a command with the inputs the agent has determined.
- `AzExtUserInputWithInputQueue`: the azure input class that service extensions uses when needing to use inputs from a `AzureUserInputQueue`.

# Benchmarking
- `AgentBenchmarkConfig`: the config object extensions use to expose benchmarks (similar to tests) to the agent extension. Benchmarks define how the agent should respond given a certain prompt. Benchmarks can be run from the chat UX only. In the future, we will hopefully be able to run benchmarks from CICD. Benchmarks are **not** called tests because benchmarks are not expected to pass 100% of the time.

# Real Usages

The other thing this change accomplishes is modifying two existing wizard steps in order to make an example integration with the "Create Function App" command possible. These steps are the:
- `LocationListStep`
   - Updates the internal context type to have a `ui: IAzureAgentInput`
   - Adds the various agent metadatas needed for calling into `IAzureAgentInput` (type enforced by the previous bullet)
- `SiteNameStep`
   - Adds an internal context type with a `ui: IAzureAgentInput`
   - Adds options to the constructor to disable validation and provide custom agent metadata
   - Adds the various agent metadatas needed for calling into `IAzureAgentInput` (type enforced by the previous bullet)

Additionally, an example integration with the "Create Function App" command can be found at this draft functions PR here: https://github.com/microsoft/vscode-azurefunctions/pull/3941

Finally, if you wish to refer to the agent extension repo to see how it uses the added types/classes in this PR, that repo can be found here: https://github.com/microsoft/vscode-azure-agent


appservice/package.json
```diff
@@ -40,9 +40,9 @@
         "@azure/core-client": "^1.7.2",
         "@azure/core-rest-pipeline": "^1.10.3",
         "@azure/storage-blob": "^12.3.0",
-        "@microsoft/vscode-azext-azureutils": "^2.0.6",
+        "@microsoft/vscode-azext-azureutils": "2.1.0",
```
bwateratmsft, line 43: Is there a reason to not continue using caret notation? i.e.:
```suggestion
        "@microsoft/vscode-azext-azureutils": "^2.1.0",
```

appservice/package.json
```diff
@@ -40,9 +40,9 @@
...
         "@microsoft/vscode-azext-github": "^1.0.0",
-        "@microsoft/vscode-azext-utils": "^2.1.0",
+        "@microsoft/vscode-azext-utils": "2.2.0",
```
bwateratmsft, line 45: Same question applies

azure/package.json
```diff
@@ -39,7 +39,7 @@
         "@azure/core-client": "^1.6.0",
         "@azure/core-rest-pipeline": "^1.9.0",
         "@azure/logger": "^1.0.4",
-        "@microsoft/vscode-azext-utils": "^2.0.0",
+        "@microsoft/vscode-azext-utils": "2.2.0",
```
bwateratmsft, line 42: Same question applies

appservice/package.json
MRayermannMSFT: No. I forgot to put it back when I stopped using the beta versions. 🙈

azure/package.json
MRayermannMSFT: Same answer applies.

bwateratmsft (approved):

utils/index.d.ts
```diff
@@ -2100,3 +2123,223 @@ export declare namespace validationUtils {
...
+ * This information should be placed in an `agentMetdata` property.
+ */
+export type ExtensionAgentMetadata = {
+    version: "1.0";
+
+    /**
+     * The VS Code command ID of a command that the extension implements which can be used to get the list
+     * of {@link WizardBasedCommandConfig}s that the extension implements and wishes to expose via the agent.
+     *
+     * Note: the function type is only for testing in the agent extension itself.
```
JasonYeMSFT, line 2280: I feel you have removed the function type here.

utils/index.d.ts
```diff
@@ -2100,3 +2123,223 @@ export declare namespace validationUtils {
...
+     */
+    getWizardCommandsCommandId: string;
+
+    /**
+     * The VS Code command ID of a command that the extension implements which can be used to run, without performing
```
JasonYeMSFT, line 2285: The distinction between "run" and "execution" isn't very clear to me. What does it mean by "without performing execution"?

utils/index.d.ts
MRayermannMSFT: Thanks!

utils/index.d.ts
MRayermannMSFT: I have updated the comments

utils/index.d.ts
```diff
@@ -2100,3 +2123,223 @@ export declare namespace validationUtils {
...
+     * - "The subscription that the Storage Account should be created in."
+     * - "The resource group that the Container App should be created in."
+     * - "The function runtime for the Function App."
+     * - "The name of the Static Web App."
+     */
+    parameterDisplayDescription: string;
+};
+
+export type AgentQuickPickItem<T extends QuickPickItem = QuickPickItem> = {
+    agentMetadata: {
```
alexweininger, line 2166: Could you add examples to the docs for these two properties?

alexweininger (approved):

utils/index.d.ts
MRayermannMSFT: Yes.

utils/index.d.ts
```diff
@@ -2166,6 +2166,9 @@ export type AgentQuickPickItem<T extends QuickPickItem = QuickPickItem> = {
     agentMetadata: {
         /**
          * If this quick pick item should not be picked by the agent.
+         *
+         * @example If an item exists as a way to link a user to some information about the other items in the quick pick, this
```
alexweininger, line 2170: You mean if the item opens a link in the browser?

appservice/src/createAppService/SiteNameStep.ts
```diff
@@ -48,12 +61,24 @@ export class SiteNameStep extends AzureNameStep<IAppServiceWizardContext> {
...
-        })).trim();
+            asyncValidationTask: async (name: string): Promise<string | undefined> => await this.asyncValidateSiteName(client, name),
+            agentMetadata: this._inputBoxOptionsAgentMetadata ?? {
+                parameterName: "siteName",
+                parameterDisplayTitle: "Site Name",
+                parameterDisplayDescription: "The name of the app service site.",
+            }
+        };
+
+        if (this._skipValidation) {
```
nturinski, line 76: I think that this might be a pain to go in and implement to other name steps later.

We don't need to do this now, but I was thinking that we could give `AzureNameStep` a `formatOptions` function that would just do this for you. Maybe something like this:

```

const options: AgentInputBoxOptions = this.formatOptions({
            prompt,
            placeHolder,
            validateInput: (name: string): string | undefined => this.validateSiteName(name),
            asyncValidationTask: async (name: string): Promise<string | undefined> => await this.asyncValidateSiteName(client, name),
            agentMetadata: this._inputBoxOptionsAgentMetadata ?? {
                parameterName: "siteName",
                parameterDisplayTitle: "Site Name",
                parameterDisplayDescription: "The name of the app service site.",
            }
        });

.......


protected formatOptions(options: AzExtInputBoxOptions ): AzExtInputBoxOptions {
  if (this._skipValidation) {
    options.validateInput = undefined;
    options.asyncValidationTask = undefined;
  }
}
```

I could see in the future that we may want to have other options that automatically get formatted/adjusted, so just keeping the name generic.

azure/package.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "@microsoft/vscode-azext-azureutils",
     "author": "Microsoft Corporation",
-    "version": "2.0.6",
+    "version": "2.1.0",
```
nturinski, line 4: Similar to my comment above, since `IResourceGroupWizardContext` is used by so many steps, this will definitely break any extension that updates.

utils/index.d.ts
```diff
@@ -2100,3 +2123,239 @@ export declare namespace validationUtils {
...
+    /**
+     * A title cased string for the parameter this quick pick is for. Will be displayed to the user.
+     *
+     * For example:
+     * - "Subscription"
+     * - "Resource Group"
+     * - "Runtime"
+     * - "Name"
+     */
+    parameterDisplayTitle: string;
```
nturinski, line 2151: Could we consolidate `parameterName` and `parameterDisplayTitle`? It's not a real big deal, but I can't really think of a scenario where they would need to be different and it just adds more boilerplate for each step.

Or maybe we can make one option, and it would just set itself equal to whatever the required one is if the property wasn't provided?

utils/index.d.ts
```diff
@@ -2100,3 +2123,239 @@ export declare namespace validationUtils {
...
+// #endregion
+
+// #region Agent integration types
+
+/**
+ * A config that describes a command that the extension implements which makes use of wizards that use
+ * an {@link IAzureAgentInput}/{@link IAzureUserInput} to get user input.
+ */
+export type WizardBasedCommandConfig = {
+    type: "wizard";
```
nturinski, line 2231: What's with this? Will this include other types in the future?

nturinski (dismissed):

utils/index.d.ts
MRayermannMSFT: I hope so, and I'd like to leave this here as part of that hope.

appservice/src/createAppService/SiteNameStep.ts
MRayermannMSFT: I think I will dump this part of the change for now. Looking at how validation is handled, it is handled by `AzExtUserInput`, which is out of the question.

utils/index.d.ts
MRayermannMSFT: I will try dropping `parameterName` for now. Looking at the agent code I *think* it should be ok. It may need to come back though.

nturinski (dismissed):

alexweininger (approved):

PR #1684, opened 2024-02-02
Meta: Let VSCode change the settings.json

bwateratmsft:
VSCode *really* wants to change these settings. May as well let it.


nturinski (approved):
🤖 Must comply

PR #1685, opened 2024-02-02
build: Support 1ES pipelines

bwateratmsft:


azure-pipelines/templates/1espackage.yml
```diff
@@ -0,0 +1,17 @@
+steps:
+  - task: Npm@1
+    displayName: "\U0001F449 Package"
```
bwateratmsft, line 3: This is the 👉 emoji. It makes it much easier to visually separate what is from our template vs. what is from 1ES templates.

azure-pipelines/templates/1espackage.yml
```diff
@@ -0,0 +1,17 @@
+steps:
+  - task: Npm@1
+    displayName: "\U0001F449 Package"
+    inputs:
+      command: custom
+      customCommand: run package
+    condition: succeeded()
+
+  - task: CopyFiles@2
```
bwateratmsft, line 9: 1ES has a publish task built in, you just have to copy the files to the publish folder

azure-pipelines/templates/package.yml
```diff
@@ -1,13 +1,13 @@
 steps:
```
MRayermannMSFT, line 1: So is this file being used anywhere anymore?

bwateratmsft: It's still being used by the old pipelines which I'd like to keep alive and working while we work on this.

MRayermannMSFT (approved):
LGTM. Probably get at least one other azcode folk to review though. 🙈

nturinski:
So for clarification, does this mean that we'll stop publishing our @microsoft packages to npm and have to use the 1ES registry or whatever the equivalent is?

bwateratmsft:
> So for clarification, does this mean that we'll stop publishing our @microsoft packages to npm and have to use the 1ES registry or whatever the equivalent is?

We still plan to publish to NPM, but we have to do release through ESRP instead of via our own accounts / our own bot account.

PR #1690, opened 2024-02-08
Standardize skipping execution of an Azure Wizard, have run quick pick respect showLoadingPrompt

MRayermannMSFT:


utils/src/wizard/NoExecuteStep.ts
```diff
@@ -0,0 +1,21 @@
...
+import { IActionContext } from '../..';
+import { UserCancelledError } from '../errors';
+import { AzureWizardExecuteStep } from './AzureWizardExecuteStep';
+
+export class NoExecuteStep<T extends IActionContext> extends AzureWizardExecuteStep<T> {
+    public priority: number = 200;
+
+    private _key: string = "NoExecute";
+    public async execute(_wizardContext: T): Promise<void> {
+        throw new UserCancelledError(this._key);
```
JasonYeMSFT, line 15: Should we create a new Error class for it?

MRayermannMSFT: I don't think for today that will be helpful, or even if the answer is yes. There's a lot of existing code which basically says "ok, no big deal, just a user cancelled error", and that's the behavior we want. If we want additional behavior, then ya, I think an answer of "yes" becomes more clear.

nturinski (dismissed):

utils/src/wizard/NoExecuteStep.ts
nturinski: It might be nice to have an additional error just for telemetry purposes.

JasonYeMSFT (dismissed):

utils/src/pickTreeItem/runQuickPickWizard.ts
```diff
@@ -16,7 +16,6 @@ export async function runQuickPickWizard<TPick>(context: types.PickExperienceCon
 
     const wizard = new AzureWizard(wizardContext, {
         hideStepCount: true,
-        showLoadingPrompt: true,
```
nturinski, line 19: Do you know where this is being called? I assume it's probably in mostly in the client extension-- we need to remember to make a note in the changelog that we need to explicitly add this property to anywhere calling it.

Or better yet, it might be better to do this to keep current behavior.
`showLoadingPrompt: wizardOptions.showLoadingPrompt ?? true`

utils/src/wizard/SilentExecuteActivityContext.ts
```diff
@@ -0,0 +1,18 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { ExecuteActivityContext } from "../..";
```
nturinski, line 6: 🤨

nturinski (approved):

PR #1697, opened 2024-02-28
appservice: Bump version and update app service sdk to 14.0.0

motm32:
The Azure Functions extension is now using v14.0.0 of the appservice sdk and the build tests are failing since there are now typing incompatibilites with the tools package.


nturinski:
Bump the appservice version too since you need to publish asap

nturinski (approved):

PR #1701, opened 2024-03-05
appsettings: Add value validation for containerized function apps 

motm32:
Related to this issue within functions: https://github.com/microsoft/vscode-azurefunctions/issues/4014

### DIFF
--- a/appsettings/src/tree/AppSettingsTreeItem.ts
+++ b/appsettings/src/tree/AppSettingsTreeItem.ts
@@ -37,7 +37,7 @@ export function validateAppSettingKey(settings: StringDictionary, client: IAppSe
 }
 
 export function validateAppSettingValue(value: string): string | undefined {
-    if (!value || value === ' ') {
+    if (!value.trim()) {
         return vscode.l10n.t('App setting values cannot be null, undefined or an empty string.');
     }
 
--- a/appsettings/src/tree/AppSettingsTreeItem.ts
+++ b/appsettings/src/tree/AppSettingsTreeItem.ts
@@ -5,6 +5,7 @@
 
 import type { StringDictionary } from '@azure/arm-appservice';
 import { AzExtParentTreeItem, AzExtTreeItem, createContextValue, IActionContext, ICreateChildImplContext, TreeItemIconPath } from '@microsoft/vscode-azext-utils';
+import * as vscode from 'vscode';
 import { ThemeIcon } from 'vscode';
 import { AppSettingsClientProvider, IAppSettingsClient } from '../IAppSettingsClient';
 import { AppSettingTreeItem } from './AppSettingTreeItem';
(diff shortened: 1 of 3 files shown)


appsettings/src/tree/AppSettingsTreeItem.ts
```diff
@@ -35,6 +36,15 @@ export function validateAppSettingKey(settings: StringDictionary, client: IAppSe
     return undefined;
 }
 
+export function validateAppSettingValue(value: string, client: IAppSettingsClient): string | undefined {
+    if (client.isContainer && value === (null || undefined || ' ')) {
```
MicroFish91, line 40: Alternative solution:

```typescript
value = value.trim();
if (client.isContainer && !value) {
  // ...
}
```

Also, just for my own clarification, but why do we only care about the value not being empty when the client is container?

motm32: In the issue they point out it doesn't cause an error for normal function apps just containerized ones. It's likely not something they would want to do but it is possible to hav an empty value for normal function apps.

nturinski: I don't think that it would actually work though:
image

Even if it doesn't, I think it's a stupid thing for users to do and am fine with blocking them on it.

motm32: Oki doke I'll change it to be for both.

appsettings/src/tree/AppSettingsTreeItem.ts
```diff
@@ -35,6 +36,15 @@ export function validateAppSettingKey(settings: StringDictionary, client: IAppSe
     return undefined;
 }
 
+export function validateAppSettingValue(value: string): string | undefined {
+    if (!value || value === ' ') {
```
MicroFish91, line 40: Is any number of empty spaces without actual content considered invalid?  If so, I think I would add the `.trim()` and let `!value` take care of it since an empty string `''` is evaluated as false in a scenario like this 🤔

motm32: Whoops! Didn't see that in your earlier comment.

MicroFish91 (approved):

appsettings/src/tree/AppSettingsTreeItem.ts
```diff
@@ -35,6 +36,15 @@ export function validateAppSettingKey(settings: StringDictionary, client: IAppSe
...
 
+export function validateAppSettingValue(value: string): string | undefined {
+    if (!value.trim()) {
+        return vscode.l10n.t('App setting values cannot be null, undefined or an empty string.');
+    }
+
+    return undefined;
+}
+
+
```
nturinski, line 47: nit: extra line break.

nturinski (approved):

nturinski (approved):

MicroFish91 (approved):

PR #1703, opened 2024-03-07
Add listSubscriptions to resourceApi typings

nturinski:
Implementation needs to be done in Resource Groups, which will be coming soon

### DIFF
--- a/utils/hostapi.d.ts
+++ b/utils/hostapi.d.ts
@@ -64,7 +64,7 @@ export interface AzureHostExtensionApi {
     /**
      *
      * @param filter Whether to filter the list returned, according to the list returned
-     * by `getTenantFilters()` and `getSubscriptionFilters()`. Optional, default true.
+     * by `getTenantFilters()` and `getSubscriptionFilters()`.
      */
     getSubscriptions: (filter: boolean) => Promise<AzureSubscription[]>;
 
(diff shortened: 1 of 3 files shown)


utils/hostapi.d.ts
```diff
@@ -61,6 +61,13 @@ export interface AzureHostExtensionApi {
      */
     registerActivity(activity: Activity): Promise<void>;
 
+    /**
+     *
+     * @param filter Whether to filter the list returned, according to the list returned
+     * by `getTenantFilters()` and `getSubscriptionFilters()`. Optional, default true.
+     */
+    getSubscriptions: (filter: boolean) => Promise<AzureSubscription[]>;
```
MicroFish91, line 69: Should this be `(filter?: boolean)`?

nturinski: Oh yeah, I guess the comment does say that. I kind of wanted to be explicit about it so I'll amend the comment.

MicroFish91 (approved):

bwateratmsft:
Shouldn't the auth package be used instead?

nturinski:
Maybe... Here's my scenario:

I need a way to list the filtered subscriptions to create my own `SubscriptionListStep` for a user to select their sub. This is happening in the Functions and Storage extension.  Since I don't think we'd want the client extensions to directly consume the auth package, I was planning on having Resources handle listing the subs out to the clients.

Resources also controls the tenant/subscription filter settings, so I thought it made more sense for it to deliver the list to the clients.

PR #1714, opened 2024-03-27
appservice: Implementation for flex deployment using new api

nturinski:
Flex deployment requirements changed quite a bit. I'll try to break it down as succulently as possible.

- Remote build should be defaulted. The only runtime where it should be true is python, I believe. I am getting confirmation from the team.
- Because we shouldn't prompt for remote build anymore (we used to ask the user if they wanted it or not), I removed the `customDeployment` message logic and reverted it back to how it used to be.
- Flex deployment API endpoint changed. It required the headers for it to accept the request now as well.
- Similar to my Functions PR, the payload for the old API version isn't compatible with flex deployment (I need a property that only exists on the new payload). That's why I'm doing the get request
- For flex deployment, we need to create a storage blob container. The endpoint takes care of it from there. We're able to get connection strings and etc. from the `functionAppConfig`

### DIFF
--- a/appservice/src/deploy/wizard/deployZip/DeployFlexExecuteStep.ts
+++ b/appservice/src/deploy/wizard/deployZip/DeployFlexExecuteStep.ts
@@ -48,7 +48,6 @@ export class DeployFlexExecuteStep extends DeployZipBaseExecuteStep {
                 const client = blobClient.getContainerClient(containerName);
                 if (!await client.exists()) {
                     await blobClient.createContainer(containerName, { access: "container" });
-
                 }
             }
         }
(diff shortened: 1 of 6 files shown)


appservice/src/deploy/wizard/deployZip/DeployFlexExecuteStep.ts
```diff
@@ -28,4 +34,65 @@ export class DeployFlexExecuteStep extends DeployZipBaseExecuteStep {
...
+        const settings: StringDictionary = await siteClient.listApplicationSettings();
+        const connectionString = settings.properties && settings.properties[connectionStringAppSettingName];
+        if (connectionString) {
+            const blobClient = BlobServiceClient.fromConnectionString(connectionString);
+            const containerName = site.properties?.functionAppConfig?.deployment.storage.value.split('/').pop();
+            if (containerName) {
+                const client = blobClient.getContainerClient(containerName);
+                if (!await client.exists()) {
+                    await blobClient.createContainer(containerName, { access: "container" });
+
```
MicroFish91, line 51: Nit: extra new line

MicroFish91 (dismissed):
LGTM!

> I'll try to break it down as succulently as possible

Can confirm, PR was extra succulent 👍

MicroFish91 (approved):

PR #1715, opened 2024-03-27
pipelines: Use new ESRP NPM release pipeline

bwateratmsft:


nturinski (approved):
Does having an approver alias mean that you have to be the one to approve something? Do you know what this is used for?

Also why do we have to change the project from AzCode to DevDiv?

bwateratmsft:
> Does having an approver alias mean that you have to be the one to approve something? Do you know what this is used for?

No, they accept an approver and owner alias but do not actually use them for anything. Approval is automatic. :shrug:

> Also why do we have to change the project from AzCode to DevDiv?

All the pipelines are moved from the ms-azuretools ADO tenant to the DevDiv one.

PR #1722, opened 2024-04-30
auth: Add AzureSessionProvider and related types

alexweininger:
Goals: improve structure of our auth package so it can be more easily consumed by partner extensions that don't necessary follow our tree view structure.

Putting this PR up so I can give feedback and ask questions easier as we iterate over this code.


auth/src/utils/resourceManagement.ts
```diff
@@ -0,0 +1,36 @@
+import { SubscriptionClient } from '@azure/arm-resources-subscriptions';
+import { TokenCredential } from '@azure/core-auth';
```
bwateratmsft, line 2: `import type` right?

auth/src/AzureSessionProvider.ts
```diff
@@ -0,0 +1,25 @@
+import { TenantIdDescription } from "@azure/arm-resources-subscriptions";
```
bwateratmsft, line 1:
```suggestion
import { type TenantIdDescription } from "@azure/arm-resources-subscriptions";
```

auth/src/VSCodeAzureSessionProvider.ts
```diff
@@ -0,0 +1,241 @@
+import { TenantIdDescription } from "@azure/arm-resources-subscriptions";
```
bwateratmsft, line 1:
```suggestion
import { type TenantIdDescription } from "@azure/arm-resources-subscriptions";
```

auth/src/AzureSessionProvider.ts
```diff
@@ -0,0 +1,25 @@
...
+    PromptIfRequired,
+}
+
+export type AzureSessionProvider = {
+    signIn(): Promise<void>;
+    signInStatus: SignInStatus;
+    tenants: DefinedTenant[];
+    isSignedInToTenant(tenantId: string): boolean;
+    signInStatusChangeEvent: Event<SignInStatus>;
+    getAuthSession(tenantId: string, behavior: GetSessionBehavior, scopes?: string[]): Promise<AzureAuthenticationSession | undefined>;
```
peterbom, line 23: I think it's worth reconsidering the parameters for `getAuthSession`. For my purposes at least, the only reason for including `scopes` in there was to specify the special `VSCODE_CLIENT_ID` value.

The way I've done it here, it's a bit weird that the special `TENANT_ID` scope is abstracted away, but consumers are expected specify `VSCODE_CLIENT_ID` explicitly.

Could we consider something like an `options` object? E.g.
```typescript
export type GetAuthSessionParams {
    tenantId: string;
    behavior: GetSessionBehavior;
    applicationClientId?: string;
}
```

If there's a need to specify unconstrained `scopes` it could be added later. However, if we add it now and it causes confusion ("what scopes do I _actually_ need to specify?"), it'll be hard to remove.

alexweininger (commented):
@peterbom could you describe how the AKS extension would utilize these changes. Would you just be instantiating a `VSCodeAzureSessionProvider` and ignoring `VSCodeAzureSubscriptionProvider`? Would you be instantiating both?

auth/src/VSCodeAzureSessionProvider.ts
```diff
@@ -0,0 +1,241 @@
...
+                case AuthScenario.SignIn:
+                    options = { createIfNone: true, clearSessionPreference: true, silent: false };
+                    break;
+                case AuthScenario.GetSessionPrompt:
+                    // the 'createIfNone' option cannot be used with 'silent', but really we want both
+                    // flags here (i.e. create a session silently, but do create one if it doesn't exist).
+                    // To allow this, we first try to get a session silently.
+                    silentFirst = true;
+                    options = { createIfNone: true, clearSessionPreference: false, silent: false };
+                    break;
```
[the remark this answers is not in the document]
peterbom: I'm _pretty_ sure I tried that. However, I think it pops up a confirmation dialog if you enable `createIfNone`, which is what I was trying to avoid with the
`silent` option. I'd be happy to be proved wrong on that though - this is one of my least favourite parts of the code.

(It also might be a feature request on the VS Code side, but I'm not sure I've got a comprehensive idea of what the options/behaviour I'm asking for would be.)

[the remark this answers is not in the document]
alexweininger: > (It also might be a feature request on the VS Code side, but I'm not sure I've got a comprehensive idea of what the options/behaviour I'm asking for would be.)

Yeah, it sounds like it could be a bug. Or at least the docstring for `createIfNone` is misleading.

auth/src/VSCodeAzureSessionProvider.ts
```diff
@@ -0,0 +1,241 @@
...
+import { NotSignedInError } from "./NotSignedInError";
+import { getConfiguredAuthProviderId, getConfiguredAzureEnv } from "./utils/configuredAzureEnv";
+import { getSubscriptionClient } from "./utils/resourceManagement";
+
+enum AuthScenario {
+    Initialization,
+    SignIn,
+    GetSessionSilent,
+    GetSessionPrompt,
+}
```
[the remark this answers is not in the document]
peterbom: Introducing this was really a ripple effect of the "silent first" approach you [commented on](https://github.com/microsoft/vscode-azuretools/pull/1722/files#r1585396238) below.

I agree, ideally we'd pass a get-session options object around, but when the receiving code converts that into two calls with different options values, the `enum` abstraction starts to look a bit better.

peterbom:
> @peterbom could you describe how the AKS extension would utilize these changes? Would you just be instantiating a `VSCodeAzureSessionProvider` and ignoring `VSCodeAzureSubscriptionProvider`? Would you be instantiating both?

@alexweininger - yes, that's what I'm thinking: instantiating a `VSCodeAzureSessionProvider` and ignoring `VSCodeAzureSubscriptionProvider`

A quick summary of my reasoning:
- Consistency of use: It enables us to use the same pattern for interacting with ARM SDKs (getting credentials and instantiating clients), regardless of whether a command was launched from a TreeView containing subscriptions or from the command palette.
- Separation of concerns: I don't feel that subscriptions are conceptually tied to sessions/credentials, and when moving away from the Azure Account extension I reworked the treeview to that it only exposed persistent properties of resources like IDs and names, and not sessions/credentials (which I worried about getting stale). Since the equivalent of the `AzureSessionProvider` gets us credentials easily I don't think we'd have a big need for an `AzureSubscriptionProvider`.

alexweininger:
> yes, that's what I'm thinking: instantiating a `VSCodeAzureSessionProvider` and ignoring `VSCodeAzureSubscriptionProvider`

Great, that's what I figured. And I completely agree with your reasoning. I think other extensions like yours will do the same.

auth/src/VSCodeAzureSubscriptionProvider.ts
```diff
@@ -27,19 +27,17 @@ export class VSCodeAzureSubscriptionProvider extends vscode.Disposable implement
...
-        const disposable = vscode.authentication.onDidChangeSessions(async e => {
-            // Ignore any sign in that isn't for the configured auth provider
-            if (e.provider.id !== getConfiguredAuthProviderId()) {
-                return;
-            }
+    private readonly sessionProvider: AzureSessionProvider;
 
-            if (await this.isSignedIn()) {
+    public constructor() {
+        const sessionProvider = new VSCodeAzureSessionProvider();
```
alexweininger, line 33: Since the session provider holds state (tenants and their signed-in statuses), I'd like to avoid having multiple instantiated within the same extension context since they could become out of sync. Currently if someone does `new VSCodeAzureSessionProvider()` and `new VSCodeAzureSubscriptionProvider()`, there will be two instances.

I don't think we need to proactively prevent instantiating multiple `VSCodeSessionProvider`s, but since this might be a common scenario, I'd like to prevent it here. A simple solution would be to make the session provider a constructor argument instead.

alexweininger:
@peterbom When you have a chance, could you add me as a contributor to https://github.com/peterbom/vscode-azuretools/tree/experiment/session-provider? That you don't have to do all the changes yourself.

auth/src/VSCodeAzureSubscriptionProvider.ts
peterbom: Agree. I was trying to avoid breaking the existing contract.

If we were going to make breaking changes, I wonder if we could go a step further and provide something like an `activateVSCodeSessionProviders` method (called by consumers from their extension's `activate` method) that creates the `VSCodeAzureXxxxProvider` instances, and registers them with the extension context (for proper disposal later, although I'm not sure how important that is). And then `getXxxxProvider` methods that return the created instances. Sort of like a singleton pattern, VS Code style.

I ended up doing something like that [here](https://github.com/Azure/vscode-aks-tools/blob/b2678a63c02add1362bd362a9c1d4d30865589ab/src/auth/azureSessionProvider.ts#L27-L36).

That might be too big of a change, but just throwing it out there for discussion.

alexweininger: IMO singleton pattern here makes sense, but I think a note in our documentation and making it an optional constructor argument could work fine for now.

If we think this is a big enough issue then we can consider breaking changes.

auth/src/AzureSessionProvider.ts
alexweininger: 1. I'm not familiar with the special `VSCODE_CLIENT_ID`? How and why are you using this?
2. And I agree, I think an options object would be good here. We can add an optional scopes property in there too.

```ts
export type GetAuthSessionParams {
    tenantId: string;
    behavior: GetSessionBehavior;
    applicationClientId?: string;
    scopes?: string[];
}
```

alexweininger: Also, there are scenarios where my team has needed to exclude the ARM scope, and add in something else. I will research into exactly what that was and report back. I want to make sure we can cover that scenario too.

peterbom: > I'm not familiar with the special VSCODE_CLIENT_ID? How and why are you using this?

I'm not _yet_, but I'm experimenting with using the session credentials to access `graph` endpoints for listing and creating applications and service principals. The [default "Visual Studio Code" application](https://github.com/microsoft/vscode/blob/4b81db67db61a03b3cf795a37ee8a7aaa4924f94/extensions/microsoft-authentication/src/AADHelper.ts#L21) does not have the required delegated permissions to do that. You can use the `VSCODE_CLIENT_ID` scope to [specify a different application](https://www.eliostruyf.com/microsoft-authentication-provider-visual-studio-code/#using-the-authentication-provider-with-your-azure-ad-app).

Speaking of that, when I said above I wasn't using the `scopes` option I think I was wrong. When creating an authentication provider needed to initialise the graph client, you need to implement a `getAccessToken` function. This takes some optional `scopes` and I'm passing them on to my `getAuthSession` function. If I didn't do that, I think the `aud` claim in the resulting access token would be wrong, and requests would be rejected by the graph endpoints.

Perhaps the scenarios you're referring to above are similar (e.g. accessing graph rather than ARM).

auth/src/utils/resourceManagement.ts
```diff
@@ -0,0 +1,36 @@
+import { SubscriptionClient } from '@azure/arm-resources-subscriptions';
```
bwateratmsft, line 1:
```suggestion
import { type SubscriptionClient } from '@azure/arm-resources-subscriptions';
```

alexweininger:
I originally wanted to try to get this PR in before we officially announce the Azure Account extension deprecation, but now I've decided to wait. Once announced, I'm expecting teams to come to us with specific asks and needs for their migration, and I'd like to make sure we cover those in this refactor.

I've gathered a list of all extensions that depend on Azure Account and I don't think as many will actually need these changes as I thought. Only 2-3 out of about 25. If my research is correct, and that's truly the case, then I'd like to hold off on putting these changes through. The added complexity of the session provider might not be worth it if only the 2-3 partners are depending on it, while 25 others only need a subscription provider.

bwateratmsft:
> I originally wanted to try to get this PR in before we officially announce the Azure Account extension deprecation, but now I've decided to wait. Once announced, I'm expecting teams to come to us with specific asks and needs for their migration, and I'd like to make sure we cover those in this refactor.
> 
> I've gathered a list of all extensions that depend on Azure Account and I don't think as many will actually need these changes as I thought. Only 2-3 out of about 25. If my research is correct, and that's truly the case, then I'd like to hold off on putting these changes through. The added complexity of the session provider might not be worth it if only the 2-3 partners are depending on it, while 25 others only need a subscription provider.

@alexweininger agreed, good plan. Those needing a session provider should also consider just directly using the account APIs provided by VSCode, these are more session-shaped.

bwateratmsft:
Closing; at this point this is very stale. We can reopen if needed, but it should go in after #2104.

PR #1723, opened 2024-05-06
auth: Implementation fo AzureSub provider that leverages federated credentials

nturinski:
I've explained this quite a bit offline, but just some context for the PR.

We are using federated credentials (service principals) in order to log in to Azure. This token is used in place of VS Code sessions and passed to our subscription client.  The resources extension will need to swap which sub provider it is using based on whether or not it is in a testing environment, but there should be no code changes beyond that.

Still need to add the README. We should translate Hossam's document from OneNote to markdown and include it as a README.

### DIFF
--- a/auth/package.json
+++ b/auth/package.json
@@ -55,6 +55,7 @@
     "dependencies": {
         "@azure/arm-resources-subscriptions": "^2.1.0",
         "@azure/core-client": "^1.9.2",
+        "@azure/core-rest-pipeline": "^1.16.0",
         "@azure/ms-rest-azure-env": "^2.0.0"
     }
 }
--- a/auth/package.json
+++ b/auth/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@microsoft/vscode-azext-azureauth",
     "author": "Microsoft Corporation",
-    "version": "2.3.0",
+    "version": "2.4.0",
     "description": "Azure authentication helpers for Visual Studio Code",
     "tags": [
         "azure",
@@ -32,6 +32,7 @@
     },
     "devDependencies": {
         "@azure/core-auth": "^1.4.0",
+        "@azure/identity": "^4.2.0",
         "@microsoft/eslint-config-azuretools": "^0.2.1",
         "@types/glob": "^8.1.0",
         "@types/html-to-text": "^8.1.0",
@@ -53,6 +54,7 @@
     },
     "dependencies": {
         "@azure/arm-resources-subscriptions": "^2.1.0",
+        "@azure/core-client": "^1.9.2",
         "@azure/ms-rest-azure-env": "^2.0.0"
     }
 }
--- a/auth/src/AzureDevOpsSubscriptionProvider.ts
+++ b/auth/src/AzureDevOpsSubscriptionProvider.ts
@@ -5,13 +5,12 @@
 
 import type { SubscriptionClient, TenantIdDescription } from '@azure/arm-resources-subscriptions';
 import type { TokenCredential } from '@azure/core-auth'; // Keep this as `import type` to avoid actually loading the package (at all, this one is dev-only)
-import { ServiceClient } from '@azure/core-client';
-import { createHttpHeaders, createPipelineRequest, type PipelineRequest } from '@azure/core-rest-pipeline';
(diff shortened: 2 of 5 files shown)


auth/package.json
```diff
@@ -32,6 +32,7 @@
     },
     "devDependencies": {
         "@azure/core-auth": "^1.4.0",
+        "@azure/identity": "^4.2.0",
```
bwateratmsft, line 35: I don't see this package being used

auth/src/AzureDevOpsSubscriptionProvider.ts
```diff
@@ -0,0 +1,233 @@
...
+        return !!this._tokenCredential;
+    }
+
+    public async signOut(): Promise<void> {
+        this._tokenCredential = undefined;
+    }
+
+    public async getTenants(): Promise<TenantIdDescription[]> {
+        return [{
+            tenantId: this._tokenCredential?.tenantId,
```
alexweininger, line 94: So there can only ever be one tenant associated with the token? I also don't see where the `tenantId` property is ever set.

alexweininger (approved):

auth/package.json
nturinski: https://github.com/microsoft/vscode-azuretools/pull/1723/files#diff-9c30a2d74591058fcf9d06dbc7074281071ad805af0619e472a83df969ac8284R202

That lazy load tho~~

nturinski: Note about the new dependencies-- they were already included in the `@azure/ms-rest-azure-env` but I didn't want to import implicitly, so just included them in the package.json. It shouldn't be introducing new dependencies.

auth/src/AzureDevOpsSubscriptionProvider.ts
```diff
@@ -0,0 +1,234 @@
...
+                        scopes: scopes || [],
+                    };
+
+                }
+            }
+        };
+    }
+
+    public onDidSignIn: Event<void> = () => { return new Disposable(() => { /*empty*/ }) };
+    public onDidSignOut: Event<void> = () => { return new Disposable(() => { /*empty*/ }) };
```
alexweininger, line 163: You could probably add support for these pretty easily

alexweininger (approved):

auth/src/AzureDevOpsSubscriptionProvider.ts
```diff
@@ -0,0 +1,234 @@
...
+            process.env.SYSTEM_TEAMFOUNDATIONCOLLECTIONURI: ${process.env.SYSTEM_TEAMFOUNDATIONCOLLECTIONURI ? "✅" : "❌"}\n
+            process.env.SYSTEM_TEAMPROJECTID: ${process.env.SYSTEM_TEAMPROJECTID ? "✅" : "❌"}\n
+            process.env.SYSTEM_PLANID: ${process.env.SYSTEM_PLANID ? "✅" : "❌"}\n
+            process.env.SYSTEM_JOBID: ${process.env.SYSTEM_JOBID ? "✅" : "❌"}\n
+            REMEMBER: process.env.SYSTEM_ACCESSTOKEN must be explicitly mapped!\n
+            https://learn.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml#systemaccesstoken
+        `);
+        }
+
+        const oidcRequestUrl = `${teamFoundationCollectionUri}${teamProjectId}/_apis/distributedtask/hubs/build/plans/${planId}/jobs/${jobId}/oidctoken?api-version=7.1-preview.1&serviceConnectionId=${serviceConnectionId}`;
```
MicroFish91, line 199: Question: How did you figure out the exact format for the `oidcRequestUrl`?  Was this referenced in the MS Learn docs somewhere or by manually inspecting a job that was spun up?

MicroFish91: Nevermind, disregard I'm assuming it's the same format as this :)
https://learn.microsoft.com/en-us/rest/api/azure/devops/distributedtask/oidctoken/create

alexweininger: Good question. We might want to add that link to a comment above this.

PR #1724, opened 2024-05-06
azure-pipelines: Add workflow identity federation to the test pipeline template

hossam-nasr:
Modify the `test.yml` pipeline template to add support for [workflow identity federation](https://learn.microsoft.com/en-us/entra/workload-id/workload-identity-federation) for nightly testing pipelines. We do this by adding two new steps, one dummy keyvault step to connect to the newly-created `AzCodeE2ETests` service connection for this purpose. The other is a "real" Key Vault step connecting to the `AzCodeE2ETestsCredsKV` key vault, to obtain the identifiers required to connect to the `AzCodeE2ETests` service connection and obtain an OIDC token in code. Then, the needed environment variables are manually propagated to the `Test` step.


azure-pipelines/templates/test.yml
```diff
@@ -1,4 +1,24 @@
 steps:
+  # This gives the AzCodeE2ETests service connection access to this pipeline.
```
alexweininger, line 2: These templates (including test.yml) are used by all of our extensions. I believe these steps will fail for most of the extensions that use them, but A. don't have the service connection setup and B. don't have tests that need a service connection.

We need a way to let extensions opt-in to these steps.

I don't have a recommended solution for this yet, but wanted to post this feedback so we can discuss.

alexweininger: Is there an environment variable we could check to conditionally add these steps? That's one idea I have

hossam-nasr: That's a good callout, Alex! That was one of my concerns too. We did talk about potentially adding a `useAzureFederatedIdentityCredentials` environment variable or something, but we haven't finalized this. I will say though, I don't think these two steps should make any pipeline fail perse? It'll just be annoying because they will need to be independently authorized to use the `AzCodeE2ETests` service connection. But beyond that, the environment variables will be propagated and nothing done with them 🤷🏽

alexweininger: > It'll just be annoying because they will need to be independently authorized to use the `AzCodeE2ETests` service connection.

I guess I was thinking it was unreasonable to do this for every extension pipeline even though they won't be used. We aren't the only team using these templates. (ex: the Docker extension) I'm not sure we want to force everyone to add it before merging.

azure-pipelines/templates/test.yml
```diff
@@ -1,4 +1,26 @@
...
+  - task: AzureKeyVault@1
+    displayName: 'Authorize AzCodeE2ETests service connection'
```
bwateratmsft, line 4: Can you prepend the `displayName` with this emoji: 👉
It's little but it helps visually distinguish which steps in the pipeline are ours vs. coming from the 1ES pipeline templates.

```suggestion
    displayName: '\U0001F449 Authorize AzCodeE2ETests service connection'
```

azure-pipelines/templates/test.yml
```diff
@@ -1,4 +1,26 @@
...
+    inputs:
+      azureSubscription: 'AzCodeE2ETests'
+      KeyVaultName: 'AzCodeE2ETestsDummyKV'
+    condition: eq($AzCode_UseAzureFederatedCredentials, true)
+    env:
+      SYSTEM_ACCESSTOKEN: $(System.AccessToken)
+
+  # This task obtains the values necessary to connect to the AzCodeE2ETests service connection from the AzCodeE2ETestsCredKV key vault.
+  - task: AzureKeyVault@2
+    displayName: 'Get AzCodeE2ETests service connection secrets'
```
bwateratmsft, line 14:
```suggestion
    displayName: '\U0001F449 Get AzCodeE2ETests service connection secrets'
```

azure-pipelines/templates/test.yml
bwateratmsft: e.g. 
image

azure-pipelines/templates/test.yml
hossam-nasr: Okay so I added the condition, but it looks like the pipeline still needs to be authenticated for the service connection regardless, which makes sense, since the condition could be a runtime variable. That said, it looks like these pipelines are only used on main in the nightly schedule, since it looks like our GitHub pipelines don't use these 1ES templates. In that case, does it make sense that these pipelines will all now need to be permitted to use the service connections? Bear in mind it's only a one time action

azure-pipelines/templates/test.yml
hossam-nasr: Done!

hossam-nasr:
Updated the main `1esmain.yml` pipeline such that it accepts a `useAzureFederatedCredentials` parameter, which is passed all the way down to the `test.yml` template, which includes the relevant key vault steps (and sets the relevant env vars) only if `useAzureFederatedCredentials` is set to `true`. I tested this on a few pipelines and confirmed that if `useAzureFederatedCredentials` is `false` (the default, basically no change on the client extensions side), the pipeline runs the same test step as before, and without prompting for permissions. If `useAzureFederatedCredentials` parameter is set to `true`, that's when the Key Vault steps are included and env vars are propagated, and that's when permissions are prompted. Since this is a template parameter and not a runtime condition, this is figured out at "compile time", no two `.yml` files needed! @alexweininger Who's the plumber now? 😎

alexweininger:
> Updated the main `1esmain.yml` pipeline such that it accepts a `useAzureFederatedCredentials` parameter, which is passed all the way down to the `test.yml` template, which includes the relevant key vault steps (and sets the relevant env vars) only if `useAzureFederatedCredentials` is set to `true`. I tested this on a few pipelines and confirmed that if `useAzureFederatedCredentials` is `false` (the default, basically no change on the client extensions side), the pipeline runs the same test step as before, and without prompting for permissions. If `useAzureFederatedCredentials` parameter is set to `true`, that's when the Key Vault steps are included and env vars are propagated, and that's when permissions are prompted. Since this is a template parameter and not a runtime condition, this is figured out at "compile time", no two `.yml` files needed!

That is epic, great work.

> Who's the plumber now? 😎

I'm happy to hand over my denim overalls and tool box  👷  🪠  🧰

azure-pipelines/1esmain.yml
```diff
@@ -1,3 +1,8 @@
+parameters:
+- name: useAzureFederatedCredentials
```
alexweininger, line 2: Can you add a comment here explaining what this variable does for someone who doesn't want to become a plumber to figure it out?

alexweininger (dismissed):
tiny feedback, otherwise looks good to go

azure-pipelines/1esmain.yml
hossam-nasr: Yes lol, done. Does it make sense now?

alexweininger (approved):

PR #1727, opened 2024-05-08
auth: Add `AzureDevOpsSubscriptionProvider` documentation and workflow identity fedration guide to the README

hossam-nasr:
[auth package] modify the README to add documentation of the newly created `AzureDevOpsSubscriptionProvider`, and include a guide on how to set up workflow identity federation on ADO to leverage that newly created class.


alexweininger (commented):
I'm gonna let Nathan approve this. My only feedback is to put it below the logs section. I'm also wondering if this should be a completely separate .md file

nturinski:
Hmm, I agree that it should be its own file given now large it is.

Hossam also mentioned offline that it would be better to let users pass in the tenant id, service principal id, and the secret via parameters so they don't have to necessarily rely on the same exact environment variables that we are. 

I agree with that idea since I'm sure people would like the option of being able to hard code the values for initial testing, acquire the values somewhere else, etc.

hossam-nasr:
If it's going to be a separate file, where do you think it should live?

alexweininger:
It can live in the root of this package.

hossam-nasr:
Updated this to have the guide in a separate file + added more screenshots for easier guidance and readability :)

auth/AzureFederatedCredentialsGuide.md
```diff
@@ -0,0 +1,173 @@
...
+
+These are:
+
+- `serviceConnectionId`: The resource ID of the service connection created in step 2, which can be found on the `resourceId` field of the URL at the address bar, when viewing the service connection in the Azure DevOps portal
+- `domain`: The `Tenant ID` field of the service connection properties, which can be accessed by clicking "Edit" on the service connection page
+- `clientId`: The `Service Principal Id` field of the service connection properties, which can be accessed by clicking "Edit" on the service connection page
+
+  ![identifier values for service connection](guide-imgs/identifier_values.jpg)
+
+Make you pass an object containing these variables for the `new AzureDevOpsServiceProvider()` constructor. These values are _not_ secrets, so they can be set as environment variables, assigned as pipeline variables in ADO, accessed and assigned using an Azure Key Vault step, or even manually hardcoded in code (not recommended).
```
alexweininger, line 173:
```suggestion
Make sure you pass an object containing these variables for the `new AzureDevOpsServiceProvider()` constructor. These values are _not_ secrets, so they can be set as environment variables, assigned as pipeline variables in ADO, accessed and assigned using an Azure Key Vault step, or even manually hardcoded in code (not recommended).
```

nturinski:
These screenshots are 🔥

alexweininger:
> These screenshots are 🔥

Yeah was going to say this too. This is really great. I'm only a tiny bit worried about keeping this up-to-date, but 🤷

alexweininger (dismissed):

nturinski (dismissed):

bwateratmsft (commented):
Make sure these images are in the .npmignore so we aren't unintentionally bundling them with the NPM package

MicroFish91 (approved):
Super useful, great job :)

hossam-nasr:
@bwateratmsft Good call out, done!

@nturinski @alexweininger @MicroFish91 This actually dismissed y'all's approval rip, can you approve again? <3

alexweininger (approved):

PR #1729, opened 2024-05-14
auth: Refactor `AzureDevOpsSubscriptionProvider` so that it accepts values as arguments

hossam-nasr:
[auth package] modify the `AzureDevOpsSubscriptionProvider` and `AzureDevOpsSubscriptionProviderFactory` so that the three values required to identify the service connection: service connection ID, domain, and client ID, are passed as arguments to the constructor, rather than defaulting to being set as environment variables in the way we currently do. This allows for more versatility and flexibility in the way this provider can be used by other parties and users wanting to do their own E2E testing

### DIFF
--- a/auth/package.json
+++ b/auth/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@microsoft/vscode-azext-azureauth",
     "author": "Microsoft Corporation",
-    "version": "2.4.0",
+    "version": "2.5.0",
     "description": "Azure authentication helpers for Visual Studio Code",
     "tags": [
         "azure",
--- a/auth/package.json
+++ b/auth/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@microsoft/vscode-azext-azureauth",
     "author": "Microsoft Corporation",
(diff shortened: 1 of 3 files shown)


alexweininger:
The most nitpick I've ever done: Can you prefix the PRs like "auth:" instead of "[auth]". We've been doing it that way and I want to stay consistent (as much as possible).

hossam-nasr:
@alexweininger lol yes done 🫡

nturinski (dismissed):
LGTM 🫡

auth/package.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "@microsoft/vscode-azext-azureauth",
     "author": "Microsoft Corporation",
-    "version": "2.4.0",
+    "version": "2.5.0",
```
nturinski, line 4: Though maybe just 2.4.1?

hossam-nasr: Yeah fair, changed to `2.4.1`

nturinski (approved):

PR #1730, opened 2024-05-16
appservice: Make createParsedSite async to retrieve defaultHostName if it doesn't exist

nturinski:
Fixes https://github.com/microsoft/vscode-azurefunctions/issues/4121

Hopefully.

The LIST call sometimes won't retrieve `defaultHostName`. This seems to be happening somewhat randomly to people, but in our testing, the GET site call will always retrieve all the properties for that site, so we can use that to populate the property.

If it doesn't work, it'll throw an error which is somewhat intentional because this property is pretty much a requirement.

This isn't _technically_ a breaking change because `new ParsedSite` will still work, it just won't throw an error if `defaultHostName` is undefined. That'll come whenever they use a command that requires it 😅

### DIFF
--- a/appservice/package-lock.json
+++ b/appservice/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "@microsoft/vscode-azext-azureappservice",
-    "version": "3.2.1",
+    "version": "3.3.0",
     "lockfileVersion": 2,
     "requires": true,
     "packages": {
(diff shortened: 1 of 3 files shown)


nturinski:
Closing in favor of: https://github.com/microsoft/vscode-azurefunctions/pull/4135

PR #1732, opened 2024-05-29
auth: Allow passing in a tenantId when running the signIntoTenant command

motm32:
This is needed to allow signing into tenants from the tenants view being added to resource groups. See this issue for more info: https://github.com/microsoft/vscode-azureresourcegroups/issues/849


alexweininger (commented):
Why don't we just use `subscriptionProvider.signIn(tenantId)` directly?

PR #1740, opened 2024-06-05
appservice: Bump version for releasing

nturinski:


alexweininger (approved):

nturinski:
Not sure why I never merged this, but we're way past this version number now so 🤷

PR #1744, opened 2024-06-10
azure-pipelines: Add parameter to manually initiate long running tests

MicroFish91:
Updates to the base templates so that we can manually trigger long running tests.

See example branch of it hooked up:
https://github.com/microsoft/vscode-azurecontainerapps/pull/689

Example:
image


MicroFish91:
Closing, decided to use the one parameter instead of two.

PR #1754, opened 2024-07-08
appservice: Fix containerized function app name validation to exclude uppercase letters 

motm32:
This is what the name validation on the portal now says: 
image

It matches the validation from container apps so using the regular expression from there.

### DIFF
--- a/appservice/src/createAppService/SiteNameStep.ts
+++ b/appservice/src/createAppService/SiteNameStep.ts
@@ -113,7 +113,7 @@ export class SiteNameStep extends AzureNameStep<SiteNameStepWizardContext> {
 
         if (name.length < siteNamingRules.minLength || name.length > siteNamingRules.maxLength) {
             return vscode.l10n.t('The name must be between {0} and {1} characters.', siteNamingRules.minLength, siteNamingRules.maxLength);
-        } else if ((!/^[a-z]([-a-z0-9]*[a-z0-9])?$/.test(name)) && this._siteFor === "containerizedFunctionApp") {
+        } else if (this._siteFor === "containerizedFunctionApp" && (!/^[a-z]([-a-z0-9]*[a-z0-9])?$/.test(name))) {
             return vscode.l10n.t("A name must consist of lower case alphanumeric characters or '-', start with an alphabetic character, and end with an alphanumeric character and cannot have '--'.");
         } else if (siteNamingRules.invalidCharsRegExp.test(name)) {
             return vscode.l10n.t("The name can only contain letters, numbers, or hyphens.");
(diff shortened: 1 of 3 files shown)


appservice/src/createAppService/SiteNameStep.ts
```diff
@@ -113,6 +113,8 @@ export class SiteNameStep extends AzureNameStep<SiteNameStepWizardContext> {
 
         if (name.length < siteNamingRules.minLength || name.length > siteNamingRules.maxLength) {
             return vscode.l10n.t('The name must be between {0} and {1} characters.', siteNamingRules.minLength, siteNamingRules.maxLength);
+        } else if ((!/^[a-z]([-a-z0-9]*[a-z0-9])?$/.test(name)) && this._siteFor === "containerizedFunctionApp") {
```
nturinski, line 116: nit: You should check the siteFor property first so the conditional can just abort if it's not a containerized app. It's probably small, but still a wasted check.

nturinski (dismissed):

nturinski (approved):
Should go ahead and bump so we can release and add to Functions asap.

appservice/src/createAppService/SiteNameStep.ts
```diff
@@ -113,6 +113,8 @@ export class SiteNameStep extends AzureNameStep<SiteNameStepWizardContext> {
 
         if (name.length < siteNamingRules.minLength || name.length > siteNamingRules.maxLength) {
             return vscode.l10n.t('The name must be between {0} and {1} characters.', siteNamingRules.minLength, siteNamingRules.maxLength);
+        } else if (this._siteFor === "containerizedFunctionApp" && (!/^[a-z]([-a-z0-9]*[a-z0-9])?$/.test(name))) {
```
MicroFish91, line 116: I think the `*` might be switched up here, otherwise I think you can still have multiple `-`'s.

There's a pretty sweet regexp you made for `ContainerAppNameStep` that might be good to copy over here.  Seems like the rules might be pretty similar

motm32: Oh lol yes I copied this from `ManagedEnvironmentNameStep` since I figured it was the same between the two but I can change it 

On another note shouldn't the regex be the same between `ContainerAppNameStep` and `ManagedEnvironmentNameStep`  since the error message is the same between the two?

MicroFish91: Yeah good catch, if the requirements are the same I think we should update that check logic

MicroFish91 (approved):

MicroFish91 (approved):

PR #1755, opened 2024-07-09
Reference the `MicroBuildTemplate` from the `1ESPipelineTemplates/`

MicroFish91:
Still working with 1ES team to get the release pipeline working again.  They mentioned that we probably shouldn't be referencing the project directly.

Similar convention shown here: https://devdiv.visualstudio.com/DevDiv/_wiki/wikis/DevDiv.wiki/40112/Publishing-to-3rd-party-package-managers

### DIFF
--- a/azure-pipelines/1es-release-npm.yml
+++ b/azure-pipelines/1es-release-npm.yml
@@ -20,8 +20,8 @@ resources:
   repositories:
     - repository: MicroBuildTemplate
       type: git
-      name: MicroBuildTemplates/MicroBuildTemplates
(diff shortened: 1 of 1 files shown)


bwateratmsft (approved):

alexweininger:
What's the status on this?

MicroFish91:
> What's the status on this?

The release pipeline should be working again!

MicroFish91:
The release pipeline is working again, I'm not going to bother with this change since it's working as-is and I don't really want to mess with it.  If need be we can always reopen and update this reference down the road

PR #1756, opened 2024-07-09
Add static methods and export `AppServicePlanRedundancyStep`

MicroFish91:
Using the new static methods on this step to help fix the upstream tests in the app service extension.

See here for reference:
https://github.com/microsoft/vscode-azureappservice/pull/2653


appservice/src/index.ts
```diff
@@ -3,36 +3,37 @@
...
-export * from './KuduModels';
-export * from './SiteClient';
-export * from './TunnelProxy';
 export * from './confirmOverwriteSettings';
 export * from './createAppService/AppInsightsCreateStep';
 export * from './createAppService/AppInsightsListStep';
 export { AppKind, WebsiteOS } from './createAppService/AppKind';
 export * from './createAppService/AppServicePlanCreateStep';
 export * from './createAppService/AppServicePlanListStep';
+export * from './createAppService/AppServicePlanRedundancyStep';
```
nturinski, line 12: nit: Do you know what's changing all of the ordering of the exports? Do you have a different linter configuration or something?

appservice/src/createAppService/AppServicePlanRedundancyStep.ts
```diff
@@ -46,25 +46,37 @@ export class AppServicePlanRedundancyStep extends AzureWizardPromptStep<IAppServ
             'eastus2euap',
         ];
 
+        location = location.replace(/\s/, "").toLowerCase();
```
nturinski, line 49: I don't see this in the previous code. What did you run into that prompted you to make this change?

nturinski (approved):

appservice/src/createAppService/AppServicePlanRedundancyStep.ts
MicroFish91: The data that gets stored for the location seems to be different than the name.  So the name of our quick pick is usually something like `East US` but will be stored as 'eastus'.  I basically wanted to enable being able to check locations by passing in either type of field, not just the latter since we typically use the name of the pick as part of the inputs list of testing

appservice/src/index.ts
MicroFish91: I assume it's from this:
image

I think at some point this got turned on by default for me as I don't ever remember explicitly setting this.  It's basically just alphabetizing the imports which I feel like is good practice anyway, but I'm happy to manually change back the other imports and save without format if that's preferred

appservice/src/createAppService/AppServicePlanRedundancyStep.ts
nturinski: Take a look at `LocationListStep`. Could you use `locationMatchesName` or `generalizeLocationName` for the same task?

appservice/src/index.ts
nturinski: Maybe just add that to the .vscode settings of the repo then? I don't have a strong preference, I'd just want us to align so we don't have to keep seeing diffs between export alphabetizing.

appservice/src/createAppService/AppServicePlanRedundancyStep.ts
MicroFish91: `generalizeLocationName` would be perfect, but it's not exported, so would need an update to azure package and release.

MicroFish91: Added a todo for this item since not currently expored

appservice/src/index.ts
MicroFish91: Reverted, maybe we can discuss standardizing formatting a bit more with the team at some later date

nturinski (approved):

Kwpumpa:
[](url)

Kwpumpa:
[Uploading Backup-codes-kwpumpa.txt…]()

Kwpumpa:
[account-statement_2024-12-01_2024-12-06_en_78834c.pdf](https://github.com/user-attachments/files/18221671/account-statement_2024-12-01_2024-12-06_en_78834c.pdf)

PR #1757, opened 2024-07-19
azure: Add wizard steps to list/create `UserAssignedIdentities` and execute role definitions

nturinski:
See an example of this being used here: https://github.com/microsoft/vscode-azurefunctions/pull/4213/files

This adds `managedIdentity` to the `ResourceWizardContext` since managed identity will be included in a lot of our resources.

The `UserAssignedIdentity` steps are for listing and creating new identities. Right now, it will always default to creating it in the resource group. That reminds me, I need to test what happens if I pick an existing resource group that already has an MI in it. I'm not even sure what behavior should be-- create a new one with a suffix or use the existing one?

The `RoleAssignmentStep` should work for both system and user assigned identities. Because the scope will be created _during_ the wizard, you need to pass a function that returns the property that you want since the context will be updated by the time the function is called. Execute priority is also 900 because in most cases, it should always be the last thing to execute.


nturinski:
I should probably revisit exposing the scope since it contains the subscription id of the resource as well. Probably can just split by `/` and take the last two indices.
`3:39:20 PM: Role assignment "StorageBlobDataContributor" created with resource "/subscriptions/9b5c7ccb-9857-4307-843b-8875e83f65e9/resourceGroups/naturinsmi22/providers/Microsoft.Storage/storageAccounts/naturinsmi22".`

azure/src/clients.ts
```diff
@@ -27,6 +29,18 @@ export async function createResourcesClient(context: InternalAzExtClientContext)
     }
 }
 
+export async function createManagedServiceIdentityClient(context: InternalAzExtClientContext): Promise<ManagedServiceIdentityClient> {
+    return createAzureClient(context, (await import('@azure/arm-msi')).ManagedServiceIdentityClient);
+}
+
+export async function createAuthorizationManagementClient(context: InternalAzExtClientContext): Promise<AuthorizationManagementClient> {
+    if (parseClientContext(context).isCustomCloud) {
+        return <AuthorizationManagementClient><unknown>createAzureClient(context, (await import('@azure/arm-authorization-profile-2020-09-01-hybrid')).AuthorizationManagementClient);
```
MicroFish91, line 38: Oh nice, so the hybrid profile version was made to support custom clouds?

alexweininger:
> TODO: Add more description on Monday 😅

Time is running out

nturinski:
> I should probably revisit exposing the scope since it contains the subscription id of the resource as well. Probably can just split by `/` and take the last two indices. `3:39:20 PM: Role assignment "StorageBlobDataContributor" created with resource "/subscriptions/9b5c7ccb-9857-4307-843b-8875e83f65e9/resourceGroups/naturinsmi22/providers/Microsoft.Storage/storageAccounts/naturinsmi22".`

New message looks like this:
image

alexweininger (commented):
Initial pass

azure/src/clients.ts
nturinski: Discussed offline, but the api-versions for custom clouds isn't kept up to date as quickly as standard Azure SDKs.

alexweininger (approved):

PR #1762, opened 2024-08-06
auth: Add a `getSessionWithScopes` to AzureAuthentication

nturinski:
`getSession` would always end up retrieving the session without a scope. When we tried to fix it via this [PR](https://github.com/microsoft/vscode-azuretools/pull/1640/files), there ended up being a lot of scope issues specifically around deployment.

However, the need for a scoped token has come up again and is a requirement for security work. The easiest workaround to all the brokenness was to just make a new function that retrieves a new session while leaving the old one version alone.

This will be called by the Resources extension so that the user will not need to log in again. Here's that PR https://github.com/microsoft/vscode-azureresourcegroups/compare/nat/scopedSessions?expand=1

### DIFF
--- a/auth/src/AzureAuthentication.ts
+++ b/auth/src/AzureAuthentication.ts
@@ -15,7 +15,7 @@ export interface AzureAuthentication {
      *
      * @returns A VS Code authentication session or undefined, if none could be obtained.
      */
-    getSession(scopes?: string[]): vscode.ProviderResult<vscode.AuthenticationSession>;
+    getSession(): vscode.ProviderResult<vscode.AuthenticationSession>;
     /**
      * Gets a VS Code authentication session for an Azure subscription.
      *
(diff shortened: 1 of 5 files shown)


bwateratmsft (dismissed):

auth/src/AzureAuthentication.ts
```diff
@@ -9,12 +9,19 @@ import type * as vscode from 'vscode';
  * Represents a means of obtaining authentication data for an Azure subscription.
  */
 export interface AzureAuthentication {
+    /**
+     * Gets a VS Code authentication session for an Azure subscription.
+     * Always uses the default scope, `https://management.azure.com/.default/.default.`
```
alexweininger, line 14: Clarify here that it will respect the `microsoft-sovereign-cloud.environment` setting.

alexweininger (approved):

PR #1763, opened 2024-08-07
azure, utils: Improve and dedupe `randomUtils`

bwateratmsft:
Random bytes are more random than just partial GUIDs. Not all GUID parts are created _as_ randomly.


utils/src/utils/randomUtils.ts
```diff
@@ -18,7 +18,13 @@ export namespace randomUtils {
...
 
     export function getRandomHexString(length: number = 6): string {
-        return crypto.randomUUID().slice(0, length);
+        if (length <= 0) {
+            throw new Error(`Length must be strictly positive`);
+        }
+
+        const array = new Uint8Array(Math.ceil(length / 2)); // Each byte is represented by 2 hex characters
+        crypto.getRandomValues(array);
+        return Buffer.from(array).toString('hex').slice(0, length);
```
bwateratmsft, line 27: @alexweininger is `Buffer` going to cause problems?

alexweininger: It could, but I think it's fine to use here since:
* Our team isn't currently planning on shipping any extensions that run in web only contexts.
* It would be up to the consuming extension to polyfill `Buffer`.

azure/src/utils/randomUtils.ts
alexweininger, line 1: The build passed so I'm guessing nothing was using this?

alexweininger (approved):

PR #1766, opened 2024-08-07
utils: Allow execute steps to contribute success or fail items for display in the activity log

MicroFish91:
This implementation will enable execute steps to contribute success or fail items (as activity children) to the activity log upon execution.  The resulting `ExecuteActivityOutput` will be handled by the Azure Wizard which has been updated accordingly.

Example of what this would look like in the activity log:

Example with `continueOnFail` set to true when a step fails:
image

Example of how it will probably look implemented in ACA:
https://github.com/microsoft/vscode-azurecontainerapps/pull/722


utils/src/wizard/AzureWizard.ts
```diff
@@ -180,16 +187,48 @@ export class AzureWizard<T extends (IInternalActionContext & Partial<types.Execu
...
+                    if (!step.options.continueOnFail) {
+                        throw e;
+                    }
+                } finally {
+                    output ??= {};
+                    this.displayActivityOutput(output, step.options);
+
                     currentStep += 1;
+                    step = steps.pop();
                 }
```
alexweininger, line 211: Could you explain the changes made here a bit? I just want to make sure we understand them fully before making changes to the wizard since it's so central to so many commands.

MicroFish91: Yeah sure, so basically all execute steps can now have the following methods: `createFailOutput` & `createSuccessOutput`.  The output return signature is basically just an object with two optional properties: `message: string` and `treeItem: AzExtTreeItem` (the tree item will be pushed in as an activity child).  These outputs indicate what needs to be displayed to the user upon successful or unsuccessful execution of an activity step.

Flow for successful execution:
- In the try block, assuming the step executes without throwing an error
- (Line 199) We try to get an output from the step (optional since the method may not exist).  Output will either be the defined success output or `undefined` if not set.
- (Line 207) If `undefined`,  output will be an empty object and `displayActivityOutput` will display nothing to the user.  
- (Line 207) If defined, `displayActivityOutput` will display any defined output message to the output log, and push the tree item into the `activityChildren` array if it has been initialized.  If any options to suppress showing are set on the execute step, these will be honored here.

Flow for failed execution:
- In the try block, assuming the step executes and throws an error
- (Line 201) We try to get an output from the step (optional since the method may not exist).  Output will be either the defined fail output or `undefined` if not set.
- (Line 202) New step option allows for us to essentially swallow errors.  In some cases, we don't necessarily want the entire wizard to stop just because a step failed.  So when a step has `continueOnFail` set, don't throw the error, but still display any defined output unless it has been explicitly set to be suppressed in the step options.
the defined success output or `undefined` if not set.
- (Line 207) If `undefined`,  output will be an empty object and `displayActivityOutput` will display nothing to the user.  
- (Line 207) If defined, `displayActivityOutput` will display any defined output message to the output log, and push any tree item into the `activityChildren` array if it has been initialized.  If any options to suppress showing are set on the execute step, these will be honored here.

Rest of the code should be basically the same.  I did use a guard clause with the `shouldExecute` to try and remove some nesting.  Let me know if that makes sense / if you have any questions!

nturinski (approved):

PR #1767, opened 2024-08-12
utils: Add `onProgress` state to show activity log creating in real-time

nturinski:
This leverages the new changes to the activity tree that will automatically create children for the `onSuccess` and `onError` states. We already had an emitter for running/progress, so it was pretty easy to add another output for in progress. I'll put up a PR in VM's that showcases the changes shortly.

### DIFF
--- a/utils/.vscode/settings.json
+++ b/utils/.vscode/settings.json
@@ -10,6 +10,6 @@
         "**/node_modules": true
     },
     "editor.codeActionsOnSave": {
-        "source.fixAll.eslint": true
+        "source.fixAll.eslint": "explicit"
     }
 }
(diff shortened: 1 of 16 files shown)


MicroFish91:
Nice! I'll try testing this out in ACA a bit as well

utils/.vscode/settings.json
```diff
@@ -10,6 +10,6 @@
         "**/node_modules": true
     },
     "editor.codeActionsOnSave": {
-        "source.fixAll.eslint": true
+        "source.fixAll.eslint": "explicit"
```
nturinski, line 13: Not sure why this changed, but I noticed some of our other extensions are like that so I just left it.

MicroFish91 (approved):

PR #1768, opened 2024-08-13
utils: Ensure JWT redaction for error messages

MicroFish91:
Ensure anything resembling a JWT is automatically redacted from errors before submitting to telemetry


utils/src/masking.ts
```diff
@@ -108,6 +108,7 @@ export function maskUserInfo(unknownArg: unknown, actionValuesToMask: string[],
     data = data.replace(/[a-z]+:\/\/\S*/gi, getRedactedLabel('url'));
     data = data.replace(/\S+(?<!(?<!\-)\basp)\.(com|org|net)\S*/gi, getRedactedLabel('url'));
     data = data.replace(/\S*(key|token|sig|password|passwd|pwd)[="':\s]+\S*/gi, getRedactedLabel('key'));
+    data = data.replace(/e[^\.\s]*\.e[^\.\s]*\.[^\.\s]+/gi, getRedactedLabel('jwt'));
```
nturinski, line 111: Can you show a reference or something about how this regex was conjured?

MicroFish91: Honestly just a read around a bit and then played with header and payload values here:
https://jwt.io/

Since the header and payload will be encoded from json data, and since json data starts with "{" it seems to always get encoded starting with an e for the first and second part.  I read around and this seems to be roughly echo'd elsewhere as well.

Thus the rule I made was.... 
<Section 1>.<Section 2>.<Section 3>

Where the rules for each are...
Section 1: Start with an e, followed by any number of values that are not '.' or '\s'
Section 2: Same as Section 1
Section 3: At least 1 value that is not '.' or '\s'

We could get even more specific probably, but my general approach wasn't to be able to identify jwt's perfectly, but rather, just be able to identify anything that appears jwt-like and remove it.

nturinski (commented):
I assumed that they were fake tokens, but I did double-check to verify that they were. I am concerned that cred-scan is going to flag us for having PAT in our code now even if it is for testing purposes.

nturinski (approved):

PR #1774, opened 2024-08-23
dev: Add ddsignfiles wrapper

alexweininger:
This file is a wrapper around the DDSignFiles.dll tool that is present in pipelines that have signing setup and enabled. It has been copied from source because it's easier than consuming an npm package. We can use this js helper to sign files with a js script in each extension instead of requiring a .csproj and building it with MSBuild. I think I'd rather stay in the nodejs world if we can.

### DIFF
--- a/dev/build/SignFiles.js
+++ b/dev/build/SignFiles.js
@@ -1,0 +1,93 @@
+// File copied from https://dev.azure.com/devdiv/Engineering/_git/MicroBuild?path=/src/NodeJS/Signing/SignFiles.js&_a=contents&version=GBmain
+
+const fs = require('fs')
+const path = require('path')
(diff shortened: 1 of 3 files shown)


bwateratmsft:
Docker and several others do not use the dev package. Can we put this in the pipelines instead?

alexweininger:
> Docker and several others do not use the dev package. Can we put this in the pipelines instead?

Rats. The pipelines folder isn't currently published as a package. Do you know if I include this script in the pipelines folder if it will be accessible within a pipeline run that depends on the templates? I'll try it out.

alexweininger:
I'm starting to think it might just be easier to use MSBuild and write a csproj file into each repo 😢. Let me try that out too.

PR #1775, opened 2024-08-23
azure-pipelines: Add steps to sign extensions with MSBuild

alexweininger:
After discussions with the team, the CI/CD design we're going for is: two pipelines 1. Build and 2. Release. Build will build/package/test AND sign. And upload the signed bits. Release will just download the bits, and release it using vsce.

In this PR:
1. Move from the 1ES template to the MicroBuild template. They are basically the same so there were little breaking changes.
2. Enable the signing toolset which is auto installed thanks to the new MicroBuild template
3. Add a new sign template that uses vsce to generate a manifest (the file that is used for signing)
4. Use MSBuild to sign the extensions

The changes in the pipelines will look for a ".azure-pipelines/SignExtension.signproj" file in the repo and will sign the extension if present. I will be adding this and the below change to each extension.

Template consumers will have to add the following to their 1esmain.yml. This is required by the MicroBuild template that we are now using. I can't figure out a way to avoid this.

```
variables:
  # Required by MicroBuild template
  - name: TeamName
    value: "Azure Tools for VS Code"
```

See https://github.com/microsoft/vscode-azureresourcegroups/pull/907 for what these changes will look like.


azure-pipelines/templates/stage-artifacts.yml
```diff
@@ -0,0 +1,17 @@
+steps:
+  - task: CopyFiles@2
+    displayName: "\U0001F449 Copy packages and vsix to staging directory"
+    inputs:
+      # Uploading the package.json so we can publish later without cloning the source
```
bwateratmsft, line 5: Can you elaborate more on this?

azure-pipelines/templates/sign.yml
```diff
@@ -0,0 +1,32 @@
+steps:
+  # Check if the SignExtension.signproj file exists and set a variable using PowerShell
+  # All other steps in this template will only run if the file exists
+  - powershell: |
+      $fileExists = Test-Path -Path "$(Build.SourcesDirectory)/SignExtension.signproj"
```
bwateratmsft, line 5: Can we stuff this .signproj into `azure-pipelines` instead of at the root of the project? There's enough gunk at the root...

azure-pipelines/templates/stage-artifacts.yml
alexweininger: In the release pipeline I'm testing, I'm not cloning the source at all. I'm just downloading and releasing artifacts from a build pipeline run. In the release pipeline I need the package.json to know what extension is being released.

azure-pipelines/templates/sign.yml
alexweininger: I will try *cries in more pipelines trial and error*

alexweininger: Got it first try 😎

bwateratmsft: :shipit:

bwateratmsft (approved):

alexweininger:
I'm going to wait to merge this until most (if not all) the consumers are ready. I'm pretty sure I have all the PRs made.

PR #1776, opened 2024-08-26
utils & azure: Add new error telemetry properties

MicroFish91:
Our team is needing to suppress the original error telemetry properties and replace them with new ones (contact me offline if you need additional context).  We don't want to break anyone still relying on the old properties, so we should keep emitting those normally as well while we migrate to using the new props.


azure/src/wizard/VerifyProvidersStep.ts
```diff
@@ -39,7 +39,18 @@ export class VerifyProvidersStep<T extends ISubscriptionActionContext> extends A
...
+                /**
+                 * @param providerError
+                 * @deprecated
+                 * Continue to emit telemetry for clients who are still using this property. You should suppress this property if you need to migrate to the new replacement.
+                 *
+                 * @param providerError2
+                 * A duplicate replacement of the `providerError` telemetry property.
+                 */
+                context.telemetry.properties.providerError = perror.message;
+                context.telemetry.properties.providerError2 = perror.message;
```
nturinski, line 53: I prefer we use v2 over just 2. I think it just makes it clearer that it's intended to replace the original, not be an additional error.

utils/src/callWithTelemetryAndErrorHandling.ts
```diff
@@ -136,7 +136,18 @@ function handleError(context: types.IActionContext, callbackId: string, error: u
...
+            /**
+             * @param errorMessage
+             * @deprecated
+             * Continue to emit telemetry for clients who are still using this property. You should suppress this property if you need to migrate to the new replacement.
+             *
+             * @param errorMessage2
+             * A duplicate replacement of the `errorMessage` telemetry property which should be used instead.
+             */
             context.telemetry.properties.errorMessage = errorData.message;
+            context.telemetry.properties.errorMessage2 = errorData.message;
```
bwateratmsft, line 149: Numbers in property names get censored in telemetry (don't ask my why, they just do). Not sure of a good solution though.

bwateratmsft: At least in the GDPR catalog.

bwateratmsft: I'll take a look for sure tomorrow. Please hold on merging until then. I have an example from Docker I can look at.

bwateratmsft (changes requested):

bwateratmsft (dismissed):

PR #1777, opened 2024-08-26
utils: Export masking utility and bump version for release

MicroFish91:

### DIFF
--- a/utils/index.d.ts
+++ b/utils/index.d.ts
@@ -1754,6 +1754,7 @@ export declare namespace AzExtFsExtra {
 }
 
 export declare function maskValue(data: string, valueToMask: string | undefined): string;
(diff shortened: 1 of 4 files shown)


utils/index.d.ts
```diff
@@ -1754,6 +1754,7 @@ export declare namespace AzExtFsExtra {
 }
 
 export declare function maskValue(data: string, valueToMask: string | undefined): string;
+export declare function maskUserInfo(unknownArg: unknown, actionValuesToMask: string[], lessAggressive?: boolean, getUsername?: () => string): string;
```
bwateratmsft, line 1757: Can you add inline documentation so it's clear what the different args do?

bwateratmsft (dismissed):

PR #1789, opened 2024-09-09
auth: Add changes to be compatible with the new tenant view 

motm32:
See [here ](https://github.com/microsoft/vscode-azureresourcegroups/pull/869)for tenant view PR. 

_Currently a work in progress will not merge until the tenant view changes in resources have been finished and fully tested with these changes_

### DIFF
--- a/auth/src/AzureSubscription.ts
+++ b/auth/src/AzureSubscription.ts
@@ -5,7 +5,8 @@
 
 import type { TokenCredential } from '@azure/core-auth';
 import type { Environment } from '@azure/ms-rest-azure-env';
-import type { AzureAuthentication } from './AzureAuthentication';
+import * as vscode from "vscode";
+import { AzureAuthentication } from './AzureAuthentication';
 
 /**
  * A type representing an Azure subscription ID, not including the tenant ID.
@@ -55,4 +56,9 @@ export interface AzureSubscription {
      * The credential for authentication to this subscription. Compatible with Azure track 2 SDKs.
      */
     readonly credential: TokenCredential;
+
+    /**
+     * The account associated with this subscription. This is optional as we only need the account if there are duplicate subscriptions.
+     */
+    readonly account?: vscode.AuthenticationSessionAccountInformation;
 }
--- a/auth/src/AzureSubscription.ts
+++ b/auth/src/AzureSubscription.ts
@@ -58,7 +58,7 @@ export interface AzureSubscription {
     readonly credential: TokenCredential;
 
     /**
-     * The account associated with this subscription. This is optional as we only need the account if there are duplicate subscriptions.
+     * The account associated with this subscription.
      */
-    readonly account?: vscode.AuthenticationSessionAccountInformation;
+    readonly account: vscode.AuthenticationSessionAccountInformation;
 }
--- a/auth/src/AzureSubscriptionProvider.ts
+++ b/auth/src/AzureSubscriptionProvider.ts
@@ -15,7 +15,7 @@ export interface AzureSubscriptionProvider {
      * Gets a list of tenants available to the user.
      * Use {@link isSignedIn} to check if the user is signed in to a particular tenant.
      *
-     * @param session - Optionally pass in a specific vscode sesssion to get tenants for.
+     * @param account - Optionally pass in a specific account to get tenants for.
      *
      * @returns A list of tenants.
      */
--- a/auth/src/AzureSubscriptionProvider.ts
+++ b/auth/src/AzureSubscriptionProvider.ts
@@ -19,7 +19,7 @@ export interface AzureSubscriptionProvider {
      *
      * @returns A list of tenants.
      */
-    getTenants(session?: vscode.AuthenticationSession): Promise<TenantIdDescription[]>;
+    getTenants(account?: vscode.AuthenticationSessionAccountInformation): Promise<TenantIdDescription[]>;
 
     /**
      * Gets a list of Azure subscriptions available to the user.
--- a/auth/src/AzureSubscriptionProvider.ts
+++ b/auth/src/AzureSubscriptionProvider.ts
@@ -3,9 +3,9 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import type { TenantIdDescription } from '@azure/arm-resources-subscriptions';
 import type * as vscode from 'vscode';
 import type { AzureSubscription } from './AzureSubscription';
-import type { TenantIdDescription } from '@azure/arm-resources-subscriptions';
 
 /**
  * An interface for obtaining Azure subscription information
@@ -15,9 +15,11 @@ export interface AzureSubscriptionProvider {
      * Gets a list of tenants available to the user.
      * Use {@link isSignedIn} to check if the user is signed in to a particular tenant.
      *
+     * @param session - Optionally pass in a specific vscode sesssion to get tenants for.
+     *
      * @returns A list of tenants.
      */
-    getTenants(): Promise<TenantIdDescription[]>;
+    getTenants(session?: vscode.AuthenticationSession): Promise<TenantIdDescription[]>;
 
     /**
      * Gets a list of Azure subscriptions available to the user.
--- a/auth/src/VSCodeAzureSubscriptionProvider.ts
+++ b/auth/src/VSCodeAzureSubscriptionProvider.ts
@@ -59,13 +59,15 @@ export class VSCodeAzureSubscriptionProvider extends vscode.Disposable implement
      * @returns A list of tenants.
      */
     public async getTenants(account?: vscode.AuthenticationSessionAccountInformation): Promise<TenantIdDescription[]> {
-        const session = await getSessionFromVSCode(undefined, undefined, { account });
-        const { client } = await this.getSubscriptionClient(undefined, undefined, session);
-
         const results: TenantIdDescription[] = [];
+        const accounts = account ? [account] : await vscode.authentication.getAccounts(getConfiguredAuthProviderId());
 
-        for await (const tenant of client.tenants.list()) {
-            results.push(tenant);
+        for (account of accounts) {
+            const { client } = await this.getSubscriptionClient(account, undefined, undefined);
+
+            for await (const tenant of client.tenants.list()) {
+                results.push(tenant);
+            }
         }
 
         return results;
@@ -212,7 +214,7 @@ export class VSCodeAzureSubscriptionProvider extends vscode.Disposable implement
      * @returns The list of subscriptions for the tenant.
      */
     private async getSubscriptionsForTenant(tenantId: string, account: vscode.AuthenticationSessionAccountInformation): Promise<AzureSubscription[]> {
-        const { client, credential, authentication } = await this.getSubscriptionClient(tenantId, undefined, undefined, account);
+        const { client, credential, authentication } = await this.getSubscriptionClient(account, tenantId, undefined);
         const environment = getConfiguredAzureEnv();
 
         const subscriptions: AzureSubscription[] = [];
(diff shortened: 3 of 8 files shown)


auth/src/AzureSubscription.ts
```diff
@@ -55,4 +56,9 @@ export interface AzureSubscription {
      * The credential for authentication to this subscription. Compatible with Azure track 2 SDKs.
      */
     readonly credential: TokenCredential;
+
+    /**
+     * The account associated with this subscription. This is optional as we only need the account if there are duplicate subscriptions.
+     */
+    readonly account?: vscode.AuthenticationSessionAccountInformation;
```
alexweininger, line 63: Why is this optional? It seems like it's always set.

auth/src/VSCodeAzureSubscriptionProvider.ts
```diff
@@ -58,8 +58,8 @@ export class VSCodeAzureSubscriptionProvider extends vscode.Disposable implement
      *
      * @returns A list of tenants.
      */
-    public async getTenants(): Promise<TenantIdDescription[]> {
-        const { client } = await this.getSubscriptionClient();
+    public async getTenants(session?: vscode.AuthenticationSession): Promise<TenantIdDescription[]> {
```
alexweininger, line 61: I think we should change the parameter to `account?: vscode.AuthenticationSessionAccountInformation` and then acquire the session within the function. That way consumers only have to list accounts and then provide them instead of listing accounts, getting the session, then passing that in.

See what I mean in the RGs PR here: https://github.com/microsoft/vscode-azureresourcegroups/pull/869/files#diff-c145c307da953aefde25e0a2d004387be9aba78ad9e596af5628ef25b4074246R57-R60


```ts
                const accounts = await vscode.authentication.getAccounts(getConfiguredAuthProviderId());
                for (const account of accounts) {
                    const session = await vscode.authentication.getSession(getConfiguredAuthProviderId(), getScopes(undefined), { account: account });
                    const tenants = await this.getTenants(session);
```

It eliminates the need to expose the `getScopes()` logic.

alexweininger: I tried this out locally and it seemed to work well.

motm32: Ok sounds good I will change it over to this logic.

auth/src/AzureSubscription.ts
motm32: I think I originally had this optional since we only wanted to pass in the account when there was a duplicate but since then we changed the logic to always pass the account. I'll make this a required property.

motm32: Actually looking further into it `AzureDevOpsSubscriptionProvider` uses a version of `AzureSubscription` without the account attached to it. So if i make it not optional it will break this and I'm not sure if that is something we would want to change.

alexweininger: Ok, I'll look into it.

auth/src/VSCodeAzureSubscriptionProvider.ts
```diff
@@ -58,7 +58,8 @@ export class VSCodeAzureSubscriptionProvider extends vscode.Disposable implement
      *
      * @returns A list of tenants.
      */
-    public async getTenants(session?: vscode.AuthenticationSession): Promise<TenantIdDescription[]> {
+    public async getTenants(account?: vscode.AuthenticationSessionAccountInformation): Promise<TenantIdDescription[]> {
+        const session = await vscode.authentication.getSession(getConfiguredAuthProviderId(), ['https://management.azure.com/.default'], { account: account });
```
alexweininger, line 62: We can't hardcode the management scope here since it needs to change based on the sovereign cloud setting.

alexweininger: Otherwise the changes are 💯

alexweininger, line 62:
```suggestion
        const session = await getSessionFromVSCode(undefined, undefined, { account });
```

auth/src/AzureSubscriptionProvider.ts
```diff
@@ -15,9 +15,11 @@ export interface AzureSubscriptionProvider {
      * Gets a list of tenants available to the user.
      * Use {@link isSignedIn} to check if the user is signed in to a particular tenant.
      *
+     * @param session - Optionally pass in a specific vscode sesssion to get tenants for.
```
bwateratmsft, line 18: Shouldn't this be:
```suggestion
     * @param account - Optionally pass in a specific account to get tenants for.
```

auth/src/VSCodeAzureSubscriptionProvider.ts
```diff
@@ -91,24 +92,28 @@ export class VSCodeAzureSubscriptionProvider extends vscode.Disposable implement
...
-
-                // If the user is not signed in to this tenant, then skip it
-                if (!(await this.isSignedIn(tenantId))) {
-                    continue;
-                }
-
-                // For each tenant, get the list of subscriptions
-                results.push(...await this.getSubscriptionsForTenant(tenantId));
             }
+
```
bwateratmsft, line 116: Nit: spurious newline added

auth/src/AzureSubscriptionProvider.ts
alexweininger: @motm32 did you see this

motm32: Nope! Shall change that now.

alexweininger:
Get the build to pass and then verify that all the changes are compatible with https://github.com/microsoft/vscode-azureresourcegroups/pull/869

motm32:
> Get the build to pass and then verify that all the changes are compatible with [microsoft/vscode-azureresourcegroups#869](https://github.com/microsoft/vscode-azureresourcegroups/pull/869) and then we can work on releasing this

I have done some prleiminary testing with the updates to this package and the resources PR but going to do some more thorough testing today. Should be hopefully ready to go sometime tomorrow if testing goes well.

auth/src/VSCodeAzureSubscriptionProvider.ts
```diff
@@ -58,8 +58,9 @@ export class VSCodeAzureSubscriptionProvider extends vscode.Disposable implement
      *
      * @returns A list of tenants.
      */
-    public async getTenants(): Promise<TenantIdDescription[]> {
-        const { client } = await this.getSubscriptionClient();
+    public async getTenants(account?: vscode.AuthenticationSessionAccountInformation): Promise<TenantIdDescription[]> {
```
alexweininger, line 61: What's the behavior when we don't pass the account arg here? I know that was the previous behavior, but I still have no idea what's supposed to happen. I mostly want to know so we can document it.

alexweininger: Follow up: Althought it'd be a breaking change, it might be worth investigating making the account arg required. Especially if the behavior of not passing it in is unknown/indeterminate.

auth/src/VSCodeAzureSubscriptionProvider.ts
```diff
@@ -237,15 +242,21 @@ export class VSCodeAzureSubscriptionProvider extends vscode.Disposable implement
...
+    private async getSubscriptionClient(tenantId?: string, scopes?: string[], session?: vscode.AuthenticationSession, account?: vscode.AuthenticationSessionAccountInformation): Promise<{ client: SubscriptionClient, credential: TokenCredential, authentication: AzureAuthentication }> {
         const armSubs = await import('@azure/arm-resources-subscriptions');
-        const session = await getSessionFromVSCode(scopes, tenantId, { createIfNone: false, silent: true });
         if (!session) {
-            throw new NotSignedInError();
+            if (account) {
+                session = await getSessionFromVSCode(scopes, tenantId, { createIfNone: false, silent: true, account });
+            } else {
+                session = await getSessionFromVSCode(scopes, tenantId, { createIfNone: false, silent: true });
+            }
```
alexweininger, line 252: I believe this would be the same, but without the if statement
```suggestion
            session = await getSessionFromVSCode(scopes, tenantId, { createIfNone: false, silent: true, account });
```

auth/src/VSCodeAzureSubscriptionProvider.ts
```diff
@@ -237,15 +242,21 @@ export class VSCodeAzureSubscriptionProvider extends vscode.Disposable implement
...
         }
 
         const credential: TokenCredential = {
             getToken: async () => {
+                if (!session) {
+                    throw new NotSignedInError();
+                }
```
alexweininger, line 259: We moved this check into this callback. Before we did it right after we fetched the session. I tried it out and both seemed to satisfy TS.

Was there a reason to put it into this callback?

auth/src/VSCodeAzureSubscriptionProvider.ts
motm32: image

auth/src/VSCodeAzureSubscriptionProvider.ts
motm32: When I had it outside the callback the build was failing for me. TS wasn't giving me any errors until I ran `npm run build` then is gave me a session may be undefined within the callback.

auth/src/VSCodeAzureSubscriptionProvider.ts
motm32: I think that was a passed behavior. I'll work on making it a required property.

auth/src/VSCodeAzureSubscriptionProvider.ts
alexweininger: Ok I'll work on this locally and get back to you

auth/src/VSCodeAzureSubscriptionProvider.ts
```diff
@@ -237,15 +242,21 @@ export class VSCodeAzureSubscriptionProvider extends vscode.Disposable implement
      *
      * @returns A client, the credential used by the client, and the authentication function
      */
-    private async getSubscriptionClient(tenantId?: string, scopes?: string[]): Promise<{ client: SubscriptionClient, credential: TokenCredential, authentication: AzureAuthentication }> {
+    private async getSubscriptionClient(tenantId?: string, scopes?: string[], session?: vscode.AuthenticationSession, account?: vscode.AuthenticationSessionAccountInformation): Promise<{ client: SubscriptionClient, credential: TokenCredential, authentication: AzureAuthentication }> {
```
alexweininger, line 245: I think we should remove the session parameter to keep it simpler and always fetch the session within this method.

```suggestion
    private async getSubscriptionClient(tenantId?: string, scopes?: string[], account?: vscode.AuthenticationSessionAccountInformation): Promise<{ client: SubscriptionClient, credential: TokenCredential, authentication: AzureAuthentication }> {
```

And when we do this I think we can keep this check:
```ts
if (!session) {
    throw new NotSignedInError();
}
```

where it was before.

auth/src/VSCodeAzureSubscriptionProvider.ts
```diff
@@ -59,13 +59,15 @@ export class VSCodeAzureSubscriptionProvider extends vscode.Disposable implement
      * @returns A list of tenants.
      */
     public async getTenants(account?: vscode.AuthenticationSessionAccountInformation): Promise<TenantIdDescription[]> {
-        const session = await getSessionFromVSCode(undefined, undefined, { account });
-        const { client } = await this.getSubscriptionClient(undefined, undefined, session);
-
         const results: TenantIdDescription[] = [];
+        const accounts = account ? [account] : await vscode.authentication.getAccounts(getConfiguredAuthProviderId());
+
+        for (account of accounts) {
```
alexweininger, line 65:
```suggestion
for await (account of account ? [account] : await vscode.authentication.getAccounts(getConfiguredAuthProviderId())) {

```

auth/src/VSCodeAzureSubscriptionProvider.ts
alexweininger: With your recent changes I think it's now possible to make this required

motm32: But account is not always passed in? That's the whole reason we are checking each account to get all the tenants?

alexweininger: Oh duh

alexweininger:
Ok I think we're done changing function signatures. Can you make sure the doc comments are all up-to-date?

alexweininger (approved):

alexweininger:
:shipit:

PR #1790, opened 2024-09-11
azure-pipelines: Fix release pipeline

alexweininger:
Fix a bug where the deployment job could've downloaded different artifacts from the verification job. To prevent this I combined them into a single job so that only one artifact download happens.


alexweininger:
It worked 😄 https://dev.azure.com/devdiv/DevDiv/_build/results?buildId=10206225&view=results

PR #1798, opened 2024-09-19
auth: Update Changelog v3.0.0

motm32:

### DIFF
--- a/auth/CHANGELOG.md
+++ b/auth/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Change Log
 
+## 3.0.0 - 2024-09-19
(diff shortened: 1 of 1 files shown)


auth/CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Change Log
 
+## 3.0.0 - 2024-09-19
+* [#1789](https://github.com/microsoft/vscode-azuretools/pull/1789) Change `getTenants` to be compatible with the new Azure Resources tenants view
```
alexweininger, line 4: Can you add a note about the potential breaking change? Basically if they want to continue getting the tenants for a single account then they need to now pass a specific account into the function. Otherwise they will now get all the tenants back for all accounts.

alexweininger (approved):

PR #1804, opened 2024-09-24
utils: Update vscode-tas-client to latest which removes axios dependency

alexweininger:


bwateratmsft (approved):
The best dependency is a removed dependency

PR #1807, opened 2024-10-29
utils & dev: Add fix for tests that fail due to having `showLoadingPrompt` set to `true`

MicroFish91:
We were seeing tests failing due to having `showLoadingPrompt` set to true.  It appears that the loading pick's `onDidHide` was being triggered when running multiple tests back-to-back, which would then cause a cancel and a premature wizard exit.

This new logic would add an `isTesting` boolean that the wizard can check internally so that we can prevent an early cancel from being issued.  This boolean would be automatically set each time we create a test action context.

### DIFF
--- a/dev/index.d.ts
+++ b/dev/index.d.ts
@@ -222,6 +222,7 @@ export declare class TestUserInput {
     public readonly onDidFinishPrompt: Event<PromptResult>;
 
     public constructor(vscode: typeof import('vscode'));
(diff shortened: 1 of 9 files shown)


alexweininger (commented):
Can you add some documentation to the property you added just to give some context about why it's needed and what it does?

alexweininger (approved):

PR #1819, opened 2024-11-18
Add account parameter to isSignedIn

### LINKED ISSUE #1809: getSubscriptions() throws if user has signed in to two accounts
1) run the script from https://github.com/Azure/bicep/pull/15403#issuecomment-2448473532 to install a version of the Bicep extension that uses the new @microsoft/vscode-azext-azureauth functionality
2) Using azureResourceGroups.login, log in to two different accounts in Azure.  (Note: I did this because I was testing a different Azure account, but it wasn't obvious how to sign out of the first one, I assumed this would do it.)
3) Create a new main.bicep file and save it with:
```bicep
resource storageaccount 'Microsoft.Storage/storageAccounts@2021-02-01' = {
  name: 'name'
  location: 'westus3'
  kind: 'StorageV2'
  sku: {
    name: 'Premium_LRS'
  }
}
```
4) in the editor's context menu for main.bicep, click "Deploy Bicep File..."
5) Click enter for default deployment name
!) Next you should see a picklist of subscriptions, but you will instead see ,in the Output window, under "Bicep Operations", this error:
> 1:57:16 PM: Deployment failed for c:\Users\stephwe\Downloads\main.bicep. You are not signed in to an Azure account. Please sign in.
EXPECTED: should work

More details:
Image
We call isSignedIn(), which finds that there's a session open for the first account and returns true.  So then we call getSubscriptions(), which calls getTenants() on each account.  The second getTenants() call throws because the second account doesn't have an open session:

Image
```
getSubscriptionClient (vscode-bicep\node_modules\@microsoft\vscode-azext-azureauth\out\src\VSCodeAzureSubscriptionProvider.js:297)
getTenants (vscode-bicep\node_modules\@microsoft\vscode-azext-azureauth\out\src\VSCodeAzureSubscriptionProvider.js:80)
getSubscriptions (vscode-bicep\node_modules\@microsoft\vscode-azext-azureauth\out\src\VSCodeAzureSubscriptionProvider.js:138)
```

alexweininger:
Fixes #1809

### DIFF
--- a/auth/src/AzureSubscriptionProvider.ts
+++ b/auth/src/AzureSubscriptionProvider.ts
@@ -42,7 +42,7 @@ export interface AzureSubscriptionProvider {
      *
      * @returns True if the user is signed in, false otherwise.
      */
-    isSignedIn(tenantId?: string): Promise<boolean>;
+    isSignedIn(tenantId?: string, account?: vscode.AuthenticationSessionAccountInformation): Promise<boolean>;
 
     /**
      * Asks the user to sign in or pick an account to use.
--- a/auth/src/VSCodeAzureSubscriptionProvider.ts
+++ b/auth/src/VSCodeAzureSubscriptionProvider.ts
@@ -108,7 +108,7 @@ export class VSCodeAzureSubscriptionProvider extends vscode.Disposable implement
                     }
 
                     // If the user is not signed in to this tenant, then skip it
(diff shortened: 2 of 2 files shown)


motm32 (commented):
Do we need to also add the account parameter to the AzureSubscriptionProvider interface?

motm32 (approved):

PR #1824, opened 2024-11-25
azure-pipelines: Ignore .vscode-test directory in CG scans

bwateratmsft:
We're getting a large number of component governance alerts because certain VSCode built-in extensions (`handlebars`, `npm`, `json`, etc.) have the same package names as NPM packages. The alerts are false on multiple levels--firstly, they're in `.vscode-test` which should not be scanned, and secondly, they aren't the actual NPM packages--just the package name and version matches a vulnerable NPM package.

This should suppress those CG alerts.

### DIFF
--- a/.azure-pipelines/1esmain.yml
+++ b/.azure-pipelines/1esmain.yml
@@ -47,6 +47,8 @@ extends:
     sdl:
       credscan:
         suppressionsFile: $(Build.SourcesDirectory)\.azure-pipelines\compliance\CredScanSuppressions.json
+      componentgovernance:
+        ignoreDirectories: $(Build.SourcesDirectory)\**\.vscode-test
     pool:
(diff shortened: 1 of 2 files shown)


.azure-pipelines/1esmain.yml
```diff
@@ -47,6 +47,8 @@ extends:
     sdl:
       credscan:
         suppressionsFile: $(Build.SourcesDirectory)\.azure-pipelines\compliance\CredScanSuppressions.json
+      componentgovernance:
+        ignoreDirectories: $(Build.SourcesDirectory)\**\.vscode-test
```
bwateratmsft, line 51: I don't know if this `**\` syntax will work. If not, we will have to explicitly exclude all of the packages' `.vscode-test` directories.

PR #1827, opened 2024-11-26
auth: Support multi-account scenarios

alexweininger:
TODO: bump version appropriately

The changes are basically adding an account parameter wherever there's a tenant parameter. The rest of the changes are fixes for issues that came up with the new multi-account scenarios. The changes in this PR have been tested by two partner teams, Bicep and SCOPE studio.

Fixes:
* Added a new call to getSubscriptionsForTenant that gets subscriptions in the users home tenant. This can mean duplicates show up in the initial result so I added logic to remove them.
* Added logic to remove duplicate subscriptions while retaining those from different accounts.
* Enhanced the `isSignedIn` method to check for a session across all accounts if no tenant or account is provided.
* Refactored the `getSubscriptionsForTenant` method to ensure it returns an empty list if the user is not signed in to the specified tenant or account.

### DIFF
--- a/auth/src/VSCodeAzureSubscriptionProvider.ts
+++ b/auth/src/VSCodeAzureSubscriptionProvider.ts
@@ -252,6 +252,7 @@ export class VSCodeAzureSubscriptionProvider extends vscode.Disposable implement
      */
     private async getSubscriptionsForTenant(account: vscode.AuthenticationSessionAccountInformation, tenantId?: string): Promise<AzureSubscription[]> {
         // If the user is not signed in to this tenant or account, then return an empty list
+        // This is to prevent the NotSignedInError from being thrown in getSubscriptionClient
         if (!await this.isSignedIn(tenantId, account)) {
             return [];
         }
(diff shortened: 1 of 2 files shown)


auth/src/VSCodeAzureSubscriptionProvider.ts
```diff
@@ -234,8 +271,7 @@ export class VSCodeAzureSubscriptionProvider extends vscode.Disposable implement
                 /* eslint-disable @typescript-eslint/no-non-null-assertion */
                 name: subscription.displayName!,
                 subscriptionId: subscription.subscriptionId!,
-                /* eslint-enable @typescript-eslint/no-non-null-assertion */
```
bwateratmsft, line 237: You can move this `eslint-enable` down instead of removing it

bwateratmsft (approved):

bwateratmsft (approved):

bwateratmsft:
> These are big changes with potential to change behavior, but there are no breaking API typings changes. Should I bump to 4.0.0 or to 3.1.0?

That's a good question. I'd lean toward 3.1.0 since there aren't strictly breaking changes.

PR #1849, opened 2024-12-06
auth: Create AzureTenant interface which includes account

alexweininger:

### DIFF
--- a/auth/src/AzureSubscriptionProvider.ts
+++ b/auth/src/AzureSubscriptionProvider.ts
@@ -3,9 +3,9 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import type { TenantIdDescription } from '@azure/arm-resources-subscriptions';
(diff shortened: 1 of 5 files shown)


motm32 (approved):

auth/src/AzureSubscriptionProvider.ts
```diff
@@ -19,7 +19,7 @@ export interface AzureSubscriptionProvider {
      *
      * @returns A list of tenants.
      */
-    getTenants(account?: vscode.AuthenticationSessionAccountInformation): Promise<TenantIdDescription[]>;
+    getTenants(account?: vscode.AuthenticationSessionAccountInformation): Promise<AzureTenant[]>;
```
alexweininger, line 22: This is technically a breaking change, but it's harmless. The break would be a compile error to anyone implementing this interface. Bump to 4.0.0 or to 3.2.0?

alexweininger: I'll add some more changes and figure it out later

PR #1856, opened 2024-12-12
auth: Fixes "Cannot read properties of undefined 'tenantId' error

motm32:
Please refer to this [rg issue](https://github.com/microsoft/vscode-azureresourcegroups/issues/966) for the error.

Also added the account property to isSignedIn() because for some reason I kept getting all of my tenants showing up in the list when I should only be getting unauthenticated tenants.

### DIFF
--- a/auth/src/utils/getUnauthenticatedTenants.ts
+++ b/auth/src/utils/getUnauthenticatedTenants.ts
@@ -4,18 +4,27 @@
 *--------------------------------------------------------------------------------------------*/
 
 import type { TenantIdDescription } from "@azure/arm-resources-subscriptions";
+import * as vscode from "vscode";
 import type { AzureSubscriptionProvider } from "../AzureSubscriptionProvider";
+import { getConfiguredAuthProviderId } from "./configuredAzureEnv";
 
 /**
  * @returns list of tenants that VS Code doesn't have sessions for
  */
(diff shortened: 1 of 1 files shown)


alexweininger:
If the problem was because we weren't passing `account` into `isSignedIn` on line 16, then would this solve it?

image

I added the account property to our tenant interface recently https://github.com/microsoft/vscode-azuretools/pull/1849

motm32:
> If the problem was because we weren't passing `account` into `isSignedIn` on line 16, then would this solve it?
> 
> image
> I added the account property to our tenant interface recently #1849

Yes lol for some reason it wasn't showing up when I first implemented it 😅

alexweininger (approved):

auth/src/utils/getUnauthenticatedTenants.ts
```diff
@@ -13,9 +13,12 @@ export async function getUnauthenticatedTenants(subscriptionProvider: AzureSubsc
     const tenants = await subscriptionProvider.getTenants();
     const unauthenticatedTenants: TenantIdDescription[] = [];
     for await (const tenant of tenants) {
-        if (!await subscriptionProvider.isSignedIn(tenant.tenantId)) {
-            unauthenticatedTenants.push(tenant);
+        if (tenant) {
```
alexweininger, line 16: This if statement is weird to me but I understand why you included it. Maybe add a comment for future us so we're not confused about why it's there and accidentally remove it.

nturinski (approved):

PR #1882, opened 2025-02-03
appservice: Update steps and validation to support new domain name label scopes

MicroFish91:
Here are the updated steps and validation for supporting the new domain label scopes, see [here](https://techcommunity.microsoft.com/blog/appsonazureblog/public-preview-creating-web-app-with-a-unique-default-hostname/4156353) for more context.

New validation required for site names:
```text
In general, here are the checks/validation that should be done depending on the customer’s selection:
 
* If global hostname selected (old format of “*.azurewebsites.net”), use these three (3) checks/validations prior to allowing customer to create site:
  1.      Global CheckNameAvailability (CNA) API (the one without providing region)
  2.      Regional CheckNameAvailability (CNA) API (the one where you need to provide region)
  3.      ARM ID Check / Check if a resource already exists with the same ARM ID
    a.      A “create” on a pre-existing resource with same ARM ID would result to an “update” of the site resource
    b.      We do this check by doing a GET on the site and if it results in:
        i.      Status 200 – this means that there is already a pre-existing site with the exact ARM ID, so you would need block user from creating the resource
        ii.      Status 404 – this means that no resource exists with the exact ARM ID, so you can allow user to create the resource
    
  Special case: If site name input (user input) is >=47 chars, ignore result of regional CNA because it inherently has a shorter character limit than Global CNA

* If regional hostname selected (new format of “*.<RegionName>.azurewebsites.net”), use these two (2) checks/validations prior to allowing customer to create site:
    1. Regional CNA
    2. ARM ID Check / Check if a resource already exists with the same ARM ID
 
"Global CNA" is the global (no region/location in the parameter) CheckNameAvailability API
"Regional CNA" is the regional (with region/location in the parameter) CheckNameAvailability API
```

Corresponding `vscode-azureappservice` PR: https://github.com/microsoft/vscode-azureappservice/pull/2703

Edit: After some experimentation, I've found that selecting for the `Tenant` domain label scope actually fails if over 43 chars long, even though the validation says it's okay.  This is the error:
image

Setting the limit to 43 chars instead seems to completely fix the issue.  Verified portal has a similar limit.

### DIFF
--- a/appservice/package.json
+++ b/appservice/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@microsoft/vscode-azext-azureappservice",
     "author": "Microsoft Corporation",
-    "version": "3.3.1",
+    "version": "3.4.0",
     "description": "Common tools for developing Azure App Service extensions for VS Code",
     "tags": [
         "azure",
--- a/appservice/package.json
+++ b/appservice/package.json
@@ -42,7 +42,7 @@
         "@azure/storage-blob": "^12.3.0",
         "@microsoft/vscode-azext-azureutils": "^3.0.0",
         "@microsoft/vscode-azext-github": "^1.0.0",
-        "@microsoft/vscode-azext-utils": "^2.5.0",
+        "@microsoft/vscode-azext-utils": "^2.5.13",
         "dayjs": "^1.11.2",
         "fs-extra": "^10.0.0",
         "p-retry": "^3.0.1",
--- a/appservice/src/createAppService/SiteNameStep.ts
+++ b/appservice/src/createAppService/SiteNameStep.ts
@@ -154,7 +154,7 @@ export class SiteNameStep extends AzureNameStep<SiteNameStepWizardContext> {
         const location: AzExtLocation = await LocationListStep.getLocation(context);
         const authToken: string = nonNullValueAndProp((await context.credentials.getToken() as { token?: string }), 'token');
 
-        // Todo: Can potentially replace with call using SDK once the update is available
+        // Todo: Can replace with call using SDK once the update is available
         const options: AzExtRequestPrepareOptions = {
             url: `https://management.azure.com/subscriptions/${context.subscriptionId}/providers/Microsoft.Web/locations/${location.name}/checknameavailability?api-version=${apiVersion}`,
             method: 'POST',
--- a/appservice/src/createAppService/SiteNameStep.ts
+++ b/appservice/src/createAppService/SiteNameStep.ts
@@ -158,7 +158,7 @@ export class SiteNameStep extends AzureNameStep<SiteNameStepWizardContext> {
             throw new Error(vscode.l10n.t('Internal Error: A location is required when validating a site name with regional CNA.'));
         }
         if (domainNameScope === DomainNameLabelScope.ResourceGroup && !resourceGroupName) {
-            throw new Error(vscode.l10n.t('Internal Error: A resource group name is required for validating a resource group scoped site with regional CNA.'));
+            throw new Error(vscode.l10n.t('Internal Error: A resource group name is required for validating this level of domain name scope.'));
         }
 
         const apiVersion: string = '2024-04-01';
--- a/appservice/src/createAppService/SiteNameStep.ts
+++ b/appservice/src/createAppService/SiteNameStep.ts
@@ -127,7 +127,7 @@ export class SiteNameStep extends AzureNameStep<SiteNameStepWizardContext> {
         return undefined;
     }
 
-    // Todo: Leave reference to GitHub comment
+    // For comprehensive breakdown of validation logic, please refer to: https://github.com/microsoft/vscode-azuretools/pull/1882#issue-2828801875
     private async asyncValidateSiteName(context: SiteNameStepWizardContext, sdkClient: WebSiteManagementClient, name: string): Promise<string | undefined> {
         name = name.trim();
 
--- a/appservice/src/createAppService/SiteNameStep.ts
+++ b/appservice/src/createAppService/SiteNameStep.ts
@@ -129,11 +129,13 @@ export class SiteNameStep extends AzureNameStep<SiteNameStepWizardContext> {
 
     // Todo: Leave reference to GitHub comment
     private async asyncValidateSiteName(context: SiteNameStepWizardContext, sdkClient: WebSiteManagementClient, name: string): Promise<string | undefined> {
-        let validationMessage: string | undefined;
+        name = name.trim();
 
+        let validationMessage: string | undefined;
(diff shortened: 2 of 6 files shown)


appservice/src/index.ts
```diff
@@ -17,23 +14,27 @@ export * from './createAppService/AppServicePlanSkuStep';
...
 export * from './deploy/getDeployNode';
+export * from './deploy/IDeployContext';
 export * from './deploy/localGitDeploy';
-export { IPreDeployTaskResult, handleFailedPreDeployTask, runPreDeployTask, tryRunPreDeployTask } from './deploy/runDeployTask';
+export { handleFailedPreDeployTask, IPreDeployTaskResult, runPreDeployTask, tryRunPreDeployTask } from './deploy/runDeployTask';
 export * from './deploy/showDeployConfirmation';
 export { disconnectRepo } from './disconnectRepo';
 export * from './editScmType';
 export { registerAppServiceExtensionVariables } from './extensionVariables';
+export * from './KuduModels';
```
nturinski, line 35: Your linter reordered some stuff. I'm okay with it because I think mine tries to organize it the same way, but we should decide on a standard.

appservice/src/index.ts
```diff
@@ -17,23 +14,27 @@ export * from './createAppService/AppServicePlanSkuStep';
 export * from './createAppService/CustomLocationListStep';
 export * from './createAppService/IAppServiceWizardContext';
 export * from './createAppService/LogAnalyticsCreateStep';
+export * from './createAppService/setLocationsTask';
+export * from './createAppService/SiteDomainNameLabelScopeStep';
```
nturinski, line 18: Don't forget to add this to the index.d.ts or else you won't be able to actually import it in the extension without typings yelling at you.

appservice/package.json
```diff
@@ -42,7 +42,7 @@
         "@azure/storage-blob": "^12.3.0",
         "@microsoft/vscode-azext-azureutils": "^3.0.0",
         "@microsoft/vscode-azext-github": "^1.0.0",
-        "@microsoft/vscode-azext-utils": "^2.5.0",
+        "@microsoft/vscode-azext-utils": "^2.5.13",
```
nturinski, line 45: Don't forget to bump the appservice version number as well.

appservice/src/createAppService/SiteNameStep.ts
```diff
@@ -122,12 +127,104 @@ export class SiteNameStep extends AzureNameStep<SiteNameStepWizardContext> {
...
         } else {
             return undefined;
         }
     }
+
+    private async asyncValidateRegionalCNA(context: SiteNameStepWizardContext, domainNameScope: DomainNameLabelScope, siteName: string, resourceGroupName?: string): Promise<string | undefined> {
+        if (!LocationListStep.hasLocation(context)) {
+            throw new Error(vscode.l10n.t('Internal Error: A location is required when validating a site name with regional CNA.'));
+        }
+        if (domainNameScope === DomainNameLabelScope.ResourceGroup && !resourceGroupName) {
```
nturinski, line 160: nit: Could probably be an else if since you only need one of these to throw an error.

appservice/src/createAppService/SiteNameStep.ts
```diff
@@ -122,12 +127,104 @@ export class SiteNameStep extends AzureNameStep<SiteNameStepWizardContext> {
...
+            throw new Error(vscode.l10n.t('Internal Error: A resource group name is required for validating this level of domain name scope.'));
+        }
+
+        const apiVersion: string = '2024-04-01';
+        const location: AzExtLocation = await LocationListStep.getLocation(context);
+        const authToken: string = nonNullValueAndProp((await context.credentials.getToken() as { token?: string }), 'token');
+
+        // Todo: Can replace with call using SDK once the update is available
+        const options: AzExtRequestPrepareOptions = {
+            url: `https://management.azure.com/subscriptions/${context.subscriptionId}/providers/Microsoft.Web/locations/${location.name}/checknameavailability?api-version=${apiVersion}`,
```
nturinski, line 170: You should probably use the environment resourceEndpointUri here since it'll probably matter for different Azure clouds.

appservice/src/createAppService/SiteNameStep.ts
```diff
@@ -122,12 +127,104 @@ export class SiteNameStep extends AzureNameStep<SiteNameStepWizardContext> {
...
+        const pipelineResponse = await client.sendRequest(createPipelineRequest(options)) as AzExtPipelineResponse;
+        const checkNameResponse = pipelineResponse.parsedBody as {
+            hostName?: string;
+            message?: string;
+            nameAvailable?: boolean;
+            reason?: string;
+        };
+
+        if (!checkNameResponse.nameAvailable) {
+            // If site name input is >=47 chars, ignore result of regional CNA because it inherently has a shorter character limit than Global CNA
```
nturinski, line 194: Should you change `validateSiteName` to check the max length based on regional and global then?

appservice/src/index.ts
MicroFish91: This is one of our projects that utilize the automatically generated declarations:
image

Confusing sometimes because not all our older projects do it that way :P

appservice/src/createAppService/SiteNameStep.ts
```diff
@@ -24,6 +26,8 @@ const siteNamingRules: IAzureNamingRules = {
     invalidCharsRegExp: /[^a-zA-Z0-9\-]/
 };
 
+const regionalCNAMaxLength: number = 43;
```
nturinski, line 29: Can you leave a little comment about all of the weird hoops you had regarding how you came to this number?

appservice/src/createAppService/SiteNameStep.ts
```diff
@@ -122,12 +134,101 @@ export class SiteNameStep extends AzureNameStep<SiteNameStepWizardContext> {
...
+            }
+            return checkNameResponse.message;
+        }
+
+        return undefined;
+    }
+
+    private async asyncValidateUniqueARMId(context: SiteNameStepWizardContext, client: WebSiteManagementClient, siteName: string, resourceGroupName?: string): Promise<string | undefined> {
+        if (!resourceGroupName) {
+            context.relatedNameTask = this.generateRelatedName(context, siteName, this.getRelatedResourceNamingRules(context));
```
nturinski, line 215: Should you be overwriting the relatedNameTask like this? What if one already exists?

nturinski (dismissed):

appservice/src/createAppService/SiteNameStep.ts
MicroFish91: Good point, I should probably use `??=` here

nturinski (approved):

PR #1889, opened 2025-02-24
azure: Changes to managed identity steps so they are compatible with functions settings conversion

motm32:
Needed to make a couple changes to make the current managed identity steps compatible with converting settings. 

Things I have changed: 

- Adding a `ResourceGroupListStep` to the `UserAssignedIdentityListStep`. Users who are converting will not have specified a resource group prior to choosing to create a managed identity
- Allow `RoleAssignmentExecuteStep` to assign multiple roles in the step. It is common not only for converting to assign more than one role
- Added more common role types and also made changes to how we get the id for the roles as that uses a users subscriptionId

TODO: 

- [ ] put up the functions PR to show how these changes are used.

### DIFF
--- a/azure/index.d.ts
+++ b/azure/index.d.ts
@@ -5,7 +5,6 @@
 
 /* eslint-disable @typescript-eslint/no-explicit-any */
 
-import { type RoleDefinition } from '@azure/arm-authorization';
 import { Identity } from '@azure/arm-msi';
 import type { ExtendedLocation, ResourceGroup } from '@azure/arm-resources';
 import type { Location } from '@azure/arm-resources-subscriptions';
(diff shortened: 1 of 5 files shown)


azure/index.d.ts
```diff
@@ -506,4 +511,69 @@ export declare const CommonRoleDefinitions: {
         readonly description: "Allows for read, write and delete access to Azure Storage blob containers and data";
         readonly roleType: "BuiltInRole";
     };
+    readonly storageBlobDataOwner: {
```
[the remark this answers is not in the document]
motm32: I'll add a link

alexweininger (approved):

azure/src/wizard/RoleAssignmentExecuteStep.ts
```diff
@@ -3,42 +3,50 @@
...
-        const scope = this.getScopeId();
-        if (!scope) {
-            throw new Error(l10n.t('No scope was provided for the role assignment.'));
-        }
-        const scopeSplitArr = scope.split('/');
-        const resourceName = scopeSplitArr[scopeSplitArr.length - 1] ?? '';
-        const resourceType = scopeSplitArr[scopeSplitArr.length - 2] ?? '';
+        const roles = this.roles();
+        if (roles) {
+            for (const role of roles) {
```
nturinski, line 31: I don't think we would want to make the whole step fail if one role fails. Maybe do a try/catch so it can continue?

azure/index.d.ts
```diff
@@ -383,14 +383,20 @@ export declare class RoleAssignmentExecuteStep<T extends IResourceGroupWizardCon
...
 
     public execute(wizardContext: T, progress: Progress<{ message?: string; increment?: number }>): Promise<void>;
     public shouldExecute(wizardContext: T): boolean;
 }
 
+export interface Role {
+    scopeId: string | undefined;
+    roleDefinitionId: string;
+    roleDefinitionName: string;
+}
```
alexweininger, line 398: Can you add doc strings to this interface?

alexweininger (dismissed):

alexweininger (approved):

motm32:
Waiting to merge as we are still discussing some role implementation details that may need to be changed.

PR #1893, opened 2025-02-27
appservice: Remove creating a container blob on flex deploy

nturinski:
Back when I first implemented flex deployment, I thought it'd be useful to create the container for the user if it didn't exist. Unfortunately, if they don't have SAS available, the get request fails and throws an error, failing the whole deployment.

Other dev tooling doesn't create the container for the user and it should be created when they create their flex app anyway. I was accounting for an edge case if they accidentally deleted their container, but was told we should keep deployment behavior consistent between tooling.


alexweininger (approved):

MicroFish91 (approved):

MicroFish91:
Sorry I just merged my other app service pr, so you'll probably have to fix the version bump in package/package-lock

MicroFish91 (approved):

PR #1913, opened 2025-03-05
appservice: Remove `globally unique` descriptors from site name prompts

MicroFish91:
Expanding this coverage to functions now as well, so just going to remove globally unique from all the prompts.  With the new domain name label scopes, names are not required to always be globally unique.


MicroFish91:
Could probably release this PR with this other PR so we only have to bump once: https://github.com/microsoft/vscode-azuretools/pull/1913

bwateratmsft (approved):

MicroFish91:
Could I get a reapprove please, I removed the version bump, going to let Nathan's PR take care of that instead.

nturinski (approved):

PR #1925, opened 2025-03-11
azure: Check `newResourceGroupName` when creating managed identity

nturinski:
If you create a new MI through the wizard, it will mess up if you also create a new resource group in the process.


MicroFish91 (dismissed):

bwateratmsft (approved):

azure/src/wizard/UserAssignedIdentityCreateStep.ts
```diff
@@ -4,7 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { ManagedServiceIdentityClient } from '@azure/arm-msi';
```
bwateratmsft, line 6: Unrelated to the PR, but we should lazy load this Azure package

nturinski: I thought we had linter rule for this... 🤔

MicroFish91 (approved):

azure/src/wizard/UserAssignedIdentityCreateStep.ts
MicroFish91: I think we do in most of the extensions, but probably haven't updated it for the shared packages

PR #1928, opened 2025-03-17
appsettings: Small changes for managed identity settings addition

motm32:
From testing I found that the `DEPLOYMENT_STORAGE_CONNECTION_STRING` is something we wouldn't want to convert. 

This also includes some changes to `AppSettingsTreeItem` so that I can add an entry point for the `Add Function App Identity Connections...` command on the top level App Settings Tree item 
image

### DIFF
--- a/appsettings/src/tree/AppSettingsTreeItem.ts
+++ b/appsettings/src/tree/AppSettingsTreeItem.ts
@@ -75,16 +75,9 @@ export class AppSettingsTreeItem extends AzExtParentTreeItem {
     }
 
     static async createAppSettingsTreeItem(context: IActionContext, parent: AzExtParentTreeItem, clientProvider: AppSettingsClientProvider, extensionPrefix: string, options?: AppSettingsTreeItemOptions): Promise<AppSettingsTreeItem> {
-        const client = await clientProvider.createClient(context);
-        const appSettings = await client.listApplicationSettings();
-        if (appSettings.properties) {
-            for (const [key, value] of Object.entries(appSettings.properties)) {
-                if (isSettingConvertible(key, value)) {
-                    options?.contextValuesToAdd?.push('convert');
-                }
-            }
-        }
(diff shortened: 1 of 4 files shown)


appsettings/src/tree/AppSettingsTreeItem.ts
```diff
@@ -74,6 +74,19 @@ export class AppSettingsTreeItem extends AzExtParentTreeItem {
         }
     }
 
+    static async createAppSettingsTreeItem(context: IActionContext, parent: AzExtParentTreeItem, clientProvider: AppSettingsClientProvider, extensionPrefix: string, options?: AppSettingsTreeItemOptions): Promise<AppSettingsTreeItem> {
```
nturinski, line 77: I know I already looked at this, but I just thought of this--

You should probably make the constructor private so that people don't accidentally create the app settings with a raw new constructor since if they did, it wouldn't have the expected behavior.

appsettings/src/tree/AppSettingsTreeItem.ts
```diff
@@ -74,6 +74,19 @@ export class AppSettingsTreeItem extends AzExtParentTreeItem {
...
+        const client = await clientProvider.createClient(context);
+        const appSettings = await client.listApplicationSettings();
+        if (appSettings.properties) {
+            for (const [key, value] of Object.entries(appSettings.properties)) {
+                if (isSettingConvertible(key, value)) {
+                    options?.contextValuesToAdd?.push('convert');
+                }
+            }
+        }
+        return new AppSettingsTreeItem(parent, clientProvider, extensionPrefix, options);
```
nturinski, line 87: I'd also add a `refreshImpl` implementation that does this logic and then call `this.refreshImpl` in the static create function.  That way if a user refreshes the node, it will update the settings.

nturinski (dismissed):

nturinski (approved):

PR #1933, opened 2025-03-19
all: Do some rounds of `npm audit fix`

bwateratmsft:
I updated `mocha` and `@types/mocha` to match what's in the dev package which significantly reduced the number of packages in use, while also removing the `npm audit` issues. Additionally, I bumped the `github` package version in preparation for releasing it, so that we can fix the audit issues in `appservice`.


utils/package.json
```diff
@@ -48,17 +48,17 @@
         "@microsoft/eslint-config-azuretools": "^0.2.1",
         "@microsoft/vscode-azext-dev": "^2.0.4",
         "@types/html-to-text": "^8.1.0",
-        "@types/mocha": "^7.0.2",
+        "@types/mocha": "^10.0.10",
         "@types/node": "^16.0.0",
         "@types/semver": "^7.3.9",
         "@types/uuid": "^9.0.1",
-        "@types/vscode": "^1.94.0",
+        "@types/vscode": "1.95.0",
```
bwateratmsft, line 55: This was already using 1.95.0, but was incorrectly using caret notation.

alexweininger (approved):

MicroFish91 (approved):
Looks good, I double checked github package

PR #1935, opened 2025-03-20
feat: Adds localization dictionaries

bk201-:
- Adds new script to build dictionary bundle
- Adds dictionaries for all packages
- Fixes strings for localization

@alexweininger I'm not sure if I should commit dictionary bundles or not. It depends on yours vision. If you don't have any plans to translate packages I would like to run l10n script before packaging and just include them into npm package instead of committing them to Git. What do you think?


bwateratmsft:
For my own education, what effect does having these bundles do downstream--i.e. in the extensions?

bk201-:
Hi @bwateratmsft! 

Our extension vscode-cosmosdb uses some packages from this repo. For instance extension show dialog with buttons, which of course have title. Since the official tool for localization `@vscode/l10n-dev` looking for strings for localization only inside `src` folder. It ignores `node_modules` for many reasons:
- it relies on imports in TS files. The import should be like `import * as vscode from 'vscode'`
- packages in node_modules often are built using webpack, where import above is transpiled to `require(vscode_1)` or `webpack_require()` or something else. Therefore `@vscode/l10n-dev` tool ignores `node_modules`

So our extension uses yours packages, your packages are prepared for translation, but the dictionary (bundle.l10n.json) doesn't exist, therefore we can't translate the strings from yours packages.

In case of these dictionaries will exist we can inject them into our dictionary.

If you already have experience with localization in required packages or know how to implement it better, we can discuss

bwateratmsft:
@tmeschter do you know if it's standard practice to include the generated `bundle.l10n.json` files in the repo?

bwateratmsft (approved):

bwateratmsft:
I'll approve it for now. If updating the bundles turns out to be a recurring issue we can revisit.

PR #1937, opened 2025-03-24
appservice: Add ability to append additional warnings to deploy confirmation message

motm32:
This is an an example of how the warning message might look like: 
image

### DIFF
--- a/appservice/src/deploy/showDeployConfirmation.ts
+++ b/appservice/src/deploy/showDeployConfirmation.ts
@@ -12,8 +12,13 @@ import { delay } from '../utils/delay';
 import { updateWorkspaceSetting } from '../utils/settings';
 import { AppSource, IDeployContext } from './IDeployContext';
 
-export async function showDeployConfirmation(context: IDeployContext, site: ParsedSite, deployCommandId: string): Promise<void> {
+export async function showDeployConfirmation(context: IDeployContext, site: ParsedSite, deployCommandId: string, warningMessages?: string[]): Promise<void> {
     const warning: string = l10n.t('Are you sure you want to deploy to "{0}"? This will overwrite any previous deployment and cannot be undone.', site.fullName);
+    let details = ''
+    if (warningMessages) {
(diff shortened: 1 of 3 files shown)


appservice/src/deploy/showDeployConfirmation.ts
```diff
@@ -12,16 +12,21 @@ import { delay } from '../utils/delay';
 import { updateWorkspaceSetting } from '../utils/settings';
 import { AppSource, IDeployContext } from './IDeployContext';
 
-export async function showDeployConfirmation(context: IDeployContext, site: ParsedSite, deployCommandId: string): Promise<void> {
+export async function showDeployConfirmation(context: IDeployContext, site: ParsedSite, deployCommandId: string, warningMessages?: string[]): Promise<void> {
```
nturinski, line 15: Update the index.d.ts to include the new typings.

appservice/src/deploy/showDeployConfirmation.ts
```diff
@@ -12,16 +12,21 @@ import { delay } from '../utils/delay';
...
     const warning: string = l10n.t('Are you sure you want to deploy to "{0}"? This will overwrite any previous deployment and cannot be undone.', site.fullName);
+    let addedWarnings = ''
+    if (warningMessages) {
+        const warningMessagesString: string = warningMessages.map((message) => `${message}`).join('\n\n');
```
nturinski, line 19: This is fine, but one of the rare times that you can use can use Array.reduce 🤣

nturinski (dismissed):

appservice/src/deploy/showDeployConfirmation.ts
alexweininger: Can't we just do `const warningMessagesString = warningMessages.join('\n\n');`

alexweininger: @motm32 do dis

motm32: I am don't worry

nturinski (approved):

PR #1941, opened 2025-03-26
utils: Use `vscode.TelemetryTrustedValue` for recording resourceId

alexweininger:

### DIFF
--- a/utils/src/registerCommand.ts
+++ b/utils/src/registerCommand.ts
@@ -3,7 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { commands, Uri } from 'vscode';
+import { commands, Uri, TelemetryTrustedValue } from 'vscode';
 import * as types from '../index';
 import { callWithTelemetryAndErrorHandling } from './callWithTelemetryAndErrorHandling';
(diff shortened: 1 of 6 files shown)


nturinski (approved):

MicroFish91 (approved):

bwateratmsft (approved):

utils/src/registerCommand.ts
```diff
@@ -78,15 +78,20 @@ async function setTelemetryProperties(context: types.IActionContext, args: unkno
...
+            context.telemetry.properties.resourceId = new TelemetryTrustedValue(resourceId);
+        }
         context.telemetry.properties.subscriptionId = (firstArg as { resource: Resource })?.resource?.subscription?.subscriptionId;
     }
 
     // handles items from v1 extensions
     for (const arg of args) {
         if (arg instanceof AzExtTreeItem) {
             try {
-                context.telemetry.properties.resourceId = arg.id;
```
bwateratmsft, line 89: Can we change the logic here to only attach subscription ID and resource ID if subscription is _defined_? The Docker extension shouldn't be attaching these IDs but it is. It does not have subscriptions associated with `AzExtTreeItem`s.

bwateratmsft (changes requested):
See inline comment

bwateratmsft (approved):

PR #1943, opened 2025-03-27
utils: Only try to record sub and resource id if item has a sub

alexweininger:
PR has two goals:
1. Don't try to record resourceId/subId telemetry for items without subscriptions (ex: docker items)
3. Stop throwing the "No Azure subscription found for this tree item." error and recording it in the telemetryError property


bwateratmsft (approved):

alexweininger:
Released in [v2.6.6](https://github.com/microsoft/vscode-azuretools/releases/tag/microsoft-vscode-azext-utils-v2.6.6)

PR #1949, opened 2025-03-28
azure: Add retry logic to user roles and change IdentityCreate priority

nturinski:
I was running into an issue where the role assignment was failing because the managed identity was just created. The retry logic should fix it, but also creating it immediately after the resource group step also makes sense since the other resources are going to be associated to it.


alexweininger (dismissed):

MRayermannMSFT (dismissed):
- Consider pulling out that `5` into a `const`?
- Consider reporting the number of retries used in telemetry?

MicroFish91 (dismissed):

alexweininger:
Piggybacking on Matt R. if it does still fail after 5 retries, make sure the error reported is helpful to the user. From quickly looking at the code I can't tell if that's the case.

MicroFish91 (approved):

PR #1950, opened 2025-03-28
utils: Add optional `configureBeforeExecute` method to execute steps

MicroFish91:

### DIFF
--- a/utils/src/wizard/AzureWizardExecuteStep.ts
+++ b/utils/src/wizard/AzureWizardExecuteStep.ts
@@ -11,11 +11,11 @@ export abstract class AzureWizardExecuteStep<T extends types.IActionContext & Pa
     public options: types.AzureWizardExecuteStepOptions = {};
 
     public abstract priority: number;
+    public configureBeforeExecute?(wizardContext: T): void | Promise<void>;
     public abstract execute(wizardContext: T, progress: Progress<{ message?: string; increment?: number }>): Promise<void>;
     public abstract shouldExecute(wizardContext: T): boolean;
 
     public createSuccessOutput?(context: T): types.ExecuteActivityOutput;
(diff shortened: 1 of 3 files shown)


bwateratmsft (approved):

alexweininger:
What is scenario that requires this? The changes look good, I just want to know the context behind them. Also for record keeping.

MicroFish91:
I've found its especially useful for steps that try to combine verification and create logic together.  A good example of this would be `ResourceGroupCreateStep`.  I have a PR coming up shortly that would convert this step to display activity outputs.  I'll add a link once the PR is posted.

Since `ResourceGroupCreateStep` currently handles both creation logic and verification logic in its `execute` method, trying to figure out what to display and when during success, fail, and progress states becomes entangled/messier.  Being able to separate the verification logic and then running the step based on whether or not the resource exists greatly simplifies what we need to think about displaying and also keeps the logic paths in each method more concise.

utils/src/wizard/AzureWizardExecuteStep.ts
```diff
@@ -11,11 +11,11 @@ export abstract class AzureWizardExecuteStep<T extends types.IActionContext & Pa
...
 
     public abstract priority: number;
+    public configureBeforeExecute?(wizardContext: T): void | Promise<void>;
     public abstract execute(wizardContext: T, progress: Progress<{ message?: string; increment?: number }>): Promise<void>;
     public abstract shouldExecute(wizardContext: T): boolean;
 
     public createSuccessOutput?(context: T): types.ExecuteActivityOutput;
     public createFailOutput?(context: T): types.ExecuteActivityOutput;
     public createProgressOutput?(context: T): types.ExecuteActivityOutput;
-
```
nturinski, line 20: nit: Why cut down a young linebreak in its prime?

nturinski (approved):

PR #1952, opened 2025-03-31
utils: Add miscellaneous activity utils

MicroFish91:
When an execute step starts, a progress item may be added to the end of the list of activity children.  If within that step we want to add additional activity children, pushing them will put them behind the current progress item which doesn't look great.  In those scenarios, it makes more sense to add the child right before the current progress item, this util handles this by splicing it one in front from the back and refreshing the tree item.


nturinski:
I know we talked about this briefly, but now that I'm thinking about this, this behavior seems kind of weird to me.

I agree that it's visually better, but I feel like inserting steps above the active node is a weird behavior that users may get confused by these nodes getting slipped in.

What does the team think about having another nested layer of children nodes on the progressing item instead?

bwateratmsft:
> What does the team think about having another nested layer of children nodes on the progressing item instead?

I'd say more children, not injecting nodes

MicroFish91:
> What does the team think about having another nested layer of children nodes on the progressing item instead?

In this idea, once the progress item completes, what's the behavior you would expect?  Let's say it succeeded, would the children remain on that step's success item, or would they be carried up and flattened one level above as children of the entire activity?

Here's an example of how I was originally doing it in container apps (this was added before we were trying to support the showing of live progress):
image

In this example, `Enable external ingress....` is being injected during `ContainerAppCreateStep`.  It is mostly an informational node rather than a subtask to track.
https://github.com/microsoft/vscode-azurecontainerapps/pull/862/files

This kind of segways into a discussion that probably needed to happen soon anyway, but what should the convention be here.  Should we omit informational nodes like this from the activity log and keep the items primarily tied 1:1 with their corresponding execute steps?  Although I kind of like being able to see that information in the activity log, it does create some messy scenarios that I'm not sure are worth the tradeoff.  Would appreciate any thoughts people have here!

PR #1954, opened 2025-03-31
utils: Show durations for activities and activity children

MicroFish91:

### DIFF
--- a/utils/src/utils/dateUtils.ts
+++ b/utils/src/utils/dateUtils.ts
@@ -3,12 +3,21 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-export namespace dateUtils {
-    export function getFormattedDurationInMinutesAndSeconds(start: Date, end: Date): string {
-        const durationSeconds: number = Math.round((end.getTime() - start.getTime()) / 1000);
+import * as dayjs from 'dayjs';
+// eslint-disable-next-line import/no-internal-modules
+import * as duration from 'dayjs/plugin/duration';
 
-        const minutes: number = Math.floor(durationSeconds / 60);
-        const seconds: number = durationSeconds - minutes * 60;
-        return minutes ? `${minutes}m ${seconds}s` : `${seconds}s`;
+dayjs.extend(duration);
+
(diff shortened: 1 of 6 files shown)


utils/src/utils/dateUtils.ts
```diff
@@ -0,0 +1,14 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+export namespace dateUtils {
+    export function getFormattedDurationInMinutesAndSeconds(start: Date, end: Date): string {
```
bwateratmsft, line 7: The `dayjs` package (which we already use) can probably do this for us.

bwateratmsft (dismissed):

utils/src/wizard/AzureWizard.ts
```diff
@@ -188,17 +189,19 @@ export class AzureWizard<T extends (IInternalActionContext & Partial<types.Execu
 
             let step: AzureWizardExecuteStep<T> | undefined = steps.pop();
             while (step) {
+                const start: Date = new Date();
```
nturinski, line 192: nit: Small detail, but this should probably go on line 205-206 where the step is executed as part of the try/finally clause,

MicroFish91: Although I agree and initially did it this way too, I think if we omit the other step methods being run, the individual step times will not sum up to the total recorded time as closely, and that could look weird to our users if they are paying close attention (more noticeable on longer commands).  For that reason, I thought it might be better to include the other methods in the calculation as well, but am happy to discuss if we want to only measure the execute method.

utils/src/utils/dateUtils.ts
MicroFish91: Let me know if this is what you had in mind or if you were preferring that I remove the `dateUtils` entirely!

bwateratmsft: This is what I was thinking: https://day.js.org/docs/en/durations/format

bwateratmsft (approved):

PR #1960, opened 2025-04-04
utils: Have `getFormattedDuration` match the `dayjs` duration input signature

MicroFish91:


bwateratmsft (approved):
This is technically a breaking change but I think it's fine, it's unlikely there are any consumers of this new function.

MicroFish91:
> This is technically a breaking change but I think it's fine, it's unlikely there are any consumers of this new function.
Yeah also I'm pretty sure we haven't bumped the version yet :)

bwateratmsft (approved):

PR #1961, opened 2025-04-07
pipelines: Enable custom publish commands and ability sign and publish multiple vsix files

### LINKED ISSUE #1938: pipelines: Modify the build pipeline so that it can do os/arch specific builds of the .vsix
For the Github Copilot for Azure extension we would like to create OS specific vsix for the extension to reduce the package size for distribution. We are currently using the azure tools pipelines for the same.

Would it be possible to update the build and release pipeline as follow : 
Build pipeline: Modify build pipeline to produce os/arch specific .vsix's
Release pipeline: Update release pipeline to release os/arch specific .vsix's to the marketplace.

msalaman:
Fixes #1938 

build template: add "vsixFileNames", enables build to sign mulitple vsix files and save signatures to directories based on their names all in one pipeline which enables multiple vsix's to be used in the publish pipeline.

publish template: add "publish commands" string array parameter. This allows consumers to fully customize the publish command, which solves a couple of limitations: 1. can add flags for publishing pre-release extensions, and platform specific extensions. It also allows for executing multiple publish commands in one pipeline which can be used to publish vsix's for different platforms all in one pipeline. Also added "publishMultipleVsixFiles" 
to bypass check that only allows for one vsix file to allow for this scenario

### DIFF
--- a/azure-pipelines/1esstages.yml
+++ b/azure-pipelines/1esstages.yml
@@ -16,7 +16,9 @@ parameters:
   - name: enableSigning
     type: boolean
     default: True
-
+  - name: vsixFileName
+    type: string
+    default: ''
 stages:
   - stage: BuildStage
     pool:
@@ -48,6 +50,7 @@ stages:
               - template: ./templates/sign.yml
                 parameters:
                   enableSigning: ${{ parameters.enableSigning }}
+                  vsixFileName: ${{ parameters.vsixFileName }}
               - template: ./templates/stage-artifacts.yml
               - template: ./templates/test.yml
                 parameters:
--- a/azure-pipelines/1esstages.yml
+++ b/azure-pipelines/1esstages.yml
@@ -47,10 +47,15 @@ stages:
               - template: ./templates/1espackage.yml
                 parameters:
                   additionalPostPackageSteps: ${{ parameters.additionalPostPackageSteps }}
-              - template: ./templates/sign.yml
-                parameters:
-                  enableSigning: ${{ parameters.enableSigning }}
-                  vsixFileNames: ${{ parameters.vsixFileNames }}
+              - ${{ if eq(join('', parameters.vsixFileNames), '') }}:
+                  - template: ./templates/sign.yml
+                    parameters:
+                      enableSigning: ${{ parameters.enableSigning }}
+              - ${{ if ne(join('', parameters.vsixFileNames), '') }}:
+                  - template: ./templates/signMultiple.yml
+                    parameters:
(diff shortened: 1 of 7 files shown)


azure-pipelines/1esstages.yml
```diff
@@ -45,9 +47,15 @@ stages:
...
                   additionalPostPackageSteps: ${{ parameters.additionalPostPackageSteps }}
-              - template: ./templates/sign.yml
-                parameters:
-                  enableSigning: ${{ parameters.enableSigning }}
+              - ${{ if eq(join('', parameters.vsixFileNames), '') }}:
+                  - template: ./templates/sign.yml
+                    parameters:
+                      enableSigning: ${{ parameters.enableSigning }}
+              - ${{ if ne(join('', parameters.vsixFileNames), '') }}:
+                  - template: ./templates/signMultiple.yml
```
bwateratmsft, line 55: We should avoid splitting signing into two different templates. There ought to be a way to do it in one template.

azure-pipelines/release-extension.yml
```diff
@@ -27,6 +27,19 @@ parameters:
...
+  # A list of publish commands to run. It defaults to publishing a single .vsix file. If you want to publish multiple .vsix files (platform specifc), you can add mutliple commands to this list.
+  - name: publishCommands
+    type: object
+    default:
+      [
+        "vsce publish --azure-credential --packagePath $(vsixFileName) --manifestPath extension.manifest --signaturePath extension.signature.p7s",
+      ]
+
+  # When true, skips check to make sure there are not multiple .vsix files found in the build directory
+  - name: publishMultipleVsixFiles
```
bwateratmsft, line 39: Can we just remove the check rather than needing to add a parameter for just it?

azure-pipelines/1esstages.yml
JasonYeMSFT: Can you always pass the vsix names to the sign.yml and have the internal logic of sign.yml to decide how to sign the vsix(s)?

azure-pipelines/release-extension.yml
JasonYeMSFT: I agree with Brandon. If the pipeline supports building/publishing single and multiple vsix at the same time we can just remove the check.

azure-pipelines/1esstages.yml
msalaman: Good point, I've merged the logic into the sign template

bwateratmsft (approved):

Issue #1958, opened 2025-04-03
Update activity logic in Azure Tools shared packages to support modernization and v2


MicroFish91:

MicroFish91:
This is basically done, see referenced PRs for more details.

PR #1962, opened 2025-04-07
utils: Update activity log logic to support the new v2 migration

MicroFish91:
Update: Pretty much ready for review now, still working on addressing the feedback already supplied.

In support of: https://github.com/microsoft/vscode-azureresourcegroups/pull/1113
Partially addresses: https://github.com/microsoft/vscode-azuretools/issues/1958

### DIFF
--- a/utils/src/activityLog/activities/ExecuteActivity.ts
+++ b/utils/src/activityLog/activities/ExecuteActivity.ts
@@ -6,9 +6,9 @@
 import * as vscode from 'vscode';
 import * as hTypes from '../../../hostapi';
 import * as types from '../../../index';
-import { activityFailContext, activityFailIcon } from '../../constants';
+import { activityErrorContext, activityFailContext, activityFailIcon } from '../../constants';
 import { ResourceGroupsItem } from '../../pickTreeItem/quickPickAzureResource/tempTypes';
-import { createGenericElement } from '../../tree/v2/createGenericElement';
+import { ActivityChildItem, ActivityChildType } from '../../tree/v2/ActivityChildItem';
 import { ActivityBase } from "../Activity";
 
 export class ExecuteActivity<TContext extends types.ExecuteActivityContext = types.ExecuteActivityContext> extends ActivityBase<void> {
@@ -35,11 +35,15 @@ export class ExecuteActivity<TContext extends types.ExecuteActivityContext = typ
                 }
 
                 return [
-                    createGenericElement({
+                    new ActivityChildItem({
                         contextValue: 'executeResult',
                         label: vscode.l10n.t("Click to view resource"),
-                        commandId: 'azureResourceGroups.revealResource',
-                        commandArgs: [resourceId],
+                        activityType: ActivityChildType.Command,
+                        command: {
+                            title: '',
+                            command: 'azureResourceGroups.revealResource',
+                            arguments: [resourceId],
+                        },
                     }),
                 ];
 
@@ -51,9 +55,10 @@ export class ExecuteActivity<TContext extends types.ExecuteActivityContext = typ
         return {
             label: this.label,
             getChildren: (_parent: ResourceGroupsItem) => {
-                const errorItemOptions: types.GenericElementOptions = {
-                    contextValue: 'executeError',
-                    label: error.message
+                const errorItemOptions: types.ActivityChildItemOptions = {
+                    label: error.message,
+                    contextValue: activityErrorContext,
+                    activityType: ActivityChildType.Error,
                 };
 
                 if (this.context.activityChildren) {
@@ -63,7 +68,7 @@ export class ExecuteActivity<TContext extends types.ExecuteActivityContext = typ
                     return activityChildren;
                 }
 
-                return [createGenericElement(errorItemOptions)];
+                return [new ActivityChildItem(errorItemOptions)];
             }
         }
     }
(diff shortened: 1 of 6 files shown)


utils/src/tree/v2/ActivityChildItem.ts
```diff
@@ -0,0 +1,36 @@
...
+    iconPath?: TreeItemIconPath;
+    initialCollapsibleState?: TreeItemCollapsibleState;
+
+    constructor(options: types.ActivityChildItemOptions) {
+        this.id = options.id;
+        this.label = options.label;
+        this.contextValue = options.contextValue;
+        this.description = options.description;
+        this.iconPath = options.iconPath;
+        this.initialCollapsibleState = options.initialCollapsibleState;
```
nturinski, line 24: Maybe we should include the ActivityStatus (it's an enum that has failed/success/running/etc.) on this as well?

That way you can just check its value instead of looking at the context value and doing the regex test on it as well.

nturinski (dismissed):
Looks good to me for the most part. I think we can always iterate on it more, ofc.

utils/src/tree/v2/ActivityChildItem.ts
```diff
@@ -0,0 +1,36 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.txt in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import { TreeItem, TreeItemCollapsibleState } from "vscode";
+import { TreeItemIconPath } from "../../..";
+import * as types from '../../../index';
+
+export class ActivityChildItem implements types.ActivityChildItemBase {
```
bwateratmsft, line 10: Shouldn't this have a `getChildren` method to fully implement `types.ActivityChildItemBase`? Otherwise did you mean for it to be `abstract`?

utils/src/tree/v2/ActivityChildItem.ts
MicroFish91: Yeah that could be nifty.  For this scenario, would you want the activity status to be a part of `ActivityChildItemBase` as well then?

utils/src/activityLog/activities/ExecuteActivity.ts
```diff
@@ -54,55 +50,57 @@ export class ExecuteActivity<TContext extends types.ExecuteActivityContext = typ
     public errorState(error: types.IParsedError): hTypes.ActivityTreeItemOptions {
         return {
             label: this.label,
-            getChildren: (parent: AzExtParentTreeItem) => {
-                const errorItem = new GenericTreeItem(parent, {
+            getChildren: (_parent: ResourceGroupsItem) => {
```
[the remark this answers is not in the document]
MicroFish91: Will reach out to discuss this one more offline

utils/src/tree/v2/ActivityChildItem.ts
nturinski: Yeah, I think that would make sense. It can be optional and default to "NotStarted" if not found, maybe?

MicroFish91:
Update:  So I ran into a new edge case with the v2 changes where `getChildren` is being called on the fail state more than once.  In the UI, this can manifest as more than one fail item showing up after refresh is called.  To combat this, I updated `errorState`'s `getChildren` to be idempotent.  Since part of the logic involves replacing the existing `getChildren` method, we want to make sure we only do this once.

I decided to add a `_hasBeenModified` property on the `ActivityChildItemBase` to track this so we don't inadvertently update more than once.  

@bwateratmsft I talked with Nathan a bit about this already on Teams but was wondering your thoughts about handling idempotency this way / if you had any thoughts on a better way to do this.  Thanks!

utils/src/tree/v2/ActivityChildItem.ts
MicroFish91: Oh I meant for `getChildren` to be optional

utils/src/tree/v2/ActivityChildItem.ts
MicroFish91: This one?

image

I was thinking some more about this, and I don't know if `NotStarted` would apply.  Execute steps only provide items once they've started running, so anything `NotStarted` just wouldn't show up I think.  Also, not all items we discussed using will necessarily go through this full lifecycle.  It makes sense for the `ActivityItem`, just not sure if it's the best representation for `ActivityChildItem`.

Is something like `activityType` more helpful?  Example:

1. 'success'
2. 'fail'
3. 'progress'
4. 'info'
5. 'error'
6. Anything else?

I assume you can infer the `activityStatus` (when it applies) based on this type?  Let me know what you guys are thinking

nturinski: Yeah, I get what you mean. This feels fine to me.

I'm not sure I 100% understand the distinction between fail and error.

MicroFish91: Oh yeah let me provide a couple images that will better help illustrate the difference by what I mean there

nturinski (dismissed):
I think I'm okay with this for the most part. I kind of want to get something we can start integrating with and see where the shortcomings are by like tomorrow or Thursday.

utils/src/activityLog/activities/ExecuteActivity.ts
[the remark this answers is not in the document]
MicroFish91: Since we probably want to push this through soon lets just revisit this in a future PR if needed.

utils/src/tree/v2/ActivityChildItem.ts
MicroFish91: image
`Create Network Security Group....` is a fail item indicating the step that failed
`Network security group cannot be created...` is an error item indicating the error message

nturinski (approved):

PR #1966, opened 2025-04-10
azure: Upgrade utils package to `v3.0.0` and `@types/vscode` to `v1.95.0`

MicroFish91:

### DIFF
--- a/azure/src/createAzureClient.ts
+++ b/azure/src/createAzureClient.ts
@@ -140,7 +140,7 @@ export class CorrelationIdPolicy implements PipelinePolicy {
 
     public async sendRequest(request: PipelineRequest, next: SendRequest): Promise<PipelineResponse> {
         const headerName = 'x-ms-correlation-request-id';
-        const id: string = (this.context.telemetry.properties[headerName] as string) ||= uuidv4();
+        const id: string = (this.context.telemetry.properties[headerName] as string | undefined) ||= uuidv4();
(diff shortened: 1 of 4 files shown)


azure/src/createAzureClient.ts
```diff
@@ -140,7 +140,7 @@ export class CorrelationIdPolicy implements PipelinePolicy {
 
     public async sendRequest(request: PipelineRequest, next: SendRequest): Promise<PipelineResponse> {
         const headerName = 'x-ms-correlation-request-id';
-        const id: string = this.context.telemetry.properties[headerName] ||= uuidv4();
+        const id: string = (this.context.telemetry.properties[headerName] as string | undefined) ||= uuidv4();
```
MicroFish91, line 143: @alexweininger Just checking, but I was getting a type error here due to the new possibility for TrustedTelemetry type, what's the recommended way to handle this, do we need to typecast?

nturinski: I say that we don't allow generic telemetry properties be TrustedTelemetryValue types. We should reserve that for only specific properties that we sign off on as trusted. Otherwise we have to deal with this annoying typing and I find that it's a bit of a security/privacy risk,

nturinski (approved):

MicroFish91:
Sorry staled the review, had to change the version bump

nturinski (approved):

PR #1967, opened 2025-04-11
utils: Replace activity progress report messages with a live timer

MicroFish91:
We had discussed adding a live activity timer and hiding progress reports from the activity log description.  This would help accomplish that.  

Corresponding Resource Groups PR: https://github.com/microsoft/vscode-azureresourcegroups/pull/1119

(Note: Progress reports would still be used commands leveraging the Azure Wizard without registering an activity, see below.)
https://github.com/microsoft/vscode-azuretools/blob/713123e274ceb0ae434395bbe80048e6780cddfa/utils/src/wizard/AzureWizard.ts#L293


utils/src/activityLog/Activity.ts
```diff
@@ -96,4 +102,16 @@ export abstract class ActivityBase<R> implements hTypes.Activity {
...
+    private startTimer(startTimeMs: number): void {
+        this.timer = setInterval(() => {
+            this.timerMessage = dateTimeUtils.getFormattedDurationInMinutesAndSeconds(Date.now() - startTimeMs);
+            this.report();
+        }, 1000);
+    }
+
+    public dispose(): void {
+        clearInterval(this.timer);
+        this.cancellationTokenSource.dispose();
```
bwateratmsft, line 115: CTS is optional, right?

```suggestion
        this.cancellationTokenSource?.dispose();
```

utils/src/activityLog/Activity.ts
```diff
@@ -80,6 +85,7 @@ export abstract class ActivityBase<R> implements hTypes.Activity {
             this._onErrorEmitter.fire({ ...this.getState(), error: e });
             throw e;
         } finally {
+            this.dispose();
```
bwateratmsft, line 88: Does `run()` always get run? If so, this object is always disposing of itself, and never needs to be disposed by the caller. In that case I wouldn't make it implement `vscode.Disposable`.

MicroFish91: Oh shoot good point, run doesn't always get run, I'll remove implementing disposable

utils/src/activityLog/Activity.ts
bwateratmsft: Actually, this class isn't the owner of this CTS, so we shouldn't be disposing it.

bwateratmsft: Nevermind this comment chain, CTS is always defined. It's optional on the interface but always present in this class.

utils/src/activityLog/activities/ExecuteActivity.ts
```diff
@@ -24,6 +24,12 @@ export class ExecuteActivity<TContext extends types.ExecuteActivityContext = typ
         }
     }
 
+    protected override report(progress?: { message?: string; increment?: number }): void {
+        const message: string | undefined = this.context.activityChildren?.length ? this.timerMessage : progress?.message;
+        this._onProgressEmitter.fire({ ...this.getState(), message });
```
MicroFish91, line 29: Made a new change here.  I decided to override the report method in the concrete class.   Two reasons for this:
1) The concrete class is what actually holds the `ExecuteActivityContext`
2) The abstract class base implementation doesn't seem coupled to the Wizard, so I didn't want to pollute its implementation with Azure Wizard-y things.

The way this behaves:
- If an activity has children, we just show a timer as the description.  We offload the responsibility of showing what's being executed to the children
- If no children, we still should indicate what's happening, so let's default back to any progress report messages in the description

utils/src/activityLog/activities/ExecuteActivity.ts
```diff
@@ -24,6 +24,12 @@ export class ExecuteActivity<TContext extends types.ExecuteActivityContext = typ
         }
     }
 
+    protected override report(progress?: { message?: string; increment?: number }): void {
+        const message: string | undefined = this.context.activityChildren?.length ? this.timerMessage : progress?.message;
```
nturinski, line 28: nit: I think a comment might be worthwhile here rationalizing why we decided to keep progress.message when there's no activity children.

nturinski (approved):

nturinski (approved):

PR #1968, opened 2025-04-15
utils: Update `ActivityChildItemBase` format

MicroFish91:
I was messing around some more with parsing the types using LM tools, and these additions to `ActivityChildItemBase` would help

### DIFF
--- a/utils/src/tree/v2/ActivityChildItem.ts
+++ b/utils/src/tree/v2/ActivityChildItem.ts
@@ -5,7 +5,6 @@
 
 import { v4 as uuidv4 } from "uuid";
 import { ProviderResult, TreeItem } from "vscode";
-import { TreeElementBase } from "../../..";
 import * as types from '../../../index';
 
 export enum ActivityChildType {
@@ -19,12 +18,14 @@ export enum ActivityChildType {
 
 export class ActivityChildItem implements types.ActivityChildItemBase {
     readonly id: string;
+    label: string;
     contextValue: string;
     activityType: ActivityChildType;
     description?: string;
 
     constructor(readonly options: types.ActivityChildItemOptions) {
         this.id = options.id ?? uuidv4();
+        this.label = options.label;
         this.activityType = options.activityType;
         this.contextValue = options.contextValue;
         this.description = options.description;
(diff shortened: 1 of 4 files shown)


utils/src/tree/v2/ActivityChildItem.ts
```diff
@@ -47,5 +48,5 @@ export class ActivityChildItem implements types.ActivityChildItemBase {
         };
     }
 
-    getChildren?(): ProviderResult<TreeElementBase[]>;
+    getChildren?(): ProviderResult<types.ActivityChildItemBase[]>;
```
bwateratmsft, line 51:
```suggestion
    abstract getChildren?(): ProviderResult<types.ActivityChildItemBase[]>;
```

MicroFish91: If marking `getChildren()` as abstract, was the intention to also make `getChildren` required? (Just because I don't know if I've ever seen an optional abstract method before).  

I'd also lean towards wanting to keep the method optional just because not all `ActivityChildItems` will have additional children (unless your preference is to always have the method defined and and return an empty list in the case of no children).

MicroFish91: Hmm, s most of the time, this class is just used as a leaf node. But I also wanted to use it for error items, which need `getChildren` so we can attach an error message.

I didn’t want to make two separate classes for parent and non-parent since they’re almost the same. But yeah, the name is probably confusing — I probably should call it `GenericActivityChildItem` instead.

MicroFish91: Oh, I think I finally understand what the miscommunication is here.  So `getChildren` does actually get set sometimes within this class with logic in the constructor.  That's why I have it optionally defined in this class and am not just leaving it to be defined in the interface it's implementing.

image

nturinski: Another thing Matt didn't mention is that `ActivityChildItem` is being instantiated in our code. It's automatically created by a type of execute step called `AzureWizardExecuteStepWithActivityOutput`.

{A24A45B7-26C5-4CEB-9306-EFE7525ACD92}


This step can have two types of children-- a pass or fail child. The pass child should be a leaf node since no further information is required. The fail child needs to be a parent because we want to append the error as its child.

bwateratmsft: Ahh, I see. Ok, in that case it does make sense to keep it.

bwateratmsft (approved):

PR #1969, opened 2025-04-17
Utils: Add an addExecuteSteps property to inject steps into the azure wizard

nturinski:
There are cases where we need to add additional steps based on the result of an execute step.

1. An execute step fails (maybe due to an access restriction) and we want to add another step to re-prompt the user
2. Role assignments aren't known until a previous execute step. This step would inject more steps based on which roles are going to be required.

Note for cases where we _expect_ errors that you will need to enable the `options.continueOnFail` flag.

Also ignore all of the other commits; I forgot to change from an old branch but it shouldn't impact the merge history.


utils/index.d.ts
```diff
@@ -1430,6 +1430,11 @@ export declare abstract class AzureWizardExecuteStep<T extends IActionContext &
      */
     public abstract shouldExecute(wizardContext: T): boolean;
 
+    /**
+     * Optionally returns AzureWizardExecuteSteps that will be injected and sorted into the Azure Wizard executeSteps array by priority
```
bwateratmsft, line 1434: This behavior could result in steps happening out of the expected order, e.g. if I inject one with priority 90, from a step with priority 100, it'll happen in reverse order of usual. We should maybe note that in the method description.

bwateratmsft (dismissed):

utils/index.d.ts
nturinski: I can note that, but if someone injects a step with a higher priority than the one that created it, I think they deserve to have some bugs.

But I do re-sort the array by priority every time a step is injected.

alexweininger (approved):

MicroFish91 (approved):

MicroFish91:
Nice, this will likely be useful for `ResourceGroupCreateStep`

PR #1977, opened 2025-04-22
utils: Add a fail-safe that will add the last reported Azure sub ID to properties

nturinski:
This adds a property to the internal telemetry reporter that will save the last reported Azure sub ID. I think it's safe to do this, even for commands that don't directly interact with Azure, because we're more interested in users that are engaged/signed in Azure even if they happened to do a local command (such as create a new project).

### DIFF
--- a/utils/src/callWithTelemetryAndErrorHandling.ts
+++ b/utils/src/callWithTelemetryAndErrorHandling.ts
@@ -237,6 +237,13 @@ function handleTelemetry(context: types.IActionContext, callbackId: string, star
                     }
                 }
             }
+            if (context.telemetry.properties.subscriptionId) {
+                // set the lastReported subscription id to the current one if it exists
+                ext._internalReporter.lastReportedAzureSubscriptionId = context.telemetry.properties.subscriptionId;
+            } else {
+                // use the last reported subscription id if no subscription id was set on the telemetry properties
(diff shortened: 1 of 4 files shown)


utils/src/callWithTelemetryAndErrorHandling.ts
```diff
@@ -237,6 +237,13 @@ function handleTelemetry(context: types.IActionContext, callbackId: string, star
                     }
                 }
             }
+            if (context.telemetry.properties.subscriptionId) {
+                // set the lastReported subscription id to the current one if it exists
+                ext._internalReporter.lastReportedAzureSubscriptionId = context.telemetry.properties.subscriptionId;
+            } else {
+                // use the last reported subscription id if no subscription id was set on the telemetry properties
+                context.telemetry.properties.subscriptionId = ext._internalReporter.lastReportedAzureSubscriptionId;
+            }
```
bwateratmsft, line 246: This seems like something that is better solved at the time of Kusto querying, rather than this. Failing that, I'd try to create a telemetry handler (https://github.com/microsoft/vscode-azuretools/blob/39c8500374abeb2c82a19d81c968ecaeb167bf38/utils/src/callWithTelemetryAndErrorHandling.ts#L97-L99) to do this if possible.

nturinski:
Talked about this with Brandon offline and am closing it.

PR #1982, opened 2025-04-25
appservice: Convert deployment steps to output activity

nturinski:
It looks a little bit jank still to be honest, but it's pretty tough because of how the steps are broken out. If I don't have progress for the Wait step, it looks like the deployment is frozen. I think this could be improved, but given our timeline, I think that this will have to do for now.

{505D7340-DEFB-469B-AAAC-C26C26C31AD6}

### DIFF
--- a/appservice/src/deploy/wizard/StopAppBeforeDeployExecuteStep.ts
+++ b/appservice/src/deploy/wizard/StopAppBeforeDeployExecuteStep.ts
@@ -3,20 +3,29 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { AzureWizardExecuteStep } from "@microsoft/vscode-azext-utils";
+import { AzureWizardExecuteStepWithActivityOutput } from "@microsoft/vscode-azext-utils";
 import { Progress, l10n } from "vscode";
-import { ext } from "../../extensionVariables";
 import { InnerDeployContext } from "../IDeployContext";
 
-export class StopAppBeforeDeployExecuteStep extends AzureWizardExecuteStep<InnerDeployContext> {
+export class StopAppBeforeDeployExecuteStep extends AzureWizardExecuteStepWithActivityOutput<InnerDeployContext> {
+    stepName: string = 'StopAppBeforeDeployExecuteStep';
+    protected getTreeItemLabel(context: InnerDeployContext): string {
+        return l10n.t('Stop app "{0}" before deployment', context.site.fullName);
+    }
+    protected getOutputLogSuccess(context: InnerDeployContext): string {
+        return l10n.t('Successfully stopped app "{0}".', context.site.fullName);
+    }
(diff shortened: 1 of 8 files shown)


appservice/src/deploy/wizard/deployZip/WaitForDeploymentToCompleteStep.ts
```diff
@@ -3,13 +3,48 @@
...
 import { waitForDeploymentToComplete } from "../../waitForDeploymentToComplete";
 
 export class WaitForDeploymentToCompleteStep extends AzureWizardExecuteStep<InnerDeployContext> {
+    public stepName: string = 'WaitForDeploymentToCompleteStep';
     public priority: number = 210;
+    public createSuccessOutput(context: InnerDeployContext): ExecuteActivityOutput {
+        return {
+            item: new ActivityChildItem({
+                contextValue: createContextValue([activitySuccessContext, context.site.id]),
+                label: l10n.t('Finished deployment pipeline.'),
```
MicroFish91, line 18: Can we make the tree item labels for these children present tense

MicroFish91:
> It looks a little bit jank still to be honest

What part, the transitions in the activity log?  I probably need to set aside some time and try to run through this to see exactly.  I already can kind of guess why since we have that whole `wait and poll` logic.  I had to do something similar in ACA that I didn't like and was hoping to revisit in the future as well.

appservice/src/deploy/wizard/deployZip/WaitForDeploymentToCompleteStep.ts
nturinski: It's a bit odd if it's present tense in this case. Deployment is kind of roughly broken into two steps-- deploying the package itself and then pinging the logs to see when it completed. 

I know generally we adopt a pattern where the tree label doesn't change throughout progress/success, but I felt like in this case, it doesn't make a lot of sense to do that.

nturinski:
Added a child node to display output channel to build step:
image

appservice/src/deploy/wizard/StopAppBeforeDeployExecuteStep.ts
```diff
@@ -3,20 +3,29 @@
...
+    }
+
     public priority: number = 100;
-    public async execute(context: InnerDeployContext, progress: Progress<{ message?: string; increment?: number }>): Promise<void> {
+    public async execute(context: InnerDeployContext, _progress: Progress<{ message?: string; increment?: number }>): Promise<void> {
         const site = context.site;
         const client = await site.createClient(context);
-
-        const stoppingApp = l10n.t('Stopping app...');
-        progress.report({ message: stoppingApp });
```
MicroFish91, line 18: I'll let you make the final call, but there's a couple reasons we might still want to include progress reports:

1. I think it fires the `onProgress` event and refreshes the activity tree item
2. For users of the wizard that aren't using activities, it provides a message to show in the `vscode.window.showProgress` box

PR #1983, opened 2025-04-28
utils: Confirm id being added to telemetry is an azure id

nturinski:
We were adding a lot of just random node ids to telemetry, like "slots" and so on. The check should ensure that only Azure IDs are being attached to the telemetry context.

### DIFF
--- a/utils/src/utils/AzureResourceIdTelemetry.ts
+++ b/utils/src/utils/AzureResourceIdTelemetry.ts
@@ -1,0 +1,18 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.txt in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import { TelemetryTrustedValue } from "vscode";
+import { IActionContext } from "../..";
+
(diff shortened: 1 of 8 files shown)


motm32 (approved):

utils/src/utils/AzureResourceIdTelemetry.ts
```diff
@@ -0,0 +1,18 @@
...
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.txt in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import { TelemetryTrustedValue } from "vscode";
+import { IActionContext } from "../..";
+
+export function setAzureResourceIdTelemetryProperties(context: IActionContext, resourceId: string): void {
+    if (isAzureResourceId(resourceId)) {
+        context.telemetry.properties.resourceId = new TelemetryTrustedValue(resourceId);
```
MicroFish91, line 11: In any given command scenario, is the assumption that we will only ever want to populate one resource id?

nturinski: Yep, this is only for actions against a specific resource. So like, if I deploy to my app, etc.

PR #1987, opened 2025-04-30
utils: `TelemetryTrustedValues` fix

nturinski:
`TelemetryTrustedValues` were automatically getting converted to strings thus why we were seeing "[object Object]" in telemetry. This skips masking those values now.


MicroFish91 (approved):

alexweininger (approved):
😎

PR #1992, opened 2025-05-01
azure: Add a `ResourceGroupVerifyStep`

MicroFish91:
### Overview
`ResourceGroupCreateStep` originally handled multiple responsibilities, so I refactored it to better separate concerns and improve clarity in the activity log. Since the step previously included verification logic, I extracted that into a new `ResourceGroupVerifyStep`. This allows the verification process to be shown more transparently as a distinct entry in the activity log.

To maintain backwards compatibility, I also kept the original verification logic in the `configureBeforeExecute` method of `ResourceGroupCreateStep`. This ensures existing behavior remains unchanged, while giving users the option to add the new `ResourceGroupVerifyStep` for improved UI visibility at any time that they wish.

### Example Screenshots by Scenario

Create with resource group not already existing:
image

Create with resource group already existing:
image

w/ 403 error:
image

w/o 403 error:
image

### DIFF
--- a/azure/src/wizard/ResourceGroupVerifyStep.ts
+++ b/azure/src/wizard/ResourceGroupVerifyStep.ts
@@ -24,8 +24,8 @@ export class ResourceGroupVerifyStep<T extends types.IResourceGroupWizardContext
     }
     protected getTreeItemLabel(context: T) {
         return context.resourceGroup?.name ?
-            l10n.t('Verify resource group "{0}" (exists)', context.resourceGroup.name) :
-            l10n.t('Verify resource group "{0}" (available)', nonNullProp(context, 'newResourceGroupName'));
+            l10n.t('Verify resource group "{0}" exists', context.resourceGroup.name) :
+            l10n.t('Verify resource group "{0}" available', nonNullProp(context, 'newResourceGroupName'));
     }
 
     public async execute(context: T, progress: Progress<{ message?: string; increment?: number }>): Promise<void> {
--- a/azure/src/wizard/ResourceGroupVerifyStep.ts
+++ b/azure/src/wizard/ResourceGroupVerifyStep.ts
@@ -14,7 +14,6 @@ import { ext } from '../extensionVariables';
 export class ResourceGroupVerifyStep<T extends types.IResourceGroupWizardContext> extends AzureWizardExecuteStepWithActivityOutput<T> {
     public priority: number = 95;
     public stepName: string = 'resourceGroupVerifyStep';
-    private error: unknown;
 
     protected getOutputLogFail = (context: T) => l10n.t('Failed to verify whether resource group with name "{0}" already exists.', nonNullProp(context, 'newResourceGroupName'));
     protected getOutputLogSuccess(context: T) {
@@ -46,7 +45,6 @@ export class ResourceGroupVerifyStep<T extends types.IResourceGroupWizardContext
                 // Continue - we might still be able to handle missing create permissions in the create step
                 this.options.continueOnFail = true;
             }
-            this.error = error;
             throw error;
         }
     }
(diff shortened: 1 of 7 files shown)


azure/src/wizard/ResourceGroupVerifyStep.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    protected getOutputLogFail = (context: T) => l10n.t('Failed to verify whether resource group with name "{0}" already exists.', nonNullProp(context, 'newResourceGroupName'));
+    protected getOutputLogSuccess(context: T) {
+        return context.resourceGroup?.name ?
+            l10n.t('Verified resource group with name "{0}" already exists.  Using existing resource group.', context.resourceGroup.name) :
+            l10n.t('Verified resource group with name "{0}" is available for creation.', nonNullProp(context, 'newResourceGroupName'));
+    }
+    protected getTreeItemLabel(context: T) {
+        return context.resourceGroup?.name ?
+            l10n.t('Verify resource group "{0}" (exists)', context.resourceGroup.name) :
```
nturinski, line 27: I think you could just remove the parens so it says Verify resource group exists and Verify resource group is available.

azure/src/wizard/ResourceGroupVerifyStep.ts
```diff
@@ -0,0 +1,84 @@
...
+            l10n.t('Verify resource group "{0}" (available)', nonNullProp(context, 'newResourceGroupName'));
+    }
+
+    public async execute(context: T, progress: Progress<{ message?: string; increment?: number }>): Promise<void> {
+        progress.report({ message: l10n.t('Checking resource group availability...') });
+
+        const newName: string = nonNullProp(context, 'newResourceGroupName');
```
nturinski, line 34: You don't need to check for `context.newResourceGroupName` || `context.resourceGroup?.name`? I thought that `newResourceGroupName` should be undefined if the user selected an existing resource group.

MicroFish91: Well it shouldn't execute if `context.resourceGroup` exists.  The original goal of this implementation was to extract the logic out of `ResourceGroupCreateStep` which also only executes when `context.resourceGroup` doesn't exist.  Do we want to expand the way that this step gets used?

nturinski: If that were the case, then I'm a little confused because there seems to be two scenarios: 1. when the resource group exists and 2. when the resource group is available.

Would 1. be when you are using an existing resource group and 2. be when you have created one?

MicroFish91: Yeah, in my opinion this is generally what should be handled on the `ResourceGroupNameStep` since we added the async validation method which allows us to check whether or not the name is a duplicate of an existing resource.

In this implementation I'm mostly just trying to preserve the existing behavior because I don't know all the design reasons that went into adding the verify logic in the execute phase.  

For example, I know we also sometimes generate related names automatically.  What I don't know is whether or not this logic helps support that in the sense that could the generated name accidentally be of an existing resource group

MicroFish91: > Would 1. be when you are using an existing resource group and 2. be when you have created one?

1 is when your `newResourceGroupName` is the duplicate of an existing resource, 2 is when you have a `newResourceGroupName` that isn't a duplicate

azure/src/wizard/ResourceGroupVerifyStep.ts
```diff
@@ -0,0 +1,84 @@
...
+    protected getOutputLogFail = (context: T) => l10n.t('Failed to verify whether resource group with name "{0}" already exists.', nonNullProp(context, 'newResourceGroupName'));
+    protected getOutputLogSuccess(context: T) {
+        return context.resourceGroup?.name ?
+            l10n.t('Verified resource group with name "{0}" already exists.  Using existing resource group.', context.resourceGroup.name) :
+            l10n.t('Verified resource group with name "{0}" is available for creation.', nonNullProp(context, 'newResourceGroupName'));
+    }
+    protected getTreeItemLabel(context: T) {
+        return context.resourceGroup?.name ?
+            l10n.t('Verify resource group "{0}" exists', context.resourceGroup.name) :
+            l10n.t('Verify resource group "{0}" available', nonNullProp(context, 'newResourceGroupName'));
```
nturinski, line 28:
```suggestion
            l10n.t('Verify resource group "{0}" is available', nonNullProp(context, 'newResourceGroupName'));
```

nturinski: Actually, I wouldn't recommend using nonNullProp here. Just in case it's empty for some weird reason, rather have the tree label just be undefined or "" instead of throwing an error.

nturinski (approved):

PR #1994, opened 2025-05-02
appservice: Migrate `AppInsightsCreateStep` and `AppServicePlanCreateStep` to contribute activity outputs

motm32:
Code based on this [PR](https://github.com/microsoft/vscode-azuretools/pull/1980)

Still working on testing all the scenarios so will merge once testing is complete.

### DIFF
--- a/appservice/src/createAppService/AppInsightsCreateStep.ts
+++ b/appservice/src/createAppService/AppInsightsCreateStep.ts
@@ -6,7 +6,7 @@
 import type { ApplicationInsightsManagementClient } from '@azure/arm-appinsights';
 import type { ResourceGroup } from '@azure/arm-resources';
 import { AzExtLocation, LocationListStep } from '@microsoft/vscode-azext-azureutils';
-import { ActivityChildItem, ActivityChildType, activityFailContext, activityFailIcon, AzureWizardExecuteStepWithActivityOutput, createContextValue, ExecuteActivityContext, ExecuteActivityOutput, nonNullProp, nonNullValueAndProp, parseError } from '@microsoft/vscode-azext-utils';
+import { ActivityChildItem, ActivityChildType, activityFailContext, activityFailIcon, ActivityOutputType, AzureWizardExecuteStepWithActivityOutput, createContextValue, ExecuteActivityContext, ExecuteActivityOutput, nonNullProp, nonNullValueAndProp, parseError } from '@microsoft/vscode-azext-utils';
 import { v4 as uuidv4 } from "uuid";
 import { l10n, MessageItem, Progress, TreeItemCollapsibleState } from 'vscode';
 import { ext } from '../extensionVariables';
@@ -79,18 +79,19 @@ export class AppInsightsCreateStep extends AzureWizardExecuteStepWithActivityOut
                 if (pError.errorType === 'AuthorizationFailed') {
                     if (!context.advancedCreation) {
                         this._skippedCreate = true;
-                    } else {
-                        ext.outputChannel.appendLog(l10n.t('Failed to create application insights "{0}".', aiName));
-                        ext.outputChannel.appendLog(l10n.t('Unable to create application insights "{0}" in subscription "{1}" due to a lack of permissions.', aiName, context.subscriptionDisplayName));
-                        this.isMissingCreatePermissions = true;
-                        this.options.continueOnFail = true;
-                        this.addExecuteSteps = () => [new AppInsightsNoCreatePermissionsStep()];
+                        return;
                     }
-                } else {
-                    ext.outputChannel.appendLog(l10n.t('Failed to create application insights "{0}".', aiName));
-                    ext.outputChannel.appendLog(l10n.t('Error: {0}', parseError(error).message));
+
+                    this.isMissingCreatePermissions = true;
+                    this.options.continueOnFail = true;
+                    this.addExecuteSteps = () => [new AppInsightsNoCreatePermissionsStep()];
+                    // Suppress generic output and replace with custom logs
+                    this.options.suppressActivityOutput = ActivityOutputType.Message;
+                    ext.outputChannel.appendLog(l10n.t('Unable to create application insights "{0}" in subscription "{1}" due to a lack of permissions.', aiName, context.subscriptionDisplayName));
+                    ext.outputChannel.appendLog(pError.message);
                     throw error;
                 }
+                throw error;
             }
         }
     }
--- a/appservice/src/createAppService/AppInsightsCreateStep.ts
+++ b/appservice/src/createAppService/AppInsightsCreateStep.ts
@@ -92,6 +92,8 @@ export class AppInsightsCreateStep extends AzureWizardExecuteStepWithActivityOut
                 }
                 throw error;
             }
+        } else {
+            this._skippedCreate = true;
         }
     }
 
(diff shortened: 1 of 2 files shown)


MicroFish91:
This is unrelated to your change but is something that I noticed.  Should there be an else where skipped create is set to true?

```typescript
if (appInsightsLocation) {
  // ...
} else {
  this._skippedCreate = true;
}

```

appservice/src/createAppService/AppInsightsCreateStep.ts
```diff
@@ -61,56 +65,91 @@ export class AppInsightsCreateStep extends AzureWizardExecuteStepWithActivityOut
...
                     if (!context.advancedCreation) {
                         this._skippedCreate = true;
                     } else {
-                        await this.selectExistingPrompt(context);
-                        this._usedExistingAppInsights = true;
+                        ext.outputChannel.appendLog(l10n.t('Failed to create application insights "{0}".', aiName));
+                        ext.outputChannel.appendLog(l10n.t('Unable to create application insights "{0}" in subscription "{1}" due to a lack of permissions.', aiName, context.subscriptionDisplayName));
+                        this.isMissingCreatePermissions = true;
+                        this.options.continueOnFail = true;
+                        this.addExecuteSteps = () => [new AppInsightsNoCreatePermissionsStep()];
```
MicroFish91, line 87: There's two small changes I would make here:

1. Throw an error in the else path that includes `.addExecuteSteps`.  This way the step fail output gets triggered.  We will still keep running the rest of the steps since `continueOnFail` was set to `true`.
2. The else branch with all the output logs ends up being mostly redundant now because the activity log setup should automatically handle displaying the error message as part of the error item while the output log will automatically display the results of `getOutputLogFail`.

One way to refactor the catch block could be to do something like:

```typescript
catch (error) {
  const pError = parseError(error);
  if (pError.errorType === 'AuthorizationFailed') {
      if (!context.advancedCreation) {
          this._skippedCreate = true;
          return;
      } else {
          ext.outputChannel.appendLog(l10n.t('Unable to create application insights "{0}" in subscription "{1}" due to a lack of permissions.', aiName, context.subscriptionDisplayName));
          this.isMissingCreatePermissions = true;
          this.options.continueOnFail = true;
          this.addExecuteSteps = () => [new AppInsightsNoCreatePermissionsStep()];
      }
  }
  throw error;
}
```

appservice/src/createAppService/AppInsightsCreateStep.ts
```diff
@@ -61,56 +65,91 @@ export class AppInsightsCreateStep extends AzureWizardExecuteStepWithActivityOut
...
-                        });
-                } else if (pError.errorType === 'AuthorizationFailed') {
+                const pError = parseError(error);
+                if (pError.errorType === 'AuthorizationFailed') {
                     if (!context.advancedCreation) {
                         this._skippedCreate = true;
                     } else {
-                        await this.selectExistingPrompt(context);
-                        this._usedExistingAppInsights = true;
+                        ext.outputChannel.appendLog(l10n.t('Failed to create application insights "{0}".', aiName));
```
MicroFish91, line 83: Nit: This `Failed to create...` log says a lot of the same things as the line below it, I would just pick one of the two to display.

appservice/src/createAppService/AppServicePlanCreateStep.ts
```diff
@@ -17,62 +19,107 @@ import { CustomLocation, IAppServiceWizardContext } from './IAppServiceWizardCon
...
+            const client: WebSiteManagementClient = await createWebSiteClient(context);
+            context.plan = await client.appServicePlans.beginCreateOrUpdateAndWait(rgName, newPlanName, await getNewPlan(context));
         } catch (e) {
             if (parseError(e).errorType === 'AuthorizationFailed') {
-                await this.selectExistingPrompt(context);
+                ext.outputChannel.appendLog(l10n.t('Failed to create app service plan "{0}".', newPlanName));
+                ext.outputChannel.appendLog(l10n.t('Unable to create app service plan "{0}" in subscription "{1}" due to a lack of permissions.', newPlanName, context.subscriptionDisplayName));
+                this.isMissingCreatePermissions = true;
+                this.options.continueOnFail = true;
+                this.addExecuteSteps = () => [new AppServicePlanNoCreatePermissionsStep()];
```
MicroFish91, line 61: Similar feedback for this step as noted from the create step

nturinski:
This guy and his guard clauses...

Should we just go ahead and adopt the `VerifyStep` pattern for all of these resources though? I guess it can be broken into 2 PRs.

appservice/src/createAppService/AppInsightsCreateStep.ts
MicroFish91, line 87: Some recommendations:

1. The log message for `Failed to create...` (line 83) sounds pretty similar to the log message immediately below it, so my preference would be to only display one of the two.
2. I would throw an error at the end of the else branch of code that has the `.addExecuteSteps`.  This way the create activity shows up as a fail with the reason attached.  The `continueOnFail` is set to true, so the wizard will keep running the rest of the steps, allowing the user to select a new one.
3. In the else branch on line 90, the messages will end up being redundant since we'll attach the error message to the activity tree item on fail, and the `getOutputLogFail` should be automatically displayed as well.

Also I made a small change to the resource group PR where I suppress the activity output message for the special case, I think that would probably be a useful pattern here as well.  

Example:

```typescript
catch (error) {
  const pError = parseError(error);
  if (pError.errorType === 'AuthorizationFailed') {
      if (!context.advancedCreation) {
          this._skippedCreate = true;
          return;
      }

      this.isMissingCreatePermissions = true;
      this.options.continueOnFail = true;
      this.addExecuteSteps = () => [new AppInsightsNoCreatePermissionsStep()];

      // Suppress generic output and replace with custom logs
      this.options.suppressActivityOutput = ActivityOutputType.Message;
      ext.outputChannel.appendLog(l10n.t('Unable to create application insights "{0}" in subscription "{1}" due to a lack of permissions.', aiName, context.subscriptionDisplayName));
      ext.outputChannel.appendLog(pError.message);
  }
  throw error;
}
```

appservice/src/createAppService/AppInsightsCreateStep.ts
```diff
@@ -61,56 +65,92 @@ export class AppInsightsCreateStep extends AzureWizardExecuteStepWithActivityOut
...
-                } else {
+
+                    this.isMissingCreatePermissions = true;
+                    this.options.continueOnFail = true;
+                    this.addExecuteSteps = () => [new AppInsightsNoCreatePermissionsStep()];
+                    // Suppress generic output and replace with custom logs
+                    this.options.suppressActivityOutput = ActivityOutputType.Message;
+                    ext.outputChannel.appendLog(l10n.t('Unable to create application insights "{0}" in subscription "{1}" due to a lack of permissions.', aiName, context.subscriptionDisplayName));
+                    ext.outputChannel.appendLog(pError.message);
                     throw error;
```
MicroFish91, line 92: nit, there's an extra `throw error` here.

appservice/src/createAppService/AppInsightsCreateStep.ts
```diff
@@ -17,36 +18,39 @@ import { IAppServiceWizardContext } from './IAppServiceWizardContext';
...
         }
         return message;
     }
-    protected getOutputLogSuccess(context: IAppServiceWizardContext): string {
-        const aiName: string = nonNullProp(context, 'newAppInsightsName');
-        let message: string = l10n.t('Successfully created application insights "{0}".', aiName);
+    protected getOutputLogFail = (context: IAppServiceWizardContext) => l10n.t('Failed to create application insights "{0}"', nonNullProp(context, 'newAppInsightsName'));
+    protected getTreeItemLabel(context: IAppServiceWizardContext): string {
+        let message: string = l10n.t('Create application insights "{0}"', nonNullProp(context, 'newAppInsightsName'));
         if (this._skippedCreate) {
```
MicroFish91, line 33: Although not related to your change, for the `if (appInsightsLocation) {}` section below, I think we should handle the condition when there isn't an app insights location.  Since nothing gets created in that case, I think we will want to set the `_skippedCreate` to true so that the skip tree item label is used, otherwise the tree item label will not match what happened.

MicroFish91 (approved):

PR #2003, opened 2025-05-13
azure: Fix output strings for resource group create step

nturinski:
Fixes https://github.com/microsoft/vscode-azureresourcegroups/issues/1141

There's no rush to get this into the release, as far as I'm concerned. We can just pick up these changes whenever the next azure utils package is published.


MicroFish91 (approved):

MicroFish91:
Although nothing wrong with adding progress messages to the output log, I generally have been avoiding it as I think it's not super useful information now that the activity log progress exists

MicroFish91 (approved):

PR #2005, opened 2025-05-14
utils: Add an `insights` property for Copilot data (formerly referred to as `commandMetadata`)

MicroFish91:
WIP.... but probably won't add much more to this initially, trying to keep the first pass fairly simple and add to it as needed

### DIFF
--- a/utils/index.d.ts
+++ b/utils/index.d.ts
@@ -1333,44 +1333,44 @@ export declare interface ExecuteActivityContext {
     activityChildren?: ActivityChildItemBase[];
 
     /**
-     * Activity context to be shared with Copilot
+     * Additional activity-related insights to be shared with Copilot
      */
-    copilot: CopilotContext;
+    activityInsights: ActivityInsights;
 }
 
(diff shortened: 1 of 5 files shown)


utils/index.d.ts
```diff
@@ -1331,6 +1331,50 @@ export declare interface ExecuteActivityContext {
      * Children to show under the activity tree item.
      */
     activityChildren?: ActivityChildItemBase[];
+
+    /**
+     * Activity insights to be shared with Copilot
+     */
+    activityInsights: ActivityInsights;
```
MicroFish91, line 1338: Originally had this as `commandInsights`, but prefixing with activity seemed more in line with the convention here

nturinski (approved):
You'll have to bump some versions. I think I'd recommend at least doing a minor bump maybe even major. Even though it's not breaking, it'll just be easier for us to track which extensions actually have the attributes enabled, etc.

nturinski (approved):

PR #2012, opened 2025-05-22
utils: Changes to AzureWizard for the new Confirmation web view

motm32:
Includes changes for adding confirmationViewProperty to each promptStep which then get pushed onto an array which can be used to populate the view. 

Also has changes related to GoBackError so we can pass in a numberOfStepsToGoBack. 

Once the ACA PR which shows how these changes are used is up I will link it here.

### DIFF
--- a/utils/index.d.ts
+++ b/utils/index.d.ts
@@ -668,6 +668,9 @@ export declare function isUserCancelledError(error: unknown): error is UserCance
 export declare class NoResourceFoundError extends Error {
     constructor(context?: ITreeItemPickerContext);
 }
+export class GoBackError extends Error {
+    constructor(numberOfStepsToGoBack?: number);
+}
 
 export type CommandCallback = (context: IActionContext, ...args: any[]) => any;
 
@@ -1290,6 +1293,13 @@ export declare class AzureWizard<T extends IActionContext & Partial<ExecuteActiv
      * @param options Options describing this wizard
      */
     public constructor(wizardContext: T, options: IWizardOptions<T>);
+    /**
+     * An array of objects with three properties
+     *      name: property which is  displayable name of the step
+     *      value: The label of the chosen value for the step
+     *      valueInContext: The name which can be used to access the value in the wizard context
+     */
+    public confirmationViewProperties: { name: string, value: string, valueInContext: string; }[];
 
     public prompt(): Promise<void>;
     public execute(): Promise<void>;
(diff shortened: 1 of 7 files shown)


utils/index.d.ts
```diff
@@ -1364,19 +1374,14 @@ export declare abstract class AzureWizardExecuteStepWithActivityOutput<T extends
     protected abstract getTreeItemLabel(context: T): string;
     /**
      * Abstract method to get the success string for the output log.
-     * Will be called once upon a successful run.
```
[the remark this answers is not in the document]
motm32: oh yes seems that way!

utils/index.d.ts
```diff
@@ -1496,6 +1506,19 @@ export declare abstract class AzureWizardPromptStep<T extends IActionContext> {
...
+     * @returns An object with three properties
+     *      name: A displayable name of the step
+     *      value: The label of the chosen value for the step
+     *      valueInContext: The name which can be used to access the value in the wizard context
+     */
+    public confirmationViewProperty?(wizardContext: T): {
+        name: string;
+        value: string;
+        valueInContext: string;
+    };
```
alexweininger, line 1520: Since we're using this in multiple places, have you considered making a `ConfirmationViewProperty` interface/type?

utils/src/wizard/AzureWizard.ts
```diff
@@ -135,10 +136,25 @@ export class AzureWizard<T extends (IInternalActionContext & Partial<types.Execu
...
+                                for (let i = 0; i < stepsToGoBack; i++) {
+                                    step = this.goBack(step);
+                                    this.confirmationViewProperties.pop();
+                                }
+                            } else {
+                                if (this.confirmationViewProperties.length > 0) {
+                                    this.confirmationViewProperties.pop();
+                                }
+                                step = this.goBack(step);
+                            }
```
alexweininger, line 157: I think it should be possible to add tests for this new go back error functionality without using a confirmation view. Just test it with a normal wizard.

utils/index.d.ts
```diff
@@ -1290,6 +1293,13 @@ export declare class AzureWizard<T extends IActionContext & Partial<ExecuteActiv
      * @param options Options describing this wizard
      */
     public constructor(wizardContext: T, options: IWizardOptions<T>);
+    /**
+     * An array of objects with three properties
+     *      name: The displayable name of the step
+     *      value: The label of the chosen value for the step
+     *      valueInContext: The name which can be used to access the value in the wizard context
+     */
+    public confirmationViewProperties: { name: string, value: string, valueInContext: string; }[];
```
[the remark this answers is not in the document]
motm32: Yes. I am not using it right now but Nathan said it would be good to have this value for other copilot scenarios we have in mind. So I just left it in.

utils/index.d.ts
```diff
@@ -1526,6 +1515,17 @@ export declare abstract class AzureWizardPromptStep<T extends IActionContext> {
...
+/**
+ * name: A displayable name of the step
+ * value: A displayable value chosen by the user (The label of the chosen value for the step)
+ * valueInContext: The name which can be used to access the value in the wizard context
+ */
+export type ConfirmationViewProperty = {
+    name: string;
+    value: string;
+    valueInContext: string;
+}
```
alexweininger, line 1527: nit: If you put the property descriptions on separate comments above each property in the object then the doc strings will show up in the hovers when people are using the interface.

utils/index.d.ts
```diff
@@ -1496,13 +1503,33 @@ export declare abstract class AzureWizardPromptStep<T extends IActionContext> {
...
+     */
+    name: string;
+    /**
+     * A displayable value chosen by the user (The label of the chosen value for the step)
+     */
+    value: string;
+    /**
+     * The name which can be used to access the value in the wizard context
+     */
+    valueInContext: string;
```
alexweininger, line 1530: I think I'd prefer a name like `propertyName`, or `propertyKey`, or even just `key`. Could be more specific and do `contextPropertyName` or similar.

alexweininger (approved):

PR #2016, opened 2025-06-04
Stop masking user info in error notification

JasonYeMSFT:


bwateratmsft (changes requested):
Actually, upon rereading the original code, I think the telemetry error message now gets sent unredacted?

PR #2017, opened 2025-06-04
dev: Add changes to GoBackError and TestInput for Utils tests

motm32:
Refer to this PR for the main changes we are testing: https://github.com/microsoft/vscode-azuretools/pull/2012

This just enables me to pass in a number to the GoBackError so we can throw the error a certain number of times. Once this is merged and release I'll update the utils PR with the tests.

### DIFF
--- a/dev/src/TestUserInput.ts
+++ b/dev/src/TestUserInput.ts
@@ -10,7 +10,7 @@ import * as types from '../index';
 export enum TestInput {
     UseDefaultValue,
     BackButton,
-    BackButtonWithNumberOfSteps
+    BackThreeSteps
 }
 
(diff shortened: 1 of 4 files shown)


dev/src/TestUserInput.ts
```diff
@@ -9,12 +9,16 @@ import * as types from '../index';
 
 export enum TestInput {
     UseDefaultValue,
-    BackButton
+    BackButton,
+    BackButtonWithNumberOfSteps
```
alexweininger, line 13: maybe indicate in the name here that it goes back 3 steps

alexweininger (approved):

PR #2021, opened 2025-06-23
azure: Add new methods for `LocationListStep`

MicroFish91:
###  Copilot Summary

Adds new utility methods to `LocationListStep` to help swap locations when a new resource requires different offerings than existing deployment resources. These utilities make it easier to pick a compatible location when resource requirements change.

- New methods assist with selecting and validating locations based on available offerings.
- Designed to reduce friction when resources need to be deployed in locations with different requirements.
- No breaking changes; utilities are additive.

### Example of Use

https://github.com/microsoft/vscode-azurefunctions/pull/4566/files

- `validateDTSConnection.ts` lines 55 - 56 (Reset and suggest location on the new context)
-  `DurableTaskSchedulerListStep` lines 50-51 (Setting provider for new location and adding step)

### DIFF
--- a/azure/index.d.ts
+++ b/azure/index.d.ts
@@ -636,7 +636,14 @@ export declare const CommonRoleDefinitions: {
         readonly roleName: "DocumentDB Account Contributor",
         readonly description: "Can manage Azure Cosmos DB accounts.",
         readonly roleType: "BuiltInRole"
-    }
+    },
+    readonly durableTaskDataContributor: {
+        name: "0ad04412-c4d5-4796-b79c-f76d14c8d402",
+        type: "Microsoft.Authorization/roleDefinitions",
+        roleName: "Durable Task Data Contributor",
+        description: "Durable Task role for all data access operations.",
+        roleType: "BuiltInRole"
+    },
 };
 /**
  * Constructs the role id for a given subscription and role name id
--- a/azure/index.d.ts
+++ b/azure/index.d.ts
@@ -117,6 +117,14 @@ export declare class LocationListStep<T extends ILocationWizardContext> extends 
     public static setLocation<T extends ILocationWizardContext>(wizardContext: T, name: string): Promise<void>;
 
     /**
+     * Resets all location and location-related metadata on the wizard context back to its uninitialized state.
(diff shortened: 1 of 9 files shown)


azure/index.d.ts
```diff
@@ -116,6 +116,23 @@ export declare class LocationListStep<T extends ILocationWizardContext> extends
...
+     */
+    public static setAutoSelectLocation<T extends ILocationWizardContext>(wizardContext: T, name: string): Promise<void>;
+
+    /**
+     * Resets all location and location-related metadata on the wizard context back to its uninitialized state.
+     * This includes clearing the selected location, cached location tasks, provider location maps, and any step-tracking flags.
+     * Use this to ensure the wizard context is fully reset before starting a new location selection process.
+     * @param wizardContext The context of the wizard
+     */
+    public static resetLocation<T extends ILocationWizardContext>(wizardContext: T): void;
```
alexweininger, line 134: after being reset, how does the wizard know to re-prompt for location?

MicroFish91: You’d need to explicitly re-add the location prompt by calling `LocationListStep.addStep()` again. The cleanest and most reliable way to do this is by creating a fresh `wizardContext`. Also, maybe I need to enhance the logic to accept an optional promptSteps argument and attempt to remove any existing location steps as part of resetting. 🤔 (lmk what you think)

Some more background on why I added these, I'm running into an issue where we can't just reuse the site location from the function app resource when creating he DTS resource because the location providers aren't the same.
By using a new wizard/wizardContext—even if we spread in properties from the existing deploy context—we can explicitly reset the location for the DTS create flow. This isolates the DTS logic from the broader deployment logic, while still reusing relevant parts of the context.

MicroFish91:
Q: Should I allow passing in of an optional `promptSteps` argument so that we can remove any existing location prompts if they exist as well?

nturinski:
> Q: For reset, should I allow passing in of an optional `promptSteps` argument so that we can remove any location prompt steps if they exist as well?

Discussed offline but basically, I'd prefer if there wasn't another argument that can manipulate the steps in a wizard from unknown/unexpected codepaths.

nturinski (approved):

PR #2029, opened 2025-07-03
azure: Use `setAutoSelectLocation` in `ResourceGroupListStep` instead of `setLocation`

MicroFish91:
Because we should always respect location for new providers that could be added to the flow


nturinski (approved):
🤷‍♂️

PR #2031, opened 2025-07-07
Azure - Add property to disallow blob public access in storage account creation

ccastrotrejo:
## Summary
Sets `allowBlobPublicAccess` to `false` by default when creating new storage accounts through the Azure extension wizard.

## Changes
- Added `allowBlobPublicAccess: false` property to the storage account creation parameters in `StorageAccountCreateStep.ts`

## Why this change?
This improves security by ensuring that newly created storage accounts have blob public access disabled by default, following Azure security best practices.

### DIFF
--- a/azure/package-lock.json
+++ b/azure/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "@microsoft/vscode-azext-azureutils",
-    "version": "0.3.9",
+    "version": "0.3.9-beta",
     "lockfileVersion": 2,
     "requires": true,
     "packages": {
         "": {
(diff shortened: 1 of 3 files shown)


nturinski:
Due to policy, this setting must be explicitly set to false in the payload or the API will throw an error.

Logic Apps team needed to update that property, but are using an old version of the `azureutils` package. It would break them to update to the version that includes this property [[reference]](https://github.com/microsoft/vscode-azuretools/pull/1866) so I suggested we just release a new beta version based off of what version they are currently using.

nturinski (approved):

PR #2036, opened 2025-07-10
azure: Allow role callback to return a promise

MicroFish91:

### DIFF
--- a/azure/index.d.ts
+++ b/azure/index.d.ts
@@ -439,7 +439,7 @@ export declare class RoleAssignmentExecuteStep<T extends IResourceGroupWizardCon
     * If the scope ID is undefined, the step will throw an error.
     * @param roles An array of roles. Each role is an object and include the ARM role definition id and name of the role definition.
     * */
-    public constructor(roles: () => Role[] | undefined, options?: { priority?: number });
+    public constructor(roles: () => Role[] | Promise<Role[]> | undefined, options?: { priority?: number });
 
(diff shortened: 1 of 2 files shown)


alexweininger (dismissed):

MicroFish91:
> do we need it to be a Promise or a function that returns a promise? `() => Promise`?

Thanks, that's a tongue twister, it should be a function that returns a function that returns a promise lol 😂

MicroFish91:
> Also what scenario do we need this for?

`RoleAssignmentExecuteStep` already takes a callback that it calls to get a list of roles.  I would like to allow these callbacks to be async functions so I can make an API call if needed.  In my case I need to check to see if the managed identity has the roles we need before making the final determination on which roles to add.

nturinski:
> Currently I don't think the role callback returns or needs to return another func

This is what Alex is referring to:

`public constructor(roles: () => Role[] | Promise<Role[]>`

In this code, you're saying the roles is either a function that returns a role array or a promise that returns roles. It should probably be this:

`public constructor(roles: () => (Role[] | Promise<Role[])>`

So it's a function that returns either a role or promise that resolves with roles.

MicroFish91:
> In this code, you're saying the roles is either a function that returns a role array or a promise that returns roles. It should probably be this:
> 
> `public constructor(roles: () => (Role[] | Promise<Role[])>`

Oh you're absolutely right, that was my intent and I did not notice that.  Oddly though, when I consumed it in Azure Functions to test it, it didn't complain at all..

image

nturinski (approved):

PR #2037, opened 2025-07-15
utils: Add `commandId` to `IActionContext` and throughout activity workflow

MicroFish91:
I'd like to automatically log the id when possible so that it can percolate up to the activity log level.  We can use this for improving telemetry.

### DIFF
--- a/utils/hostapi.d.ts
+++ b/utils/hostapi.d.ts
@@ -276,6 +276,11 @@ export interface Activity {
     endTime?: Date;
 
     /**
+     * Any associated command (or callback) ID
+     */
+    commandId?: string;
+
+    /**
      * Activity / Command attributes to be shared with Copilot
      */
     attributes?: ActivityAttributes;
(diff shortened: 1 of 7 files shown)


MicroFish91:
> I don't know if we actually have to copy over the property. The context is shared and you should have access to the `callbackId` regardless (as long as it was called with `callWithTelemetryAndErrorHandling`)

Is this explicitly set in a way that we can always rely on it being there without making additional changes? I don't see it typed, but I've seen it on the context occasionally when I inspect it during debug scenarios

MicroFish91:
> I don't know if we actually have to copy over the property. The context is shared and you should have access to the `callbackId` regardless (as long as it was called with `callWithTelemetryAndErrorHandling`)

Would you prefer removing `IHandlerContext` and just adding the callback id to the `IActionContext` typings?

nturinski:
Yeah, and just call it `callbackId` in the typings too, I think. Just leave it consistent with how it's been so it doesn't break anything that may have been using it.

MicroFish91:
Actually didn't need to remove it, I was able to just let it be optional in `IActionContext`, while `IHandlerContext` explicitly makes it required.

nturinski (approved):

PR #2040, opened 2025-07-24
utils: Fix extra nodes appearing when going back

motm32:
See https://github.com/microsoft/vscode-azurecontainerapps/issues/913 for more info. 

This adds a way to specify the number of extra children added in a step so when a go back error is thrown we can ensure all children are being removed from the activityChildren array.


alexweininger (approved):
In the future it'd be best to have the activity children know what steps they came from so they can easily be cleared when needed. This current implementation relies on the step knowing or tracking how many activity children it adds which might not always be easy.

alexweininger (approved):

PR #2041, opened 2025-07-25
utils: Use a unique id to remove added activity log children

motm32:
Refer to this: https://github.com/microsoft/vscode-azurecontainerapps/pull/914

### DIFF
--- a/utils/index.d.ts
+++ b/utils/index.d.ts
@@ -1257,6 +1257,7 @@ export interface ActivityChildItemBase extends TreeElementBase {
     activityType: ActivityChildType;
     contextValue?: string;
     description?: string;
+    stepId?: string;
     getChildren?(): ProviderResult<ActivityChildItemBase[]>;
 }
 
@@ -1269,6 +1270,7 @@ export type ActivityChildItemOptions = {
     description?: string;
     iconPath?: TreeItemIconPath;
     tooltip?: string | MarkdownString | undefined;
+    stepId?: string;
     initialCollapsibleState?: TreeItemCollapsibleState;
     /**
      * If set to true, will initialize `getChildren` with an empty array.
@@ -1544,11 +1546,11 @@ export declare abstract class AzureWizardPromptStep<T extends IActionContext> {
     public id?: string;
 
     /**'
-     * Optional number used to determine how many extra children were added in a step.
-     * This value is checked in the go back function to pop off any extra children.
+     * Optional unique id used to identify the step when searching for added activity children.
+     * This is used to ensure the added children are only coming from the specific step when removed using goBack()
      */
 
-    public addedNumberOfActivityChildren?: number | undefined;
+    public uniqueId?: string;
 
     /**
      * Prompt the user for input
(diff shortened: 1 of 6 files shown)


utils/index.d.ts
```diff
@@ -1544,11 +1546,11 @@ export declare abstract class AzureWizardPromptStep<T extends IActionContext> {
...
 
     /**'
-     * Optional number used to determine how many extra children were added in a step.
-     * This value is checked in the go back function to pop off any extra children.
+     * Optional unique id used to identify the step when searching for added activity children.
+     * This is used to ensure the added children are only coming from the specific step when removed using goBack()
      */
 
-    public addedNumberOfActivityChildren?: number | undefined;
+    public uniqueId?: string;
```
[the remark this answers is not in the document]
alexweininger: > Also, can we automatically assign the prompt step with a uuid on instantiation if one isn't already provided

I was thinking the same thing, also how can we make sure the id actually makes it onto the activity children. Is there any way we can enforce that or no?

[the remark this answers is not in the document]
motm32: @MicroFish91 I originally was going to use the id prop but then in the meeting you mentioned that might not be unique which is why I made the uuid like alex suggested. 

@alexweininger My plan was to just add the id to whereever a new tree item was being injected into the array like this: 
image
This is only a problem when we are adding extra nodes, so I figured it would be ok to just add it manually since it isn't used everywhere. Not sure if we would rather add it to each node up to you guys.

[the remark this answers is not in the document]
motm32: Ahh ok yeah I was never intending to use the stepName. Seems like from the description the id should be unique, so I'll just use that.

utils/src/wizard/AzureWizardPromptStep.ts
```diff
@@ -15,7 +16,10 @@ export abstract class AzureWizardPromptStep<T extends types.IActionContext> impl
     public propertiesBeforePrompt!: string[];
     public prompted!: boolean;
     public id?: string;
-    public addedNumberOfActivityChildren?: number | undefined;
+
+    constructor(options?: { id?: string }) {
```
nturinski, line 20: I guess my comment about using a uuid all the time more applies to here, not on stepId.

motm32: Yeah I don't mind making the id required and just assigning a uuid to it. Not sure where id was being used before and if assigning a uuid to it could break anything though.

MicroFish91: > I guess my comment about using a uuid all the time more applies to here, not on stepId.

Yeah agreed, I think it's probably okay / good to make the prompt step id required, but I wouldn't want to make the activity child require the step id because not all of them actually need it.

utils/index.d.ts
```diff
@@ -1538,17 +1540,10 @@ export declare abstract class AzureWizardPromptStep<T extends IActionContext> {
     public supportsDuplicateSteps: boolean;
 
     /**
-     * Optional id used to determine if this step is unique, for things like caching values and telemetry
+     * A id used to determine if this step is unique, for things like caching values and telemetry
      * If not specified, the class name will be used instead
```
MicroFish91, line 1544: Actually might want to update this comment by deleting this line - it says if the id is not specified the class name will be used instead which isn't true anymore

MicroFish91 (approved):

PR #2054, opened 2025-08-13
azure: Ensure `ResourceGroupVerifyStep` only gets run once per wizard

MicroFish91:
Related: https://github.com/microsoft/vscode-azurefunctions/issues/4642


alexweininger (commented):
Does this value need to be reset if the user changes the resource group name?

MicroFish91:
> Does this value need to be reset if the user changes the resource group name?

I don't think so.  This gets handled in the execute phase so we wouldn't have to worry about any go back / undo logic.

PR #2055, opened 2025-08-19
utils: Support authentication challenges

alexweininger:
Will consume https://github.com/microsoft/vscode-azureresourcegroups/pull/1205 once that is released.

Only code changes are in utils/src/utils/credentialUtils.ts making it accept the new challenge object

These changes will be consumed by the azureutils package

Working towards: https://github.com/microsoft/vscode-azureresourcegroups/issues/1200

### DIFF
--- a/utils/package-lock.json
+++ b/utils/package-lock.json
@@ -9,7 +9,7 @@
             "version": "3.4.0",
             "license": "MIT",
             "dependencies": {
-                "@microsoft/vscode-azureresources-api": "file:../../vscode-azureresourcegroups/api/microsoft-vscode-azureresources-api-2.5.1.tgz",
+                "@microsoft/vscode-azureresources-api": "^2.6.0",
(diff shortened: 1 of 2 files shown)


utils/index.d.ts
```diff
@@ -169,15 +169,15 @@ export interface AzExtServiceClientCredentialsT2 {
      * @param options - The options used to configure any requests this
      *                TokenCredential implementation might make.
      */
-    getToken(scopes?: string | string[], options?: any): Promise<any | null>;
```
bwateratmsft, line 172: Shouldn't we keep `options?` even though we aren't using it? The `TokenCredential` interface has it in its `getToken()` method. https://learn.microsoft.com/en-us/javascript/api/@azure/identity/tokencredential?view=azure-node-latest

alexweininger: Sure

bwateratmsft (approved):

PR #2057, opened 2025-08-19
auth: support challenge when getting sessions from VS Code

alexweininger:
These changes will be consumed along with the other shared package updates by the Azure Resources extension.

Working towards: https://github.com/microsoft/vscode-azureresourcegroups/issues/1200


auth/src/utils/isAuthenticationSessionRequest.ts
```diff
@@ -0,0 +1,5 @@
+import * as vscode from 'vscode';
```
bwateratmsft, line 1: Copyright header needed

PR #2063, opened 2025-08-21
azure-pipelines: Set `signWithProd` to `true`

bwateratmsft:
/cc @JasonYeMSFT @alexweininger

PR #2064, opened 2025-08-21
azure-pipelines: Add signWithProd for compliance reasons

nturinski:
I suspect that is all we need to do to be compliant?


bwateratmsft:
Bruh #2063

PR #2072, opened 2025-08-27
utils: Add support for copilot user input 

motm32:
This adds a new `CopilotUserInput` which is essentially just a Wrapper class of several `vscode.window` methods that handle user input. 

I will add a PR soon in ACA that shows how this can be used but in essence if context.ui is set to a CopilotUserInput then anytime showQuickPick, showInputBox, etc. is called it will prompt copilot for an answer instead of showing the user. 

If copilot is unable to generate a response or the response it gives does not match one of the quick picks as `InvalidInputError` is thrown and this will switch the context.ui back to an AzExtUserInput and the quick picks will work as normal instead of polling Copilot. 

canPickMany case has not been fully implemented yet since ACA doesn't have any canPickMany quick picks so I would need to test this using rg. For now I am just throwing it back to the user to make the choice.


MicroFish91 (commented):
This is sick, love what you've got cooking up

utils/src/copilot/copilot.ts
```diff
@@ -0,0 +1,120 @@
...
+            // Consume the stream and collect fragments
+            for await (const fragment of request.text) {
+                fragments.push(fragment);
+            }
+
+            // Combine fragments into a single string
+            const responseText = fragments.join("");
+            const cleanedResponse = extractJsonString(responseText);
+            // Attempt to parse the response as JSON
+            console.log(cleanedResponse);
```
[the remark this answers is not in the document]
motm32: Was still going over some stuff since this was still in draft but this and the above function have been removed since I don't need them anymore. Some of the code in this file were coming from the github copilot for azure teams code that I modified.

utils/src/copilot/copilot.ts
```diff
@@ -0,0 +1,95 @@
...
+
+async function selectMostPreferredLm(): Promise<vscode.LanguageModelChat | undefined> {
+    const lms = await vscode.lm.selectChatModels({ vendor: "copilot" });
+    const lmsInPreferredOrder = (lms || [])
+        .filter((lm) => languageModelPreference.some((pref) => pref.family === lm.family && pref.vendor === lm.vendor))
+        .sort((a, b) => languageModelPreference.findIndex((pref) => pref.family === a.family && pref.vendor === a.vendor) - languageModelPreference.findIndex((pref) => pref.family === b.family && pref.vendor === b.vendor));
+    return lmsInPreferredOrder?.at(0);
+}
+
+export function createPrimaryPromptToGetSingleQuickPickInput(picks: string[], relevantContext?: string): string {
```
nturinski, line 25: I'm not too concerned about how optimal this is yet. I think when we do this for other commands, we should probably have a way to pass custom prompts since it may not be a generic one shoe fits all scenario.

utils/index.d.ts
```diff
@@ -1001,6 +1001,38 @@ export interface AzExtUserInputOptions {
     stepName?: string;
 }
 
+/**
+ * Wrapper class of several `vscode.window` methods that handle user input.
+ */
+export declare class AzExtUserInput implements IAzureUserInput {
```
nturinski, line 1007: Why do we need to export this now?

utils/src/userInput/CopilotUserInput.ts
```diff
@@ -0,0 +1,118 @@
...
+            throw new InvalidInputError();
+        }
+
+        this._onDidFinishPromptEmitter.fire({ value: pick });
+        return pick;
+    }
+
+    public showOpenDialog(): Promise<vscodeTypes.Uri[]> {
+        // Throw this back to the user
+        throw new InvalidInputError();
```
nturinski, line 42: How jarring is it when the error gets thrown back and the user has to take control? I'm not sure if there's a great way to do it, but just curious what it currently looks like.

utils/src/userInput/CopilotUserInput.ts
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    public async showInputBox(options: vscodeTypes.InputBoxOptions): Promise<string> {
+        if (options.prompt) {
+            const primaryPrompt: string = createPrimaryPromptForInputBox(options.prompt, this._relevantContext);
+            const response = await doCopilotInteraction(primaryPrompt)
+            const jsonResponse: string = JSON.parse(response) as string;
+            this._onDidFinishPromptEmitter.fire({ value: jsonResponse });
+            return jsonResponse;
+        } else if (options.value) {
```
nturinski, line 67: Should this be an else if?

I thought it'd be more like:
- Prompt Copilot
- If response is invalid, then check if (options.value)
- If it exists, return that
- If response is valid, just return that

utils/src/userInput/CopilotUserInput.ts
nturinski: Does the webview that says Copiloting is generating the picks close?

motm32: I don't think so at the moment. Since the webview lives only in the aca extension there isn't that good of a way to dispose of it here. I may can add a listener on the ACA side to see if I can dispose of it anytime that error is thrown. The view does get disgarded once the picks are all finished but yeah I totally forgot about that.

utils/src/userInput/CopilotUserInput.ts
```diff
@@ -0,0 +1,123 @@
...
+        if (options.prompt) {
+            try {
+                const primaryPrompt: string = createPrimaryPromptForInputBox(options.prompt, this._relevantContext);
+                const response = await doCopilotInteraction(primaryPrompt)
+                const jsonResponse: string = JSON.parse(response) as string;
+                this._onDidFinishPromptEmitter.fire({ value: jsonResponse });
+                return jsonResponse;
+            } catch {
+                // if copilot is unable to provide a response, fall back to the default value if it exists
+                if (options.value) {
```
nturinski, line 70: Shouldn't this be outside of `if(options.prompt)`?  If there's no prompt, shouldn't we default to `options.value`?

nturinski (approved):

nturinski (approved):

PR #2082, opened 2025-09-22
utils: Update vscode typings and change TreeIcon types

nturinski:
Upgrading the typings in preparation of MFA

### DIFF
--- a/utils/index.d.ts
+++ b/utils/index.d.ts
@@ -188,7 +188,7 @@ export interface ISubscriptionContext {
     isCustomCloud: boolean;
 }
 
-export type TreeItemIconPath = string | Uri | { light: string | Uri; dark: string | Uri } | ThemeIcon;
+export type TreeItemIconPath = vscodeTypes.IconPath | string;
 
 
 /**
--- a/utils/index.d.ts
+++ b/utils/index.d.ts
@@ -188,7 +188,7 @@ export interface ISubscriptionContext {
     isCustomCloud: boolean;
 }
 
-export type TreeItemIconPath = vscodeTypes.IconPath | string;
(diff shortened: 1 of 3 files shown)


utils/index.d.ts
```diff
@@ -188,7 +188,7 @@ export interface ISubscriptionContext {
     isCustomCloud: boolean;
 }
 
-export type TreeItemIconPath = string | Uri | { light: string | Uri; dark: string | Uri } | ThemeIcon;
+export type TreeItemIconPath = vscodeTypes.IconPath | string;
```
bwateratmsft, line 191: Do we know that `string` is still allowed?

bwateratmsft: When I hit this issue last month on the registries package I excluded string. We might want to do the same...

nturinski: I'm fine with getting rid of it, but I did want to confirm that I did check that it was still allowed:

```
/**
* The icon path or {@link ThemeIcon} for the tree item.
* When `falsy`, {@link ThemeIcon.Folder Folder Theme Icon} is assigned, if item is collapsible otherwise {@link ThemeIcon.File File Theme Icon}.
* When a file or folder {@link ThemeIcon} is specified, icon is derived from the current file icon theme for the specified theme icon using {@link TreeItem.resourceUri resourceUri} (if provided).

iconPath?: string | IconPath;
```

bwateratmsft: Ah, gotcha. I'm fine either way then--your call.

nturinski: I'd rather stay in sync with what you did, plus I can't really think of a good reason that they kept it except backwards compatibility.

MicroFish91 (approved):

PR #2104, opened 2025-10-23
auth: totally rewrite subscription providers

bwateratmsft:
A lot of the assumptions we made when originally creating this package are no longer true.

- We have the `getAccounts()` API now
- We need to support multi-account scenarios
- Subscriptions can "belong" to multiple tenants
- Calling `getSession()` can often result in `onDidChangeSessions` being fired unexpectedly, which made things like `isSignedIn()` problematic (leading to event loops)

This PR aims to address all of it. It also adds result caching by default, optional cancellation by token, and parallelizes requests when calling `getAvailableSubscriptions()`.

It is very much a breaking change, so major versions will need to be updated.

### DIFF
--- a/auth/src/providers/VSCodeAzureSubscriptionProvider.ts
+++ b/auth/src/providers/VSCodeAzureSubscriptionProvider.ts
@@ -258,14 +258,14 @@ export class VSCodeAzureSubscriptionProvider extends AzureSubscriptionProviderBa
      * Clears all caches related to the given signed-out account.
      * @param accountId The ID of the signed-out account
      */
-    private onDidSignOutAccount(accountId: string): void {
-        const lowerAccountId = accountId.toLowerCase();
-        this.accountCache.delete(lowerAccountId);
-        this.tenantCache.delete(lowerAccountId);
-        this.subscriptionCache.forEach((_, key) => {
-            if (key.startsWith(`${lowerAccountId}/`)) {
-                this.subscriptionCache.delete(key);
-            }
-        });
-    }
+    // private onDidSignOutAccount(accountId: string): void {
+    //     const lowerAccountId = accountId.toLowerCase();
+    //     this.accountCache.delete(lowerAccountId);
+    //     this.tenantCache.delete(lowerAccountId);
+    //     this.subscriptionCache.forEach((_, key) => {
+    //         if (key.startsWith(`${lowerAccountId}/`)) {
+    //             this.subscriptionCache.delete(key);
+    //         }
+    //     });
+    // }
 }


auth/src/contracts/AzureAccount.ts
```diff
@@ -0,0 +1,8 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import type * as vscode from 'vscode';
+
+export type AzureAccount = Readonly<vscode.AuthenticationSessionAccountInformation>;
```
bwateratmsft: This is just a shortcut type because `vscode.AuthenticationSessionAccountInformation` is long and we use it all over

auth/src/providers/AzureSubscriptionProviderBase.ts
```diff
@@ -0,0 +1,338 @@
...
+
+        disposables?.push(disposable);
+        return disposable;
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public async signIn(tenant?: TenantIdAndAccount, options: SignInOptions = DefaultSignInOptions): Promise<boolean> {
+        this.silenceRefreshEvents();
```
bwateratmsft:
```suggestion
        if (!options.silent) {
            // Suppress without timeout, until sign in is done--it can take a while when done interactively
            this.suppressRefreshSuggestedEvents = true;
        } else {
            this.silenceRefreshEvents();
        }
```

auth/src/providers/AzureSubscriptionProviderBase.ts
```diff
@@ -0,0 +1,338 @@
...
+        } catch (err) {
+            // Cancellation is not actually supported by vscode.authentication.getAccounts, but just in case it is added in the future...
+            this.remapLogRethrow(err, options.token);
+        }
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public async getUnauthenticatedTenants(account: AzureAccount, options?: GetOptions): Promise<AzureTenant[]> {
```
bwateratmsft:
```suggestion
    public async getUnauthenticatedTenants(account: AzureAccount, options?: Omit<GetOptions, 'all'>): Promise<AzureTenant[]> {
```

auth/src/providers/AzureSubscriptionProviderBase.ts
```diff
@@ -0,0 +1,338 @@
...
+
+        return unauthenticatedTenants;
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public async getTenantsForAccount(account: AzureAccount, options: GetOptions): Promise<AzureTenant[]> {
+        try {
+            this.logger?.debug(`auth: Fetching tenants for account '${account.id}'`);
```
bwateratmsft: Maybe use label instead? Or a somewhat filtered one like `b***r@m***.com`?

auth/src/providers/AzureSubscriptionProviderBase.ts
```diff
@@ -0,0 +1,338 @@
...
+            this.remapLogRethrow(err, options.token);
+        }
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public async getSubscriptionsForTenant(tenant: TenantIdAndAccount, options: GetSubscriptionsOptions): Promise<AzureSubscription[]> {
+        try {
+            this.logger?.debug(`auth: Fetching subscriptions for account '${tenant.account.id}' and tenant '${tenant.tenantId}'`);
```
bwateratmsft: Same--labels instead of IDs?

auth/src/providers/AzureSubscriptionProviderBase.ts
```diff
@@ -0,0 +1,338 @@
...
+    }
+
+    const controller = new AbortController();
+    const disposable = token.onCancellationRequested(() => {
+        disposable.dispose();
+        controller.abort();
+    });
+
+    return controller.signal;
+}
```
bwateratmsft: Move to utils?

auth/src/providers/VSCodeAzureSubscriptionProvider.ts
```diff
@@ -0,0 +1,191 @@
...
+        if (options.noCache) {
+            this.accountCache.clear();
+        }
+
+        // If needed, refill the cache
+        if (this.accountCache.size === 0) {
+            const accounts = await super.getAccounts(options);
+            if (accounts.length === 0) {
+                this.logger?.debug('auth: No accounts found');
+                throw new NotSignedInError();
```
bwateratmsft: Special logic to throw `NotSignedInError` if no accounts are listed at all. Should we move it up to the base class?

auth/src/providers/VSCodeAzureSubscriptionProvider.ts
```diff
@@ -0,0 +1,191 @@
...
+        if (options.noCache) {
+            this.subscriptionCache.delete(cacheKey);
+        }
+
+        // If needed, refill the cache
+        if (!this.subscriptionCache.has(cacheKey)) {
+            const subscriptions = await super.getSubscriptionsForTenant(tenant, options);
+            this.subscriptionCache.set(cacheKey, subscriptions);
+            this.logger?.debug(`auth: Cached ${subscriptions.length} subscriptions for account '${tenant.account.id}' and tenant '${tenant.tenantId}'`);
+        }
```
bwateratmsft:
```suggestion
        } else {
            this.logger?.debug(`auth: Using cached subscriptions for account '${account.id}' and tenant '${tenant.tenantId}'`);
        }
```

auth/src/providers/VSCodeAzureSubscriptionProvider.ts
```diff
@@ -0,0 +1,191 @@
...
+     * Gets the account filters that are configured in `azureResourceGroups.selectedSubscriptions`. To
+     * override the settings with a custom filter, implement a child class with filter methods overridden.
+     *
+     * If no values are returned by `getAccountFilters()`, then all accounts will be scanned for subscriptions.
+     *
+     * @returns A list of account IDs that are configured in `azureResourceGroups.selectedSubscriptions`.
+     */
+    protected getAccountFilters(): Promise<string[]> {
+        // TODO: implement account filtering based on configuration if needed
+        // TODO: cannot have PII in the settings so it has to be an ID--test that
```
bwateratmsft: I did, ID is ok. Still not sure how to impl this without using a new setting entirely

bwateratmsft:
@danegsta suggested we switch to support proper pagination a la `PagedAsyncIterableIterator`. Thoughts?

auth/src/utils/Limiter.ts
```diff
@@ -0,0 +1,52 @@
...
+interface ILimitedTaskFactory<T> {
+    factory: () => Promise<T>;
+    c: (value: T | Promise<T>) => void;
+    e: (error?: any) => void; // eslint-disable-line @typescript-eslint/no-explicit-any
+}
+
+export class Limiter<T> {
+
+    private runningPromises: number;
+    private maxDegreeOfParalellism: number;
```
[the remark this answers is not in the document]
bwateratmsft: This is known but I actually do not want to change it, so as to exactly match the source.

auth/src/providers/AzureDevOpsSubscriptionProvider.ts
```diff
@@ -0,0 +1,181 @@
...
+    /**
+    * The resource ID of the Azure DevOps federated service connection,
+    *   which can be found on the `resourceId` field of the URL at the address bar
+    *   when viewing the service connection in the Azure DevOps portal
+    */
+    serviceConnectionId: string,
+    /**
+     * The `Tenant ID` field of the service connection properties
+     */
+    tenantId: string,
```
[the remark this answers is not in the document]
bwateratmsft: ...it *was* updated...

PR #2107, opened 2025-10-23
eslint, dev: bye bye

bwateratmsft:
Step 1 in the removal of the shared eslint and dev packages


dev/index.d.ts
bwateratmsft: What's left in here is what I'll copy over to the utils index.d.ts file

PR #2119, opened 2025-11-13
auth: Completely rewrite auth package

bwateratmsft:
Starting a new PR because the previous one (#2104) was getting unwieldy with the number of comments and commits.

---

A lot of the assumptions we made when originally creating this package are no longer true.

- We have the `getAccounts()` API now
- We need to support multi-account scenarios
- Subscriptions can "belong" to multiple tenants
- Calling `getSession()` can often result in `onDidChangeSessions` being fired unexpectedly, which made things like `isSignedIn()` problematic (leading to event loops)

This PR aims to address all of it. It also adds result caching by default, optional cancellation by token, and parallelizes requests when calling `getAvailableSubscriptions()`.

It is very much a breaking change, so major versions will need to be updated.

### DIFF
--- a/auth/src/providers/VSCodeAzureSubscriptionProvider.ts
+++ b/auth/src/providers/VSCodeAzureSubscriptionProvider.ts
@@ -17,9 +17,6 @@ import { AzureSubscriptionProviderBase } from './AzureSubscriptionProviderBase';
 const ConfigPrefix = 'azureResourceGroups';
 const SelectedSubscriptionsConfigKey = 'selectedSubscriptions';
 
-// TODO: cache clearing on account sign out
-// Maybe requires https://github.com/microsoft/vscode/issues/276997
-
 /**
  * Extension of {@link AzureSubscriptionProviderBase} that adds caching of accounts, tenants, and subscriptions,
  * as well as filtering and deduplication according to configured settings. Additionally, promise


auth/src/providers/AzureSubscriptionProviderBase.ts
```diff
@@ -0,0 +1,417 @@
...
+            const accounts = await this.getAccounts(options);
+
+            for (const account of accounts) {
+                this.throwIfCancelled(options.token);
+                tenantListPromises.push(tenantListLimiter.queue(async () => {
+                    try {
+                        if (tenantsProcessed >= maximumTenants) {
+                            this.logForAccount(account, `Skipping account because maximum tenants of ${maximumTenants} has been reached`);
+                            return;
+                        }
```
bwateratmsft: As implemented, there is no guarantee that these 10 tenants are included in the tenant filters. As a result, if (for example) the first 10 tenants are not in the filter, then no subscriptions will ever be returned. Need to fix this.

bwateratmsft: Maybe? Actually I am having a hard time determining by reading! Tells me the code needs refactoring to be clearer.

auth/src/utils/Limiter.ts
```diff
@@ -0,0 +1,57 @@
...
+    e: (error?: any) => void;
+}
+
+export class Limiter<T> {
+
+    private runningPromises: number;
+    private maxDegreeOfParalellism: number;
+    private outstandingPromises: ILimitedTaskFactory<T>[];
+
+    constructor(maxDegreeOfParalellism: number) {
```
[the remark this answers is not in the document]
bwateratmsft: This one is actually left on purpose because I don't want to change from original source.

auth/src/contracts/AzureSubscriptionProviderRequestOptions.ts
```diff
@@ -0,0 +1,132 @@
...
+    promptIfNeeded?: boolean;
+};
+
+/**
+ * Default options for signing in to a tenant
+ */
+export const DefaultSignInOptions = {
+    clearSessionPreference: false,
+    promptIfNeeded: true,
+} as const satisfies SignInOptions;
```
alexweininger: fancy 😮‍💨

PR #2132, opened 2025-11-21
appservice: Fix workflow standard SKU names

### LINKED ISSUE #2130: AppServicePlanListStep failing to create new ASP: "The parameter SKU.Name has an invalid value"
We're seeing a regression in the past month in the Azure Logic Apps (Standard) extension where we consistently get an error when calling `AppServicePlanListStep` and "Create new App Service plan" due to **"The parameter SKU.Name has an invalid value"**. This is happening when selecting any sku WS1/2/3.

I've also tried manually setting the `context.newPlanSku` to bypass the `AppServicePlanSkuStep` but still seeing the same issue. Example SKU:
```typescript
context.newPlanSku = {
        name: 'Workflow Standard WS1',
        tier: 'Workflow Standard',
        size: 'WS1',
        family: 'WS',
        capacity: 1
}
```

andrew-eldridge:
Fixes #2130

### DIFF
--- a/appservice/src/createAppService/AppServicePlanSkuStep.ts
+++ b/appservice/src/createAppService/AppServicePlanSkuStep.ts
@@ -3,14 +3,14 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import type { GeoRegion, SkuDescription } from '@azure/arm-appservice';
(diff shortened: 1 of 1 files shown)


nturinski (approved):
Looks good to me. We should probably make all of the quickpicks by leveraging the `label` property instead of the `name`.

PR #2136, opened 2025-12-02
utils: Changes to enable re running commands with copilot

motm32:
This PR includes a couple of changes: 

1. Edits to copilot prompts to extract subscription and activity log child items
2. Edits to copilot user input to match up the picks based on label and subscription instead of the entire item (this makes the command run much faster)
3. Wrapper to executeCommand so we can pass in additional context 
4. Add a run generic prompt step which preforms similarly to runQuickPickWizard but doesn't require the step to be a quick pick step.

Some of these changes are used in RG and others in the ACA extension.

### DIFF
--- a/utils/tsconfig.json
+++ b/utils/tsconfig.json
@@ -3,7 +3,6 @@
     "compilerOptions": {
         "outDir": "dist/esm",
         "rootDir": ".",
-        "declaration": false,
-        "skipLibCheck": true
+        "declaration": false
     },
 }
--- a/utils/index.d.ts
+++ b/utils/index.d.ts
@@ -8,6 +8,7 @@
 import type { Environment } from '@azure/ms-rest-azure-env';
 import type { AzExtResourceType, AzureResource, AzureSubscription, ResourceModelBase } from '@microsoft/vscode-azureresources-api';
 import type * as duration from 'dayjs/plugin/duration';
+import type * as vscode from 'vscode';
 import type * as vscodeTypes from 'vscode';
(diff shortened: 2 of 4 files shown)


utils/tsconfig.json
```diff
@@ -3,6 +3,7 @@
     "compilerOptions": {
         "outDir": "dist/esm",
         "rootDir": ".",
-        "declaration": false
+        "declaration": false,
+        "skipLibCheck": true
```
motm32: For some reason the `@vscode-jsonrpc/node` which the github copilot sdk has a dependency on causes issues. From what I have looked into the package does not have a proper package map so when we try to build the cjs it causes issues. I think this is mainly because the github copilot sdk is currently only using esm. If that is gonna cause problems for us I can revert this back to use the vscode llm. 

Not sure if that is also the reason the tests are now failing.

bwateratmsft: We have seen a number of issues where the CJS and ESM types conflict but so far we've been able to just put a targeted `@ts-ignore` in the code. Is that possible here?

utils/src/copilot/copilot.ts
```diff
@@ -63,35 +54,61 @@ export function createPrimaryPromptForWorkspaceFolderPick(folders: readonly vsco
...
-
-            // Combine fragments into a single string
-            const responseText = fragments.join("");
-            const cleanedResponse = extractJsonString(responseText);
-            return cleanedResponse;
-        } catch {
-            throw new InvalidCopilotResponseError();
-        }
+export async function doGithubCopilotInteraction(primaryPrompt: string, relevantContext?: string): Promise<string> {
+    const session = await getCopilotSession(relevantContext);
```
alexweininger: `relevantContext` is ignored if `session` already exists due to the logic on line 68. Might want to redesign this so that `session` isn't a global variable.

PR #2141, opened 2025-12-04
auth: Upgrade to shared eng pkg

bwateratmsft:


auth/src/utils/configuredAzureEnv.ts
```diff
@@ -17,42 +17,36 @@ enum CloudEnvironmentSettingValue {
     Custom = 'custom',
 }
 
+class ExtendedEnvironment extends azureEnv.Environment {
+    public constructor(parameters: azureEnv.EnvironmentParameters, public readonly isCustomCloud: boolean) {
+        super(parameters);
+    }
+}
```
bwateratmsft: The issue here is that `azureEnv.Environment` is a class, not an interface--so when we were doing `...azureEnv.Environment.SomeCloud`, it was taking the prototype information away. Not ideal.

Making an extended class allows us to include the `isCustomCloud` property, and somewhat simplify the code below at the same time.

PR #2152, opened 2025-12-09
appservice: Move to new eng package

bwateratmsft:


appservice/src/SiteClient.ts
```diff
@@ -456,13 +466,13 @@ export class SiteClient implements IAppSettingsClient {
...
 
-    public async vfsPutItem(context: IActionContext, data: string | ArrayBuffer, url: string, rawHeaders?: {}): Promise<AzExtPipelineResponse> {
+    public async vfsPutItem(context: IActionContext, data: string | ArrayBuffer, url: string, rawHeaders?: Record<string, string>): Promise<AzExtPipelineResponse> {
         const client: ServiceClient = await createGenericClient(context, this._site.subscription);
         const headers = createHttpHeaders(rawHeaders);
         return await client.sendRequest(createPipelineRequest({
             method: 'PUT',
             url,
-            body: typeof data === 'string' ? data : data.toString(),
+            body: typeof data === 'string' ? data : data.toString(), // TODO: This isn't right
```
bwateratmsft: TODO: Needs to be fixed

appservice/src/utils/kuduUtils.ts
```diff
@@ -50,7 +50,7 @@ export function convertQueryParamsValuesToString(queryParams: { [key: string]: u
         if (typeof queryParams[key] === 'object') {
             // skip objects, the ServiceClient will handle them
         } else {
-            queryParams[key] = String(queryParams[key]);
+            queryParams[key] = String(queryParams[key]); // TODO: this isn't right
```
bwateratmsft: TODO: Needs to be fixed

appservice/src/SiteClient.ts
bwateratmsft: Might need input from @nturinski on this one. If the input is an `ArrayBuffer` (or `Uint8Array`), then the `.toString()` output will be something like `'10,20,30,255'`, for `[10, 20, 30, 255]`. I doubt this is what the service is expecting, but it has been like this for a couple years.

alexweininger (approved):

appservice/src/SiteClient.ts
```diff
@@ -456,13 +466,13 @@ export class SiteClient implements IAppSettingsClient {
...
 
-    public async vfsPutItem(context: IActionContext, data: string | ArrayBuffer, url: string, rawHeaders?: {}): Promise<AzExtPipelineResponse> {
+    public async vfsPutItem(context: IActionContext, data: string | ArrayBuffer | Uint8Array, url: string, rawHeaders?: Record<string, string>): Promise<AzExtPipelineResponse> {
         const client: ServiceClient = await createGenericClient(context, this._site.subscription);
         const headers = createHttpHeaders(rawHeaders);
         return await client.sendRequest(createPipelineRequest({
             method: 'PUT',
             url,
-            body: typeof data === 'string' ? data : data.toString(),
+            body: typeof data === 'string' ? data : new TextDecoder('utf-8').decode(data),
```
bwateratmsft: The old code worked, but somewhat by luck. VSCode's FS interfaces pass around `Uint8Array`, which is what we are passing into this method. At runtime, in the desktop VSCode, it's actually a Node `Buffer`, which implements `.toString(encoding?: ...)`, with UTF-8 being the default encoding. So, in the past we were taking the byte array and encoding it to UTF-8 before upload.

Thing is, `Uint8Array` (the base type also shared by web) does _not_ do that; it has just `.toString()` which writes unencoded, comma-separated `Uint8`s.

The better solution is to use `TextDecoder` to be explicit about what we're trying to do. It supports `ArrayBuffer`, `Uint8Array`, and `Buffer` all the same. `TextDecoder` is more technically correct, the best kind of correct.

appservice/src/SiteClient.ts
```diff
@@ -150,12 +152,14 @@ export class SiteClient implements IAppSettingsClient {
     }
 
     public async stop(): Promise<void> {
+        // eslint-disable-next-line @typescript-eslint/no-unused-expressions
```
nturinski: Do these lint errors go away if we return the value? In general, this eslint rule is kind of annoying since we use this pattern pretty often.

appservice/src/utils/kuduUtils.ts
```diff
@@ -43,13 +43,13 @@ export async function ignore404Error(context: IActionContext, callback: () => Pr
...
-export function convertQueryParamsValuesToString(queryParams: { [key: string]: unknown }): { [key: string]: string } {
+export function convertQueryParamsValuesToString(queryParams: { [key: string]: string | boolean | number | object | undefined | null | bigint }): { [key: string]: string } {
     const keys: string[] = Object.keys(queryParams);
 
     for (const key of keys) {
         if (typeof queryParams[key] === 'object') {
             // skip objects, the ServiceClient will handle them
-        } else {
+        } else if (queryParams[key] !== undefined && queryParams[key] !== null) {
             queryParams[key] = String(queryParams[key]);
```
nturinski: I doubt it matters too much, but the behavior changes a bit here. If the value was undefined/null, we'd still define `queryParams[key]` as "undefined". Not sure if completely skipping those keys is going to cause any issue.

appservice/src/utils/workspace.ts
```diff
@@ -40,9 +40,9 @@ export async function selectWorkspaceFile(context: IActionContext, placeHolder:
 export async function selectWorkspaceItem(context: IActionContext, placeHolder: string, options: vscode.OpenDialogOptions): Promise<vscode.WorkspaceFolder | vscode.Uri> {
     let folder: vscode.WorkspaceFolder | undefined;
     if (vscode.workspace.workspaceFolders) {
-        const folderPicks: IAzureQuickPickItem<vscode.WorkspaceFolder | undefined>[] = await Promise.all(vscode.workspace.workspaceFolders.map((f: vscode.WorkspaceFolder) => {
+        const folderPicks: IAzureQuickPickItem<vscode.WorkspaceFolder | undefined>[] = vscode.workspace.workspaceFolders.map((f: vscode.WorkspaceFolder) => {
```
nturinski: Not really sure why this was a promise.all to begin with. I wonder if the map function used to have something async in it.

appservice/src/SiteClient.ts
bwateratmsft: What we should really be doing is if/else here, not this. But yeah, we could return it too.

Issue #2129, opened 2025-11-19
AppServicePlanListStep - performance improvement


ccastrotrejo:
We use the `AppServicePlanListStep` on the logic apps vscode extension when trying to create a new resource in azure.
However we have noticed that the `AppServicePlanListStep` take a long time to load the quickpicks. There are some users that it takes more than 2 minutes.

I did a quick time measurement on our side and the `const epPlan: AppServicePlan | undefined = await tryGetAppServicePlan(client, nonNullProp(plan, 'resourceGroup'), nonNullProp(plan, 'name'));` its the one function that takes a long time to be executed (I understand it does make API calls).

<img width="524" height="579" alt="Image" src="https://github.com/user-attachments/assets/8b4b0f0e-76c7-4190-a05e-6f6d79c12628" />

Is there something we can do to improve the performance on getting the quick picks?

nturinski:
> I noticed it's doing a fetch for every single app service plan

Looking into it, we only do it for EP and WS plans. We don't personally ever create these plans for users (I think you can select it in advanced create) but I'm assuming that they are getting used primarily by Logic Apps based on [this issue.](https://github.com/microsoft/vscode-azuretools/issues/2130)

I'm all for improving the perf, but I don't think it impacts most of our Functions/App Service users.

I'm not really sure why we're doing this either:
```
if (isNewSiteLinux === isPlanLinux) {
                if (location && !LocationListStep.locationMatchesName(location, plan.location)) {
                    hasFilteredLocations = true;
                } else {
                    picks.push({
                        id: plan.id,
                        label: nonNullProp(plan, 'name'),
                        description: plan.sku?.name,
                        data: plan
                    });
                }
            }
```

Seems like we only filter Linux plans based on location? 😕

nturinski:
To improve perf, we came up with this solution:

Start retrieving the app service plans (including the `tryGetAppServicePlan` to get the Linux OS info) on the first step of the wizard and have it run in the background while the user is filling out the rest of the wizard.

Then we just do all the filtering based on their prompts when they actually get to the `AppServicePlanListStep`

PR #2158, opened 2025-12-12
Improve `AppServicePlanListStep` performance

MicroFish91:
Based on this idea: https://github.com/microsoft/vscode-azuretools/issues/2129#issuecomment-3582921250

I still need to actually test it out though to see if this fixes all the perf issues


bwateratmsft (approved):
This helps but there's gotta be a better way...

PR #2163, opened 2026-01-08
eng: Small changes to make eng package compatible for the ACA migration

motm32:
These small changes are for the web view migration work I am doing in ACA. Just basically exposing the isAutoDebug, isAutoWatch constants so I can use them in the ACA web view esbuild config and also allowing for other esbuild config names to be added to the ignore list. If you want a closer look to how these changes are used here is the draft [PR](https://github.com/microsoft/vscode-azurecontainerapps/pull/1005)


eng/package.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "@microsoft/vscode-azext-eng",
     "author": "Microsoft Corporation",
-    "version": "1.0.0-alpha.8",
+    "version": "1.0.0-alpha.9",
```
bwateratmsft: `npm i` to get the lockfile too

bwateratmsft (approved):

PR #2171, opened 2026-01-22
utils: Remove escape-string-regexp dependency

alexweininger:
I was working in here and figured I might as well remove the dependency and implement it ourselves since we only use it once.


alexweininger:
> Where does that impl come from?

From the original package, should I add a comment?

PR #2186, opened 2026-02-09
auth: fix: Fix copy all optional properties in ExtendedEnvironment constructor

alexweininger:
Fixes https://github.com/microsoft/vscode-azureresourcegroups/issues/1360

Root cause: The @azure/ms-rest-azure-env Environment constructor has a longstanding bug — it uses if (this.prop) { this.prop = parameters.prop; } for optional properties, but since this.prop is undefined at that point, the condition is always false and the copy never happens. This drops 11 of 18 properties including storageEndpointSuffix.

Auth v6 introduced ExtendedEnvironment extends Environment which triggers this bug by calling super(Environment.AzureCloud). The static Environment.AzureCloud object is unaffected (it's a plain object literal), but any new Environment(...) instance loses the optional properties.

Fix: Added Object.assign(this, parameters) after super(parameters) in ExtendedEnvironment's constructor in vscode-azuretools/auth/src/utils/configuredAzureEnv.ts.

### DIFF
--- a/auth/CHANGELOG.md
+++ b/auth/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change Log
 
+## 6.0.0-alpha.2 - 2026-02-09
+
+* Fix copy all optional properties in `ExtendedEnvironment` constructor
+
 ## 6.0.0 - 2025-12-15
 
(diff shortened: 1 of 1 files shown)


bwateratmsft (approved):
Nice! This is why we alpha...

bwateratmsft:
> I've tested and verified that this does fix the linked issue - Human Alex

🤨

PR #2193, opened 2026-02-11
azure: Change the typings of the created clients

bwateratmsft:
This would have build-time effects so I'm revving the major...shouldn't have any runtime impact, other than throwing a better error when using Azure Stack + `createRoleDefinitionsItems()`.

### DIFF
--- a/azure/index.d.ts
+++ b/azure/index.d.ts
@@ -5,12 +5,12 @@
 
 /* eslint-disable @typescript-eslint/no-explicit-any */
 
-import { AuthorizationManagementClient, type RoleDefinition } from '@azure/arm-authorization';
-import { Identity, ManagedServiceIdentityClient } from '@azure/arm-msi';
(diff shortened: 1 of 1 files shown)


azure/src/wizard/RoleAssignmentExecuteStep.ts
```diff
@@ -81,7 +81,18 @@ class SingleRoleAssignmentExecuteStep<T extends types.IResourceGroupWizardContex
...
             const principalId = nonNullValueAndProp(wizardContext.managedIdentity, 'principalId');
 
-            await amClient.roleAssignments.create(scope, guid, { roleDefinitionId, principalId });
+            await amClient.roleAssignments.create(
+                scope,
+                guid,
+                {
+                    roleDefinitionId, // Regular SDK wants this
+                    principalId, // Regular SDK wants this
+                    properties: { // Azure Stack SDK wants this instead
```
MicroFish91: So does this mean the regular SDK doesn't care about having the Stack SDK properties present (in that shape) and vice versa?

PR #2267, opened 2026-04-24
webview: Fix button theme

alexweininger:
Map Fluent UI v9 theme tokens to VS Code theme variables, ensuring accurate and non-intrusive theming across webview components. The main improvements are a clear separation between theme-wide and Button-specific token overrides, and the addition of targeted unit tests to verify these mappings. This prevents Button color overrides from unintentionally affecting other components and improves maintainability.


alexweininger:
Reason for this PR is that our fluent buttons used in our webviews weren't following the vscode buttons theme exactly. It was close but not perfect since it was based on the wrong theme tokens not specific to buttons.

bwateratmsft (approved):

PR #2268, opened 2026-04-28
webview: fix production 

motm32:
Forgot to change the path's so that production also get's the webview assests. Tested this with container apps and it seems to work fine. 

Checkout this commit in the ACA PR to see the extension changes: https://github.com/microsoft/vscode-azurecontainerapps/pull/1068/changes/772e714fa017e5063fa2c415f62b4d07f86c62de

### DIFF
--- a/webview/src/extension/extensionVariables.ts
+++ b/webview/src/extension/extensionVariables.ts
@@ -7,8 +7,30 @@ import { type ExtensionContext } from "vscode";
 
 export namespace ext {
     export let context: ExtensionContext;
+    /**
+     * Absolute path to the directory containing the webview assets (`views.js` and `views.css`).
+     *
+     * When unset, the controller falls back to
+     * `<extensionPath>/node_modules/@microsoft/vscode-azext-webview/dist`, which only works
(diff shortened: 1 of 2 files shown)


webview/src/extension/extensionVariables.ts
```diff
@@ -7,8 +7,30 @@ import { type ExtensionContext } from "vscode";
...
+     */
+    webviewAssetsDir?: string;
+}
+
+export function registerWebviewExtensionVariables(
+    context: ExtensionContext,
+    options?: RegisterWebviewExtensionVariablesOptions,
+): void {
     ext.context = context;
+    ext.webviewAssetsDir = options?.webviewAssetsDir;
```
bwateratmsft: The pattern the other packages use here is to define an interface like `WebviewExtensionVariables` that has a required value for `webviewAssetsDir`, and use that as the input type for `registerWebviewExtensionVariables`, so that the caller must provide a value for `webviewAssetsDir` when registering.

I don't feel strongly one way or the other but something to consider.

nturinski (approved):

PR #2313, opened 2026-05-28
webview: Removed Generate with Copilot in webview

gavin-aguiar:
## Summary

Removes the JSON-pipeline "Generate Project" path from the Template Gallery AI tab and consolidates everything into a single, chat-first entry point. The AI tab now always hands the user off to Copilot Chat with a grounded prompt instead of attempting to scaffold files directly from a strict-JSON model response.


webview/src/extension/TemplateGalleryController.ts
nturinski, line 145: Don't want to block for this, but shouldn't use _ for private functions.

nturinski (approved):

PR #2314, opened 2026-06-02
auth: add optional createIfNone to getSessionWithScopes

alexweininger:
## Summary

Adds an optional `options` parameter to `AzureAuthentication.getSessionWithScopes`. Passing `{ createIfNone: true }` allows an interactive consent prompt when no session for the requested scopes has been granted yet, instead of failing silently.

This enables callers to eagerly obtain consent for a **non-management audience** (e.g. the App Service audience used for Kudu/SCM deployments) before it is first needed. Default behavior is unchanged: a plain scope list is still acquired silently, and challenge (MFA step-up) requests still always prompt.

## Why

Part of the fix for microsoft/vscode-azurefunctions#5073, where deploying an Azure Function App / App Service app fails with *"You are not signed in to an Azure account"* because the App Service audience token is acquired silently and first-time consent never prompts.

This is the **auth-package half** of the fix. The companion change in `@microsoft/vscode-azext-azureappservice` calls this new option at the Kudu credential chokepoint. The two changes are decoupled (the appservice side uses a structural cast), so there is no npm version coupling.

## Changes

- `getSessionWithScopes` now accepts `options?: GetSessionWithScopesOptions` with `createIfNone?: boolean`.
- New exported `GetSessionWithScopesOptions` interface.
- Version bump to `6.1.0-alpha.2`.

## Release

This ships inside the **Azure Resources extension** (`ResourceGroupsSubscriptionProvider extends VSCodeAzureSubscriptionProvider`), which must consume the bumped auth package and release for the runtime behavior to take effect.

Refs microsoft/vscode-azurefunctions#5073

### DIFF
--- a/auth/src/providers/AzureSubscriptionProviderBase.ts
+++ b/auth/src/providers/AzureSubscriptionProviderBase.ts
@@ -396,16 +396,21 @@ export abstract class AzureSubscriptionProviderBase implements AzureSubscription
                     return session;
                 },
                 getSessionWithScopes: async (scopeListOrRequest, options) => {
-                    this.silenceRefreshEvents();
                     // A challenge (e.g. an MFA step-up) must always be able to prompt so the user can
                     // satisfy it. For a plain scope list we stay silent by default, but allow callers to
                     // opt in to an interactive consent prompt via `options.createIfNone` (used, for example,
                     // to consent to the App Service audience before a deployment).
                     // See https://github.com/microsoft/vscode-azurefunctions/issues/5073
                     const createIfNone = isAuthenticationWwwAuthenticateRequest(scopeListOrRequest) || !!options?.createIfNone;
+                    if (createIfNone) {
+                        // Interactive consent can take a while, so suppress without timeout until it is
+                        // done, then silence for a bit longer afterwards (same pattern as `signIn`).
+                        this.suppressRefreshSuggestedEvents = true;
+                    } else {
+                        this.silenceRefreshEvents();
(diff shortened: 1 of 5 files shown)


auth/src/providers/AzureSubscriptionProviderBase.ts
```diff
@@ -395,12 +395,19 @@ export abstract class AzureSubscriptionProviderBase implements AzureSubscription
...
+                    // A challenge (e.g. an MFA step-up) must always be able to prompt so the user can
+                    // satisfy it. For a plain scope list we stay silent by default, but allow callers to
+                    // opt in to an interactive consent prompt via `options.createIfNone` (used, for example,
+                    // to consent to the App Service audience before a deployment).
+                    // See https://github.com/microsoft/vscode-azurefunctions/issues/5073
+                    const createIfNone = isAuthenticationWwwAuthenticateRequest(scopeListOrRequest) || !!options?.createIfNone;
+                    const session = await getSessionFromVSCode(scopeListOrRequest, tenant.tenantId, { ...(createIfNone ? { createIfNone: true } : { silent: true }), account: tenant.account });
+                    if (createIfNone) {
+                        // Interactive consent can take a while, so silence refresh events for a bit longer.
+                        this.silenceRefreshEvents();
```
bwateratmsft, line 414: Need to silence before and after, same as `signIn` does. But, it may be better to just use `signIn`.

bwateratmsft (approved):

PR #2315, opened 2026-06-02
appservice: eagerly consent to App Service audience before Kudu/SCM calls

### LINKED ISSUE #5073: Bug Report: Azure Functions Extension v1.21.2 - "You are not signed in to an Azure account" on Deploy
Environment

- VS Code: 1.122.1
- Azure Functions Extension: 1.21.2 (broken) / 1.21.0 (working)
- Azure Resources Extension: 0.12.5
- OS: Windows
- Runtime: Python 3.12.3
-  Azure Function Hosting Plan: Flex Consumption

Description:
After updating the Azure Functions extension to v1.21.2, deploying to any Function App from VS Code fails with the error:
Error: You are not signed in to an Azure account. Please sign in.
The error occurs after the extension successfully scans all project files, exactly at the moment it attempts to initiate the upload to Azure.

What was ruled out:
- Azure CLI authentication is valid and working
- Azure Functions Core Tools deploy works correctly (func azure functionapp publish)
- VS Code correctly resolves account, tenant and subscription (confirmed in Azure Account logs)
- Creating new Function Apps in Azure from VS Code works correctly
- The issue occurs with all Function Apps, not a specific one
- Re-login, VS Code reinstall, token cache clearing, and "microsoft-authentication.implementation": "classic" setting did not resolve the issue

Root cause
The issue is specific to the deploy flow in v1.21.2 of the extension. It is not an authentication problem — all other Azure operations work correctly with the same credentials. For instance, when trying to create a Function remotely in the Azure portal, there were no error whatsoever, however, when deploying an already existing function, I got that annoying notification prompting me to log in (after I was already logged in)

Workaround
Downgrading the extension to v1.21.0 resolves the issue immediately.

Steps to reproduce:
1. Install Azure Functions extension v1.21.2
2. Open any Azure Functions project in VS Code
3. Right-click on the project → Deploy to Function App
4. Observe the error in the Azure Functions output channel

Hopefully, this issue is fixed for the next release. For the time being, I share the solution I found after a few hours trying to understand why the extension stopped working out of the blue.

alexweininger:
## Summary

Fixes microsoft/vscode-azurefunctions#5073: deploying an Azure Function App / App Service app fails with **"You are not signed in to an Azure account. Please sign in."**

## Root cause

Since `@microsoft/vscode-azext-azureappservice` 5.0.0, all Kudu/SCM operations route through the **App Service audience** token (`https://<appservice>/.default`). `getAppServiceCredentials` acquires that token **silently** via `createCredentialsForScopes`. Silent acquisition never prompts for first-time consent, so if the user has not yet consented to the App Service audience, the call fails.

## Fix

Before creating the (silent) credential, `getAppServiceCredentials` now eagerly calls `getSessionWithScopes(scopes, { createIfNone: true })`, allowing a **one-time interactive consent prompt**. Once granted, consent is cached and later calls resolve silently.

This is the single chokepoint for **all** Kudu/SCM operations — deploy, log streaming, tunnel/remote-debug — so one change covers them all.

The `authentication` object is read directly from `ISubscriptionContext`, which is now properly typed (`@microsoft/vscode-azext-utils` 4.1.1 exposes `authentication?: AzureAuthentication` from the resources API). The call is still optional-chained (`?.`) so it degrades gracefully on contexts that do not provide it — or on older runtime hosts whose `authentication` object predates `getSessionWithScopes`.

## Dependency / release note

- Bumps `@microsoft/vscode-azext-utils` to `^4.1.1` (which exposes the `authentication` typing).
- The `createIfNone` option this relies on is added in `@microsoft/vscode-azext-azureauth` `6.1.0` (companion PR #2314), which ships its `getSessionWithScopes` implementation via the Azure Resources extension at runtime.

This appservice change is consumed by both the Azure Functions and App Service extensions.

## Changes

- Eager `getSessionWithScopes(scopes, { createIfNone: true })` in `getAppServiceCredentials`.
- Retire the local `SubscriptionAuthentication` shim/cast in favor of the real `ISubscriptionContext.authentication` typing; bump utils dep to `^4.1.1`.
- Version bump to `5.0.2`.

Fixes microsoft/vscode-azurefunctions#5073

### DIFF
--- a/appservice/src/utils/appServiceEnvironment.ts
+++ b/appservice/src/utils/appServiceEnvironment.ts
@@ -42,7 +42,8 @@ export function getAppServiceScopes(environment: Environment): string[] {
  * Declared locally because this package does not depend on the resources API typings.
  */
 interface SubscriptionAuthentication {
-    getSessionWithScopes(scopeListOrRequest: string[], options?: { createIfNone?: boolean }): Promise<unknown>;
+    // Optional because older hosts may provide an `authentication` object without this method.
+    getSessionWithScopes?(scopeListOrRequest: string[], options?: { createIfNone?: boolean }): Promise<unknown>;
 }
 
 /**
@@ -56,10 +57,11 @@ export async function getAppServiceCredentials(subscription: ISubscriptionContex
     // may not have consented to yet. Eagerly acquire a session here allowing an interactive prompt,
     // so the subsequent silent credential acquisition — and the Kudu operation itself — don't fail
     // with "You are not signed in to an Azure account". This is a one-time consent: once granted it
-    // is cached and resolves silently on later calls. See
-    // https://github.com/microsoft/vscode-azurefunctions/issues/5073
+    // is cached and resolves silently on later calls. The method itself is optional-chained so we
+    // degrade gracefully on older hosts whose `authentication` object predates `getSessionWithScopes`.
+    // See https://github.com/microsoft/vscode-azurefunctions/issues/5073
     const authentication = (subscription as ISubscriptionContext & { authentication?: SubscriptionAuthentication }).authentication;
-    await authentication?.getSessionWithScopes(scopes, { createIfNone: true });
+    await authentication?.getSessionWithScopes?.(scopes, { createIfNone: true });
 
     return {
         credentials: await subscription.createCredentialsForScopes(scopes) as TokenCredential,
--- a/appservice/src/utils/appServiceEnvironment.ts
+++ b/appservice/src/utils/appServiceEnvironment.ts
@@ -37,11 +37,30 @@ export function getAppServiceScopes(environment: Environment): string[] {
 }
 
 /**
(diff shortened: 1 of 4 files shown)


appservice/src/utils/appServiceEnvironment.ts
```diff
@@ -36,12 +36,31 @@ export function getAppServiceScopes(environment: Environment): string[] {
     return [`${resourceUrl}/.default`];
 }
 
+/**
+ * Minimal shape of the `authentication` object that the Azure Resources host spreads onto every
+ * subscription context (see `createSubscriptionContext` in `@microsoft/vscode-azext-utils`).
+ * Declared locally because this package does not depend on the resources API typings.
+ */
+interface SubscriptionAuthentication {
```
alexweininger: this is temporary until I release the typings changes in the resources api

nturinski (dismissed):

MicroFish91 (dismissed):

bwateratmsft (approved):

PR #2321, opened 2026-06-05
azure: preserve authentication in parseClientContext

alexweininger:
parseClientContext deliberately copies only the known subscription fields onto the action context ("copy over just the subscription info"), which silently dropped the newly-exposed `authentication` object. Any client created via this path therefore lost the ability to prompt for scope-specific consent — defeating the App Service audience consent fix for microsoft/vscode-azurefunctions#5073 on those code paths. Copy `authentication` through as well.

Requires @microsoft/vscode-azext-utils ^4.1.1 (which exposes ISubscriptionContext.authentication); bumps azureutils to 4.2.1.

### DIFF
--- a/azure/src/createAzureClient.ts
+++ b/azure/src/createAzureClient.ts
@@ -32,7 +32,7 @@ export function parseClientContext(clientContext: InternalAzExtClientContext): I
             isCustomCloud: subscription.isCustomCloud,
             // Carry `authentication` through so scope-specific consent (e.g. the App Service audience
             // prompt for vscode-azurefunctions#5073) isn't silently dropped on contexts routed through here.
-            authentication: subscription.authentication
+            authentication: subscription.authentication,
         });
     } else {
         return clientContext;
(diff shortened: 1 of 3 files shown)


azure/src/createAzureClient.ts
```diff
@@ -29,7 +29,10 @@ export function parseClientContext(clientContext: InternalAzExtClientContext): I
             tenantId: subscription.tenantId,
             userId: subscription.userId,
             environment: subscription.environment,
-            isCustomCloud: subscription.isCustomCloud
+            isCustomCloud: subscription.isCustomCloud,
+            // Carry `authentication` through so scope-specific consent (e.g. the App Service audience
+            // prompt for vscode-azurefunctions#5073) isn't silently dropped on contexts routed through here.
+            authentication: subscription.authentication
```
bwateratmsft: Nit: include a trailing comma so the next time doesn't modify this line

bwateratmsft (dismissed):

azure/src/createAzureClient.ts
alexweininger: @copilot do this please

bwateratmsft (approved):

PR #2332, opened 2026-06-09
azure: widen azext-azureauth range to ^6.1.0-alpha.2

alexweininger:
azureutils declared `@microsoft/vscode-azext-azureauth@^6.0.0-alpha.6`, whose tuple (6.0.0) cannot be satisfied by the published 6.1.0-alpha.2 under npm's prerelease semver rules, so consumers ended up with a duplicate nested auth copy. Widen the range to ^6.1.0-alpha.2 so the single 6.1.0-alpha.2 dedupes.

No version bump. I'll ship this together with https://github.com/microsoft/vscode-azuretools/pull/2331

### DIFF
--- a/azure/package-lock.json
+++ b/azure/package-lock.json
@@ -20,7 +20,7 @@
                 "@azure/core-client": "^1.10.1",
                 "@azure/core-rest-pipeline": "^1.23.0",
                 "@azure/logger": "^1.3.0",
-                "@microsoft/vscode-azext-azureauth": "^6.0.0-alpha.6",
+                "@microsoft/vscode-azext-azureauth": "^6.1.0-alpha.2",
                 "@microsoft/vscode-azext-utils": "^4.1.1",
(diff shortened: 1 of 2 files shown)


azure/package-lock.json
```diff
@@ -1255,20 +1284,38 @@
...
+                "@azure/core-auth": "^1.10.1",
+                "@azure/core-rest-pipeline": "^1.23.0",
+                "@azure/identity": "^4.13.1",
                 "@azure/ms-rest-azure-env": "^2.0.0"
             },
             "engines": {
                 "vscode": "^1.106.0"
             }
         },
+        "node_modules/@microsoft/vscode-azext-azureauth/node_modules/@azure/arm-resources-subscriptions": {
```
bwateratmsft, line 1302: This will probably have some merge conflicts with #2331 (if not a real merge conflict, then an implicit one where it is bundling things that it won't actually use)

bwateratmsft (approved):

PR #2334, opened 2026-06-10
auth: Add DI-friendly `./next` entrypoint for azureauth (TokenCredential input)

bwateratmsft:
## Summary

Introduces a modern, dependency-injected `./next` entrypoint for `@microsoft/vscode-azext-azureauth` and turns the legacy `.` entrypoint into a thin wrapper around it. Bumps the package to **`7.0.0-alpha.1`**.

> **Draft / version note:** opened as a draft. The major bump is provisional — if we confirm there are no breaking changes to the **existing public `.` / `./azdo` API**, I'll back this off to a minor (`6.1.0-alpha.x`). See "Breaking-change assessment" below.

## What's new in `./next`
- **Dependency injection of `vscode`** — the injected namespace is used both for the authentication namespace and for the configuration lookup, enabling testing and DI.
- **Input `TokenCredential`** via a credential factory; the base provider only ever uses the injected factory (no hard-coded credential).
- **`VsCodeExtensionCredential`** — a `TokenCredential` whose `vscode` authentication namespace is injectable.
- **`./next/testing` entrypoint** — `createAzureDevOpsCredential` wrapping `@azure/identity`'s `AzurePipelinesCredential`, lazily imported so `@azure/identity` isn't always bundled.
- **No `@azure/ms-rest-azure-env`** — replaced by a structural `EnvironmentLike` with built-in cloud constants.
- **`AzureSubscription` exposes only its `TokenCredential`** — `getSession`/`getSessionWithScopes` are gone from the `./next` contract.
- **`@azure/logger` logger** instead of `vscode.OutputChannel`; callers route it to an output channel if desired.
- **Built-in `bearerTokenAuthenticationPolicy`** replaces the custom `MFAChallengePolicy`, exposed as reusable `createChallengeBearerTokenPolicy` / `applyChallengeBearerTokenPolicy` so other Azure clients can reuse the same MFA/challenge handling.

## Legacy `.` and `./azdo` entrypoints
Kept as thin wrapper/re-export layers over `./next`, preserving their existing public surface: localized errors (`NotSignedInError`), the `ms-rest-azure-env`-based `Environment`, and the per-subscription `authentication` member. The `./azdo` `AzureDevOpsSubscriptionProvider` still extends the legacy base, so its subscriptions keep the legacy shape (per-subscription `authentication` derived from the federated DevOps credential) in addition to the new `credential`.

## Tests
Extensive new unit coverage under `test/next/` (credential, environment resolution, challenge policy, token-expiry parsing, end-to-end challenge retry via a fake HTTP client). **239 tests passing**; `tsc`, `eslint --max-warnings 0`, and the full esm+cjs build are green.

## Breaking-change assessment (to decide major vs. minor)
- The **legacy `.` entrypoint** is intended to be fully backward-compatible (wrappers preserve types, localization, and members).
- The **`./azdo` entrypoint** is also backward-compatible: `AzureDevOpsSubscriptionProvider` still returns legacy `AzureAccount`/`AzureSubscription` shapes with the per-subscription `authentication` member.
- `./next`, `./next/testing`, and the reusable policy exports are **new** API (only these may break, since they didn't exist before).
- ⚠️ The sibling `azure` package still depends on `^6.1.0-alpha.2`; if we keep the major bump, its dependency range needs updating. Left untouched pending the major-vs-minor decision.

## Commits
~3,470 insertions / ~1,138 deletions across 42 files; ~26% of additions are tests. The later commits in this branch focus on deduplicating the legacy entrypoint into thin `./next` wrappers to keep the net change down.


bwateratmsft:
Closing for now. Rethinking things.

PR #2355, opened 2026-06-29
webview: Add  entryPoint  to Template Gallery create-project message …

gavin-aguiar:
## Description

The Template Gallery has two ways to create a project: the "Use Template" button on a tile (quick create, skips details) and the "Create Project" button on the template details view. Both currently emit an identical  createProject  message, so consuming extensions can't tell which path a user took. This change lets them attribute creates to the originating UI.

### Changes

• Add  ProjectCreationEntryPoint = 'card' | 'details'  and an optional  entryPoint  field on  CreateProjectMessage  ( types.ts ).
•  TemplateGalleryView  now tags creation at the source: the tile button sends  'card' , the details view's "Create Project" sends  'details' .
• Thread  entryPoint  through  TemplateGalleryController._handleCreateProject  into the abstract  createProject(template, language, location, entryPoint?) .
• Export  ProjectCreationEntryPoint  from the package index for consumers.

### DIFF
--- a/webview/package-lock.json
+++ b/webview/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "@microsoft/vscode-azext-webview",
-    "version": "1.1.0",
+    "version": "1.2.0",
     "lockfileVersion": 3,
     "requires": true,
     "packages": {
         "": {
             "name": "@microsoft/vscode-azext-webview",
-            "version": "1.1.0",
+            "version": "1.2.0",
             "license": "MIT",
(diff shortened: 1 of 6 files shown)


nturinski (commented):
You'll need to bump the package.json version and run npm install (to update the package-lock.json) so that we can release these changes.

gavin-aguiar:
> You'll need to bump the package.json version and run npm install (to update the package-lock.json) so that we can release these changes.

Done.

PR #2357, opened 2026-07-01
utils: surface troubleshooting guidance for certificate trust errors

alexweininger:
## Problem

When an operation fails because the Node.js extension host cannot build a trusted TLS certificate chain (for example `unable to get local issuer certificate` / `SELF_SIGNED_CERT_IN_CHAIN`), the extension shows only the raw OpenSSL error. On developer machines this is almost always a corporate proxy or SSL-inspection appliance whose root CA is not trusted by Node. The fix is documented (`NODE_EXTRA_CA_CERTS`), but the error itself gives the user no hint that this is the cause or where to look.

Related to microsoft/vscode-azureappservice#2899 and microsoft/vscode-azurefunctions#5124.

## Change

- Adds a shared classifier `isCertificateTrustError(parsedError)` (`utils/src/utils/certificateTrustError.ts`) that recognizes the known Node/OpenSSL certificate-chain error codes and their human-readable messages.
- Wires it into the central `handleError` path in `callWithTelemetryAndErrorHandling`. When an error is classified as a certificate trust failure and display is not suppressed, the extension now:
  - appends a short explanatory hint to the error notification and the output channel, and
  - adds a **Learn more** button that opens the troubleshooting documentation.

Because this lives in the shared `@microsoft/vscode-azext-utils` error path, **every** extension that uses `callWithTelemetryAndErrorHandling` gets the improved guidance automatically, with no per-extension changes.

## Scope / design notes

- Classification is intentionally limited to certificate-chain errors (codes + known OpenSSL strings) to avoid false positives on unrelated network failures. Broad proxy-connection classification is out of scope.
- No public API change: the classifier is internal and unit-tested via direct import; `index.d.ts` is untouched.
- The troubleshooting link points at the canonical Resource Groups README troubleshooting section that the other extensions' READMEs already link to.

## Validation

- `tsc -p ./ --noEmit` clean, `eslint --max-warnings 0` clean.
- `vscode-test`: 249 passing, including 5 new tests for the classifier.

### DIFF
--- a/utils/src/callWithTelemetryAndErrorHandling.ts
+++ b/utils/src/callWithTelemetryAndErrorHandling.ts
@@ -12,6 +12,8 @@ import { parseError } from './parseError';
 import { cacheIssueForCommand } from './registerReportIssueCommand';
 import { IReportableIssue, reportAnIssue } from './reportAnIssue';
 import { AzExtUserInput } from './userInput/AzExtUserInput';
+import { certificateTroubleshootingLink, isCertificateTrustError } from './utils/certificateTrustError';
+import { openUrl } from './utils/openUrl';
 import { limitLines } from './utils/textStrings';
 
 const maxStackLines: number = 3;
@@ -169,8 +171,15 @@ function handleError(context: types.IActionContext, callbackId: string, error: u
         }
 
         if (!context.errorHandling.suppressDisplay) {
+            const isCertError: boolean = isCertificateTrustError(unMaskedErrorData);
+            const certHint: string = l10n.t('This can happen behind a corporate proxy or SSL-inspection appliance whose certificate authority is not trusted by VS Code. See the troubleshooting steps for how to trust it.');
+
             // Always append the error to the output channel, but only 'show' the output channel for multiline errors
             ext.outputChannel.appendLog(l10n.t('Error: {0}', unMaskedErrorData.message));
+            if (isCertError) {
+                ext.outputChannel.appendLog(certHint);
+                ext.outputChannel.appendLog(l10n.t('Troubleshooting: {0}', certificateTroubleshootingLink));
+            }
 
             let notificationMessage: string;
             if (unMaskedErrorData.message.includes('\n')) {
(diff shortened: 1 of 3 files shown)


utils/src/utils/certificateTrustError.ts
```diff
@@ -0,0 +1,63 @@
...
+ * Message fragments that map to the same certificate trust failures. Some SDKs surface the
+ * human-readable OpenSSL string rather than the error code.
+ */
+const certificateErrorMessages: readonly string[] = [
+    'unable to get local issuer certificate',
+    'unable to verify the first certificate',
+    'self signed certificate',
+    'self-signed certificate',
+    'certificate has expired',
+];
```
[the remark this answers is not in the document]
alexweininger: Good catch, agreed. Removed 'certificate has expired' from the classifier in 8a109f45; an expired leaf cert or bad local clock isn't an untrusted corporate root CA, so keeping it out avoids surfacing the proxy/CA guidance where it doesn't apply.

utils/src/callWithTelemetryAndErrorHandling.ts
```diff
@@ -169,8 +171,15 @@ function handleError(context: types.IActionContext, callbackId: string, error: u
         }
 
         if (!context.errorHandling.suppressDisplay) {
+            const isCertError: boolean = isCertificateTrustError(unMaskedErrorData);
+            const certHint: string = l10n.t('This can happen behind a corporate proxy or SSL-inspection appliance whose certificate authority is not trusted by VS Code. See the troubleshooting steps for how to trust it.');
```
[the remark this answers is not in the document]
alexweininger: Agreed, reworded in 8a109f45 to attribute the failure to the Node.js extension host being unable to build a trusted chain (rather than 'VS Code'), and to mention NODE_EXTRA_CA_CERTS.

utils/src/utils/certificateTrustError.ts
```diff
@@ -0,0 +1,64 @@
...
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import type { IParsedError } from '../../index';
+
+/**
+ * Default troubleshooting documentation for certificate trust failures. All Azure extensions
+ * link their README troubleshooting sections to this canonical location.
+ */
+export const certificateTroubleshootingLink: string = 'https://github.com/microsoft/vscode-azureresourcegroups/blob/main/README.md#troubleshooting';
```
alexweininger: turn this into aka.ms link

alexweininger: here's the aka.ms link I created: aka.ms/AA11ri61

alexweininger: Done in 6cec7b39 — certificateTroubleshootingLink now points at https://aka.ms/AA11ri61 instead of the branch-specific README URL, so it survives README moves/renames. Updated the doc comment to match.

bwateratmsft (approved):

